PR #100, opened 2014-10-15
NullReferenceException bugfix for null arguments

postromantic:
Nasty bug introduced by @Florian in 0dfa3468 fixed. Unit test added


christianrondeau:
Follow-up on @postromantic request, let us know what your plans for a release with this particular fix on NuGet!

PR #103, opened 2014-10-16
Introducing JintIteropException 

auz34:
Our users owns javascript only and we provide them some API which can be utilized inside Javascript. Once something is wrong with the script it is important to localize place from which exception comes. If it is inside the script debugging should be done by user themselves. If excpetion has been thrown somewhere inside interop methods it is important to be aware about it since debugging procedures are completely different in this case.


sebastienros:
Merge this other branch to provide the minimal set of features first #326 

Thanks

PR #109, opened 2014-10-28
Improve JsValue performance by removing Nullable overhead.

ala53:
This is in reference to an issue filed yesterday.

All nullable fields in JsValue are now not Nullables, just being initialized to a default value.
Additionally, ObjectInstances also cache methods now. (By the way, the file ObjectInstance .cs had a space in it, which I removed.


postromantic:
StructLayout changes in this PR introduced a new bug: #139

PR #115, opened 2014-11-17
Allow non-empty return statements without a function body

eamodio:
Allows for better backward compatibility with Jint v1 (and seems like it should be valid?)
return; is still disallowed to keep inline with the ECMA spec tests

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -1183,5 +1183,18 @@ namespace Jint.Tests.Runtime
 
             Assert.Equal(value, result);
         }
+
+        [Fact]
+        public void ShouldAllowNonEmptyReturnWithoutFunctionBody()
+        {
+            RunTest(@"return 'hello';");
+            Assert.Equal("hello", _engine.GetCompletionValue().ToObject());
+        }
+
+        [Fact]
+        public void ShouldThrowParserExceptionWhenEptyReturnWithoutFunctionBody()
+        {
+            Assert.Throws<ParserException>(() => RunTest(@"return;"));
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


eamodio:
Doh! Didn't realize a rebase would have wiped out the previous comments. Sorry about that.
I fixed up the changes from the review - hopefully should be ready now.

sebastienros:
I prefer to follow the spec on this, and not allow such constructs, The recommendation is to handle this case inside your logic. Thanks though for the effort.

eamodio:
@sebastienros How about if it was under a configuration flag to the engine? Not on by default? While I understand the desire to follow the spec, it seems to me that given Jint is "typically" (maybe I'm way off here) something embedded into some larger app, it would seem like having "scripts" (programs) that have no functions but return a value is quite important.

Otherwise each script that wanted to do that would need to be wrapped in
```
(function() { /* real script here */ })();
```

But that still falls down in certain cases if the "real" script part doesn't have an explicit return itself.

Is there another workaround I'm missing?

fvaneijk:
We are currently using jint in a larger app and defiantly have the need to allow return (with value) from non functions.  I would be fine with a configuration flag.  In addition we have added some private/application specific features to  the JavaScript language that could be made conditional with such a flag.

honestegg:
@eamodio I had a similar need in our app. Our users can write scripts and sometimes they use a return statement and sometimes they don't. I found that you can use the Tolerant flag for the parser to allow a return statement outside of a function.
```csharp
var engine = new Engine();
engine.Execute("return 0;", new ParserOptions
{
    Tolerant = true
});
```
Does this solve the issue for you?

eamodio:
@honestegg Wow - that definitely works to get around the issue. Unfortunately it is a very brute force approach (as everything else that is under that "tolerant" umbrella I want to be an error). Maybe there is a class of tolerance? A parser option to control just a single (or range of) "tolerant" exceptions?

PR #116, opened 2014-11-17
Added unit tests around creating "namespaced" object for the script

eamodio:
I've gotten the functionality I wanted to work in one set of test (but it seems far more complex than maybe it should be)
And another similar scenario doesn't work

I'm sure there is plenty wrong with this, but am submitting this pull request as a conversation around this.

### DIFF
--- a/Jint.Tests/Jint.Tests.csproj
+++ b/Jint.Tests/Jint.Tests.csproj
@@ -64,6 +64,7 @@
     <Compile Include="Runtime\Domain\IPerson.cs" />
     <Compile Include="Runtime\Domain\Person.cs" />
     <Compile Include="Runtime\Domain\Shape.cs" />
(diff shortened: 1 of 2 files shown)


eamodio:
@sebastienros Sorry for the long delay.

What I was attempting to do was to expose a "runtime" environment within a script, within a nested namespace.

So I wanted to be able to access the C# DoWork method as ixp.data.doWork in the JS environment. And then also be able to update the ixp.data.doWork method to point to another function (in this case DoWork2 - the use-case around this, is suppose doWork is really a closure with some internal state that needs to be updated).

The first few tests (ShouldAllowMethodNamespacing*) succeed in that regard but seems like more work than it should be to get the desired functionality.

The other tests were attempts to get similar functionality, but through different (easier) means, but without complete success.

Hopefully that makes a little more sense -- if not please let me know.

PR #119, opened 2014-12-11
Return error location with JavaScriptException.

yambaypaul:
Add line and column location of error returning JavaScriptException.


djMax:
+1 on this... Debugging is a black hole right now in Jint.

doronguttman:
In most cases you can get it as shown below, but I would also +1 to have it built into the JavaScriptException itself:
private void GetLocation(out int? line, out int? column)
        {
            var location = _engine.GetLastSyntaxNode().Location;
            if (location == null)
            {
                line = null;
                column = null;
            }
            else
            {
                line = location.Start.Line;
                column = location.Start.Column;
            }
        }

djMax:
Ooh that's good info. Can you get a stack similarly?

PR #121, opened 2014-12-27
allow parsing of incomplete programs

Tvede-dk:
Added the ability to parse incomplete programs, and added a new element to allow external code to detect the end of the incomplete program. Parameters have been added with default values to make it compatible with excisting code, and to keep it simple.


sebastienros:
Thanks, but as discussed on the related issues I don't want to support this scenario.

PR #130, opened 2015-02-05
implemented conversion with newtonsoft json package

npenin:
Hi,

I implemented it because it is very easy to build/parse json from newtonsoft.json nuget package and collaboration with jint and newtonsoft seems a great advantage to me.


fvaneijk:
Any way you could provide some unit tests?

npenin:
Sure, is there a preferred place to put them ?
Already existing jint test project ? if yes, new file or existing file ?

fvaneijk:
I would suggest a new file at: ...\Jint.Tests\Runtime\NewtonSoftJson.cs  if the extension becomes a standard part of Jint else put them in a new assembly of your choice.

npenin:
here they are ;)

fvaneijk:
thanks

sebastienros:
I don't understand what the scope of the change is. Dare explaining ?

PS: @npenin and I wrote the first version of Jint together ;) Welcome back

npenin:
Sure. WebAPI implements has Newtonsoft.Json  as a default wrapper. When you have for example a dynamic parameter in your web api, you actually get a JToken. As I am using this dynamic parameter directly in Jint, I needed this convesion. 
I could also have implemented a Web API Json converter using jint, but implementing a conversion to the "official" converter seemed more straight foreward to me and would open potentially more features as it provides also a json linq interface to build and read your json object.

npenin:
Any chance you merge it in this week release ?

sebastienros:
Actually accepting this would have Jint introduce external dependencies. Can we make this a plugin or at least find a way to accept such plugins ? For instance there is already a way to extend how objects are converted, did you try this?

npenin:
I did it as a plugin as you say. I had just updated jint to introduce a missing AsDate to be consistent with other conversions. The whole conversion and dependency to Newtonsoft.Json is in another project. Don't you know me enough to know that I hate dirty code ? ;)

sebastienros:
Cool, then please create another PR with just As/IsDate()

Issue #124, opened 2015-01-08
Issues with string.Format vargs


wqweto:
This returns empty string:

            var engine = new Jint.Engine(o => o.AllowClr());
            var result = engine.Execute("System.String.Format('{0}', 42)")
                .GetCompletionValue().ToObject();

This returns "42"

            var engine = new Jint.Engine(o => o.AllowClr());
            var result = engine.Execute("System.String.Format('{0}', 42, 0)")
                .GetCompletionValue().ToObject();

Jint version 2.4.0.0

PR #132, opened 2015-02-18
bugfix: be able to pass single number as params array

postromantic:
**Bug**: Cannot pass single number to the .NET method as params object[] argument.
**Reason**: ArrayConstructor has JS behaviour for array construction: if there is only one argument and it's number, then it's length of array (not an array with one element)
**Solution**: Replace ArrayContructor with low level JS array creation in params resolution resolver method.

_Please update NuGet version with this fix._


wqweto:
Could this be a resolution for issue #124?

postromantic:
Good point, @wqweto. Yes, #124 is the same issue.

Issue #139, opened 2015-02-24
bug: cannot load Jint assembly in partial trusted IIS environment


postromantic:
Got IIS with Medium trust level. Jint 2.1.0.0 worked fine, but latest Jint failed to load with:

    System.TypeLoadException: Could not load type 'Jint.Native.JsValue' from assembly 'Jint, Version=2.4.0.0, Culture=neutral, PublicKeyToken=2e92ba9c8d81157f' because objects overlapped at offset 8 and the assembly must be verifiable.

postromantic:
PullRequest created: #140

sebastienros:
I accepted the PR but you know that partial trust is no more "supported" by IIS? The idea is that it's not a valid way to ensure security in an app, so you should just use Full trust and add correct security.

postromantic:
I know IIS doesn't support partial trust anymore. It's just a repro for the problem with explicit field offsets overlapping.

PR #140, opened 2015-02-24
FieldOffset fix to avoid TypeLoad exception in partial trust IIS level

postromantic:
Bugfix for #139: set +4 field offset to avoid overlapping

### DIFF
--- a/Jint/Native/JsValue.cs
+++ b/Jint/Native/JsValue.cs
@@ -85,7 +85,7 @@ namespace Jint.Native
         [FieldOffset(8)]
         private readonly ObjectInstance _object;
 
-        [FieldOffset(8)]
+        [FieldOffset(12)]
         private readonly string _string;
 
         [FieldOffset(16)]


ala53:
Unfortunately, this bugfix will only work on x86.

Can you put a patch up that does something like this:

```
        [FieldOffset(0)]
        private readonly bool _bool;

        [FieldOffset(0)]
        private readonly double _double;

        [FieldOffset(8)]
        private readonly ObjectInstance _object;

        [FieldOffset(8 + sizeof(IntPtr))]
        private readonly string _string;

        [FieldOffset(8 + sizeof(IntPtr) * 2)]
        private readonly Types _type;
```

That keeps us aligned to memory boundaries and handles different pointer sizes.

ala53:
Sebastien, I think its a bad idea to merge it because if you run on an x64 os, the pointers will be 8 bytes and will cause an overlap with the _type variable (which will fail).

postromantic:
@ala53, i agree, we should discuss it. 
#139 is x86 environment, so in theory assembly verification in x64 partial trusted env will fail.

__What's the reason of using explicit field offsets here?__ Avoid memory fragmentation? I don't see unmanaged code interaction where it's really important.
What about to use C++ union (FieldOffset=0 for all) https://msdn.microsoft.com/en-us/library/acxa5b99.aspx?f=255&MSPPError=-2147217396 ? Or use StructLayout.Sequental instead of explicit to let .NET align it in memory automatically?

ala53:
We had an explicit layout as an optimization: we overlapped the two objects (string and ObjectInstance) which has now been reverted and also overlapped the bool and double. FieldOffset=0 for all won't work because .Net disallows overlapping memory pointers for objects and structs (already tried it).

postromantic:
Yep, pure union doesn't work. I see 3 possible solutions:
1. Use Explicit struct layout with the following offsets: 0 0 8 16 20. There will be 'gaps' on x86, no overlapping.
2. Use Sequental struct layout and remove all field offset attributes. .NET will align fields automatically
3. Do not use struct layout attributes
What do you think?

ala53:
We could maybe use nested structs:

In pseudocode:
```
[StructLayout]
public struct base {
[FO(0)]
public double d;
[FO(0)]
public bool b;
[FO(0)]
public int i;

public Types t;

[FO(sizeof double + sizeof Types)]
private base__internal bi;

public string s { get {return bi.s;}}

//and so on

private struct base__internal {
public string s;
public object o;
}

}
```

However, we'd need to be very careful to lay everything out on memory boundaries or we will lose any performance benefit we hope to gain.

christianrondeau:
Agreed on the negligible gain; I believe you'll need to release a separate 32 bit and 64 bit versions of Jint to use that optimization; I personally suggest reverting that commit.

postromantic:
PR #150 created to remove that optimization. I think Jint should be one solid PCL (without x86/x64 versions).

christianrondeau:
Thanks @postromantic

PR #145, opened 2015-02-25
IsNaN() fix

fvaneijk:


sebastienros:
Can you undo the FieldSet change ?

fvaneijk:
reverted the field offset change in JsValue.cs but it now shows a merge conflict, the file differences seem correct though...

sebastienros:
Maybe you'd better create a new pull request

PR #150, opened 2015-03-04
Make Jint x86 and x64 compliant

postromantic:
Explicit struct layout field offsets removed. 
See #140 for discussion

### DIFF
--- a/Jint/Native/JsValue.cs
+++ b/Jint/Native/JsValue.cs
@@ -18,7 +18,6 @@ using System.Runtime.InteropServices;
 
 namespace Jint.Native
 {
-    [StructLayout(LayoutKind.Explicit)]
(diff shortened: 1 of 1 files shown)


fvaneijk:
This seems like the best and simplest approach.

sebastienros:
Isn't there a way to have different offsets based on a precompilation directive. I am sure it is provided as part of the build, and we could generate the different targets in the nuget package too. I remember this change improved the perf quite a bit and it would be sad to loose it.

postromantic:
@sebastienros, originally this change was checked in with other caching/optimization code.Teamcity tests execution time of this pull request is even faster than with FieldOffsets. Do you have any memory profiler tests on that?

sebastienros:
Then it's all good !

PR #171, opened 2015-04-10
Implement Jint.Runtime.NamespaceCache*

jmorgannz:
Class to help cache known namespaces, removing some assumptions Jint.Runtime.Interop.NamespaceReference has to make about type information.

Also helps provide more specific exceptions in certain situations; for example, with this change it becomes obvious fairly quickly if you are missing an assembly in your ClrSecurity setup - whereas before it was giving errors about generics, and you really had no idea why.

### DIFF
--- a/Jint.Tests/Runtime/InteropTests.cs
+++ b/Jint.Tests/Runtime/InteropTests.cs
@@ -1247,5 +1247,19 @@ namespace Jint.Tests.Runtime
             Assert.Equal(Nested.ClassWithStaticFields.Readonly, "Readonly");
         }
 
+        [Fact]
+        public void ShouldThrowRelevantExceptionOnUndefinedNamespace()
+        {
+            RunTest(@"
+                try {
+                    var Jint = importNamespace('Jint');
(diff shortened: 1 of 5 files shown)


sebastienros:
Doesn't compile apparently.

jmorgannz:
Sorry, my bad. I manually merged the files I changed on a machine without build tools, so missed a file reference.

jmorgannz:
Turns out this change breaks generic support. I'm new to xUnit so has been a bit of a learning curve around running the test suite to ensure nothing is broken.
Will investigate and patch if possible, and perhaps add a unit test for the scenario this change is intended to cover too.

jmorgannz:
There we go. 

Just a couple of comments: 
- Not sure the added complexity this patch adds is worth it given it only affects what type of exception is thrown in a given situation
- Not sure how you would go about referencing a generic type definition that is nested in another generic type definition - but suspect there will be issues here - even without this patch?

sebastienros:
Thanks for the effort, but I am afraid the improvement is not worth the complexity.

jmorgannz:
Yeah I figured that would be the case.
Was just one attempt to solve ambiguity between generics/namespaces and
proper object reference checking - maybe theres a simpler solution.

-Jesse

PR #175, opened 2015-04-29
Implemented the ability to convert non null values from JavaScript to Nu...

joncloud:
...llable<T>.  Implemented the ability to call C# methods that have optional parameters.


sebastienros:
Could you add a few unit tests to show what is expected and that there is no regression?
Thanks

PR #177, opened 2015-04-29
Feature/ignore method casing

csabakiss:
I updated the ObjectWrapper so PascalCased .NET methods can be called from JavaScript using camelCase. This can be really useful, to avoid conflicts caused by the different naming conventions.

### DIFF
--- a/Jint.Tests/Runtime/InteropTests.cs
+++ b/Jint.Tests/Runtime/InteropTests.cs
@@ -126,6 +126,21 @@ namespace Jint.Tests.Runtime
         }
 
         [Fact]
+        public void CanInvokeObjectMethodsWithPascalCase()
+        {
+            var p = new Person
+            {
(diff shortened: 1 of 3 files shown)


genielabs:
it is not allowing lower case with one-word-only methods.
For example I have:

    hg.Program.Pause(1);
    var f = hg.Program.InputField("Some.Field");

that cannot be called like:

    hg.program.pause(1);
    var f = hg.program.inputField("Some.Field");

but it can be called as:

    hg.Program.Pause(1);
    var f = hg.Program.inputField("Some.Field");

(note **inputField** is working)

genielabs:
https://github.com/sebastienros/jint/pull/178/files#diff-6b9d78ce3a57ce031987cbfdd492ff27L63
improved to support properties as wel.

PR #178

PR #180, opened 2015-05-13
Fix #179 toInt32/toUInt32 handling

### LINKED ISSUE #179: Issue with twiddle bitwise operator?
This code in browserify's Buffer module:

```start = ~~start```

In "regular javascript", if start is 58 at the beginning, it should be 58 at the end. In Jint, it's -1.

### PR DESCRIPTION
**Pull request by djMax**
This is mostly advisory because I realize it also includes another PR to add error location (which I find super useful) in addition to fixing #179 - where ToInt32 and ToUint32 are using C# casting operators which do not behave as Ecma Script demands for various inputs. You can see this by running "~~58" through the parser, which should come back as 58 but instead comes back as -1.

All the relevant changes are in TypeConverter. ToUint16/ToInt16 may need similar treatment.

Comments appreciated. And if you merge @yambaypaul 's changes, then I guess this would be a real PR :)

### DIFF
--- a/Jint.Tests/Runtime/Domain/ClassWithStaticFields.cs
+++ b/Jint.Tests/Runtime/Domain/ClassWithStaticFields.cs
@@ -15,4 +15,38 @@
             Setter = "Setter";
         }
     }
+
+    public class Nested
+    {
+        public class ClassWithStaticFields
+        {
+            public static string Get = "Get";
+            public static string Set = "Set";
+
+            public static string Getter
+            {
+                get { return "Getter"; }
+            }
+
+            public static string Setter
+            {
+                get
+                {
+                    return _setter;
+                }
+                set
+                {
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by fatalerrorx** [comment]
You can't do num == double.NaN need to use double.IsNan

**Comment by djMax** [comment]
Fixed, thanks.

**Comment by sebastienros** [comment]
I just added a unit for this test case and it works as expected:

```csharp
[Theory]
[InlineData(-59d, "58")]
[InlineData(58d, "~~58")]
public void ShouldInterpretUnaryExpression(object expected, string source)
{
    var engine = new Engine();
    var result = engine.Execute(source).GetCompletionValue().ToObject();
    Assert.Equal(expected, result);
}
```

sebastienros:
Could you please re-create this PR with just the necessary changes?
Thanks

PR #185, opened 2015-05-28
added as/is date

npenin:
As discussed in my previous PR, here is the changes on JsValue only to be more consistent, and enable newtonsoft.json (or other ?) conversion.

### DIFF
--- a/Jint/Native/JsValue.cs
+++ b/Jint/Native/JsValue.cs
@@ -102,6 +102,12 @@ namespace Jint.Native
         {
             return IsObject() && AsObject() is ArrayInstance;
         }
+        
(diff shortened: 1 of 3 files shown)


Jint/Native/JsValue.cs
```diff
@@ -161,6 +167,16 @@ public ArrayInstance AsArray()
         }
 
         [Pure]
+        public DateInstance AsDate()
+        {
+            if (!IsArray())
```
sebastienros: Copy-paste issue, and for the sake of catching it, please add a unit test.

npenin:
Any chance to get a new nuget package including this PR ?

PR #187, opened 2015-05-28
New import* keyword

Willy-Kimura:
Hi Sebastienros,

Thanks for this great scripting tool! There's the new **"import"** keyword in the new JavaScript standard which can be really great and handy if you actually translate it to Jint. The **"import"** keyword can be implemented as a scripting library importer (for importing external scripts to the currently edited script) like in html pages:

```
script src="mylibrary.js"
```
translated to Jint: 
```
import "mylibrary.js" ,"tools.js";

//import function myFunction() from external JS script
import "myFunction()" from "mylibrary.js";
```
...or as an array representation:
```
import ["mylibrary.js", "tools.js"];
```

It can also be used as a *Class* importer keyword to help in reducing multiple code reuses while calling internally-defined functions in C#/VB, just like the ```using``` keyword in C# or the ```imports``` keyword in VB.NET. All this is so as to help in building large and highly scalable external scripts. Will be really grateful!

### DIFF
--- a/Jint/Jint.csproj
+++ b/Jint/Jint.csproj
@@ -155,6 +155,7 @@
     <Compile Include="Parser\ParserExtensions.cs" />
     <Compile Include="Parser\ParserOptions.cs" />
     <Compile Include="Parser\Position.cs" />
+    <Compile Include="Parser\SourceReader.cs" />
     <Compile Include="Parser\State.cs" />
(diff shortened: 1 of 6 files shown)


sebastienros:
Sorry, closing this PR as it is failing the unt tests, and also contains some non-related changes in the Parser. Can you also open a bug to start discussing the subject with other members?

Thanks

PR #197, opened 2015-06-22
Allow nullable delegate parameters to be excluded

honestegg:
Allow nullable delegate parameters to be excluded from the JS function call


honestegg:
BTW -- this bug was introduced in version 2.5.0. When you have some time, I would be grateful for a new Nuget package because we can't upgrade to 2.5 as it is.

PR #198, opened 2015-06-23
Perf/flags

sebastienros:
Analyzing perf improvements and simplification to be done on flags


sebastienros:
In the spec, defineOwnProperty, Return false, if the [[Enumerable]] field of Desc is present and the [[Enumerable]] fields of current and Desc are the Boolean negation of each other.

It actually checks that the field is present or not, there is a unit test for that, so it needs to be settable, even as false.

PR #209, opened 2015-07-04
Adding Out parameters for .net methods

fvaneijk:
As an example in InteropTests.cs:

```
    [Fact]
    public void ShouldCallMethodWithOutArg1()
    {
        _engine.SetValue("a", new A());

        RunTest(@"
            var str = null;
            a.Call17(str);
            assert('out string' == str);
        ");
    }
```

In A.cs:
        public void Call17(out string x)
        {
            x = "out string";
        }

### DIFF
--- a/Jint.Tests/Runtime/Domain/A.cs
+++ b/Jint.Tests/Runtime/Domain/A.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Timers;
 using Jint.Native;
 
 namespace Jint.Tests.Runtime.Domain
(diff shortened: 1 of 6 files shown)


sebastienros:
The changes in JsValue are raising a red flags. I think there should be no changes there, but in the method wrappers only. Maybe there needs to be another kind of JsValue which would be composed of a standard one and handle the out parameter.

fvaneijk:
I'll sync with your latest on the master and see where it ends up while taking your suggestions into account.

fvaneijk:
Can you take another look and see what else should be done?

PR #217, opened 2015-08-02
cancel script execution

andrei-tatar:
Added ability to cancel execution using a cancellation token.
Useful in many situations.


sebastienros:
Can you please add a unit test?

SebastianStehle:
I think this is obsolete now: https://github.com/sebastienros/jint/blob/dev/Jint/Constraints/CancellationConstraint.cs

Should be closed.

PR #232, opened 2015-10-01
 Catching CLR exceptions in JavaScript code

Novakov:
Now it is possible for JavaScript code to catch exceptions thrown by CLR methods.

### DIFF
--- a/Jint/Native/Function/ScriptFunctionInstance.cs
+++ b/Jint/Native/Function/ScriptFunctionInstance.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using Jint.Native.Object;
 using Jint.Parser;
 using Jint.Runtime;
@@ -94,6 +95,16 @@ namespace Jint.Native.Function
 
                     if (result.Type == Completion.Throw)
                     {
+                        if (result.Value.HasValue && result.Value.Value.IsObject())
+                        {
+                            var e = result.Value.Value.ToObject() as Exception;
+                            if (e != null)
+                            {
+                                e.Rethrow();
+                                throw new InvalidOperationException("Clr is not working");
+                           }
+                        }
+
                         JavaScriptException ex = new JavaScriptException(result.GetValueOrDefault());
                         ex.Location = result.Location;
                         throw ex;
--- a/Jint/Runtime/StatementInterpreter.cs
+++ b/Jint/Runtime/StatementInterpreter.cs
@@ -19,7 +19,30 @@ namespace Jint.Runtime
 
         private Completion ExecuteStatement(Statement statement)
         {
-            return _engine.ExecuteStatement(statement);
+            try
+            {
+                return _engine.ExecuteStatement(statement);
+            }
+            catch(JavaScriptException)
+            {
+                throw;
+            }
+            catch (RecursionDepthOverflowException)
+            {
+                throw;
+            }
+            catch (StatementsCountOverflowException)
+            {
+                throw;
+            }
+            catch (TimeoutException)
+            {
+                throw;
+            }
+            catch (Exception e)
+            {                
+                return new Completion(Completion.Throw, JsValue.FromObject(_engine, e.InnerException), null);
+            }
         }
 
         public Completion ExecuteEmptyStatement(EmptyStatement emptyStatement)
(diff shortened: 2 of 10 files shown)


Jint/Native/Function/ScriptFunctionInstance.cs
```diff
@@ -94,6 +95,16 @@ public override JsValue Call(JsValue thisArg, JsValue[] arguments)
 
                     if (result.Type == Completion.Throw)
                     {
+                        if (result.Value.HasValue && result.Value.Value.IsObject())
```
sebastienros: you can use `result.Value.TryCast<Exception>()` for the full block

Jint/Native/Function/ScriptFunctionInstance.cs
```diff
@@ -94,6 +95,16 @@ public override JsValue Call(JsValue thisArg, JsValue[] arguments)
...
+                        {
+                            var e = result.Value.Value.ToObject() as Exception;
+                            if (e != null)
+                            {
+                                e.Rethrow();
```
sebastienros: how is that different from just `throw;` ?

Jint/Native/Function/ScriptFunctionInstance.cs
```diff
@@ -94,6 +95,16 @@ public override JsValue Call(JsValue thisArg, JsValue[] arguments)
...
+                                throw new InvalidOperationException("Clr is not working");
```
sebastienros: > CLR is not working

That's a funny message, not sure I would grasp the meaning ;)

Jint/Native/Function/ScriptFunctionInstance.cs
Novakov: Simple `throw e; ` will replace stack trace. `Rethrow` method tries to preserve it.

Jint/Native/Function/ScriptFunctionInstance.cs
Novakov: `Rethrow` method will always throw exception, so this line will never be executed.

Jint/Native/Function/ScriptFunctionInstance.cs
sebastienros: `throw e;` will replace the stacktrace, right, but `throw;` won't, hence my comment.

Jint/Native/Function/ScriptFunctionInstance.cs
Novakov: It looks like that `TryCast` will cast `ObjectInstance` to `Exception` not object itself.

Jint/Native/Function/ScriptFunctionInstance.cs
sebastienros: Ah, now I get the joke! Well, please remove the line then, even more since apparently you just need to use `throw;` and it will display a warning otherwise.

Jint/Native/Function/ScriptFunctionInstance.cs
Novakov: This code is not inside `catch` block, so `throw` cannot be used.

Jint/Runtime/StatementInterpreter.cs
```diff
@@ -19,7 +19,30 @@ public StatementInterpreter(Engine engine)
 
         private Completion ExecuteStatement(Statement statement)
         {
-            return _engine.ExecuteStatement(statement);
+            try
```
sebastienros: Can we refactor all Jint's exceptions to inherit from a common one, like `JintException`, then catch this one a throw on it, and fallback to you last catch otherwise?

Novakov: Sure, will do.

Jint/Native/Function/ScriptFunctionInstance.cs
sebastienros: I see ... let me enable turbo brain mode, might take some time to warm up

Jint/Runtime/StatementInterpreter.cs
Novakov: Done in 40b4c2776d1bd60b2c27b7c5cb74aadd109943a4

fatalerrorx:
You should only be catching TargetInvocationException not Exception. That way you don't need to catch and rethrow JintException's and TimeoutExeption. Also ExecuteStatementList already has a tc by adding tc to ExecuteStatement your making jint slower.

If your using ExceptionDispatchInfo.Capture shouldn't it be inside the catch clause? Otherwise how can it save the stack trace etc if you call it later via the ReThrow method.

What do you think?

sebastienros:
Merge this other branch to provide the minimal set of features first #326 

Thanks

PR #235, opened 2015-10-16
Performance boosted by around 20% (SunSpider)

nilproject:


sebastienros:
The title is very promising, let's see how much time the unit tests take. In the meantime do you know which of the changes actually improve it? Might be as simple as the `foreach`, which actually could be done in more places. The issue is that the PR contains completely independent changes, and it's hard to grasp the reason or impact between each of them. Mind commenting the code on this PR?

nilproject:
On my laptop sunspider performed around 1 minute 40 seconds instead of 2 minutes for original version. Change foreach to for and As<T>() to "as" not produced a noticeable increase. Main increase was due to changes in the size of the JsValue due to the combination of fields. This solution reduces number of operations for transfer JsValue

sebastienros:
I can confirm locally. I tried to replace all `foreach` loops and the changes are not significant. I am trying the `JsValue` refactoring now.

sebastienros:
I just created a new PR with the JsValue changes to see if this only isn't breaking the tests

nilproject:
Look at second commit

sebastienros:
Can you tell me more about your JavaScript interpreter? And also, if you want to help on performance, here is a PR that you might want to work on, it shows what the current biggest bottleneck is: Array methods perf. https://github.com/sebastienros/jint/pull/212

nilproject:
I tried to replace core of Array with my implementation (SparseArray) but performance not changed. I tried to replace MruPropertyCache2 with my StringMap2, but it's has no effect. I think, than bottleneck in other place. But i'm not sure.
My engine builded with another architecture. Instead of using pattern Visitor (StatementInterpreter), every AST-node can evaluate by self. It work fast because virtual functions work faster then switch-case. Every object contains field "attributes". It allow to remove PropertyDescriptor. Also my JSValue is class, not struct. All nodes of expressions store the "tempContainer" - i do not create new object for every step of evaluation. But my engine single-thread. Multithread evaluation of one part of code can lead to unpredictable consequences. Engine creates descriptor for all variables. This descriptor implement cache and manage life-cycle of variable. Depending on many parameters, js-functions can be invoked one of many methods (eg pseudo-inline).

sebastienros:
I tried everything for the Array too, that's why I know that the real issue is in the `ArrayPrototype` methods directly, like `Splice`, .... They are just following the spec, which is not the optimal algorithm, and does not take any advantage of the internal structure of the data.

There are cool ideas. I should try the virtual approach someday. But performance is not a priority for Jint, it's fast enough to be comparable to other engines, but the key is conformance to the spec here, which is the biggest advantage with Jint.

sebastienros:
I have mentioned this PR from the one I merged, containing only the JsValue optimization. Thanks a lot for the trick.

PR #236, opened 2015-10-16
Compacting JsValue fields

sebastienros:


sebastienros:
Based on https://github.com/sebastienros/jint/pull/235

Issue #249, opened 2015-11-21
Error creating Delegate as Action<int>


qnnnnez:
using System;
    using Jint;
    namespace Test
    {
        class Program
        {
            static void ExecuteAction(Action<int> action, int number)
            {
                action(number);
            }
            static void Main(string[] args)
            {
                var engine = new Engine();
                engine.SetValue("executeAction", new Action<Action<int>, int>(ExecuteAction));
                engine.Execute(@"executeAction(function(n){System.Console.WriteLine(n);}, 1);");
            }
        }
    }

When I executed the program, an exception was raised.

    An unhandled exception of type 'System.ArgumentException' occurred in System.Core.dll
    Additional information: 类型为“System.Int32”的表达式不能用于方法“Jint.Native.JsValue FromObject (Jint.Engine, System.Object)”的类型为“System.Object”的参数

I'm sorry for the language, but I think it says that we cannot pass a `System.Int` argument to `JsValue.FromObject`.

I think it is `DefaultTypeConverter.Convert` that caused the problem, but I have no idea about how to fix it, as I am not good at C#.

I wonder whether it is a bug, and if it is not, how can I pass a `Action<int>` to a method written in C#. Thanks a lot.

qnnnnez:
Well, I tried to fix it, and it seems to work.

PR #250, opened 2015-11-21
boxing before calling JsValue.FromObject

qnnnnez:
fix: https://github.com/sebastienros/jint/issues/249


sebastienros:
Thanks. It would be great to add a unit test in EngineTests to ensure that the bug is not regressed.

PR #269, opened 2016-02-18
Fix a error that occurs when creating an instance of value type without parameters

Taritsyn:
When creating an instance of value type without parameters the following error occurs:

	No public methods with the specified arguments were found.


sebastienros:
Thanks!

PR #276, opened 2016-03-03
Date should honor the timezone rules for daylight savings

eamodio:
Currently jint isn't honoring timezone rules when converting dates to a string. For example if the local timezone is EST, but the date is in the past or future when daylight savings would be observed jint doesn't account for that.


sebastienros:
Awesome, thanks.

PR #277, opened 2016-03-05
Support JsValue.ToObject on Arguments

eamodio:


eamodio:
Not sure why the above failed -- looks like a TeamCity issue (as it was extremely slow earlier)

PR #278, opened 2016-03-05
Line/column for exceptions in nested functions weren't correct

eamodio:
Given the following script:

``` JS
function test() {
    o.boom();
}
test();
```

Previously the exception would indicate the error was on the `test();` line rather than the `o.boom();` line.


sebastienros:
Can you fix the commits to only include the actual change and not the formatting fixes?

Thanks

sebastienros:
It was actually stupid to ask for removing the spacing changes.

PR #280, opened 2016-03-05
Don't include Locals in Globals when gathering DebugInformation

eamodio:


sebastienros:
Could you ELI5 what the issue is? Thanks.

eamodio:
This was just to avoid have duplicate information in the Global vs Local debug objects.

PR #281, opened 2016-03-06
Allow return statements without a function body (with setting)

eamodio:
Allows for better backward compatibility with Jint v1
Note since return; is allowed we break some of the ECMA spec tests (when the parser setting is used)


sebastienros:
I have already commented on this feature in the issues. This is not something I want to include by default. The solution I recommend is to create some extension method in your project that uses the existing parser APIs to parse a Body instead.

PR #284, opened 2016-03-14
Debugger improvements

pvginkel:
This pull request contains a number of improvements to better support debugging. An example of a project that uses these changes can be found at https://github.com/pvginkel/JintDebugger.

Besides a few bug fixes, the largest change is that sources can now be identified. Every time a source is loaded, a new instance of the `Script` class is created. The purpose of this class is to uniquely identify a source file (or eval or function). This is then used by the debugger to show the source to the user. Also, before, breakpoints were not tied to a specific source. The `Breakpoint` class now also has a property for the `Script` instance to ensure that the breakpoint is actually of the correct file.

### DIFF
--- a/Jint/Native/Json/JsonParser.cs
+++ b/Jint/Native/Json/JsonParser.cs
@@ -837,16 +837,13 @@ namespace Jint.Native.Json
                 {
                     Range = new int[0], 
                     Loc = 0,
-
+                    Source = new Script(options?.Source, code)
                 };
+
+            _engine.InvokeSourceLoaded(_extra.Source);
 
             if (options != null)
             {
-                if (!System.String.IsNullOrEmpty(options.Source))
-                {
-                    _extra.Source = options.Source;
-                }
-
                 if (options.Tokens)
                 {
                     _extra.Tokens = new List<Token>();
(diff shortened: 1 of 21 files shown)


Jint/Native/Json/JsonParser.cs
```diff
@@ -879,7 +876,7 @@ private class Extra
         {
             public int? Loc;
             public int[] Range;
-            public string Source;
+            public Script Source;
```
sebastienros: If it's a filename then just call it Filename, if it's really source code, is it the full loaded script or just the portion represented by Loc/Range ?

Jint/Parser/JavascriptParser.cs
```diff
@@ -62,11 +63,17 @@ public class JavaScriptParser
 
 
         public JavaScriptParser()
+            : this(null)
+        {
+        }
+
+        public JavaScriptParser(Engine engine)
```
sebastienros: This is a big issue for me, I'd really like to keep the parser independent from the engine.

fvaneijk: Could the SourceLoaded event can be a member of the parser instead?

Jint/Native/Json/JsonParser.cs
fvaneijk: Agreed the naming could be better.  Code and Source members of Script may be redundant.  They are used to track/compare breakpoints in required/included files.

sebastienros:
Closing in favor of https://github.com/sebastienros/jint/pull/852

Issue #285, opened 2016-03-17
Date.prototype.ToISOString returning negative time values pre 1970


mcshaz:
If i run the following test:
 ```
[Fact]
public void TestToDateStrings()
{
	var testDates = new[] { new DateTime(2000, 1, 1), new DateTime(2000, 1, 1, 0, 15, 15, 15), new DateTime(1900, 1, 1, 0, 15, 15, 15), new DateTime(1999, 6, 1, 0, 15, 15, 15) };
	foreach (var tzId in new[] { "Pacific Standard Time", "New Zealand Standard Time", "India Standard Time" })
	{
		TimeZoneInfo tzi = TimeZoneInfo.FindSystemTimeZoneById(tzId);
		var engine = new Engine(ctx => ctx.LocalTimeZone(tzi));
		foreach (var d in testDates)
		{
			var td = new DateTimeOffset(d, tzi.GetUtcOffset(d));
			engine.Execute(
				string.Format("var d = new Date({0},{1},{2},{3},{4},{5},{6});", td.Year, td.Month - 1, td.Day, td.Hour, td.Minute, td.Second, td.Millisecond));
			Assert.Equal(td.ToString("ddd MMM dd yyyy HH:mm:ss 'GMT'zzz"), engine.Execute("d.toString();").GetCompletionValue().ToString());
			Assert.Equal(td.UtcDateTime.ToString("yyyy-MM-dd'T'HH:mm:ss.fff'Z'"), engine.Execute("d.toISOString();").GetCompletionValue().ToString());
		}
	}
}
```
It fails on the pacific timezone with the message:
```
Assert.AreEqual failed. 
Expected:<1900-01-01T08:15:15.015Z>. 
  Actual:<1900-01-01T-16:-45:-45.-985Z>.
```
A pull request, with all tests passing, is accompanying. I have not included the test above (although the test above is passing), but would suggest something similar, but more fitting with the impressive testing framework, be included, including NZ (UTC+13 over the southern hemisphere summer can provide some edge case failures), a country on :30 or :15 timezones (India in this case), Northern and southern hemisphere zones and dates in the daylight saving period for both hemispheres (i.e. Jan & July), as well, of course, as the pre-1970 date.

PR #287, opened 2016-03-17
fix Date ToISOString pre 1970

mcshaz:
fix issue #285. All tests passing


sebastienros:
Looks good, thanks. Would you mind adding a unit test for each case ?
Thanks

mcshaz:
I am not familiar with xunit, but I have tried to add appropriate tests so that Date.prototypes toISOString and toString are producing 'valid' output. I have not tested toUTCString, as another test was already dealing with this. I suspect toString is not under the ECMA spec, but it is still probably worth testing that it returns something sensible.

PR #289, opened 2016-03-23
Adding AST Walker

fvaneijk:
Adds ability to walk the ast produced by the parser.  This can be useful to extract certain structural data from the js code enabling building tools to aid during script editing and analysis.

As an example a toString on a function will print the function...

function test(a,b)
{
  return a + b;
}

print(test.toString());

--------output---------
function test(a, b) 
{ 
return a + b; 
}


anyolejoe:
Great - thanks for adding that.

sebastienros:
I appreciate the idea and the effort but this will not be part of Jint, I'd prefer to see it as a separate project. The risk being that every time a new feature is added I would have to support how to serialize it, and also people would start reporting all the little issues with this feature. This would just add more noise and time to spend that I can't afford.

If you decide to create this project, I am more than willing to add a reference to it in the readme.

PR #291, opened 2016-03-28
Fix toInt32 and toUint32

djMax:
I feel like I've made this PR before, but according to github it seems I haven't, and it doesn't seem master has these updates. I can barely remember what it's fixing, but the money commit is "Fix implementation of ToInt32 and ToUInt32 to match ECMA-262 9.5 and 9.6"

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -1160,6 +1160,28 @@ namespace Jint.Tests.Runtime
             RunTest(@"
                 assert(true === isNaN());
             ");
-        }    
+        }
+
+        [Theory]
+        [InlineData(4d, 0, "4")]
+        [InlineData(4d, 1, "4.0")]
+        [InlineData(4d, 2, "4.00")]
+        [InlineData(28.995, 2, "29.00")]
+        [InlineData(-28.995, 2, "-29.00")]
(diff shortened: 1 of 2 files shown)


sebastienros:
I also remember this code. But what's for sure is that it will require a unit test to showcase the issue and prevent any regression ;)

djMax:
Found the old convo. #180  - but you said your test (which I still see in the tests) work. So I'm trying to figure out what was failing exactly and why this change fixed it.

djMax:
The test calls ToObject() which ends up with a double, which is why it's not a direct match I think. But I duplicated the tests with ToInt32 and I still don't get the behavior.

djMax:
I've added a couple tests. They weren't failing before unfortunately. BUT, I do think this implementation is much more in line with the actual ECMA standard, so I guess I'm hoping to prey on your desire for matching the spec. :)

sebastienros:
I will try to repro the bug to understand the issue. But I am pretty sure the double casting is to explictly solve this. At the same time I am super careful with this method as it's used all the way, internally everything is a `double`, and also everything has to be taken as an integer so conversions appear everywhere.

sebastienros:
Even though it looks more in line with the spec, the performance optimization that was made can't be removed. The double cast has the same behavior in the end.

PR #294, opened 2016-04-02
CancellationScope for cancelling script execution

xenolightning:
Added CancellationScope, allowing externally triggered cancellation of script execution.

Example usage:
```
var cts = new CancellationTokenSource();

using (new CancellationScope(cts.Token))
{
    var engine = new Engine();

    //This script would never end.
    engine.Execute("while(true){}");
}

//Some other thread can call
cts.Cancel();

```

Took the lead from the existing `StrictModeScope`, it seems like scoping is a better solution than adding it to the engine. It allows for single executions to be cancelled, and following executions on the same engine to be run. Not a very common scenario, but could be useful.

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -1638,5 +1638,43 @@ namespace Jint.Tests.Runtime
                 assert(String(a) === String(b));
             ");
         }
+
+        [Fact]
+        public void RegExpResultIsMutable()
+        {
+            RunTest(@"
+                var match = /quick\s(brown).+?(jumps)/ig.exec('The Quick Brown Fox Jumps Over The Lazy Dog');
+                var result = match.shift();
+                assert(result === 'Quick Brown Fox Jumps');
+            ");
+        }
+
+        [Fact]
+        public void RegExpSupportsMultiline()
+        {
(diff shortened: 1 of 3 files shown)


xenolightning:
The scope could be pulled into the execute method, and the token passed as an optional parameter. That might slightly restrict the usage of the scope, and not require user knowledge of the internal workings.

The token can then be shared across executions if needed.

sebastienros:
Is there a way to alter your commits so that we can actually see what the changes are? In the diff is appears like the whole files are new and I can't understand the changes.
Also I would greatly appreciate a unit test to ensure it's doesn't regress. Not with a while loop but a counter and ensure the counter doesn't hit a specific value.
Thanks

xenolightning:
Hmm, that's really weird, not sure why it thinks there's been deletions and additions. It should be about 20 additional lines + a class. I'll redo the changes.

Can you explain the counter test? It seems like something that isn't testable. The cancellation has to be externally triggered, most probably a cancel after in tests. The only other way I can think of, would be to have a callback from inside the running script to cancel itself, which seems weird.

Going back to the cancel after to check a counter value. This would be dependent on how fast the script runs, and where the cancellation is processed. 

There is no way to precisely ensure where the cancellations happen, only that they do happen.

PR #296, opened 2016-04-02
Implement object.setPrototypeOf to allow inheritance to work properly

djMax:
Without this, babel transpiles will not work. I'm sure you'll want tests - I'm a little unclear how to add them. Is there a version of ECMA tests you originally used updated for ES6? All I found was this, which is a very different format:

https://github.com/tc39/test262


djMax:
Is this fixed another way these days? If so I (or you) can close this PR.

PR #300, opened 2016-05-09
Fix for JSON.parse handling non-string values

Jos635:
According to [the spec](http://www.ecma-international.org/ecma-262/5.1/#sec-15.12.2), JSON.parse(text) should use ToString(text) to get the string value. The previous implementation would throw an error when JSON.parse was called with a non-string value.

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -876,7 +876,56 @@ namespace Jint.Tests.Runtime
             ");
         }
 
-        [Fact]
+		[Fact]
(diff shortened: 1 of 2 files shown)


sebastienros:
I am not sure `object.ToString()` is the way to do what the spec mentions as ToString. You should instead use `TypeConverter.ToString()`

sebastienros:
Also, would you mind adding a unit test that would ensure we are not regressing the feature? Thanks

Jos635:
Ah I see, I thought ToString() returned the same thing as JavaScript's toString, my bad. Fixed. I've also added a unit test.

PR #307, opened 2016-05-27
Added support for long as a return value when applicable

grisha-kotler:
e.g. 7 - long, 7.0 - double


sebastienros:
What problem are you trying to solve?
I am not very glad to add this change as numbers in javascript are double values.

sebastienros:
And currently your changes are breaking 50 unit tests.

PR #309, opened 2016-06-30
fix #144

### LINKED ISSUE #144: callback from c# to js failed
tests case throw ArgumentException
“System.Linq.Expressions.Expression`1[System.Action`1[System.String]]” 
can not convert to 
“System.Action`1[System.String]”
=======================
      public static void Main(String[] args)
        {
            var engine = new Engine(cfg => cfg.AllowClr());
            engine.SetValue("log", new Action<object>( delegate(Object obj) { Console.WriteLine(obj); } ));
            engine.SetValue("testCallBack", new System.Action<string, System.Action<string>>(delegate(string url, System.Action<string> cb) { cb("js callback from c# " + url); }));
            engine.Execute(@"  testCallBack('a',function(v){ log(v)  }) ");
            Console.Read();
        }
======reason ==========
DefaultTypeConverter will convert ScriptFunctionInstance.Call to Expression.Lambda but not a System.Action

regexing:
This fixes https://github.com/sebastienros/jint/issues/144.

### DIFF
--- a/Jint/Runtime/Interop/DefaultTypeConverter.cs
+++ b/Jint/Runtime/Interop/DefaultTypeConverter.cs
@@ -95,7 +95,7 @@ namespace Jint.Runtime.Interop
                                                     @vars),
                                                 jsValueToObject), Expression.Empty());
 
-                        return Expression.Lambda(callExpresion, new ReadOnlyCollection<ParameterExpression>(@params));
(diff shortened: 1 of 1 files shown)


sebastienros:
Can you add a simple unit test to ensure we don't regress?

PR #311, opened 2016-07-19
A number of interop related performance optimizations

sergeystoma:
Hey, 

A project that I'm working on uses Jint to let users write their own business rules to extend functionality. Rules tend to be pretty light on the actual javascript beyond simple math or comparisons with a string manipulation sprinkled around, but instead spend most of the time calling various .NET objects. We've been running into some heavy server load recently, so we took a look if there are any quick wins anywhere, and Jint's interop was the one taking the huge chunk of the time.

I've spent a couple of days optimizing a few things here and there, taking a rather conservative approach without changing significantly the way Jint work. So far I've got interop speed boost of 4-6 times depending on the script. Jint.Benchmark runs about 15% faster, mostly due to conversion of JsValue to ref type. All tests appear to be passing, and we've been using optimized branch in production as well...

Summary of changes:

**[Potentially Breaking?]** JsValue as reference type - removes some excess allocations related to boxing and unboxing, saves a bit of time on reducing copy-by-value penalty
JsValue.FromObject - replaces a cascade of checks with a direct type lookup. Good when you have a lot of different native .NET types which used to always hit worse case scenario of not matching any other type
FieldInfoDescriptor, PropertyInfoDescriptor - **a bit of ugly temporary solution**: fields and property accessors could technically be shared among all instances of the same .NET type. However PropertyDescriptor “owns" the final value, and has to provide it without knowing the object instance that it belongs to. Thus both of these objects used to store Target object reference, thus preventing their reuse. Removed the Target and replaced with ISharedDescriptor workaround… I might revisit this later.

Interop… this is where the biggest chunk of changes is...

DefaultTypeConverter - learns and remembers type->type conversion methods for faster lookups.
**[Potentially Breaking - Portability issues?]** FieldProxy/MethodProxy/PropertyProxy - replaces reflection based accessors with expressions emit-based generated proxies. I think this is quite portable, with exception of iOS /Xamarin (handled as a special case in proxies), but untested
MethodGroup - encapsulates overloaded method resolution
**[Potentially Breaking - see Object.hasOwnProperty note]** TypeInteropDescriptor - works as a pseudo-prototype for .NET types and caches a lot of the information that was otherwise replicated with each ObjectWrapper. Needs to be revisited so it could work as a true prototype. One of the consequences of caching member info separately from ObjectWrapper instances is that Object.hasOwnProperty doesn’t work anymore...

A few optimizations in random places that shaved off a bit of time each or removed chunks of memory allocations (e.g. s1.ToLower()[0] == s2.ToLower()[0] in ObjectWrapper.EqualsIgnoreString)

### DIFF
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -42,6 +42,30 @@ namespace Jint
 
         // cache of types used when resolving CLR type names
         internal Dictionary<string, Type> TypeCache = new Dictionary<string, Type>();
+		// cache of learned type conversions.
+		internal Dictionary<Type, Func<Engine, object, JsValue>> TypeMappers = new Dictionary<Type, Func<Engine, object, JsValue>>()
+		{
+			{ typeof(bool), (Engine engine, object v) => new JsValue((bool)v) },
+			{ typeof(byte), (Engine engine, object v) => new JsValue((byte)v) },
+			{ typeof(char), (Engine engine, object v) => new JsValue((char)v) },
+			{ typeof(DateTime), (Engine engine, object v) => engine.Date.Construct((DateTime)v) },
+			{ typeof(DateTimeOffset), (Engine engine, object v) => engine.Date.Construct((DateTimeOffset)v) },
+			{ typeof(decimal), (Engine engine, object v) => new JsValue((double)(decimal)v) },
+			{ typeof(double), (Engine engine, object v) => new JsValue((double)v) },
+			{ typeof(Int16), (Engine engine, object v) => new JsValue((Int16)v) },
+			{ typeof(Int32), (Engine engine, object v) => new JsValue((Int32)v) },
+			{ typeof(Int64), (Engine engine, object v) => new JsValue((Int64)v) },
+			{ typeof(SByte), (Engine engine, object v) => new JsValue((SByte)v) },
+			{ typeof(Single), (Engine engine, object v) => new JsValue((Single)v) },
+			{ typeof(string), (Engine engine, object v) => new JsValue((string)v) },
+			{ typeof(UInt16), (Engine engine, object v) => new JsValue((UInt16)v) },
+			{ typeof(UInt32), (Engine engine, object v) => new JsValue((UInt32)v) },
+			{ typeof(UInt64), (Engine engine, object v) => new JsValue((UInt64)v) },
+			{ typeof(JsValue), (Engine engine, object v) => (JsValue)v },			
+			{ typeof(System.Text.RegularExpressions.Regex), (Engine engine, object v) => engine.RegExp.Construct(((System.Text.RegularExpressions.Regex)v).ToString().Trim('/')) }
+		};
+		// cache of CLR type descriptors indexed by type object instance.
+		internal Dictionary<Type, TypeInteropDescriptor> TypeInteropCache = new Dictionary<Type, TypeInteropDescriptor>();
 
         internal JintCallStack CallStack = new JintCallStack();
 
@@ -515,10 +539,10 @@ namespace Jint
                     
                     if (desc.IsDataDescriptor())
                     {
-                        return desc.Value.Value;
+                        return desc.Value;
                     }
 
-                    var getter = desc.Get.Value;
+                    var getter = desc.Get;
                     if (getter == Undefined.Instance)
                     {
                         return Undefined.Instance;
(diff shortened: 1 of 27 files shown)


sebastienros:
All tests are passing on the CI, that's awesome, and there are no changes to the tests which means the breaking changes are not a huge impact and most apps will still work after the changes, or the changes will be trivial to apply. The perf improvements also show in the CI. That's some great results here, thanks a lot!

I will comment on the code now for you to make some simple changes. Once we agree on the comments then I'll merge.

Jint/Engine.cs
```diff
@@ -42,6 +42,30 @@ public class Engine
 
         // cache of types used when resolving CLR type names
         internal Dictionary<string, Type> TypeCache = new Dictionary<string, Type>();
+		// cache of learned type conversions.
+		internal Dictionary<Type, Func<Engine, object, JsValue>> TypeMappers = new Dictionary<Type, Func<Engine, object, JsValue>>()
```
sebastienros: What do you think about making it a static concurrent dictionary. That would make the list reusable across engine instances, which is ideal in a server scenario. This applies also to the other caches in this class.

Jint/Engine.cs
```diff
@@ -42,6 +42,30 @@ public class Engine
 
         // cache of types used when resolving CLR type names
         internal Dictionary<string, Type> TypeCache = new Dictionary<string, Type>();
+		// cache of learned type conversions.
```
sebastienros: Please fix the indentation. There are lots of occurrences like this where you replaced spaces with tabs.

Jint/Runtime/Interop/DefaultTypeConverter.cs
```diff
@@ -11,7 +11,7 @@ namespace Jint.Runtime.Interop
     public class DefaultTypeConverter : ITypeConverter
     {
         private readonly Engine _engine;
-        private static readonly Dictionary<string, bool> _knownConversions = new Dictionary<string, bool>();
+        private static readonly Dictionary<Tuple<Type, Type>, Func<Engine, ITypeConverter, object, IFormatProvider, object>> _knownConversions = new Dictionary<Tuple<Type, Type>, Func<Engine, ITypeConverter, object, IFormatProvider, object>>();
```
sebastienros: I don't like Tuples. Please create a custom class/struct.

Jint/Runtime/Interop/DefaultTypeConverter.cs
```diff
@@ -41,182 +41,265 @@ public virtual object Convert(object value, Type type, IFormatProvider formatPro
...
-                        }
-                        var tmpVars = new Expression[@params.Length];
-                        for (var i = 0; i < @params.Count(); i++)
+                        return (Engine engine, ITypeConverter typeConverter, object value, IFormatProvider f) =>
                         {
-                            var param = @params[i];
-                            if (param.Type.IsValueType)
+                            var genericArguments = type.GetGenericArguments();
+
+                            var @params = new ParameterExpression[genericArguments.Count()];
```
sebastienros: I don't like variable names with `@`. Please choose another name, like `parameters`, `variables`, ...

sebastienros: I realize it's not part of your changes... feel free to leave it if you feel you have done enough.

Jint/Engine.cs
sergeystoma: I'll give this a try, our rules server is already multithreaded, so it would be a good test

fatalerrorx:
JsValue struct -> class is a massive speed boost. Nearly 50% on some tests.

sebastienros:
I don't see any update, if you want I can apply the changes I asked for and still have you as the author of the commit.

sergeystoma:
@sebastienros Hey, sorry. Got overwhelmed with client work lately - I was planning to get back to changes this week, fix the recent merge conflicts and implement proposed changes.

sebastienros:
That's fine, take your time. I might try some of the changes individually too.

sebastienros:
Would it be possible to split this PR into specific ones? Like one for the JsValue as a class, and others for the improvements? I'd like to see the impact in terms of perfs of the JsValue one. It's hard to follow the related changes otherwise.

sebastienros:
Already extracted the `JsValue` changes here: https://github.com/sebastienros/jint/pull/324

sebastienros:
I have merged the changes that include the JsValue refactoring and caching the conversions.
Would be good to see your other changes in another PR though.

lahma:
Closing as relevant bits should be in and on top of that there have been numerous tweaks to improve performance. Please open an new PR if there's something still missing. Thank you!

Issue #11, opened 2014-02-10
IEnumerable should be converted to ArrayInstance


sebastienros:
This would allow not only CLR arrays to be used on Array functions, but lists also for instance.

PR #330, opened 2016-09-09
IEnumerables regression

mattbrailsford:
Re-applying the bug fix in issue #11 as the commit 5680d3b8b74d558c60ada017f33bcfad2e8bf72d reverted the fix

### DIFF
--- a/Jint/Extensions/TypeInferenceExtensions.cs
+++ b/Jint/Extensions/TypeInferenceExtensions.cs
@@ -1,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Jint.Extensions
+{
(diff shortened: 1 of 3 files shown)


sebastienros:
It fails unit tests.

Can you add your own unit test too to show what you are trying to achieve?

mattbrailsford:
Yea, sorry, noticed that. How do you run the unit tests locally? Doesn't
seem to integrate with the vs test runner.

sebastienros:
It "just works" on my machine, I don't do anything special. You can disable the EcmaScript test project to make things faster as they shouldn't be impacted.
Maybe you need an extension in VS.Net to run them, the XUnit runner.

mattbrailsford:
No probs. I'll take another look.

PR #333, opened 2016-09-19
Properly compile Lambdas when the target method is looking for a Delegate

briman0094:
If a Delegate is passed as a function to JavaScript, and that Delegate requires another Delegate (callback) as a parameter, Jint will try to pass the JS callback function to the Delegate as a LambdaExpression instead of compiling it to a Delegate, causing an Exception to be thrown. This checks to see if the target method wants a Delegate type instead of a Lambda, and compiles the Lambda to a Delegate if so.

### DIFF
--- a/Jint/Runtime/Interop/DelegateWrapper.cs
+++ b/Jint/Runtime/Interop/DelegateWrapper.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Globalization;
 using System.Linq;
+using System.Linq.Expressions;
 using Jint.Native;
 using Jint.Native.Function;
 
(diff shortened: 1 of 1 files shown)


sebastienros:
Would you mind adding a simple unit test showing the new behavior?

fatalerrorx:
This may be related to another PR 'fix #144' #309

briman0094:
@fatalerrorx thanks! #309 seems like a more elegant fix; I didn't think to look in DefaultTypeConverter for some reason

sebastienros:
Do you think we should close this PR as you prefer #309 ?

briman0094:
@sebastienros that's fine...it's probably a cleaner place to implement `.Compile()`, so we can use that one.

PR #334, opened 2016-09-20
Add function prototype to DelegateWrapper

briman0094:
Add the Engine.Function.PrototypeObject prototype to DelegateWrapper so it is populated with standard properties such as `call(...)`, `apply(...)`, and `bind(...)`. This allows for things such as currying of C# functions just as is possible with JS functions, or passing them as callbacks with the parameters already supplied.

My own tests did not reveal any breaking characteristics of this change but if necessary I can write unit tests


sebastienros:
All the tests pass, I'd just like another one at least to show the feature and ensure we don't regress. Thanks.

PR #335, opened 2016-09-28
Call object instance methods directly from an object

monoblaine:
Suppose that you've executed the following script and you want to call `obj.getFoo` passing some POCOs.

```js
var obj = {
    foo: 5,
    getFoo: function (someDotNetObject, someOtherObject) {
        var foo = this.foo;
        // do something with foo, someDotNetObject and someOtherObject
        var result = foo;

        return result;
    }
};
```

We want to call `getFoo`

```cs
var obj = Engine.GetValue("obj").AsObject();
var getFoo = obj.Get("getFoo");
```

If the type of the passed value was `int`, `double`, `string` or some `ObjectInstance`; we could've simply done this:

```cs
var arguments = new JsValue[] { 3, "bar" };
var result = getFoo.Invoke(obj, arguments);
```

But we want to pass some .NET objects so we have to do this:

```cs
var arguments = new object[] { someDotNetObject, someOtherObject };
var result = getFoo.Invoke(obj, arguments.Select(x => JsValue.FromObject(Engine, x)).ToArray());
```

It would be much better if we could call `getFoo` directly from `obj`:

```cs
var result = obj.CallMethod("getFoo", arguments);
```

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -852,6 +852,32 @@ namespace Jint.Tests.Runtime
             Assert.Throws<ArgumentException>(() => x.Invoke(1, 2));
         }
 
+        [Fact]
+        public void ShouldInvokeAFunctionValueThatBelongsToAnObject()
+        {
+            RunTest(@"
+                var obj = { foo: 5, getFoo: function (bar) { return 'foo is ' + this.foo + ', bar is ' + bar; } };
+            ");
+
+            var obj = _engine.GetValue("obj").AsObject();
+            var getFoo = obj.Get("getFoo");
+
+            Assert.Equal("foo is 5, bar is 7", _engine.Invoke(getFoo, obj, new object[] { 7 }).AsString());
+        }
+
+        [Fact]
+        public void ShouldNotInvokeNonFunctionValueThatBelongsToAnObject()
+        {
+            RunTest(@"
+                var obj = { foo: 2 };
+            ");
+
+            var obj = _engine.GetValue("obj").AsObject();
+            var foo = obj.Get("foo");
+
+            Assert.Throws<ArgumentException>(() => _engine.Invoke(foo, obj, new object[] { }));
+        }
+
         [Theory]
         [InlineData("0", 0, 16)]
         [InlineData("1", 1, 16)]
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -877,25 +877,25 @@ namespace Jint.Tests.Runtime
             ");
         }
 
-		[Fact]
-		public void JsonParserShouldUseToString()
-		{
-			RunTest(@"
+        [Fact]
+        public void JsonParserShouldUseToString()
+        {
+            RunTest(@"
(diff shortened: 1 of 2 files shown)


sebastienros:
Why not add an `Invoke` overload that take objects and converts them ?

monoblaine:
Do you mean an `Invoke` overload within the `JsValue` class? If so, yes that's also what I thought of first. However, a `JsValue` object does not have any reference to the related `Engine` instance while the `ObjectInstance` object does.

We can still create an `Invoke` overload and pass the related `Engine` instance as an argument, but it seems like an ugly hack.

sebastienros:
No, not what I meant. And actually what I meant already exists, look there: https://github.com/sebastienros/jint/blob/master/Jint/Engine.cs#L672

sebastienros:
Which means it could also be extended to accept a `JsValue` on top of the property name. Just in case you want to invoke the result of a function.

monoblaine:
> No, not what I meant. And actually what I meant already exists, look there: https://github.com/sebastienros/jint/blob/master/Jint/Engine.cs#L672

Yes, I'm well aware of that method. You might have noticed that [I've actually used a part from that method in my proposed method](https://github.com/sebastienros/jint/pull/335/commits/d99a79d72ba0b62ab93b325a238383f05a89610f#diff-d6e42f1248b104068929cf23dbc58dddR51).

But that method alone cannot be used to call a js function defined somewhere other than the global scope.

> Which means it could also be extended to accept a JsValue on top of the property name. Just in case you want to invoke the result of a function.

So you mean we should create another `Invoke` overload in `Engine.cs`. (which solves the problem I've just mentioned.)

Would something like this be OK?

```cs
/// <summary>
/// Invoke the current value as function.
/// </summary>
/// <param name="propertyName">The arguments of the function call.</param>
/// <returns>The value returned by the function call.</returns>
public JsValue Invoke(string propertyName, params object[] arguments)
{
    return Invoke(propertyName, null, arguments);
}

/// <summary>
/// Invoke the current value as function.
/// </summary>
/// <param name="propertyName">The name of the function to call.</param>
/// <param name="thisObj">The this value inside the function call.</param>
/// <param name="arguments">The arguments of the function call.</param>
/// <returns>The value returned by the function call.</returns>
public JsValue Invoke(string propertyName, object thisObj, object[] arguments)
{
    var value = GetValue(propertyName);

    return Invoke(value, thisObj, arguments);
}

/// <summary>
/// Invoke the current value as function.
/// </summary>
/// <param name="value">The function to call.</param>
/// <param name="thisObj">The this value inside the function call.</param>
/// <param name="arguments">The arguments of the function call.</param>
/// <returns>The value returned by the function call.</returns>
public JsValue Invoke(JsValue value, object thisObj, object[] arguments)
{
    var callable = value.TryCast<ICallable>();

    if (callable == null)
    {
        throw new ArgumentException("Can only invoke functions");
    }

    return callable.Call(JsValue.FromObject(this, thisObj), arguments.Select(x => JsValue.FromObject(this, x)).ToArray());
}
```

sebastienros:
Yes, exactly.

monoblaine:
I've pushed a new commit. The pr is no more polluting the `ObjectInstance` class.

Issue #270, opened 2016-02-24
dotnet core compatibility


ciel:
Is there any plan to update the references for JINT to be compatible with the new .NET Core system?

sebastienros:
It's there, use the `dnx` branch. At some point I will make a single project targeting all platforms, but I am waiting for the platform support story to be final.

ciel:
Thank you so much for making this wonderful library. I'm so excited to see it is still being supported and worked on.

jonjo-manywho:
We'd love to use Jint in a .NET Core project! Are there any updates to this now that it's officially released?

ciel:
Yeah, I'm also looking forward to it.

ciel:
Is it at least possible to use Jint with a `net451` import?

sebastienros:
If you can't wait RavenDb has ported it already to Core. It would have been better they contribute to this repository directly, but they might have other constraints.

ciel:
How can we help?

sebastienros:
Targeting a specific branch will let me accept PRs from different committers even if there are issues.

darkcamper:
Hi. I've just finished porting Jint to NetStandard 1.0 and it's currntly passing all tests except for  CommonScripts.ThreeDRayTrace (and I don't really know what's happening).

The bulk of the work has been replacing many changed or non-existent methods of System.Reflection

It's in pull request #337. This is my secound github contribution and I'm not reallly used to git and pull requests and stuff yet, so if I've made any mistake just tell me and I'll try to fix it.

sebastienros:
Done, and it's on Nuget. Targets PCL, Net Standard 1.3 and .NET 45. Thanks @darkcamper for your help.

darkcamper:
I don't think my contribution has been really useful 😅 . Thanks for sorting out everything and releasing it so quickly! (I'll make a good use of it)

sebastienros:
It's called motivation.

PR #337, opened 2016-10-02
NetStandard 1.0 port

darkcamper:
Jint ported to NetStandard 1.0 for issue #270


sebastienros:
I merged it too quickly, but that's fine as it's on a specific branch.

First I can't compile if I just open the solution.
Then I am adding some comments on the changes.

Jint/Native/JsValue.cs
```diff
@@ -279,60 +279,44 @@ public Types Type
...
+
                 if (converter.TryConvert(value, out result))
                 {
                     return result;
                 }
             }
 
-            var typeCode = System.Type.GetTypeCode(value.GetType());
-            switch (typeCode)
+            Dictionary<Type, Func<object, JsValue>> jsValueConverters = new Dictionary<Type, Func<object, JsValue>>()
```
sebastienros: Just use Convert.ToTypeCode to get the type code. Get the switch back then.

darkcamper: Have you been able to compile yet? If not, you might be missing [DotNetCore tools for VS 2015 ](https://download.microsoft.com/download/5/5/7/557D02A5-C3D0-4EF6-A570-4F75CD0DA5BF/DotNetCore.1.0.1-VS2015Tools.Preview2.0.2.exe)

About Convert.ToTypeCode, I've found no information about it on MSDN :( . I had to replace Type.GetTypeCode with a dictionary because I couldn't find a way with NetStandard 1.0 to get the type code from a type (although I've just noticed that I forgot to refactor the local dictionary to a static field as I intended)

Anyway if you have some method to recover the TypeCode that works with NetStandard 1.0 I'll revert the change and get rid of that ugly dictionary

sebastienros:
Look at this framework I made on dotnet code, it's using `Convert.GetTypeCode`
https://github.com/sebastienros/yessql/blob/4e9f0de89a2419ed0bd4c41c2d1779dbe917d840/src/YesSql.Core/Sql/ISqlDialect.cs#L157

No I am not missing the tooling for VS, actually I work in the dotnet core team (ASP.NET) 😃 

I think I will create a specific `project.json` files to drive the different targets. Like what Json.NET is doing.

darkcamper:
Oops! I didn't know you were working in the dotnet core team 😅 

Anyway, I might still be mistaken but I've made some tests and Convert.GetTypeCode only works from Netstandard 1.5 onwards (it was introduced in System.Reflection.TypeExtensions 4.1.0.0 contract, it seems).

I've been doing some work to add support for PCL Profile 328 (Net40), so old projects could still use the library after its porting to netstandard is complete. But if you are going to change the projects structure I believe it's going to be better to wait for your changes so there are less conflicts. What do you think?

By the way, are you still unable to compile the solution? I've re-downloaded the netcore branch from your repository to look for problems and I've been able to compile it without errors 😕

sebastienros:
I have made some progress on the multi-platform target, not yet done, now applying your changes related to reflection.

sebastienros:
I pushed my changes, it's overriding yours. Something that is not working so far is building the other satellite projects which target only 4.5. I will investigate this next.

sebastienros:
I also used this package https://www.nuget.org/packages/System.Reflection.TypeExtensions/ that is a set of poly-fills for reflection and solve most of the ones you had to make.

FlorianRappl:
Quick question on this one (maybe invalid): When do you think a multi-platform (i.e., including .NET core) release (available via NuGet) of Jint can be made?

I know there is `Jint-netcore`, but I would love to have the official one from you.

BTW: Kudos to your efforts on this project. Jint is a masterpiece! 👍

darkcamper:
So, are you going to set NETStandard 1.3 as the minimum for Jint?

sebastienros:
@darkcamper right now this is the easiest solution. Do you think that could be an issue? The minimum dotnet framework would be 4.5. But it would still be compatible with portable targets.

@FlorianRappl As soon as we have resolved the current build issues I will release it. My dear colleague and packaging master Pranav should help me by the latest tomorrow.

darkcamper:
Mmm no, I don't think that'd be an issue. As you've said, old projects targeting 4.0 could still use the PCL target (and targeting 1.3 avoids many nasty headaches)

sebastienros:
Everything should be good now on the netcore branch. A single unit test is not passing on netstandard. It's related to reflection.

svaurea:
where is the .net standard compatible jint version published ?

Tirraon:
> where is the .net standard compatible jint version published ?

There is 'net40;net45;net451;netstandard1.3;netstandard2.0' version on nuget.
Jint3.0 is in preview and support only 'net45;netstandard2.0'.

PR #339, opened 2016-10-16
Sign an Jint version 2.10 with a Strong Name

Taritsyn:
`Jint.dll` in the Jint version 2.10 package not signed with strong name.


sebastienros:
Thanks, I'll update the package later.

PR #345, opened 2016-10-31
fixed issue where the line number for exceptions in nested function …

andreykorolev:
…wasn't correct


andreykorolev:
This fix is the same as #278 but without changes to line feeds and spaces

andreykorolev:
When it will be available in nuget?

andreykorolev:
could you please publish it to nuget?

sebastienros:
Sorry for the delay, I just pushed a new version.

I will definitely try to find a way to integrate it with a custom myget feed for CI builds.

Issue #91, opened 2014-09-29
Suggestion: special syntax to call .NET indexers


h15ter:
Current .NET indexers support in JINT is a bit messy. 
1) It doesn't support multi-argument indexers
2) It's not strongly-typed (non-deterministic)
- let's say there are this[DateTime date], this[string key] and this[int index]
- there is javascript, clrObj['123'] = 'whatever'
- which indexer will be invoked?
  3) There is a quite high risk that .NET indexer will mess up with some javascript object's own property

I our case we had to write proxies/wrappers with no indexers around all our .NET types being exposed to javascript, because we don't feel the current indexer implementation is safe. Looks like some special syntax to access .NET syntax would be more reliable rather than 'clrObj[index]', maybe something like clrObj.getItem(param1, param2...) and clrObj.setItem(param1, param2...)

merakod:
This has actually been one of our biggest issues after upgrading from Jint 0.9.2
We might try to fix this..

sebastienros:
Can you refresh my mind and explain how it was behaving in Jint 0.9 ?

lahma:
Closing this as the resolution is quite smart in 3.x series and one can always expose extra object methods to do the heavy lifting in CLR code.

PR #349, opened 2016-11-11
handle the common string to int conversion 

AllNamesRTaken:
that otherwise removes the possibility to have overloaded indexers with both int and string

### DIFF
--- a/Jint/Runtime/Interop/DefaultTypeConverter.cs
+++ b/Jint/Runtime/Interop/DefaultTypeConverter.cs
@@ -222,8 +222,23 @@ namespace Jint.Runtime.Interop
 
             if (canConvert)
             {
-                converted = Convert(value, type, formatProvider);
-                return true;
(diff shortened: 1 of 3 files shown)


merakod:
Btw, this is related to https://github.com/sebastienros/jint/issues/91

Jint/Runtime/Interop/DefaultTypeConverter.cs
```diff
@@ -222,8 +222,23 @@ public virtual bool TryConvert(object value, Type type, IFormatProvider formatPr
 
             if (canConvert)
             {
-                converted = Convert(value, type, formatProvider);
-                return true;
+                //handle the common string to int conversion that otherwise removes the possibility to have overloaded indexers with both int and string 
+                var isStringValue = value is string;
```
sebastienros: Couldn't we just use Convert.ChangeType() when we detect types are different? This would then also support other integral types

sebastienros:
I think this is fixed by https://github.com/sebastienros/jint/pull/390

PR #354, opened 2016-11-20
Add Chrome debug protocol implementation

miz-online:
Hello,
I really like the Jint project because it enables me to include a javascript engine into my own projects.
What I wanted to achieve is, that I wanted to debug the code in an easy way.
Because of that, I implemented a chrome debugging protocol compliant debug websocket server, that allows to use the chrome dev tools on Jint engines.
To enable that, I had to slightly change the debugger implementation, mostly to get more information, hook into additional events (parsing scripts, exceptions) and enable expression execution without debugging (to allow watches while breaking in an breakpoint).
For details please see the modified readme.md (on the bottom)
What is not done so far are unit tests. I want to do them as well, but wanted to give you the option to give your comments to this addition.
It would really be great if you would like to add this implementation to the 'official' Jint package.
best regards and keen about responses or questions
Michael
PS: one unit test fails on my PC, with the original master code as well as with my changes (3D raytrace test)

### DIFF
--- a/Jint.DebugAgent/ChromeDebugProtocolServer.cs
+++ b/Jint.DebugAgent/ChromeDebugProtocolServer.cs
@@ -1,0 +1,203 @@
+﻿using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Net;
+using System.Net.WebSockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Jint.DebugAgent.Domains;
+using Newtonsoft.Json;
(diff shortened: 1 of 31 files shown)


sebastienros:
Here is a project working on re-implementing it with the recent improvements made to the debugger support

https://github.com/Jither/Jint.DevToolsProtocol

lahma:
Thanks for building the foundation @miz-online , we'll try to make sure you are credited for the original work! Sorry for taking so long to process, hopefully the end result will benefit you too.

PR #363, opened 2017-02-06
try to update jint

gyrodrill:
(it is not neccessary to update it, because the jint 1.0.dll in jist could satisfy my need:)


sebastienros:
If it's fixing an issue, could you add a unit test that demonstrates it?

gyrodrill:
Sorry for bordering, it is just a wrong operation.

PR #366, opened 2017-02-16
CLR instances should compare on their inner value

nielsbrinch2:
Case: When comparing two Guids, they should be considered equal if there are indeed equal. However, they are being treated as two separate objects (which they are) and so they are not considered equal.

The issue with this is the developer would expect them to be equal. The solution proposed is to check if they are CLR instances and then check on their inner value.

This commit includes a unit test that covers the test scenario. The changes in ExpressionInterpreter.Equal do not solve this problem.

Attempts at solving the problem (such as calling ToString) has resulted in multiple EcmaTests to fail. I need help.

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -1863,5 +1863,22 @@ namespace Jint.Tests.Runtime
 				}
             ");
 		}
+
+        [Fact]
+        public void ShouldCompareInnerValueOfClrInstances()
+        {
+            var engine = new Engine();
+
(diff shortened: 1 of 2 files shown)


sebastienros:
Thanks for the effort, I adapted your code and created this other PR which should fix the issue. I will merge once the unit tests have passed on the CI.

https://github.com/sebastienros/jint/pull/367

nielsbrinch2:
Thanks a lot. I reviewed your code and I see a lot of decisions that I am nowhere near to being able to have made, so I think you making it instead was a very good decision.

Issue #362, opened 2017-01-25
JSON.stringify breaks floats


mazhuravlev:
`engine.Execute(@"var o = {z: 99.20 * 0.23, x: "dfdf"}; return JSON.stringify(o);");`
returns 
`{"z":22.,816000000000003,"x":"dfdf"}`
There is a unneeded extra dot before the comma.

sebastienros:
So, was it a non-issue?

mazhuravlev:
Duplicate of [Number toFixed() incorrectly rounds fractions and negative numbers #158](https://github.com/sebastienros/jint/issues/158)
Still doesn't work in version 2.10.3

sebastienros:
thanks, will fix

RhinoDevel:
I guess this is related - I reproduced this behavior with v2.10.3 and 2.9.1:

    Jint.Engine e = new Jint.Engine();
    const string n = "n";

    //e.SetValue(n, "{\"a\":53.6841659}"); // => Leads to same problem.
    e.SetValue(n, "[53.6841659]");

    Jint.Native.JsValue val = e.Execute("var o = JSON.parse(n);JSON.stringify(o);").GetCompletionValue();

    Console.WriteLine(val.AsString()); // => Output: "[53.,6841659]"

RhinoDevel:
Pull request here: https://github.com/sebastienros/jint/pull/369

PR #369, opened 2017-03-16
Fix bug in number to string conversion.

RhinoDevel:
If V8 FastDtoa cannot convert a number, the "old" DToA implementation can fail, if the system Jint is running on does not use "." (dot) as separator between integer and fractional part of the number.

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -1880,5 +1880,16 @@ namespace Jint.Tests.Runtime
 
 			Assert.True(result);
         }
+	
+        [Fact]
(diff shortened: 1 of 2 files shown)


RhinoDevel:
This fixes the following issue: https://github.com/sebastienros/jint/issues/362

sebastienros:
Thanks, would you mind adding a test that would fail without this fix?
For instance in EngineTests.cs, you'll find many example in this class.

RhinoDevel:
@sebastienros : Is there a release planned in the near future? I am asking, because this bugfix is necessary for my project and I would prefer to update to an official jint version via NuGet.. :-) Nevermind, if currently not possible.

sebastienros:
Done

PR #374, opened 2017-04-21
Catching exceptions optionally

sebastienros:
Some users want more options to catch exceptions.
With this PR you need to call `options.CatchClrExceptions()` to trigger the conversion from CLR exceptions to JS errors.
You can also optionally pass a predicate to this method to decide when to catch them or not dynamically, like for instance to filter specific exception types.

/cc @mgentile @wslthree_twitter @danbopes


danbopes:
I like this :thumbsup:. Solves the best of both worlds.

ringls:
can you publish to nuget?thank you.

PR #375, opened 2017-04-24
Fixed concurrency issue with introducing new TypeMapper to Engine.Typ…

maximburyak:
…eMappers

### DIFF
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using Jint.Native;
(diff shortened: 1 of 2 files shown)


sebastienros:
Thanks for fixing this, such an obvious issue I always warn people against .... shame on me.

Right now it doesn't compile
Also can you explain why the change with the `KeyValuePair`?

About performance now, assuming CDs are slower that Ds, and knowing that reads are much more frequent than writes, what about using an `ImmutableDictionary` ? This way we only pay a price on adds, and never on reads, while keeping it as fast as the standard dictionary.

sebastienros:
Forget about it, it would require referencing other nuget packages. just fix the compilation issues.

maximburyak:
Looking more closely at the compilation output, I see that the System.Collections.Concurrent namespace is not supported in the portable library, closing this PR and will send a new one shortly

PR #381, opened 2017-05-01
Invoke ExtensionMethods

windischb:
I have added code to invoke extension methods

### DIFF
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -1,6 +1,9 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
 using Jint.Native;
 using Jint.Native.Argument;
 using Jint.Native.Array;
@@ -65,6 +68,8 @@ namespace Jint
         };
 
         internal JintCallStack CallStack = new JintCallStack();
+
+        internal Dictionary<Type, List<MethodInfo>> ExtensionMethods = new Dictionary<Type, List<MethodInfo>>();
 
         public Engine() : this(null)
         {
@@ -162,6 +167,39 @@ namespace Jint
             ClrTypeConverter = new DefaultTypeConverter(this);
             BreakPoints = new List<BreakPoint>();
             DebugHandler = new DebugHandler(this);
+
+            if (Options._ExtensionMethodsTypes.Any())
+            {
+                foreach (var extensionMethodType in Options._ExtensionMethodsTypes)
+                {
+                    foreach (var methodInfo in extensionMethodType.GetExtensionMethods())
+                    {
+                        var firstParamType = methodInfo.GetParameters().First().ParameterType;
+
+                        if (firstParamType == typeof(String))
+                        {
+                            String.PrototypeObject.FastAddProperty(methodInfo.Name, new MethodInfoFunctionInstance(this, new MethodInfo[] { methodInfo }), true, false, true);
+                            continue;
+                        }
+
+                        if (firstParamType.IsArray || firstParamType.GetInterfaces().Contains(typeof(IEnumerable)))
+                        {
+                            Array.PrototypeObject.FastAddProperty(methodInfo.Name, new MethodInfoFunctionInstance(this, new MethodInfo[] { methodInfo }), true, false, true);
+                            continue;
+                        }
+
+                            
+                        if (ExtensionMethods.ContainsKey(firstParamType))
+                        {
+                            ExtensionMethods[firstParamType].Add(methodInfo);
+                        }
+                        else
+                        {
+                            ExtensionMethods.Add(firstParamType, new List<MethodInfo>() { methodInfo });
+                        }
+                    }
+                }
+            }
         }
 
         public LexicalEnvironment GlobalEnvironment;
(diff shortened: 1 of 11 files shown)


Jint/ReflectionExtensions.cs
```diff
@@ -60,6 +83,22 @@ internal static MethodInfo GetMethodInfo(this Delegate d)
...
+            return type.GetMethods(BindingFlags.Public | BindingFlags.Static)
+                .Where(m => m.IsExtensionMethod());
+        }
+
+        internal static Boolean IsExtensionMethod(this MethodInfo methodInfo)
+        {
+            return methodInfo.IsDefined(typeof(ExtensionAttribute), true);
+        }
+
+        internal static bool IsIEnumerable(this Type type)
```
sebastienros: Not used, can be removed.

Jint/ReflectionExtensions.cs
```diff
@@ -26,11 +29,31 @@ internal static bool IsValueType(this Type type)
...
```
sebastienros: Not used, can be removed.

Jint/Engine.cs
```diff
@@ -162,6 +167,39 @@ public Engine(Action<Options> options)
             ClrTypeConverter = new DefaultTypeConverter(this);
             BreakPoints = new List<BreakPoint>();
             DebugHandler = new DebugHandler(this);
+
+            if (Options._ExtensionMethodsTypes.Any())
```
sebastienros: Can you move it to a private method?

Also it's quite some expensive code to run on each engine instance. At some point we might want to cache the dictionary statically (and keeping it thread-safe).

windischb: Concurrent Dictionary is not available in PCL...
Do you think it's necessary to use a lock for keeping the Dictionary thread-safe?
Or is it fine with just a static Dictionary?

Jint/Engine.cs
```diff
@@ -162,6 +166,54 @@ public Engine(Action<Options> options)
...
+        {
+            if (ExtensionMethodTypeCache.Any())
+            {
+                foreach (var cachedType in ExtensionMethodTypeCache)
+                {
+                    if (cachedType.Key == typeof(String))
+                    {
+                        foreach (var methodInfo in cachedType.Value)
+                        {
+                            String.PrototypeObject.FastAddProperty(methodInfo.Name, new MethodInfoFunctionInstance(this, new MethodInfo[] { methodInfo }), true, false, true);
```
sebastienros: Probably not. Don't change the string prototype object. This is different from just supporting extension methods.

Jint/Engine.cs
```diff
@@ -162,6 +166,54 @@ public Engine(Action<Options> options)
             ClrTypeConverter = new DefaultTypeConverter(this);
             BreakPoints = new List<BreakPoint>();
             DebugHandler = new DebugHandler(this);
+
+            BuildExtensionMethodTypeCache();
+            RegisterCachedExtensionMethods();
+        }
+
+        internal Dictionary<Type, List<MethodInfo>> ExtensionMethodTypeCache = new Dictionary<Type, List<MethodInfo>>();
```
sebastienros: The solution is to have two MethodInfo dictionaries. One that is static and reused across engine instances (a cache) and one that contains the available methods from this engine instances.
So you build the second one by looking if the method infos are already available in the cache, then copy a reference to your engine local dictionary. If not you add it in both.
And to preserve thread-safety you `lock` the code that reads and writes to the cache dictionary.

Jint/Engine.cs
```diff
@@ -162,6 +166,54 @@ public Engine(Action<Options> options)
...
+                                ExtensionMethodTypeCache.Add(firstParamType, new List<MethodInfo>() { methodInfo });
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        private void RegisterCachedExtensionMethods()
+        {
+            if (ExtensionMethodTypeCache.Any())
```
sebastienros: unnecessary call to `.Any()` as the foreach will do it for you.

Jint/Engine.cs
```diff
@@ -40,7 +43,7 @@ public class Engine
         public ITypeConverter ClrTypeConverter;
 
         // cache of types used when resolving CLR type names
-        internal Dictionary<string, Type> TypeCache = new Dictionary<string, Type>();
+        internal static readonly Dictionary<string, Type> TypeCache = new Dictionary<string, Type>();
```
sebastienros: Mistake? It has nothing to do with this PR right?

windischb:
I have tried to implement the Dictionaries and to preserve thread-safety...
It ended up with 3 Dictionaries now.

1. static Dictionary<Type, List<MethodInfo>> for caching classes which have extension methods and the MethodInfos from this class.

2. static Dictionary<Type, List<MethodInfo>> for caching types and the available MethodInfos for this type.
For example all extension methods for class Person

3. instance Dictionary<Type, List<MethodInfo>> which contains all required Methodinfos from 2. Dictionary

I'm not sure if this is what you want...

And another thing is that i can't use extension methods on Strings, if i do not change the string prototype.
If i run the Test which i have build for StringExtensions, i always get the message: Object has no method 'Reverse'

pseabury:
@sebastienros - Any guidance on @windischb last comments?  I think with extension methods I could potentially switch from Roslyn (too heavy, buggy) to jint.  This one looks close to merge-able.

boennhoff:
I am also interested in this addition into upstream. How can I take over? Should I just create my own branch and cherry-pick those commits?

boennhoff:
I took over here: PR #616

PR #385, opened 2017-05-27
Implemented unescape function on global object.

girishjjain:
unescape function is deprecated but supported by browsers and hence, would be good to have Jint support the same.


sebastienros:
The build is broken, hard to tell why ...

sebastienros:
Not sure it makes sense to have all these cases in the Enginetest. There is a folder with all conformance tests, they are just js files with a C# test file to point to them. Maybe you could do the same here. Just adapting the method `sameValue` to its local equivalent.

girishjjain:
Found the failure reason, string class in netstandard1.0 doesn't implement IEnumerable<char> interface. I've made a change to convert the string to char array using ToCharArray() method. That takes care of compilation error. 

On a side note, I didn't notice this error in my local development because I am using VS 2017 and the first time I opened project, VS complained about not being able to understand target frameworks so I manually edited it and removed portable frameworks. 

About tests, I will take a look at the approach you described. May need some help with that.

girishjjain:
@sebastienros - I tried adding a class for unescape tests with methods pointing to all the test javascript files available here: https://github.com/tc39/test262/blob/master/test/annexB/built-ins/unescape

I get error - ReferenceError: assert is not defined

These tests are not defined in the same manner as tests already out there. Some of the other tests also depend on another javascript file called propertyHelper.js. Looks to be a rabbit hole with other nested dependencies on more javascript files defined in test/harness folder as part of test262 repo. 

I don't want to rewrite those javascript tests to align with test approach followed by Jint project. I know its not pretty to have all the tests in code as so many attributes but that would be better than not having tests. I can fix the build error. What do you suggest?

girishjjain:
Hi @sebastienros - Let me know what you think about leaving the unit tests in as xunit attributes. As I said, it's not pretty but at least there will be code coverage. Thanks.

girishjjain:
Hi @sebastienros - Gentle reminder! Waiting for your input.

PR #386, opened 2017-05-28
Implemented unescape function on global object.

girishjjain:
unescape function is deprecated but supported by browsers and hence, would be good to have Jint support the same.


girishjjain:
Just noticed comment on previous pull request.

PR #390, opened 2017-06-14
Fixes Multiple Indexer and DefaultTypeConverter TryConvert

albyrock87:
Hello,

Jint crashes if used in combination with a mixed type indexed class.

The bug is inside DefaultTypeConverter: it should not crash inside the "TryConvert" method.

Thank you


albyrock87:
Thanks :)

PR #399, opened 2017-07-06
Support local date time on native clr

albyrock87:
Hi,

when working with DateTime, we may expect in output a DateTime with DateTimeKind.Local.

Actually the Engine always returns a DateTimeKind.Utc.

I added an option to change the behaviour.

Thanks

### DIFF
--- a/Jint/Options.cs
+++ b/Jint/Options.cs
@@ -14,6 +14,7 @@ namespace Jint
         private bool _allowDebuggerStatement;
         private bool _debugMode;
         private bool _allowClr;
+        private bool _clrSetLocalDateTimeKind;
         private readonly List<IObjectConverter> _objectConverters = new List<IObjectConverter>();
         private int _maxStatements;
         private int _maxRecursionDepth = -1; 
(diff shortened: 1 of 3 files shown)


Jint/Options.cs
```diff
@@ -43,6 +44,17 @@ public Options Strict(bool strict = true)
         }
 
         /// <summary>
+        /// Uses DateTimeKind.Local when updating native value
+        /// </summary>
+        /// <param name="clrSetLocalDateTimeKind"></param>
+        /// <returns></returns>
+        public Options ClrSetLocalDateTimeKind(bool clrSetLocalDateTimeKind = true)
```
sebastienros: Can we change it to `UseLocalTimeZone(bool useLocalTimeZone = true)` ?

sebastienros: Hum, then it wouldn't be obvious that's for DateTime.

`UseLocalDateTime(bool useLocalDateTime = true)`

sebastienros:
Fixed

PR #408, opened 2017-08-25
Better error from scripts.

ayende:
- Will track the identifier that cause an undefined / null access
- Will report the call stack of the error to the user


sebastienros:
Looks good. Could not find any breaking change either.

Issue #409, opened 2017-08-25
Null propagation?


ayende:
I'm 99% sure that this is going to be rejected, but our usage of Jint requires us to allow users to do things like:

```
return a.foo.bar.baz;
```

Where any of the above may be missing. 

The actual change is tiny:
https://github.com/ravendb/jint/commit/0470590a43d28dd5813b4fc1361388f219b0d02f

But it allows much nicer integration, and obviously this is protected by a feature flag.

sebastienros:
This is something we use a lot in Orchard actually, using dynamic objects. We call it the Nil behavior.

But you are right this is not acceptable as is. Can we find a solution where we could change some logic by configuration, and you could plug your own? Like when the property is resolved? I will take a look too.

sebastienros:
The idea is that I don't want to have to support unsupported behaviors ... and looking at your changes they are pretty limited. Can you find a solution when this piece of code you changed in Engine.cs can be replaced based on configuration, like externalizing from Engine? The default implementation will be the same, then you can plug yours easily.

Something likely called `IReferenceResolver`.

PR #410, opened 2017-08-25
Allow host to handle unknown references

ayende:
See #409 

This handle null reference errors as well as methods calls.

### DIFF
--- a/Jint.Tests/Runtime/NullPropogation.cs
+++ b/Jint.Tests/Runtime/NullPropogation.cs
@@ -23,7 +23,7 @@ namespace Jint.Tests.Runtime
                 return value.IsNull() || value.IsUndefined();
             }
 
-            public bool TryGetCallable(Engine engine, Reference reference, out JsValue value)
+            public bool TryGetCallable(Engine engine, object reference, out JsValue value)
             {
                 value = new JsValue(
                     new ClrFunctionInstance(engine, (thisObj, values) => thisObj)
--- a/Jint.Tests/Runtime/NullPropogation.cs
+++ b/Jint.Tests/Runtime/NullPropogation.cs
@@ -1,0 +1,128 @@
+﻿using Jint.Native;
+using Jint.Native.Object;
+using Jint.Parser;
+using Jint.Runtime;
+using Jint.Runtime.Interop;
+using Jint.Runtime.References;
+using Xunit;
+
+namespace Jint.Tests.Runtime
+{
+    public class NullPropogation
+    {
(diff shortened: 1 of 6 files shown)


Jint/Runtime/ExpressionIntepreter.cs
```diff
@@ -851,7 +852,17 @@ public JsValue EvaluateCallExpression(CallExpression callExpression)
 
             if (!func.IsObject())
             {
-                throw new JavaScriptException(_engine.TypeError, r == null ? "" : string.Format("Property '{0}' of object is not a function", (callee as Reference).GetReferencedName()));
+                var reference = (Reference) callee;
```
sebastienros: This fails some tests with `System.InvalidCastException: Unable to cast object of type 'Jint.Native.JsValue' to type 'Jint.Runtime.References.Reference'.`

sebastienros: This might have been hidden in the previous code because it was already throwing an exception. I assume the correct behavior is to throw a TypeError if it's not a reference.

ayende: Yes, I assumed that this is correct because it would NRE otherwise. Will fix in a bit.

ayende: Nope, there was no NRE, there is a check for `r == null` there.

sebastienros:
FYI in ES6 this kind of pluggable behavior is part of the language, so you can dynamically change it.

sebastienros:
The dev branch is available on myget here https://www.myget.org/feed/Packages/jint
Published on nuget.org once merged in master, not special condition to do it, just based on time and velocity of changes.

ayende:
Thanks

Jint.Tests/Runtime/NullPropogation.cs
```diff
@@ -0,0 +1,128 @@
...
+using Jint.Native.Object;
+using Jint.Parser;
+using Jint.Runtime;
+using Jint.Runtime.Interop;
+using Jint.Runtime.References;
+using Xunit;
+
+namespace Jint.Tests.Runtime
+{
+    public class NullPropogation
```
jokin: Could this typo be fixed to NullPropagation ?

Jint.Tests/Runtime/NullPropogation.cs
```diff
@@ -0,0 +1,128 @@
...
+    {
+        public class NullPropgationReferenceResolver : IReferenceResolver
```
jokin: And this one to NullPropagationReferenceResolver

PR #412, opened 2017-08-27
Updating ES6 to current code

ayende:
This is for review only, there are still a few tests that fail that I haven't gotten to.
This work includes mostly.
- rebasing of dev into es6, so we'll have all the latest changes.
- updating the project to .NET Standard 2.0
- fixing some minor todos regarding caching of arguments and literals with Esprima


sebastienros:
Let's talk a little bit before. The es6 branch has been stale for a while, so it might be easier to just reapply the changes from master. I can think of the `Symbol` support and the new unit tests organization.

Also, what do you think about changing the strategy to make the es6 branch the main one, and add es6 features as they are implemented. As es6 is a superset of es5.1 it should not be an issue and it would gather more contributions. 

We can also keep two main branches, one with 5.1 only features, and one with ES6, and any bug fix or improvement would need to be done on ES5.1, and new ES6 feature support would be done on the ES6.

My preference goes to having only one main branch with incremental new features (Symbols, arrow functions, ...). And create dev branches if we need to work on different ES6 features in parallel.

We can chat on skype if you need: sebastienros

ayende:
Closing this PR and opening a new one against dev, which is what I actually rebased this against.
Added you on Skype.
I think that parallel branches in this manner would work well, yes.

PR #413, opened 2017-08-28
Arrow functions, es6 merged to dev

ayende:
This PR does:

- Rebased ES6 branch into dev, so all the work done in both branches is preserved.
- Update projects to .NET standard 2.0
- Initial implementation for for arrow functions.

Note on this, this is probably still pretty raw. In particular, I haven't handled `HoistingScope` properly for arrow functions and `arguments` handling is probably missing as well. 
Next stage is to try to run this through https://github.com/tc39/test262/tree/master/test/language/expressions/arrow-function and see how much stuff this breaks :-)


ayende:
The PR will probably fail to build because it needs a more up to date Esprima package, I added a PR for that as well.

sebastienros:
I have integrated this work in the dev branch, but not the arrow function yet. Getting all the things in place (CI, tests) before working mostly on es6 features.

Issue #420, opened 2017-09-28
getHours on Date object returns wrong on date older than 1970


binish:
`new Date(1969,0,1,19,45).getHours()` returns -5 instead of 19

PR #431, opened 2017-11-01
Update HourFromTime to handle pre 1/1/1970 #420

WestDiscGolf:
Update to how the calc for milliseconds is done. Went with DateTime processing as DateTimeOffset from milliseconds method is not available in all targetted platforms.

@sebastienros I was unsure where to put the unit test, please advise :-)


sebastienros:
Please put the new tests beside the existing ones for dates, here https://github.com/sebastienros/jint/blob/dev/Jint.Tests/Runtime/EngineTests.cs#L1831

And look at the current CI job that failed because of regressions.

Thanks!

WestDiscGolf:
Investigating about the test failures now (thought I'd run them all locally first *sigh*)

WestDiscGolf:
Tests moved and updated for all tests to pass :)

sebastienros:
Looking at your changes I realized the issues had to happen also on minutes and seconds. So I looked into the root cause and found the explanation. I created another PR.

sebastienros:
Thanks for your work though.

WestDiscGolf:
Np! Glad it was of some help even if it didn't make the cut ;-)

Thanks for the library!

Issue #429, opened 2017-10-22
Missing support for String.prototype.padStart() and padEnd()


aviv86:

aviv86:
startsWith() and endsWith() are not supported as well

WestDiscGolf:
Would the suggested implentations [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith) and [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) be enough if implemented? Or would the more "fullfilled" pollyfills be required?

sebastienros:
@WestDiscGolf if you intend to implement the other two ones, and anything else after all, please follow the reference implementation from the specification, and not MDN. It will be easier to understand. But feel free to use the test cases from MDN.

WestDiscGolf:
@sebastienros sure thing! 👍

PR #434, opened 2017-11-03
padStart and padEnd #429

WestDiscGolf:
Thought I'd have a go at another issue on the list #429 

Let me know what you think :-)

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -2,14 +2,12 @@
 using System.Globalization;
 using System.IO;
 using System.Reflection;
-using System.Threading;
 using Jint.Native.Number;
 using Jint.Parser;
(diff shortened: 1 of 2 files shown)


sebastienros:
It looks good. As a general rule I try to follow the specification document that provide pseudo code for the implementation. I assume you did it as I can see similarities.

Jint is supposed to follow ES5.1 only, but I have started to accept more ES6 functions as it's not breaking the contract. So what you are doing is fine. There is an es6 branch, but I am thinking of just iterating on the main dev/master branch instead and add features as they are implemented. We already parse es6 JS in dev/master after all.

WestDiscGolf:
They were based on the links in the doc comments of the methods. I found once I'd implemented them that the majority of the functionality was the same so refactored it into a common method for them both to call.

Do you require me to close this pull request and re-submit it onto the es6 branch?

PR #448, opened 2017-12-18
Optimize array handling memory usage

lahma:
I did first pass on investigating arrays and they memory profile. This is done on top of the benchmark update pull request.

Here are some quick wins from caching common strings, avoiding boxing etc. We are now 40% faster for the famous array test case and use almost 80% less memory (which brings the speed).

*Before*

|   Method | N | ReuseEngine |        Mean |      Error |    StdDev |       Gen 0 |     Gen 1 |     Gen 2 |  Allocated |
|--------- |-- |------------ |------------:|-----------:|----------:|------------:|----------:|----------:|-----------:|
|     **Jint** | **1** |       **False** | **2,081.36 ms** |  **89.654 ms** | **5.0656 ms** | **440500.0000** | **8750.0000** | **2250.0000** | **1784.51 MB** |
|     **Jint** | **1** |        **True** | **2,071.57 ms** | **118.701 ms** | **6.7068 ms** | **441333.3333** | **8916.6667** | **2000.0000** | **1784.37 MB** |

*After*

|   Method | N | ReuseEngine |       Mean |      Error |    StdDev |      Gen 0 |     Gen 1 |     Gen 2 | Allocated |
|--------- |-- |------------ |-----------:|-----------:|----------:|-----------:|----------:|----------:|----------:|
|     **Jint** | **1** |       **False** | **1,201.2 ms** | **137.316 ms** | **7.7586 ms** | **93166.6667** | **4416.6667** | **1250.0000** | **388.35 MB** |
|     **Jint** | **1** |        **True** | **1,187.4 ms** |  **49.602 ms** | **2.8026 ms** | **94000.0000** | **4750.0000** | **1500.0000** | **388.21 MB** |


Jint/Native/Array/ArrayInstance.cs
```diff
@@ -120,14 +113,14 @@ public override bool DefineOwnProperty(string propertyName, PropertyDescriptor d
 
                 if (_array.Count < oldLen - newLen)
                 {
-                    var keys = _array.Keys.ToArray();
+                    var keys = new List<uint>(_array.Keys);
```
ayende: You need the `keys` instance because you might be mutating the keys, but it never escape this method.
This can be cached using a thread static instance that is shared globally. On each call, we clear the instance, then fill it, then actually use the value. It will mean that you'll have much fewer allocations

Jint/Native/Number/Dtoa/CachePowers.cs
```diff
@@ -52,17 +50,15 @@ internal CachedPower(ulong significand, short binaryExponent, short decimalExpon
...
             const int kQ = DiyFp.KSignificandSize;
             double k = System.Math.Ceiling((alpha - e + kQ - 1)*Kd1Log210);
             int index = (GrisuCacheOffset + (int) k - 1)/CachedPowersSpacing + 1;
             CachedPower cachedPower = CACHED_POWERS[index];
 
-            cMk.F = cachedPower.Significand;
-            cMk.E = cachedPower.BinaryExponent;
-            Debug.Assert((alpha <= cMk.E + e) && (cMk.E + e <= gamma));
-            return cachedPower.DecimalExponent;
+            //System.Diagnostics.Debug.Assert((alpha <= cMk.E + e) && (cMk.E + e <= gamma));
```
ayende: Did you intend to remove the assert?

Jint/Native/Array/ArrayInstance.cs
lahma: Yes I was wondering about thread safety and reuse here. I guess multiple simultaneous parser instances can be common (cannot use plain static). Need to adjust for thread-static one then. Same goes for FastDtoa.NumberToString .

Jint/Native/Number/Dtoa/CachePowers.cs
lahma: Mostly for being a bit lazy. It used calculations using mutable state that no longer exists with DiyFp being a struct, I think I can return them with some tweaks.

lahma:
I've incorporated @ayende 's suggestion, but I will open separate smaller pull requests in hopes of getting tweaks merged.

PR #449, opened 2017-12-19
Cache common int to string conversions

lahma:
Now a separate lookup that is used for integers. Memory usage drops by third when using latest array performance benchmark.

``` ini
BenchmarkDotNet=v0.10.11, OS=Windows 10 Redstone 3 [1709, Fall Creators Update] (10.0.16299.125)
Processor=Intel Core i7-6820HQ CPU 2.70GHz (Skylake), ProcessorCount=8
Frequency=2648437 Hz, Resolution=377.5812 ns, Timer=TSC
.NET Core SDK=2.1.2
  [Host]     : .NET Core 2.0.3 (Framework 4.6.25815.02), 64bit RyuJIT
  Job-LPPWMI : .NET Core 2.0.3 (Framework 4.6.25815.02), 64bit RyuJIT

InvocationCount=4  LaunchCount=1  TargetCount=3  
UnrollFactor=4  WarmupCount=3  
```
### Before

| Method |  N | ReuseEngine |    Mean |   Error |   StdDev |       Gen 0 |      Gen 1 |      Gen 2 | Allocated |
|------- |--- |------------ |--------:|--------:|---------:|------------:|-----------:|-----------:|----------:|
|   **Jint** | **20** |       **False** | **4.169 s** | **2.139 s** | **0.1209 s** | **784750.0000** | **40500.0000** | **17500.0000** |   **3.18 GB** |
|   **Jint** | **20** |        **True** | **4.090 s** | **1.327 s** | **0.0750 s** | **786750.0000** | **40250.0000** | **18500.0000** |   **3.17 GB** |

### After

| Method |  N | ReuseEngine |    Mean |    Error |   StdDev |       Gen 0 |      Gen 1 |      Gen 2 | Allocated |
|------- |--- |------------ |--------:|---------:|---------:|------------:|-----------:|-----------:|----------:|
|   **Jint** | **20** |       **False** | **2.586 s** | **0.2367 s** | **0.0134 s** | **504000.0000** | **27500.0000** | **13500.0000** |   **2.06 GB** |
|   **Jint** | **20** |        **True** | **2.663 s** | **1.9397 s** | **0.1096 s** | **504000.0000** | **24000.0000** | **11750.0000** |   **2.06 GB** |

### DIFF
--- a/Jint/Native/Array/ArrayInstance.cs
+++ b/Jint/Native/Array/ArrayInstance.cs
@@ -1,5 +1,5 @@
 ﻿using System.Collections.Generic;
-using System.Linq;
+using System.Threading;
 using Jint.Native.Object;
 using Jint.Runtime;
 using Jint.Runtime.Descriptors;
@@ -8,8 +8,11 @@ namespace Jint.Native.Array
 {
     public class ArrayInstance : ObjectInstance
     {
+        // cache key container for array iteration for less allocations
+        private static readonly ThreadLocal<List<uint>> keyCache = new ThreadLocal<List<uint>>(() => new List<uint>());
+
         private readonly Engine _engine;
-        private IDictionary<uint, PropertyDescriptor> _array = new MruPropertyCache2<uint, PropertyDescriptor>();
+        private readonly Dictionary<uint, PropertyDescriptor> _array = new Dictionary<uint, PropertyDescriptor>();
         private PropertyDescriptor _length;
 
         public ArrayInstance(Engine engine) : base(engine)
@@ -17,13 +20,7 @@ namespace Jint.Native.Array
             _engine = engine;
         }
 
-        public override string Class
-        {
-            get
-            {
-                return "Array";
-            }
-        }
+        public override string Class => "Array";
 
         /// Implementation from ObjectInstance official specs as the one
         /// in ObjectInstance is optimized for the general case and wouldn't work
@@ -120,14 +117,16 @@ namespace Jint.Native.Array
 
                 if (_array.Count < oldLen - newLen)
                 {
-                    var keys = _array.Keys.ToArray();
+                    var keys = keyCache.Value;
+                    keys.Clear();
+                    keys.AddRange(_array.Keys);
                     foreach (var key in keys)
                     {
                         uint keyIndex;
                         // is it the index of the array
                         if (IsArrayIndex(key, out keyIndex) && keyIndex >= newLen && keyIndex < oldLen)
                         {
-                            var deleteSucceeded = Delete(key.ToString(), false);
+                            var deleteSucceeded = Delete(TypeConverter.ToString(key), false);
                             if (!deleteSucceeded)
                             {
                                 newLenDesc.Value = new JsValue(keyIndex + 1);
@@ -219,7 +218,7 @@ namespace Jint.Native.Array
         {
             foreach(var entry in _array)
             {
-                yield return new KeyValuePair<string, PropertyDescriptor>(entry.Key.ToString(), entry.Value);
+                yield return new KeyValuePair<string, PropertyDescriptor>(TypeConverter.ToString(entry.Key), entry.Value);
             }
 
             foreach(var entry in base.GetOwnProperties())
@@ -289,8 +288,12 @@ namespace Jint.Native.Array
 
         public static bool IsArrayIndex(JsValue p, out uint index)
         {
-            index = ParseArrayIndex(TypeConverter.ToString(p));
+            return IsArrayIndex(TypeConverter.ToString(p), out index);
+        }
 
+        private static bool IsArrayIndex(string p, out uint index)
+        {
+            index = ParseArrayIndex(p);
             return index != uint.MaxValue;
 
             // 15.4 - Use an optimized version of the specification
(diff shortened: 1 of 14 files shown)


Jint/Native/Array/ArrayInstance.cs
```diff
@@ -8,22 +8,19 @@ namespace Jint.Native.Array
 {
     public class ArrayInstance : ObjectInstance
     {
+        // cache key container for array iteration for less allocations
+        private static readonly ThreadLocal<List<uint>> keyCache = new ThreadLocal<List<uint>>(() => new List<uint>());
+
         private readonly Engine _engine;
-        private IDictionary<uint, PropertyDescriptor> _array = new MruPropertyCache2<uint, PropertyDescriptor>();
+        private readonly Dictionary<uint, PropertyDescriptor> _array = new Dictionary<uint, PropertyDescriptor>();
```
lahma: MruPropertyCache2 does key.Equals(newKey) checks that will cause boxing with uints, for this use case normal Dictionary<K,V> should be fine, not accessing via interface removes virtual dispatch

Jint/Native/Array/ArrayInstance.cs
```diff
@@ -289,8 +288,12 @@ public override void RemoveOwnProperty(string p)
 
         public static bool IsArrayIndex(JsValue p, out uint index)
         {
-            index = ParseArrayIndex(TypeConverter.ToString(p));
+            return IsArrayIndex(TypeConverter.ToString(p), out index);
+        }
 
+        private static bool IsArrayIndex(string p, out uint index)
```
lahma: better signature when called with string, does not cause automatic JsValue conversion

Jint/Native/Array/ArrayPrototype.cs
```diff
@@ -383,22 +382,27 @@ private JsValue Splice(JsValue thisObj, JsValue[] arguments)
...
                 var fromPresent = o.HasProperty(from);
                 if (fromPresent)
                 {
                     var fromValue = o.Get(from);
-                    a.DefineOwnProperty(k.ToString(), new PropertyDescriptor(fromValue, true, true, true), false);
+                    a.DefineOwnProperty(TypeConverter.ToString(k), new PropertyDescriptor(fromValue, true, true, true), false);
                 }
             }
 
-            var items = arguments.Skip(2).ToArray();
```
lahma: removeing Linq version which can be notoriously bad, especially on full framework where .ToArray is not optimized to predict wanted size

Jint/Runtime/TypeConverter.cs
```diff
@@ -246,7 +249,32 @@ public static uint ToUint32(JsValue o)
         /// <returns></returns>
         public static ushort ToUint16(JsValue o)
         {
-            return (ushort)(uint)ToNumber(o);
+            return (ushort) (uint) ToNumber(o);
+        }
+
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static string ToString(long i)
```
lahma: separate signatures to catch common integers, preventing wrapping to JsValue overload

Jint/Runtime/TypeConverter.cs
```diff
@@ -246,7 +250,40 @@ public static uint ToUint32(JsValue o)
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static string ToString(uint i)
+        {
+            return i >= 0 && i < intToString.Length
+                ? (intToString[i] = intToString[i] ?? i.ToString())
+                : i.ToString();
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static string ToString(char c)
```
lahma: char separately to own array as char can have implicit conversion to integer values and would produce work result based on indexing

lahma:
Hopefully now the sweet spot for minimal changes for handling the common string/char caching. I added some comments to clarify some decisions. If we'd get this merged I could continue with the next effort.

/cc @ayende @sebastienros

Jint/Native/Array/ArrayInstance.cs
```diff
@@ -120,14 +117,16 @@ public override bool DefineOwnProperty(string propertyName, PropertyDescriptor d
...
                 {
-                    var keys = _array.Keys.ToArray();
+                    var keys = keyCache.Value;
+                    keys.Clear();
+                    keys.AddRange(_array.Keys);
                     foreach (var key in keys)
                     {
                         uint keyIndex;
                         // is it the index of the array
                         if (IsArrayIndex(key, out keyIndex) && keyIndex >= newLen && keyIndex < oldLen)
```
sebastienros: Is it possible that we don't need to call `IsArrayIndex` at all here? `keys` is a list of `uint`, so the end result will always be a valid `uint`.

Jint/Runtime/TypeConverter.cs
```diff
@@ -246,7 +250,40 @@ public static uint ToUint32(JsValue o)
...
+        {
+            return c >= 0 && c < charToString.Length
+                ? (charToString[c] = charToString[c] ?? c.ToString())
```
sebastienros: I believe this pattern is not thread safe as the array is shared, and any thread could write to it if it's the first to encounter the number. Unless you would initialize it completely in the static constructor and just read it here.

10 thousand integers of 5 digits max to string would take around 100KB of memory I assume.

sebastienros:
LGTM minus comments.

lahma:
@sebastienros I've done the changes you suggested, thanks for the quick review and good suggestions. I made the string cache a tad smaller now as we are allocation eagerly - probably can grow it later if needed.

Let me know if you want more changes.

PR #450, opened 2017-12-19
Use struct for DiyFp

lahma:
By converting DiyFp to struct we can lower memory traffic by  about 30% with small performance gain. I'm using the array test again here.

I have also internalized the Dtoa bits so that no poor soul takes dependency on them, I believe they are not something that library should offer as part of the API.

/cc @sebastienros @ayende 

### Before

| Method |  N | ReuseEngine |    Mean |   Error |   StdDev |       Gen 0 |      Gen 1 |      Gen 2 | Allocated |
|------- |--- |------------ |--------:|--------:|---------:|------------:|-----------:|-----------:|----------:|
|   **Jint** | **20** |       **False** | **4.169 s** | **2.139 s** | **0.1209 s** | **784750.0000** | **40500.0000** | **17500.0000** |   **3.18 GB** |
|   **Jint** | **20** |        **True** | **4.090 s** | **1.327 s** | **0.0750 s** | **786750.0000** | **40250.0000** | **18500.0000** |   **3.17 GB** |

### After

| Method |  N | ReuseEngine |    Mean |    Error |   StdDev |       Gen 0 |      Gen 1 |      Gen 2 | Allocated |
|------- |--- |------------ |--------:|---------:|---------:|------------:|-----------:|-----------:|----------:|
|   **Jint** | **20** |       **False** | **4.006 s** | **0.9027 s** | **0.0510 s** | **537000.0000** | **36750.0000** | **17000.0000** |   **2.19 GB** |
|   **Jint** | **20** |        **True** | **3.987 s** | **0.1880 s** | **0.0106 s** | **539500.0000** | **40250.0000** | **18500.0000** |   **2.19 GB** |

### DIFF
--- a/Jint/Native/Number/Dtoa/CachePowers.cs
+++ b/Jint/Native/Number/Dtoa/CachePowers.cs
@@ -32,8 +32,7 @@ using System.Diagnostics;
 
 namespace Jint.Native.Number.Dtoa
 {
-
-    public class CachedPowers
+    internal class CachedPowers
     {
         private const double Kd1Log210 = 0.30102999566398114; //  1 / lg(10)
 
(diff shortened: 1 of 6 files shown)


Jint/Native/Number/Dtoa/CachePowers.cs
```diff
@@ -52,17 +51,16 @@ internal CachedPower(ulong significand, short binaryExponent, short decimalExpon
         }
 
 
-        internal static int GetCachedPower(int e, int alpha, int gamma, DiyFp cMk)
+        internal static (short, DiyFp) GetCachedPower(int e, int alpha, int gamma)
```
sebastienros: Would there be less allocations if you used two `out` values instead of a `Tuple` ?

lahma: ValueTuple is a struct so it wont' allocate memory. I did quick check on the out alternative and benchmarkdotnet showed both ways to allocate 0B and ValueTuple version being slightly faster (can't explain that).

PR #452, opened 2017-12-22
Reduce memory allocations

lahma:
This time around again trying to reduce memory allocations to give virtual machine some breathing room. Most of the changes should have positive changes all around.

* remove LINQ from hot paths
* use List<T> where possible to get rid of memory allocations when using foreach
* cache empty completions

Next I'm going to tackle generally ObjectInstance and it's memory usage, trying to optimize MruPropertyCache2 etc.

### Before (dev branch)

|    Method |   N |    Mean |    Error |   StdDev |       Gen 0 |      Gen 1 | Allocated |
|---------- |---- |--------:|---------:|---------:|------------:|-----------:|----------:|
| Benchmark | 500 | 1.094 s | 0.0049 s | 0.0071 s | 243554.1667 | 15075.0000 | 997.15 MB |

### After (this pull request)

|    Method |   N |    Mean |    Error |   StdDev |       Gen 0 |      Gen 1 | Allocated |
|---------- |---- |--------:|---------:|---------:|------------:|-----------:|----------:|
| Benchmark | 500 | 1.057 s | 0.0070 s | 0.0105 s | 204179.1667 | 50595.8333 | 905.73 MB |

### Before

| Method |  N |    Mean |    Error |   StdDev |       Gen 0 |      Gen 1 |      Gen 2 | Allocated |
|------- |--- |--------:|---------:|---------:|------------:|-----------:|-----------:|----------:|
|   Jint | 20 | 2.656 s | 0.1310 s | 0.0074 s | 244000.0000 | 20250.0000 | 10000.0000 |   1.04 GB |

### After

| Method |  N |    Mean |   Error |   StdDev |       Gen 0 |      Gen 1 |      Gen 2 | Allocated |
|------- |--- |--------:|--------:|---------:|------------:|-----------:|-----------:|----------:|
|   Jint | 20 | 2.588 s | 3.788 s | 0.2140 s | 217750.0000 | 21000.0000 | 10250.0000 | 956.95 MB |

### DIFF
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -444,7 +444,7 @@ namespace Jint
             }
         }
 
-        public object EvaluateExpression(Expression expression)
+        public object EvaluateExpression(INode expression)
         {
             _lastSyntaxNode = expression;
 
@@ -679,7 +679,8 @@ namespace Jint
         /// <summary>
         /// Invoke the current value as function.
         /// </summary>
-        /// <param name="propertyName">The arguments of the function call.</param>
+        /// <param name="propertyName">The name of the function to call.</param>
+        /// <param name="arguments">The arguments of the function call.</param>
         /// <returns>The value returned by the function call.</returns>
         public JsValue Invoke(string propertyName, params object[] arguments)
         {
@@ -868,7 +869,7 @@ namespace Jint
             for (var i = 0; i < variableDeclarations.Count; i++)
             {
                 var variableDeclaration = variableDeclarations[i];
-                var declarations  = (List<VariableDeclarator>) variableDeclaration.Declarations;
+                var declarations  = variableDeclaration.Declarations;
                 foreach (var d in declarations)
                 {
                     var dn = d.Id.As<Identifier>().Name;
--- a/Jint/Jint.csproj
+++ b/Jint/Jint.csproj
@@ -1,13 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-
   <PropertyGroup>
     <NeutralLanguage>en-US</NeutralLanguage>
     <TargetFramework>netstandard2.0</TargetFramework>
     <AssemblyOriginatorKeyFile>Jint.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
   </PropertyGroup>
-
   <ItemGroup>
-    <PackageReference Include="Esprima" Version="1.0.0-beta-1000" /> 
+    <PackageReference Include="Esprima" Version="1.0.0-beta-1022" />
   </ItemGroup>
 </Project>
--- a/Jint/Native/Function/FunctionConstructor.cs
+++ b/Jint/Native/Function/FunctionConstructor.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Linq;
 using Esprima;
 using Esprima.Ast;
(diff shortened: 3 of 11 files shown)


Jint/Engine.cs
```diff
@@ -761,7 +765,12 @@ public JsValue GetValue(JsValue scope, string propertyName)
         }
 
         //  http://www.ecma-international.org/ecma-262/5.1/#sec-10.5
-        public void DeclarationBindingInstantiation(DeclarationBindingType declarationBindingType, IEnumerable<FunctionDeclaration> functionDeclarations, IEnumerable<VariableDeclaration> variableDeclarations, FunctionInstance functionInstance, JsValue[] arguments)
+        public void DeclarationBindingInstantiation(
+            DeclarationBindingType declarationBindingType,
+            IList<FunctionDeclaration> functionDeclarations,
```
lahma: IList gives us indexing but not allocation free foreacher, so using indexing

Jint/Engine.cs
```diff
@@ -849,14 +860,19 @@ public void DeclarationBindingInstantiation(DeclarationBindingType declarationBi
...
 
             // process all variable declarations in the current parser scope
-            foreach (var d in variableDeclarations.SelectMany(x => x.Declarations))
+            for (var i = 0; i < variableDeclarations.Count; i++)
             {
-                var dn = d.Id.As<Identifier>().Name;
-                var varAlreadyDeclared = env.HasBinding(dn);
-                if (!varAlreadyDeclared)
+                var variableDeclaration = variableDeclarations[i];
+                var declarations  = (List<VariableDeclarator>) variableDeclaration.Declarations;
```
lahma: this is a repeating pattern, we know Esprima and that it produces List<T>, we can cast to it can get non-virtual dispatch + allocation free foreach

Jint/Engine.cs
ayende: Is there a reason to _not_ use `List<T>` here?

ayende: In particular, I'm thinking about bound check elimintations.

Jint/Engine.cs
ayende: Might be worth just changing it there, no? That would give also skip the type checks here.

Jint/Engine.cs
lahma: The problem here is (like other places too) that we are dealing with public API. Changing to concrete implementations might not be the prettiest thing, even though best performing. I think @sebastienros  can be the judge of this.

Jint/Engine.cs
lahma: I have a pending pull request https://github.com/sebastienros/esprima-dotnet/pull/13 that optimizes things, but does not change the API. Same thing here about setting the public API to concrete implementations. I guess streaming API like IEnumerable will not be supported in the near future and everything is materialized as is.

Forced cast is pretty much safe though as unit tests cover well all these and upgrading to different kind of Esprima would show up.

lahma: I've created separate pull request to change Ast to have concerete types: https://github.com/sebastienros/esprima-dotnet/pull/14

lahma: Concrete type PR went in, but requires NuGet release to be used here.

sebastienros: Pushed on Nuget, should appear soon.

lahma:
I've upgraded to latest Esprima and cleaned the casts, removed Linq from DefaultTypeConverter too. I unfortunately missed the HoistingScope members but they have an API cleanup PR pending that fixes them too.

Can we merge at this point?

PR #454, opened 2017-12-22
Make ObjectInstance MruPropertyCache2 dictionary lazy

lahma:
This change add commonly accessed prototype and constructor fields directly to ObjectInstance. Dictionary is initialized if these two fields are not enough to describe the object. On top of memory usage changes so that PR needs to be addressed first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -160,4 +160,4 @@ project.lock.json
 .build
 
 .idea
-BenchmarkDotNet.Artifacts
+BenchmarkDotNet.Artifacts*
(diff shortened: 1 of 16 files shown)


sebastienros:
I am currently refactoring all the Environment classes and ObjectInstance to follow the ES6 spec, this will give me some headache when I merge 🚑

lahma:
What if you rebase overwriting with your changes and we re-implement the ObjectInstance part later if there's a TODO comment?

sebastienros:
Don't worry, I'll make it work.

PR #456, opened 2017-12-23
Cache common JsValue allocations

lahma:
Here we get a really nice reduction of allocations and a speed boost by catching all sorts of implicit JsValue allocations. I generally don't like implicit operators but it was quite easy to redirect them to use caching. 

As JsValue is immutable and attached to certain object, we can easily pair them lazily and reuse. Also caching common int, double and char maps.

### Before 

|    Method |   N |    Mean |    Error |   StdDev |       Gen 0 |      Gen 1 | Allocated |
|---------- |---- |--------:|---------:|---------:|------------:|-----------:|----------:|
| Benchmark | 500 | 1.214 s | 0.0755 s | 0.1083 s | 202687.5000 | 55312.5000 |  903.2 MB |

### After

|    Method |   N |    Mean |    Error |   StdDev |       Gen 0 |      Gen 1 | Allocated |
|---------- |---- |--------:|---------:|---------:|------------:|-----------:|----------:|
| Benchmark | 500 | 1.131 s | 0.0086 s | 0.0123 s | 189937.5000 | 54187.5000 |  817.6 MB |

### DIFF
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -45,22 +45,22 @@ namespace Jint
 
         internal static Dictionary<Type, Func<Engine, object, JsValue>> TypeMappers = new Dictionary<Type, Func<Engine, object, JsValue>>()
         {
-            { typeof(bool), (Engine engine, object v) => new JsValue((bool)v) },
-            { typeof(byte), (Engine engine, object v) => new JsValue((byte)v) },
-            { typeof(char), (Engine engine, object v) => new JsValue((char)v) },
+            { typeof(bool), (Engine engine, object v) => (bool) v ? JsValue.True : JsValue.False },
+            { typeof(byte), (Engine engine, object v) => JsValue.FromInt((byte)v) },
+            { typeof(char), (Engine engine, object v) => JsValue.FromChar((char)v) },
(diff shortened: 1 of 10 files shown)


lahma:
@sebastienros does this look mergeable? I've kept additions internal so shouldn't cause too much confusion.

sebastienros:
I'd like to keep this change for later. i might refactor JsValue to be a base class for ObjectInstance too making everything a JsValue.

lahma:
Could you perhaps refactor after merge? The API does not actually change as the cache helpers are internal. When refactoring can then easily reflect against the "ideal" numbers.

With all these PRs the problem is not having previous steps merged causing a lot of noise in profiling. This JsValue caching does a big difference with 10% allocation reduction.

lahma:
As I found out that Literal.Cacheable is never set to true, the optimization path never kicks in. I've changed cacheable to ignored as the values are literals. Even better performance now:

|    Method |   N |    Mean |    Error |   StdDev |       Gen 0 |      Gen 1 | Allocated |
|---------- |---- |--------:|---------:|---------:|------------:|-----------:|----------:|
| Benchmark | 500 | 1.058 s | 0.0096 s | 0.0140 s | 187437.5000 | 62437.5000 |  812.9 MB |

PR #472, opened 2018-01-27
Expose _statementCount

xclud:
I am using Jint in a project and i need to show my customers how many CPU instructions they have consumed.

I had to edit the project on my local machine. Then i thought it's a useful enhancement to the original project.

Thank you for the great Jint.

### DIFF
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -295,6 +295,17 @@ namespace Jint
             _statementsCount = 0;
         }
 
+        /// <summary>
+        /// Gets current statement count.
+        /// </summary>
+        public int StatementCount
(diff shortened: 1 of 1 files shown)


Jint/Engine.cs
```diff
@@ -295,6 +295,17 @@ public void ResetStatementsCount()
             _statementsCount = 0;
         }
 
+        /// <summary>
+        /// Gets current statement count.
+        /// </summary>
+        public int StatementCount
```
lahma: Should this be `ExecutedStatementCount`?

xclud: Yours is a better name.

PR #482, opened 2018-02-12
add memory usage limit option to Engine

0xFireball:


Jint/Engine.cs
```diff
@@ -370,6 +383,17 @@ public Completion ExecuteStatement(Statement statement)
                 throw new TimeoutException();
             }
 
+#if NETCOREAPP2_0
+            if (Options._MemoryLimit > 0)
+            {
+                var memoryUsage = GC.GetAllocatedBytesForCurrentThread() - _initialMemoryUsage;
+                if (memoryUsage > Options._MemoryLimit)
+                {
+                    throw new MemoryLimitExceededException();
```
ayende: I would suggest adding some details here. Script has allocated X but is limited to Y

Jint/Runtime/MemoryLimitExceededException.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Jint.Runtime
+{
+    public class MemoryLimitExceededException : Exception
+    {
+        public MemoryLimitExceededException() : base("The memory limit has been exceeded.")
```
ayende: I would recommend using standard exception ctors, instead.

Jint/Engine.cs
```diff
@@ -372,6 +385,17 @@ public Completion ExecuteStatement(Statement statement)
                 throw new TimeoutException();
             }
 
+#if NETCOREAPP2_0
+            if (Options._MemoryLimit > 0)
+            {
+                var memoryUsage = GC.GetAllocatedBytesForCurrentThread() - _initialMemoryUsage;
```
pranavkm: Might be a bit more cumbersome, but you could avoid cross-compiling by reflecting on the method. This should make it work in all platforms that support the API.

`Func<int> GetAllocatedBytesForCurrentThread = _methodInfo != null ? Delegate.CreateDelegate() : long.MaxValue;`

Jint/Engine.cs
```diff
@@ -385,16 +400,14 @@ public Completion ExecuteStatement(Statement statement)
                 throw new TimeoutException();
             }
 
-#if NETCOREAPP2_0
             if (Options._MemoryLimit > 0)
             {
-                var memoryUsage = GC.GetAllocatedBytesForCurrentThread() - _initialMemoryUsage;
+                var memoryUsage = GetAllocatedBytesForCurrentThread() - _initialMemoryUsage;
                 if (memoryUsage > Options._MemoryLimit)
```
ayende: This will _always_ be triggered if the method is not available.

Jint/Engine.cs
```diff
@@ -76,6 +74,20 @@ public class Engine
...
+        {
+            var methodInfo = typeof(GC).GetMethod("GetAllocatedBytesForCurrentThread");
+
+            if (methodInfo != null)
+            {
+                GetAllocatedBytesForCurrentThread =  (Func<long>)Delegate.CreateDelegate(typeof(Func<long>), null, methodInfo);
+            }
+            else
+            {
+                GetAllocatedBytesForCurrentThread = () => long.MaxValue;
```
ayende: Why not returning 0 here?

Jint/Engine.cs
```diff
@@ -247,6 +258,8 @@ internal ReferencePool ReferencePool
         }
         #endregion
 
+        private static Func<long> GetAllocatedBytesForCurrentThread;
```
pranavkm: `static readonly`

Jint/Engine.cs
```diff
@@ -384,6 +411,22 @@ public Completion ExecuteStatement(Statement statement)
...
+                {
+                    var memoryUsage = GetAllocatedBytesForCurrentThread() - _initialMemoryUsage;
+                    if (memoryUsage > Options._MemoryLimit)
+                    {
+                        throw new MemoryLimitExceededException($"Script has allocated {memoryUsage} but is limited to {Options._MemoryLimit}");
+                    }
+                }
+                else
+                {
+                    throw new PlatformNotSupportedException("The current platform doesn't support MemoryLimit.");
```
pranavkm: 👏

Issue #436, opened 2017-11-17
JSON.parse with reviver is not supported


BinishPeter:
Nothing happens when JSON.parse is invoked with reviver.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse

sebastienros:
https://github.com/sebastienros/jint/blob/dev/Jint/Native/Json/JsonInstance.cs#L43

PR #484, opened 2018-02-22
fixed bug #436

pranayrauthu:
modified JSON.parse according to spec.

### DIFF
--- a/Jint/Native/Json/JsonInstance.cs
+++ b/Jint/Native/Json/JsonInstance.cs
@@ -1,5 +1,6 @@
 ﻿using Jint.Native.Object;
 using Jint.Runtime;
+using Jint.Runtime.Descriptors;
 using Jint.Runtime.Interop;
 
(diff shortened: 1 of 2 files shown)


Jint/Native/Json/JsonInstance.cs
```diff
@@ -36,10 +38,93 @@ public void Configure()
...
+            {
+                var valueAsObject = value.AsObject();
+                if (valueAsObject.Class == "Array")
+                {
+                    var valAsArray = value.AsArray();
+                    var i = 0;
+                    var arrLen = valAsArray.GetLength();
+                    while (i < arrLen)
+                    {
+                        var newValue = AbstractWalkOperation(valAsArray, i.ToString());
```
lahma: ideally always call TypeConverter.ToString(i), it has cached lookups for strings, this i.ToString() has been repeated three times here

PR #486, opened 2018-03-07
JsString multiple concatination fails

pitMobile:
If you concatenate strings more than once JSString (ConcatenatedString) delivers a wrong value, just the first concatenation works, overwriting ToObject in ConcatenatedString resolves this issue

### DIFF
--- a/Jint/Native/JsString.cs
+++ b/Jint/Native/JsString.cs
@@ -164,6 +164,17 @@ namespace Jint.Native
 
                 return _value;
             }
+            
+            [Pure]
+            public override object ToObject()
+            {
+                if (_dirty)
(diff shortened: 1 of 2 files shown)


Jint/Native/JsString.cs
```diff
@@ -164,6 +164,17 @@ public override string AsString()
 
                 return _value;
             }
+            
+            [Pure]
+            public override object ToObject()
+            {
+                if (_dirty)
```
lahma: maybe this should just call AsString() to remove duplication?

sebastienros:
Can I ask you to add a unit test also?
Thanks

PR #498, opened 2018-04-24
Allow creating an array instance based on existing PropertyDescriptor array/dictionary

maximburyak:


Jint/Native/Array/ArrayInstance.cs
```diff
@@ -541,7 +568,7 @@ private bool TryGetDescriptor(uint index, out PropertyDescriptor descriptor)
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal void WriteArrayValue(uint index, PropertyDescriptor desc)
+        public void WriteArrayValue(uint index, PropertyDescriptor desc)
```
lahma: WriteArrayValue is quite counter-intuitive API-wise, thus it was set as internal to allow direct calls. Would this be the case to just expose the indexer that has already been requested and delegate it to this method?

maximburyak: Yeah, that's a mistake. The Ctors already does the job, and yes, it's unsafe to expose  that function, without having a mechanism updating the 'length' automatically.

lahma:
LGTM

PR #509, opened 2018-06-05
Fixed self referencing objects to CLR transfer

### LINKED ISSUE #507: Recursive ToObject calls for graph
ObjectWrapper implementation tries to recursively create graph in file https://github.com/sebastienros/jint/blob/rel/3.0-beta/Jint/Native/Object/ObjectInstance.cs#L923

Problem occurs when I want to pass a self referencing parent child relation to c# method.

To recreate scenario,

```c#
        [Fact]
        public void Recursive() {
            var engine = new Engine();

            engine.Global.Put("printer", 
                  Native.JsValue.FromObject(engine, NodePrinter.Instance), true);

            engine.Execute("var p = { parent: null, child: null };");
            engine.Execute("var c = { parent: p, child: null };");
            engine.Execute("p.child = c;");
            engine.Execute("printer.print(p);");
        }

        public class NodePrinter {

            public static NodePrinter Instance = new NodePrinter();

            public void Print(Native.JsValue node) {
                var a = node.AsObject();
                var b = node.AsObject();
                System.Diagnostics.Debug.WriteLine($"{a},{b}");
            }

        }

```
This creates infinite recursion, this issue is fixed in pull request #509

ackava:
Issue fixed
#507 

Added unit test for the same.


ackava:
@sebastienros Can you please look into this pull request?

PR #515, opened 2018-06-19
Harmonize exception throwing

lahma:
This is mostly about cleanup and making exception throwing consistent. Now there's ExceptionHelper that does most of the throwing. This will also add possibility for JIT to inline better as there are no throw statements in function bodies.

### DIFF
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -405,38 +405,7 @@ namespace Jint
 
         public Completion ExecuteStatement(Statement statement)
         {
-            if (_maxStatements > 0 && _statementsCount++ > _maxStatements)
-            {
-                ThrowStatementsCountOverflowException();
-            }
-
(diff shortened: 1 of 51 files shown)


Jint/Engine.cs
```diff
@@ -717,13 +722,7 @@ public void PutValue(Reference reference, JsValue value)
             else
             {
                 var baseValue = reference.GetBase();
-                if (!(baseValue is EnvironmentRecord record))
```
sebastienros: I assume that if there is this test it's because it was in the spec. Maybe there is not code path for it, but I wouldn't remove it, unless you found out it could never be true.

lahma: But as you can see the code right after it does forced cast so the invariant is present. It just changes from ambiguous ArgumentNullException to maybe even better InvalidCastException?

sebastienros: Sorry, I thought it was throwing a JS exception. It's just dotnet one.

PR #523, opened 2018-08-05
Added attributes for including/ignoring fields/properties/constructors/methods

Mattiemus:
This change includes the ability to specify which parts of a class should be made visible when using either the TypeReference class or ObjectWrapper.

Classes can be decorated with the new JintClass attribute which will specify whether methods should be discovered via an OptIn or OptOut mode. OptIn discovery requires all fields to be annotated with an attribute to be visible from JS, where OptOut includes all fields by default, but allows removal via the JintIgnore attribute. By default undecorated classes will use the OptOut discovery mode, so will retain existing behavior.

    [JintClass(DiscoveryModes.OptIn)]
    public class Example
    {
        [JintConstructor]
        public Example(int test)
        {
        }

        [JintField]
        public int TestField;

        [JintProperty]
        public int TestProperty { get; set; }

        [JintMethod]
        public void TestMethod()
        {
        }

        public void NotVisibleMethod()
        {
            // This method will not be visible within javascript
        }
    }

Alternatively OptOut can be used to include all fields by default

    [JintClass] // Or [JintClass(DiscoveryModes.OptOut)] - OptOut is default to retain existing behavior
    public class Example
    {
        public Example(int test)
        {
        }

        public int TestField;

        public int TestProperty { get; set; }

        public void TestMethod()
        {
        }

        [JintIgnore]
        public void NotVisibleMethod()
        {
            // This method will not be visible within javascript
        }
    }

### DIFF
--- a/Jint.Tests/Runtime/Domain/OptInTestClass.cs
+++ b/Jint.Tests/Runtime/Domain/OptInTestClass.cs
@@ -1,0 +1,27 @@
+﻿using Jint.Runtime.Interop;
+
+namespace Jint.Tests.Runtime.Domain
+{
+    [JintClass(DiscoveryModes.OptIn)]
+    public class OptInTestClass
+    {
+        public int TestRemovedField;
+
+        [JintField]
+        public int TestField;
+
+        public int TestRemovedProperty { get; set; }
+
+        [JintProperty]
+        public int TestProperty { get; set; }
+
+        public void TestRemovedMethod()
+        {
+        }
+
+        [JintMethod]
+        public void TestMethod()
+        {
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


sebastienros:
I suggest you create an interface like `IReflectionHandler` that would have the methods you provide in the Type helper, and a default implementation that reads everything like it's doing today. They you can create another implementation that will be opt-in instead an use the attributes. This will allow for anyone else to implement their own implementation. You can also make it a collection of `IReflectionHandler` and the result would be a struct containing the member info, and a flag that states if it's allowed or not. Then in the configuration we can _add_ multiple handlers. One could be opt-in, the second one could be based on attributes, and a third one would then just remove elements based on names (it's just an example of how extensible this would be).

henriquegnetzka:
Hey guys! Why wasn't this merged? I'm really struggling with this (preventing "one class" from reflecting)...

PR #526, opened 2018-08-14
System.Guid compatible

marcusalexandre:
Compatible with ValuesObjects, such as Guids


sebastienros:
Could you write a small unit test please (and thank you)? In EngineTests probably.

sebastienros:
This part of the code has changed a lot

PR #530, opened 2018-08-21
net45 support.

### LINKED ISSUE #528: [Jint 3.0] net45 support
Hi,

we want to use esprima and jint 3 in an net45 project. Can we add net45 suppoort?
I can send a PR. See sebastienros/esprima-dotnet#33

Regards
Michael

ViceIce:
fixes #528

Required for build: [VisualOn.Esprima.1.0.0-beta-1036-01.nupkg.zip](https://github.com/sebastienros/esprima-dotnet/files/2305579/VisualOn.Esprima.1.0.0-beta-1036-01.nupkg.zip)

Build from sebastienros/esprima-dotnet#34


Jint/Jint.csproj
```diff
@@ -1,12 +1,13 @@
...
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net45</TargetFrameworks>
     <AssemblyOriginatorKeyFile>Jint.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
     <LangVersion>latest</LangVersion>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Esprima" Version="1.0.0-beta-1036" />
+    <PackageReference Include="VisualOn.Esprima" Version="1.0.0-beta-1036*" />
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
```
lahma: This should probably have Condition that checks that only if net45

lahma: And if I remember right ValueTuple can bring a lot of trouble to consuming side prior to 4.7.2 due to versioning, is this really used somewhere...

lahma: Ah I see, might be a future improvement to remove dependency by using internal struct to pass values inside Dtoa.

Jint/ArrayExt.cs
```diff
@@ -0,0 +1,18 @@
+
+namespace System
+{
+    internal static class ArrayExt
```
sebastienros: Can't we make it conditional to NET45 and use the same extension method name as netstandard? `Emtpy<string>()`

ViceIce: This is currently not possible. There is no way to define static extension methods.

ViceIce: Just for reference: https://github.com/dotnet/roslyn/issues/996 :smile:

lahma:
I rebased the branch and fixed some issues with the rebase, then fixed couple issues more and removed the ValueTuple dependency. Let's see how it fares in CI.

PR #539, opened 2018-09-06
Add some missing new Math methods in ECMAScript 6

lahma:


Jint/Native/Math/MathInstance.cs
```diff
@@ -650,6 +653,89 @@ private static JsValue Tan(JsValue thisObject, JsValue[] arguments)
...
+            {
+                return +0;
+            }
+            else if (NumberInstance.IsNegativeZero(x))
+            {
+                return -0;
+            }
+            else if (double.IsPositiveInfinity(x))
+            {
+                return double.PositiveInfinity;
```
sebastienros: Don't we have constants for these?

lahma: We don't have a ECMA test for this, but I tweaked the code to follow same pattern as the two above to return value as-is if it's either positive or negative infinity.

PR #541, opened 2018-09-09
ES6 Map, Set and initial Iterators

lahma:
First implementation where most tests pass and one can use new Map(), new Set() and use them pretty much as expected. Some shortcomings listed in skipped.json.


Jint/Native/Function/FunctionInstance.cs
```diff
@@ -15,23 +15,35 @@ public abstract class FunctionInstance : ObjectInstance, ICallable
         private const string PropertyNameLength = "length";
         private const int PropertyNameLengthLength = 6;
         protected PropertyDescriptor _length;
+
+        private const string PropertyNameName = "name";
+        private const int PropertyNameNameLength = 4;
```
sebastienros: I assume for performance reasons. Isn't the `==` operator on string cleaver enough to already apply this?

Jint/Native/JsValue.cs
```diff
@@ -142,6 +144,37 @@ public ArrayInstance AsArray()
...
+                return null;
+            }
+
+            if (!(value is ICallable callable))
+            {
+                ExceptionHelper.ThrowArgumentException("The value is not iterable");
+                return null;
+            }
+
+            var obj = (ObjectInstance) callable.Call(this, System.Array.Empty<JsValue>());
```
sebastienros: There is a `Arguments.Empty` field for this (same thing in the end, just more meaningful)

Jint/Native/Array/IArrayLike.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Jint.Native.Array
+{
+    internal interface IArrayLike
```
sebastienros: Looking at the spec, an "array-like" object is from TypedArray, and here only String implements it. 
https://www.ecma-international.org/ecma-262/6.0/#sec-typedarray-objects

Also in the Array.from you don't use Iterator at all, is that intentionally left for later?
https://www.ecma-international.org/ecma-262/6.0/#sec-array.from

sebastienros: I see that what they call "array-like" is any object that has a "length" property and can have properties like `'0'`, ... So yes String is an array-like object. But maybe the interface should actually expose the `Length` and `this[int]` members? Might be easier to implement the spec that expects array-like objects.

Jint/Native/Function/FunctionInstance.cs
lahma: I measured it with BDN at it showed to be a bit faster when we don't need a null check before we do length check, so a real micro-optimization that might be something to remove at some point.

Jint/Native/Array/IArrayLike.cs
lahma: I started to check Array items and was overwhelmed with the stuff that is needed. So I was thinking maybe process this part first and when PR lands continue with adding missing bits to strings and Array.

PR #543, opened 2018-09-10
.NET 4.5 support and more ES6 features

sebastienros:


sebastienros:
FYI I never squash from dev to release or master. I just squash PRs to dev.

PR #544, opened 2018-09-13
how to get source  from function  in jint

MustafaKAPLAN:
example:
function test(a,b){ return +b;}

var src= test.toString();
i want to ---> stc ==>'function test(a,b){ return +b;}' 
please help
in c#
      var v=engine.getproperty('test');
     var src= ??????  from v


sebastienros:
Not sure what you are doing here, I assume it was a mistake.
If you have a question feel free to join on the Gitter forum, or file an issue.

Issue #383, opened 2017-05-16
new Date(2027,0,1).setMonth(-1) does not go backward in time


christianrondeau:
Based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setMonth, if the month is outside the range (not between 0 and 11), it will change the year accordingly (and potentially adjust for missing day)

Simple use case:

```
var myDate = new Date(2019,0,1);
myDate.setMonth(-1);
```

Here, myDate is 2019/01/01, but it should be 2018/12/01. Any negative value will be silently ignored.

WestDiscGolf:
Unsure if this is related, but playing in the chrome console, the return value of setMonth is in milliseconds so will need converting back into a date?

![image](https://user-images.githubusercontent.com/374343/32294905-94d6f688-bf3f-11e7-9e50-a3154a187e8c.png)

WestDiscGolf:
Had a go at writing a fix for this, but can't quite put my finger on the solution. However I have written a unit test to prove the problem (although looking at this again, I'm sure it could also be improved).

```
[Theory]
[InlineData("new Date(2019, 0, 1)", 1, 12, 2018)]
[InlineData("new Date(1970, 0, 1)", 1, 12, 1969)]
[InlineData("new Date(1958, 0, 1)", 1, 12, 1957)]
public void ShouldSetMonthToPreviousYear(string startDate, int day, int month, int year)
{
      var code = $"var myDate = {startDate}; myDate.setMonth(-1);";

      var engine = new Engine();
      var result = engine.Execute(code).GetValue("myDate").AsDate().ToDateTime();

      Assert.Equal(day, result.Day);
      Assert.Equal(month, result.Month);
      Assert.Equal(year, result.Year);
}
```

majood:
Any news about this issue? Should be regarded as a bug since all major browsers support it?

WestDiscGolf:
Interestingly I've just rebased my branch (as it was almost a year out of date) and re-run my unit test (from above) on the updated code from the past year. Both the first 2 tests (2019 -> 2018 and 1970 -> 1969) both work. However the last one 1958->1957 still breaks but with the value of 2nd December 1957. So almost there ;-)

Looking through the changes in `DatePrototype` it appears than the change that Victor did in commit 43b57529 has resolved part of the issue, but not all. The change was in relation to #467.

sebastienros:
Can you create a PR with only a failing unit test in it?

PR #548, opened 2018-11-01
Add in failing unit test for #383

WestDiscGolf:
As discussed in #383


sebastienros:
See https://github.com/sebastienros/jint/pull/549

PR #555, opened 2018-11-17
ES6 array functions

lahma:
Missing Array functions

Skipping ones related to

* realms
* constructor functions
* proxies
* flat and flatMap (experimental and there's plenty of stuff here without them too)

### DIFF
--- a/Jint.Tests.Ecma/EcmaTest.cs
+++ b/Jint.Tests.Ecma/EcmaTest.cs
@@ -251,6 +251,14 @@ namespace Jint.Tests.Ecma
                     continue;
                 }
 
+                if (sourceFile.Skip
+                    && (sourceFile.Reason == "part of new test suite"
+                        || sourceFile.Reason.IndexOf("configurable", StringComparison.OrdinalIgnoreCase) > -1))
+                {
+                    // we consider this obsolete and we don't need to process at all
+                    continue;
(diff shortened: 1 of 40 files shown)


sebastienros:
Looks great so far. I will want to check why these tests need to be skipped though.

lahma:
Yep, most of them have same magic number `-4294967294`, for example:

```
function testcase() {
    var obj = { 1: "true", 2: "2", length: "-4294967294" };

    return Array.prototype.indexOf.call(obj, "true") === 1 &&
    Array.prototype.indexOf.call(obj, "2") === -1;
}
runTestCase(testcase);
```

If you run the code in Chrome console, you'll see it evaluates to false. Just another day in JavaScript land with need for such a test case 😉

lahma:
I cross-referenced tests by name and marked ones present in new suite under Array (that are being run already) as skipped. The other skips follow quite the same pattern. object.length can now be 2^53-1 so many old tests tried to prove functions to throw error and actually caused a hang (both in Jint and Chrome) due to huge amount of data being traversed as millions of items was suddenly valid.

lahma:
Good idea. I added logic to ignore test cases that are obsolete (invalid now like `configurable: true` or part of new test suite), now there are only 48 reported as skipped.

PR #560, opened 2018-12-02
ES6 string functions

lahma:
* ignoring test cases related to unimplemented symbols
* Ignoring matchAll which is proposal


Jint/Runtime/TypeConverter.cs
```diff
@@ -294,10 +294,22 @@ internal static string ToString(double d)
         }
 
         /// <summary>
-        /// http://www.ecma-international.org/ecma-262/5.1/#sec-9.8
+        /// http://www.ecma-international.org/ecma-262/6.0/#sec-topropertykey
+        /// </summary>
+        public static string ToPropertyKey(JsValue o)
```
sebastienros: 👍

PR #563, opened 2018-12-09
WIP Spread, rest and default

lahma:


Jint/Runtime/Interpreter/Expressions/JintNewExpression.cs
```diff
@@ -21,14 +22,22 @@ protected override void Initialize()
...
-
-            BuildArguments(_jintArguments, arguments);
+            JsValue[] arguments;
+            if (_hasSpreads)
+            {
+                arguments = BuildArgumentsWithSpreads(_jintArguments);
+            }
+            else
+            {
+                arguments = _engine._jsValueArrayPool.RentArray(_jintArguments.Length);
```
sebastienros: Is there a 'release' operation on these?

Jint/Native/JsValue.cs
```diff
@@ -168,8 +168,9 @@ internal IIterator GetIterator(Engine engine)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         internal bool TryGetIterator(Engine engine, out IIterator iterator)
         {
-            if (!(this is ObjectInstance oi)
-                || !oi.TryGetValue(GlobalSymbolRegistry.Iterator._value, out var value)
+            var objectInstance = TypeConverter.ToObject(engine, this);
+
+            if (!objectInstance.TryGetValue(GlobalSymbolRegistry.Iterator._value, out var value)
```
sebastienros: should Iterator._value be internal and look like `Iterator.Value` if that's the expected usage?

sebastienros:
Just one test for now? WIP and you are working on the default suite?

Jint/Native/JsValue.cs
lahma: That might make it look a bit prettier, now it's just direct field access.

Jint/Runtime/Interpreter/Expressions/JintNewExpression.cs
lahma: Yes it's there at the end. Custom build array (from List.ToArray) can also be returned, it's currently put back to pool only if size is in the range of 1...4.

lahma:
I started with this as it a bit muddy to do other features when they might use spread for testing, for example the arrow functions. Ideally we need the actual ECMA suite, but didn't find any particular "folder" to include. So I created couple obvious test cases for now.

lahma:
Scratch that, expression suite has array, call and new. I'll add those too.

PR #564, opened 2018-12-13
atob and btoa are fairly standard functions

AsyncVoid:


sebastienros:
Sorry but I can't find anything like this in the spec. You will have to add it in your own application.

If there is something you need to make it easier to extend existing JS types from your application, please let me know.

Ghostbird:
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa

There are fairly standard indeed. I ran into this today.

VollRahm:
I ran into this today too

lahma:
Related https://twitter.com/addaleax/status/1384553868726607872?s=19

Ghostbird:
> Related https://twitter.com/addaleax/status/1384553868726607872?s=19

I agree, that atob and btoa are unsuitable for string encoding. In our case I just added it to our Jint engine instance as a function that calls `Convert.ToBase64String` after ensuring that the input array data is valid bytes. It's used to save thumbnails IIRC.

KoalaBear84:
This seems to work fine, at least for my usecase. Which is great :)

```csharp
Func<string, string> atob = str => Encoding.UTF8.GetString(Convert.FromBase64String(str));
JintEngine.SetValue("atob", atob);
```

Ghostbird:
> This seems to work fine, at least for my usecase. Which is great :)
> 
> ```cs
> Func<string, string> atob = str => Encoding.ASCII.GetString(Convert.FromBase64String(str));
> JintEngine.SetValue("atob", atob);
> ```
> 
> Looks like it's better to use ASCII instead of UTF8.

I think you shouldn't use string encoding at all with `btoa`/`atob`. If you need to encode strings as strings, there are far better methods available. The only thing we use it for, is to encode binary data `byte`s.

KoalaBear84:
Do you have an alternative option? I don't have any idea what else i could do. If I look to the Mozilla link you posted it also should be ASCII, or UTF-16. ASCII worked fine, and only tested UTF-8, so not sure if UTF-16 would have worked.

I didn't continue because this worked for my use case.

Ghostbird:
I finally realised the big issue with `btoa` / `atob`. They're string to string conversion functions. Which is madness. I realised this after we got some unexpected behaviour when we cross-referenced our `btoa` to my browser's built-in btoa. The odd thing was, that it was the _browser_ that was producing "incorrect" values. Turns out that if you feed a byte array into `btoa` in javascript, it's coerced to a string first, which adds encoding complications, and then it runs `btoa` as string-to-string function on it.

We removed our `btoa` from our jint globals and replaced it with a global `toBase64` function that will explicitly demand a `Uint8ClampedArray` as input.

PR #566, opened 2018-12-19
Improve for loops, improve reported reference errors

lahma:
While investigating some RavenDB trouble found out:

* out reference error can be missing the actual name that was tried to be referenced
* some for semantics have changed in the latest compliance suite, namely what is returned on break

Added related tests from 262 suite.

### DIFF
--- a/Jint.Tests.Test262/Test262Test.cs
+++ b/Jint.Tests.Test262/Test262Test.cs
@@ -21,7 +21,6 @@ namespace Jint.Tests.Test262
         private static readonly Dictionary<string, string> _skipReasons =
             new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
-
         static Test262Test()
         {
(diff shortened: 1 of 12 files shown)


Jint.Tests.Test262/Test262Test.cs
```diff
@@ -67,20 +66,42 @@ protected void RunTestCode(string code, bool strict)
...
+                catch (JavaScriptException j)
+                {
+                    lastError = TypeConverter.ToString(j.Error);
+                }
+                catch (Exception e)
+                {
+                    lastError = e.ToString();
+                }
 
-            Assert.Null(lastError);
```
sebastienros: Shouldn't the `negative` check be around this line? The rest is identical in both conditions

lahma: Indeed, you caught me copy-pasting, shame on me. Changed to only check varying part.

PR #570, opened 2018-12-22
ES6 String Literals

lahma:


Jint/Native/Number/NumberPrototype.cs
```diff
@@ -234,15 +235,15 @@ public static string ToBase(long n, int radix)
                 return "0";
             }
 
-            var result = new StringBuilder();
+            var result = StringBuilderPool.GetInstance();
```
sebastienros: Would it make sense to have the pool return a disposable instance and release it on disposal?

Jint/Runtime/Interpreter/Expressions/JintTemplateLiteralExpression.cs
```diff
@@ -0,0 +1,57 @@
...
+                _expressions[i] = Build(_engine, exp);
+            }
+
+            _initialized = true;
+        }
+
+        private JsString BuildString()
+        {
+            var sb = StringBuilderPool.GetInstance();
+            for (var i = 0; i < _templateLiteralExpression.Quasis.Count; i++)
```
sebastienros: Is there a number of fragments under which a concat is faster than using a builder?

lahma: Tried concatenation (+=) against builder and the are pretty much same speed-wise. After 2 placeholders string builder's allocation profile gets better.

lahma:
Rebased and cleaned StringBuilderPool, now using dispose pattern.

PR #573, opened 2018-12-26
Fix regression in StrictlyEqual

lahma:
A regression sneaked in at one point, tracked it down and added the base64 encoding test to protect from reappearance.


lahma:
I'm merging this as this is just a revert to old functionality and contains a new test case.

sebastienros:
ok

Issue #459, opened 2018-01-05
ToPrecision() not being evaluated as expected


peadarMc:
Hello,
I am currently looking at Jint as an alternative to MsScriptControl.

I was doing some preliminary testing and found that the ToPrecision javascript method isn't returning the values I would expect to see. Below is the console app I used for testing

        static void Main(string[] args)
        {
            var script1 = @"
            var num = 133.714;
            var n = num.toPrecision(2);
            n;
            ";

            var jsEngine = new Engine();
            string scriptResult;

            var ores = jsEngine.Execute(script1.ToString()).GetCompletionValue();
            scriptResult = ores.ToString(); // Jint is returning 133.7 here

            ScriptControl sc = new ScriptControl { Language = "Javascript" };
            string scriptResult2;
            object ores2 = sc.Eval(script1.ToString());
            scriptResult2 = ores2.ToString(); // Script control returns 130 
        }

peader:
Anyone have a chance to look at this? The above example can just as easily be run in a browser console for comparison with Jint.

lahma:
This has been now fixed in dev, ECMAScript test suite passes also for toPrecision.

PR #574, opened 2018-12-29
ES6 Number APIs

lahma:
* quite a lot of work as missing Dtoa functionality now ported to finally fix the issues with toPrecision (#459)
* the Dtoa code is intentionally C style syntax for it to be easier to diff with v8 code base, so maybe we can be less strict about stylistic issues
* converted builder to be simpler to follow v8 code base and moved formatting to prototype
* unified string builder pool to work like other pools, sharing dtoa builder instance via engine, like pools
* new tests included & passing


sebastienros:
I wonder if we can do better with the SB pool. Now we need to pass (and have) the engine for most type conversions because of it. First question, do we really need a pool of instance, or just one reusable instance, as in "are there recursive calls" that will issue multiple SBs?

Should/could the SBs be shared by all Jint engines, using a static pool? I assume it matters in a server scenario only where concurrency is more an issue, but at the same time it could also be an improvement as all threads could reuse the same SB overtime. Another option would be to have a static pool of SB pools, to reduce concurrency by reusing the pools across engine instances.

PR #578, opened 2019-01-04
ES6 destructuring

lahma:
Decided to tackle this in interim as this gives more test coverage again, also paying debt by releasing tests from ignore.


sebastienros:
LGTM

How much do you follow the spec compared to your instincts/tests?

lahma:
I had the spec there while implementing, but I did some shortcuts to make it perform better (spec always has the IsFunction, IsReference, etc metafunctions that are faster to follow by ideology, not creating/calling them). Here the change is quite isolated to expression handling (the new class), but when the time comes for let/const I think by-the-book might be easier.

Some-hundred new tests are passing, I'd say it's quite well according to spec 😉

lahma:
And yes, the tests marked with "Esprima problem" are mostly about object spread which Esprima doesn't handle and is not part of the ES6, object spread is in draft version.

sebastienros:
The hidden question was "did you just write it from tests", you answered, thanks.

PR #582, opened 2019-01-07
NullReferenceException relating to IObjectWrapper fixed

steffen-liersch:
This contribution fixes a NullReferenceException relating to IObjectWrapper. I found the issue in context with a custom ObjectWrapper based on ObjectInstance implementing IObjectWrapper.

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -1799,6 +1799,12 @@ namespace Jint.Tests.Runtime
         }
 
         [Fact]
+        public void RegExpPrototypeToString()
+        {
+            RunTest("assert(RegExp.prototype.toString() === '//');");
+        }
+
+        [Fact]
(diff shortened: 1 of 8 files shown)


sebastienros:
Can you create a new PR targetting the dev branch? And also add a unit test that will check for the bug?

steffen-liersch:
File Jint/Runtime/ExpressionIntepreter.cs is no more present in the dev branch. Furthermore the issue seems not to apply to the dev branch.

sebastienros:
Then you should use the dev branch, until we merge to master. The package is available on myget.

steffen-liersch:
At the moment the dev branch is no option for me due to the following main reasons:

- I have a function to load a module similar to function require in Node.js. With the dev branch the only way I see to get the context of the calling module (for resolving the module name against the path of the calling module) is to use the debug callbacks.
- Furthermore it seems as if line number in the case of an error are not implemented yet.

PR #585, opened 2019-01-11
Fixes delegates created from expressions.

kevinarthurackerman:
Fixes bug where delegates generated from expressions contain a closure as their first argument, causing mapping of parameters to fail.

### DIFF
--- a/Jint/Runtime/Interop/DelegateWrapper.cs
+++ b/Jint/Runtime/Interop/DelegateWrapper.cs
@@ -1,9 +1,9 @@
 using System;
 using System.Globalization;
 using System.Reflection;
-using System.Linq;
 using Jint.Native;
 using Jint.Native.Function;
+using System.Linq;
 
 namespace Jint.Runtime.Interop
 {
--- a/Jint/Runtime/Interop/DelegateWrapper.cs
+++ b/Jint/Runtime/Interop/DelegateWrapper.cs
@@ -38,11 +38,16 @@ namespace Jint.Runtime.Interop
         public override JsValue Call(JsValue thisObject, JsValue[] jsArguments)
         {
             var parameterInfos = _d.Method.GetParameters();
-
+            
+#if NETFRAMEWORK
             if (parameterInfos.Length > 0 && parameterInfos[0].ParameterType.FullName == "System.Runtime.CompilerServices.Closure")
             {
-                parameterInfos = parameterInfos.Skip(1).ToArray();
+                var reducedLength = parameterInfos.Length - 1;
+                var reducedParameterInfos = new ParameterInfo[reducedLength];
+                Array.Copy(parameterInfos, 1, reducedParameterInfos, 0, reducedLength);
+                parameterInfos = reducedParameterInfos;
             }
+#endif
 
             int delegateArgumentsCount = parameterInfos.Length;
             int delegateNonParamsArgumentsCount = _delegateContainsParamsArgument ? delegateArgumentsCount - 1 : delegateArgumentsCount;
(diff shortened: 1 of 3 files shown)


Jint/Runtime/Interop/DelegateWrapper.cs
```diff
@@ -37,6 +38,12 @@ public DelegateWrapper(Engine engine, Delegate d)
         public override JsValue Call(JsValue thisObject, JsValue[] jsArguments)
         {
             var parameterInfos = _d.Method.GetParameters();
+
+            if (parameterInfos.Length > 0 && parameterInfos[0].ParameterType.FullName == "System.Runtime.CompilerServices.Closure")
```
sebastienros: Is typeof possible here?

lahma: Have your tried wrapping the code with
```
#if NETFRAMEWORK
... the code ...
#endif
```

Jint/Runtime/Interop/DelegateWrapper.cs
```diff
@@ -37,6 +38,12 @@ public DelegateWrapper(Engine engine, Delegate d)
...
+            {
+                parameterInfos = parameterInfos.Skip(1).ToArray();
```
lahma: Skip is a bit nasty as it uses LINQ and doesn't perform that well, especially on full framework, but this is fortunately quite isolated case.

Jint/Runtime/Interop/DelegateWrapper.cs
kevinarthurackerman: I had only done it on the addition before - I just applied it to both and now it works, but it is basically ignoring the test case, which is probably not desirable. Does it get tested with both flags?

Jint/Runtime/Interop/DelegateWrapper.cs
kevinarthurackerman: I updated the code to more permanently use Array.Copy, it looks like it is about 0.1 ticks faster on average, but is a fair amount more complicated. Alternatively I could use ArraySegment<> (or Span<> in C# 7), but that would change the signature downstream and make the non-NETFRAMEWORK case less performant.

Jint/Runtime/Interop/DelegateWrapper.cs
lahma: Could you please add this line to appveyor.yml which is in the root folder:

```
- dotnet test .\Jint.Tests\Jint.Tests.csproj -c Release -f net452
```

That should include the test in CI builds.

Jint/Runtime/Interop/DelegateWrapper.cs
lahma: Yes it was bit of nitpicking, in Jint codebase there should be as little `using System.Linq;` as possible, which I see you kept 😉 LINQ is much wiser on .NET Core and can be used more safely, but it still adds some penalty - especially on allocation side. So ideally following the Roslyn ideology of not using it at all if possible.

Thanks for considering this too.

Jint/Runtime/Interop/DelegateWrapper.cs
kevinarthurackerman: Done 👍

Jint/Runtime/Interop/DelegateWrapper.cs
kevinarthurackerman: Fixed lol.

I'm used to code bases where an extra object allocation is the least of it's sins 😄

Issue #592, opened 2019-01-29
RegExp.prototype.toString() throws System.NullReferenceException


steffen-liersch:
Hi!

RegExp.prototype.toString() throws a System.NullReferenceException. The issue exists in

- Jint 2.11.10
- Jint 3.0.0-beta-1353
- dev branch

The exception is caused by ToRegExpString. Field Flags is null for the desribed case:

```
        private static JsValue ToRegExpString(JsValue thisObj, JsValue[] arguments)
        {
            var regExp = thisObj.TryCast<RegExpInstance>();

            return "/" + regExp.Source + "/"
                + (regExp.Flags.Contains("g") ? "g" : "")
                + (regExp.Flags.Contains("i") ? "i" : "")
                + (regExp.Flags.Contains("m") ? "m" : "")
                ;
        }
```

Best regards,
Steffen

steffen-liersch:
I've created pull request #593 to the dev branch.

PR #593, opened 2019-01-29
System.NullReferenceException caused by RegExp.prototype.toString() fixed

steffen-liersch:
...see issue #592


sebastienros:
Could you please add a unit test that repros the issue?

PR #595, opened 2019-02-13
Latest Esprima beta changes

AmateurMemetics:
This update is a pre-requisite for adding arrow functions.

Unfortunately as I'm not completely familiar with the code / intentions behind all of the changes in Esprima, I may have gone about this in the wrong way (in particular I do recall reading that `Esprima.Ast.List` is not supposed to be exposed?).

I'd be happy to take any comments on board. Once this is in, the change to add arrow functions is very minor.

### DIFF
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -255,7 +255,7 @@ namespace Jint
             }
 
             ClrTypeConverter = new DefaultTypeConverter(this);
-            BreakPoints = new List<BreakPoint>();
+            BreakPoints = new System.Collections.Generic.List<BreakPoint>();
             DebugHandler = new DebugHandler(this);
         }
 
@@ -301,7 +301,7 @@ namespace Jint
         public event DebugStepDelegate Step;
         public event BreakDelegate Break;
         internal DebugHandler DebugHandler { get; private set; }
-        public List<BreakPoint> BreakPoints { get; private set; }
+        public System.Collections.Generic.List<BreakPoint> BreakPoints { get; private set; }
 
         internal StepMode? InvokeStepEvent(DebugInformation info)
         {
@@ -762,8 +762,8 @@ namespace Jint
         //  http://www.ecma-international.org/ecma-262/5.1/#sec-10.5
         internal bool DeclarationBindingInstantiation(
             DeclarationBindingType declarationBindingType,
-            List<FunctionDeclaration> functionDeclarations,
-            List<VariableDeclaration> variableDeclarations,
+            Esprima.Ast.List<FunctionDeclaration> functionDeclarations,
+            Esprima.Ast.List<VariableDeclaration> variableDeclarations,
             FunctionInstance functionInstance,
             JsValue[] arguments)
         {
@@ -847,7 +847,7 @@ namespace Jint
             return canReleaseArgumentsInstance;
         }
 
-        private void AddFunctionDeclarations(List<FunctionDeclaration> functionDeclarations, EnvironmentRecord env, bool configurableBindings, bool strict)
+        private void AddFunctionDeclarations(Esprima.Ast.List<FunctionDeclaration> functionDeclarations, EnvironmentRecord env, bool configurableBindings, bool strict)
         {
             var functionDeclarationsCount = functionDeclarations.Count;
             for (var i = 0; i < functionDeclarationsCount; i++)
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -429,8 +429,8 @@ namespace Jint
             {
                 DeclarationBindingInstantiation(
                     DeclarationBindingType.GlobalCode,
-                    program.HoistingScope.FunctionDeclarations,
-                    program.HoistingScope.VariableDeclarations,
+                    ref program.HoistingScope.FunctionDeclarations,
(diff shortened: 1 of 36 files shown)


Jint/Engine.cs
```diff
@@ -762,8 +762,8 @@ public JsValue GetValue(JsValue scope, string propertyName)
         //  http://www.ecma-international.org/ecma-262/5.1/#sec-10.5
         internal bool DeclarationBindingInstantiation(
             DeclarationBindingType declarationBindingType,
-            List<FunctionDeclaration> functionDeclarations,
-            List<VariableDeclaration> variableDeclarations,
+            Esprima.Ast.List<FunctionDeclaration> functionDeclarations,
```
lahma: these should be passed as refs as Esprima's list is a struct

Jint/Engine.cs
```diff
@@ -847,7 +847,7 @@ public JsValue GetValue(JsValue scope, string propertyName)
             return canReleaseArgumentsInstance;
         }
 
-        private void AddFunctionDeclarations(List<FunctionDeclaration> functionDeclarations, EnvironmentRecord env, bool configurableBindings, bool strict)
+        private void AddFunctionDeclarations(Esprima.Ast.List<FunctionDeclaration> functionDeclarations, EnvironmentRecord env, bool configurableBindings, bool strict)
```
lahma: this should be passed as ref as Esprima's list is a struct

Jint/Runtime/Environments/DeclarativeEnvironmentRecord.cs
```diff
@@ -384,7 +384,7 @@ internal static JsValue HandleAssignmentPatternIfNeeded(IFunction functionDeclar
             return defaultValue;
         }
 
-        internal void AddVariableDeclarations(List<VariableDeclaration> variableDeclarations)
+        internal void AddVariableDeclarations(Esprima.Ast.List<VariableDeclaration> variableDeclarations)
```
lahma: this should be passed as ref as Esprima's list is a struct

Jint/Runtime/Interpreter/JintStatementList.cs
```diff
@@ -15,12 +15,12 @@ private class Pair
...
         private readonly Engine _engine;
         private readonly Statement _statement;
-        private readonly List<StatementListItem> _statements;
+        private readonly Esprima.Ast.List<StatementListItem> _statements;
 
         private Pair[] _jintStatements;
         private bool _initialized;
 
-        public JintStatementList(Engine engine, Statement statement, List<StatementListItem> statements)
+        public JintStatementList(Engine engine, Statement statement, Esprima.Ast.List<StatementListItem> statements)
```
lahma: statements probably should be a ref param

Jint/Runtime/Interpreter/Statements/JintSwitchBlock.cs
```diff
@@ -8,11 +9,11 @@ namespace Jint.Runtime.Interpreter.Statements
     internal sealed class JintSwitchBlock
     {
         private readonly Engine _engine;
-        private readonly List<SwitchCase> _switchBlock;
+        private readonly Esprima.Ast.List<SwitchCase> _switchBlock;
         private JintSwitchCase[] _jintSwitchBlock;
         private bool _initialized;
 
-        public JintSwitchBlock(Engine engine, List<SwitchCase> switchBlock)
+        public JintSwitchBlock(Engine engine, Esprima.Ast.List<SwitchCase> switchBlock)
```
lahma: switchBlock probably should be a ref param

Happypig375:
I think using the `in` modifier instead of `ref` would be better since it guarantees the list cannot be reassigned.

lahma:
But as Esprima list is not read-only I believe a defensive copy would be made.

AmateurMemetics:
I agree with @lahma. I've just been reading this blog post to get a better grasp on the `in` modifier in C#: https://blogs.msdn.microsoft.com/seteplia/2018/03/07/the-in-modifier-and-the-readonly-structs-in-c/ - which concludes that a defensive copy will be made for structs which are not read-only. 

I'll go ahead and update this PR with `ref` keywords.

Jint/Runtime/Interpreter/Statements/JintSwitchBlock.cs
```diff
@@ -8,11 +9,11 @@ namespace Jint.Runtime.Interpreter.Statements
     internal sealed class JintSwitchBlock
     {
         private readonly Engine _engine;
-        private readonly List<SwitchCase> _switchBlock;
+        private readonly Esprima.Ast.List<SwitchCase> _switchBlock;
         private JintSwitchCase[] _jintSwitchBlock;
         private bool _initialized;
 
-        public JintSwitchBlock(Engine engine, List<SwitchCase> switchBlock)
+        public JintSwitchBlock(Engine engine, in Esprima.Ast.List<SwitchCase> switchBlock)
```
AmateurMemetics: Used an `in` modifier here because `_switchBlock` is `readonly`

AmateurMemetics:
Is there anything stopping this from being merged in?

lahma:
Sorry to keep you waiting. @sebastienros makes the final call and I believe he's been a bit under the weather, hopefully back soon to review.

Issue #111, opened 2014-11-07
Consecutive calls of Math.random() always return same value 


seedsnj:
E.g. Calling  the custom random()  function that concats 4 values returned by Math.random()

```
function random() {
return Math.random() + "-" + Math.random() + "-" + Math.random() + "-" + Math.random();
};
```

returns 
0.2160154353901816-0.2160154353901816-0.2160154353901816-0.2160154353901816

Note, in this sample,  the sub-strings separated by ‘-‘ are  same. 

In code,  MathInstance.cs appears to construct a new .NET Random object without using a different seed,  thus the random sequence will be determined by the default seed, which is the current system time. Refer to the MSDN page http://msdn.microsoft.com/en-us/library/system.random%28v=vs.110%29.aspx

```
private static JsValue Random(JsValue thisObject, JsValue[] arguments)
        {
            return new Random().NextDouble();
        }
```

In the short sample code,  it is highly possible that Jint executes the 4 Math.random()  calls in the same milliseconds interval, and returns same random value.  

 This problem is described in  the Stack Overflow page http://stackoverflow.com/questions/4855756/random-number-generation-same-number-returned 

> It happens because the Random object is initialized from Environment.GetTickCount, which is a millisecond timer. So if you call the Random constructor twice within the same millisecond, you're going to get the same initial value.

Version: Jint 2.4.0
Potential fix:   use  a static .NET Random object in MathInstance.cs
Severity: high

PR #596, opened 2019-02-20
make Random() thread safety

sxul:
System.Random isn't thread-safe. When multi-threaded sharing instance, Random.Next() will always returns zero.

### DIFF
--- a/Jint/Native/Math/MathInstance.cs
+++ b/Jint/Native/Math/MathInstance.cs
@@ -9,7 +9,7 @@ namespace Jint.Native.Math
 {
     public sealed class MathInstance : ObjectInstance
     {
-        private static readonly Random _random = new Random();
+        [ThreadStatic] private static Random _random;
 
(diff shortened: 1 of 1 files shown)


lahma:
Would it make more sense to just make the field non-static? It's constructed for each Engine only once anyway.

sxul:
> 
> 
> Would it make more sense to just make the field non-static? It's constructed for each Engine only once anyway.

#111

lahma:
Instance field would be reused and would return different values on each invocation?

sxul:
[https://docs.microsoft.com/en-us/dotnet/api/system.threadstaticattribute?view=netframework-4.7.2](ThreadStaticAttribute)

> Indicates that the value of a **static field is unique for each thread**.

lahma:
I understand how the attribute works, it just has little meaning when Jint's Engine itself is not thread safe.

PR #597, opened 2019-02-23
added full support for interop with methods that have default values

boennhoff:
Hi jint devs,
as mentioned on gitter, I was looking for a way to interop with C# methods that make use of default values.  After some hints from @sebastienros, I implemented this behaviour myself: Default Values should now be fully supported, including a unit test.

Even if it works, I am not totally happy, especially with the `JsValue.FromObject()` part. What is the best way to convert/cast the default value? Or is this maybe not the best place for the code?

### DIFF
--- a/Jint/Runtime/TypeConverter.cs
+++ b/Jint/Runtime/TypeConverter.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Globalization;
-using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using Esprima.Ast;
@@ -495,9 +494,7 @@ namespace Jint.Runtime
             }
         }
 
-        private static readonly Engine _engine = new Engine();
-
-        public static IEnumerable<Tuple<MethodBase, JsValue[]>> FindBestMatch<T>(T[] methods, Func<T, bool, JsValue[]> argumentProvider) where T : MethodBase
+        public static IEnumerable<Tuple<MethodBase, JsValue[]>> FindBestMatch<T>(Engine engine, T[] methods, Func<T, bool, JsValue[]> argumentProvider) where T : MethodBase
         {
             System.Collections.Generic.List<Tuple<T, JsValue[]>> matchingByParameterCount = null;
             foreach (var m in methods)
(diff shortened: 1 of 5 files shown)


Jint/Runtime/TypeConverter.cs
```diff
@@ -494,6 +495,8 @@ public static void CheckObjectCoercible(Engine engine, JsValue o)
             }
         }
 
+        private static readonly Engine _engine = new Engine();
```
lahma: you should add new parameter `Engine engine` and pass `_engine`from the current call sites. It should be the same engine that is processing request.

Jint/Runtime/TypeConverter.cs
```diff
@@ -522,6 +525,25 @@ public static void CheckObjectCoercible(Engine engine, JsValue o)
                     matchingByParameterCount = matchingByParameterCount ?? new List<Tuple<T, JsValue[]>>();
                     matchingByParameterCount.Add(new Tuple<T, JsValue[]>(m, arguments));
                 }
+                else if (parameterInfos.Length > arguments.Length)
+                {
+                    // check if we got enough default values to provide all parameters (or more in case some default values are provided/overwritten)
+                    var defaultValuesCount = parameterInfos.Count(x => x.HasDefaultValue);
```
lahma: We try to avoid usage of LINQ, could you change this to old fashioned `foreach` that increments counter, so we don't have single LINQ usage for just this.

Jint/Runtime/TypeConverter.cs
```diff
@@ -522,6 +525,25 @@ public static void CheckObjectCoercible(Engine engine, JsValue o)
...
+                        // create missing arguments from default values
+                        var argsWithDefaults = new List<JsValue>(arguments);
+                        for (var i = arguments.Length; i < parameterInfos.Length; i++)
+                        {
+                            var param = parameterInfos[i];
+                            var value = JsValue.FromObject(_engine, param.DefaultValue);
+                            argsWithDefaults.Add(value);
+                        }
+
+                        matchingByParameterCount = matchingByParameterCount ?? new List<Tuple<T, JsValue[]>>();
```
lahma: this needs to be `new System.Collections.Generic.List<Tuple<T, JsValue[]>>();` as it now breaks compilation due to Esprima having also the List<T>

boennhoff: Added full qualified namespace to my List call, but not to the lower one, as it is a plain copy of code a few lines above. Since I can't reproduce the issue of not compiling, I hope that fixes it.

Jint/Runtime/TypeConverter.cs
boennhoff: Done. But would you elaborate on the non-LINQ usage? Performance issues?

Jint/Runtime/TypeConverter.cs
boennhoff: Too easy, done :-)

boennhoff:
hmmm, never did the github review process, what do I need to trigger now?

lahma:
The build still fails, you could try to do a rebase, I believe Appveyor does merge from dev, which explains the build failure.

Jint/Runtime/TypeConverter.cs
lahma: Yes, LINQ is always slower and we try to avoid it in Jint.

Jint/Runtime/TypeConverter.cs
boennhoff: After rebasing the describes problem was obvious...

boennhoff:
sooo, push forcing has a reason... :-)

boennhoff:
Thanks for the easy introduction to the fork workflow on github.

Don't you have complains about the object conversion part? If the code would sit later in the processing, we could avoid the arguments/parameter conversion at all. But I am not yet able to say something about the impact on performance, or if there even is a better place for it.

lahma:
@sebastienros is this good to go?

PR #601, opened 2019-02-25
Implement arrow functions

AmateurMemetics:
This PR fully implements arrow functions.

As part of getting the test suite to pass, I've done a little playing around with where we call `SetFunctionName`.

It boils down to this test:

```
const { cover = (function () {}), xCover = (0, function() {})  } = {};
assert.sameValue(cover.name, 'cover');
assert.notSameValue(xCover.name, 'xCover');
```
And a similar one with array binding patterns.

I'd consider these relevant changes to the PR as this area did need modifying to work with `JintArrowFunctionExpressions` anyway.

### DIFF
--- a/Jint.Benchmark/Jint.Benchmark.csproj
+++ b/Jint.Benchmark/Jint.Benchmark.csproj
@@ -22,6 +22,7 @@
     <PackageReference Include="BenchmarkDotNet" Version="0.11.2" />
     <ProjectReference Include="..\Jint\Jint.csproj" />
     <PackageReference Include="Jurassic" Version="3.0.0-alpha2" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.1" />
     <PackageReference Include="NiL.JS.NetCore" Version="2.5.1200" />
   </ItemGroup>
 </Project>
--- a/Jint.Tests.Ecma/TestCases/alltests.json
+++ b/Jint.Tests.Ecma/TestCases/alltests.json
@@ -16825,8 +16825,8 @@
     source: "ch13/13.2/13.2-14-s.js"
   },
   {
-    skip: false,
-    reason: "",
+    skip: true,
+    reason: "function length is configurable",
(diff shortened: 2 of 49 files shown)


lahma:
I hope I'm not opening Pandora's box here, but have you tried what happens if you add the more extensive arrow function test suite? Test case below should enable them, they are not run at the moment.

```
using Xunit;

namespace Jint.Tests.Test262
{
    public class ExpressionTests : Test262Test
    {
        [Theory(DisplayName = "language\\expressions\\arrow-function")]
        [MemberData(nameof(SourceFiles), "language\\expressions\\arrow-function", false)]
        [MemberData(nameof(SourceFiles), "language\\expressions\\arrow-function", true, Skip = "Skipped")]
        protected void ArrowFunction(SourceFile sourceFile)
        {
            RunTestInternal(sourceFile);
        }
    }
}
```

These usually bring out the corner cases of spec 🙂

AmateurMemetics:
Hadn't spotted that. There _are_ a few (163) failures, I'll get back to you once I've taken a look through them & spotted any immediate bugs.

AmateurMemetics:
Updated with my latest work. This is now a WIP branch that I expect will need some help to get 100% of the tests passing.

lahma:
@AmateurMemetics @sebastienros almost there, only two test cases failing, both share the same reason. I'm running out of time for now so if either one of you could help and check with fresh eyes. Just can't figure out where the last scope isolation belongs.

*language/expressions/arrow-function/scope-paramsbody-var-open.js*

```
var x = 'outside';
var probeParams, probeBody;

(function(_ = probeParams = function() { return x; }) {
  var x = 'inside';
  probeBody = function() { return x; };
}());

assert.sameValue(probeParams(), 'outside');
assert.sameValue(probeBody(), 'inside');
```

```
Assert.Null() Failure
Expected: (null)
Actual:   Test262Error: Expected SameValue(«inside», «outside») to be true
```

kevinroast:
whoop whoop! this is great work guys!

PR #606, opened 2019-03-05
Don't allow changes to shared Undefined property descriptor

### LINKED ISSUE #594: Unable to define object properties by assignment
I'm having trouble getting some object properties created by assignment, looking for some ideas on what's going on.  This is an example of what's happening in my code, (I don't have a standalone test case yet).  When the following code is run in Jint right after engine creation, it works correctly.  However, when I run the same code later -- in response to some user input -- it fails to define the properties unless I use Object.defineProperty().  Note it doesn't throw any errors, it just silently fails to define the properties.

```
var obj = {};
// obj = {}

obj.testVar1 = 1;
// obj = {}

obj["testVar2"] = 2;
// obj = {}

Object.defineProperty(obj, "testVar3", { value: 3 });
// obj = { testVar: 3 }
```

Maybe I'm missing something, but shouldn't defining object properties by assignment always work?  Or is there something in ES5 that makes it only _sometimes_ work?

Thanks for any help, this is boggling me.

lahma:
This is basically same as changing runtime's string.Empty, hilarity ensues.

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -1009,6 +1009,16 @@ namespace Jint.Tests.Runtime
             Assert.Throws<ArgumentException>(() => _engine.Invoke(foo, obj, new object[] { }));
         }
 
+        [Fact]
+        public void ShouldNotAllowModifyingSharedUndefinedDescriptor()
+        {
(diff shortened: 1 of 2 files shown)


sebastienros:
Fixes #594

PR #607, opened 2019-03-05
Unsealed engine for customized extensions

steffen-liersch:


steffen-liersch:
This change is included in PR #612 .

PR #613, opened 2019-03-14
Add 'load(...)' for REPL

atifaziz:
This PR adds and enables loading scripts in the REPL app via a new `load` function that takes a single argument that's the path of the script file to load and execute.


atifaziz:
Merge? :wave:

PR #616, opened 2019-03-18
Added support for invoking Extension Methods

boennhoff:
Hello again,
I took over PR #381, or rather continued it in my own branch. Most code still goes back to @windischb, I just made it work against the current dev branch.

BUT, there is still the issue, that it won't work on "strings" directly. So the support ist not yet complete (one unit test fails), but it's already usable.

Whats your thoughts about this and code quality in general? Don't look too much into the several commits, I might squash them...


boennhoff:
Well, thankfully I improved the tests and recognized that actually the extension methods are matching without filtering for the right method name. m(
Any hints?

I commented the StringExtension UnitTest, because I would like to look at it as an advanced feature. So for now I would concentrate on Extension Methods on objects. Any doubts?

boennhoff:
I fixed the problem with non-matching method names. It should now works against object extensions.

boennhoff:
Is there really no interest for this?

sebastienros:
Any update on what doesn't work with it? (string for instance).

boennhoff:
Sorry, for taking so long for an answer. Actually the question is easier answered the other way round: Extension Methods are fully supported when called on _objects_, when called on _primitives_ it doesn't work, like `"string".SomeExtensionMethod()`. But `var x = "string"; x.SomeExtensionMethod();` works.

Within the original PR #381 you already complained about the way @windischb implemented it for strings, see [here](https://github.com/sebastienros/jint/pull/381#discussion_r114577683). So I also doN#t know where to start doing the implementation for primitives.

But i personally came to the conclusion, that for the moment it is better to just have that feature, then no support for ExtensionMethods at all. Most of the times they are called on objects either way, right?

A follow up PR or issue could be created, I just need an advise on where to implement and I could have a look. If interest is available i also fix this branch again.

windischb:
Any news on that?
Would be great if this feature can be integrated in the near future...

@boennhoff 
thanks for take over and try to get this implemented correctly!

lahma:
@boennhoff @windischb is it OK if I find the time that I'll squash this to single commit and rebase against current dev? I think this is quite close and currently it's quite PITA to get this rebased the keeping commits as there's some back and forth movement...

windischb:
For me, it's totally ok!
But @sebastienros  mentioned on my Pull Request that the string prototype object shouldn't be changed.
So i'm not sure anymore if this is a good way to implement that feature.
I have created a fork which don't change the string prototype and seems to be a cleaner way, at least from my perspective.
But i dont want to spam you with another pull request regarding this topic!
Maybe it's worth to take a look on that approach if it fits better...?

['https://github.com/sebastienros/jint/compare/dev...windischb:ExtensionMethods'](https://github.com/sebastienros/jint/compare/dev...windischb:ExtensionMethods)

lahma:
OK so we have already three takes on this 🙂 I'll try to stitch these together and yes, we probably shouldn't touch the actual ECMA constructs but find ways to override and and extend.

boennhoff:
I am fine as well @lahma. I really would appreciate getting this feature upstream, so is this the plan after rebasing?

lahma:
Yes I'd like very much to get this finalized as it's been pending such a long time, Sébastien makes the final call.

PR #625, opened 2019-05-17
fixes #615 improve engine creation performance

### LINKED ISSUE #615: Improve engine creation performance
While doing some memory profiling I found that the main cost was during the initialization of the constructor objects when we create many `ClrFunctionInstance` and `PropertyDescriptor` instances like here:

https://github.com/sebastienros/jint/blob/dev/Jint/Native/Date/DateConstructor.cs#L77

I assume these objects should be immutable (they can still be replaced) and we should be able to mark them static and reuse them.

Snapshot of 50 requests in Orchard Core:

image

lahma:
* we already have lazy initialization support for arguments instance, so I derived a solution where objects are initialized lazily
* first access to features initializes objects, so if you don't use Map, it won't allocate or initialize
* no performance impact on earlier benchmarks

## Before

|      Method |     Mean |     Error |    StdDev |   Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------ |---------:|----------:|----------:|--------:|-------:|------:|----------:|
| EngineBenchmark | 56.49 us | 0.0473 us | 0.0442 us | 26.9775 | 0.1221 |     - | 110.51 KB |
| EvaluationBenchmark  | 791.4 us | 0.7033 us | 0.6578 us | 138.6719 |     - |     - | 570.16 KB |
| UncacheableExpressionsBenchmark|  225.8 ms | 0.1986 ms | 0.1858 ms | 30333.3333 | 333.3333 |     - | 122.57 MB |

## After

* memory allocations reduced 92%
* minimal initialization and simple script run reduced 95%

|      Method |     Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------ |---------:|----------:|----------:|-------:|------:|------:|----------:|
| EngineBenchmark | 2.745 us | 0.0093 us | 0.0078 us | 2.1248 |     - |     - |   8.72 KB |
|   EvaluationBenchmark| 796.4 us | 1.711 us | 1.601 us | 138.6719 |     - |     - | 570.16 KB |
| UncacheableExpressionsBenchmark|  223.4 ms | 0.1680 ms | 0.1571 ms | 30333.3333 | 333.3333 |     - | 122.57 MB |

### DIFF
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -173,50 +173,6 @@ namespace Jint
             // Because the properties might need some of the built-in object
             // their configuration is delayed to a later step
 
-            Global.Configure();
-
-            Object.Configure();
-            Object.PrototypeObject.Configure();
-
-            Symbol.Configure();
-            Symbol.PrototypeObject.Configure();
-
-            Function.Configure();
-            Function.PrototypeObject.Configure();
-
-            Array.Configure();
-            Array.PrototypeObject.Configure();
-
-            Map.Configure();
-            Map.PrototypeObject.Configure();
-
-            Set.Configure();
-            Set.PrototypeObject.Configure();
-
-            Iterator.Configure();
-            Iterator.PrototypeObject.Configure();
-
-            String.Configure();
-            String.PrototypeObject.Configure();
-
-            RegExp.Configure();
-            RegExp.PrototypeObject.Configure();
-
-            Number.Configure();
-            Number.PrototypeObject.Configure();
-
-            Boolean.Configure();
-            Boolean.PrototypeObject.Configure();
-
-            Date.Configure();
-            Date.PrototypeObject.Configure();
-
-            Math.Configure();
-            Json.Configure();
-
-            Error.Configure();
-            Error.PrototypeObject.Configure();
-
             // create the global environment http://www.ecma-international.org/ecma-262/5.1/#sec-10.2.3
             GlobalEnvironment = LexicalEnvironment.NewObjectEnvironment(this, Global, null, false);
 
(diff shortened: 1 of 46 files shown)


sebastienros:
I think it should be a huge gain for ravendb also. Will benchmark it on Orchard tomorrow, because we don't even use all these objects.

lahma:
@sebastienros did you ever find the time to compare performance in Orchard?

sebastienros:
I remember I checked and it was faster but don't recall the numbers.

lahma:
Great, thanks for the update 🙂

sebastienros:
if you like perf, there is so much to do in Orchard. It's allocating like crazy, which is not surprising because it does stuff ;) but which also means there are many ways to improve it.

lahma:
Maybe at some point, probably requires quite a lot studying to understand hot paths and important scenarios...

Jint/Engine.cs
```diff
@@ -59,32 +80,35 @@ public sealed class Engine
         internal readonly ArgumentsInstancePool _argumentsInstancePool;
         internal readonly JsValueArrayPool _jsValueArrayPool;
 
-        public ITypeConverter ClrTypeConverter;
+        public readonly ITypeConverter ClrTypeConverter;
```
ViceIce: Why was this done? It is now no longer possible to override with a custom `ITypeConverter`

lahma: Just an oversight. Maybe submit a PR with a test case that protects from happening again?

ViceIce: See #654  for pr

PR #631, opened 2019-06-22
Number.toPrecision throws between 26 & 100

### LINKED ISSUE #599: Number.toPrecision throws between 26 & 100
The buffer in dtoa is only ever allocated 25 digits, so `(3).toPrecision(n)` where `n` is between 26 and 100 throws an index out of bounds exception.

This can be fixed be reallocating the buffer in `DtoaNumberFormatter` if the fast case doesn't succeed (L65). However, this seems like a pretty complex rabbit hole and I'd appreciate more eyes on this.

I'm planning on tackling this after I merge in arrow functions (which will enable tests that fail due to this issue).

lahma:
fixes #599

### DIFF
--- a/Jint.Tests/Runtime/NumberTests.cs
+++ b/Jint.Tests/Runtime/NumberTests.cs
@@ -1,0 +1,61 @@
+using System;
+using Jint.Runtime;
+using Xunit;
+
+namespace Jint.Tests.Runtime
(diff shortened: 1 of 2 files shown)


lahma:
Merged as trivial fix, has test coverage and follows V8 logic.

Issue #635, opened 2019-07-18
[Bug] function immidiately called


MeikelLP:
I want to implement a (for now) basic modding interface for my Unity3D game.

I want to define a mod in `module.exports` styled manner:

```js
module.exports = {
  name: 'Mod1',
  init () {
    registry.Add(new BlockPrototype('Dirt', 'assets/dirt.png'))
    registry.Add(new BlockPrototype('Grass', 'assets/grass.png'))
  }
}
```

And then call the individual phase while mod-loading (i.e. `init`). But sadly the function `init` is instantly called when I `Engine.Execute` this script block. This should not happen as I only define this function.

I tried to workaround this issue with defining the function separately as named `function init () {}` and then assign it but that didn't work as well.

lahma:
What verion of Jint? Running such script with latest preview version doesn't give an error even though `registry` is not defined.

MeikelLP:
I'm not at work right now so I can't say for sure. I directly downloaded from nuget today. Should be `3.0.0-beta-1525`. I'm adding more source code (especially C#) tomorrow.

MeikelLP:
I cannot reproduce this bug anymore. I don't know how it fixed itself :/

PR #637, opened 2019-07-24
[WIP] Added support for anonymous objects.

MeikelLP:
As I mentioned in https://github.com/sebastienros/jint/issues/635#issuecomment-512963364 I cannot assign an anonymous object `{}` to a typed object. This can be very annoying if you want to have 'friendly scripting support' and its more idiomatic in JS as well.

You can now assign an anonymous object to a typed property/field of a given object:

```cs
public void ShouldParseAnonymousToTypeObject(){
      var obj = new Wrapper();
      var engine = new Engine().SetValue("x", obj);
}

private class Wrapper
{
     public Testificate Test { get; set; }
}

private class Testificate
{
     public Func<int, int, int> Init { get; set; }
}
```

```js
x.test = {
    name: 'Testificate',
    init (a, b) {
        return a + b
    }
}
```

See `EngineTests.ShouldParseAnonymousToTypeObject` for the full test.

* [x] <del>WIP as I don't know how I shall handle namings in JS <-> C#. Currently I'm doing only `ToLower` because it's easy for this test. Because I use the `ExpandoObject` as a `IDictionary<string, object>` the names are case sensitive. Shall I implement case-insensitivity for this or are there other conventions?</del> -> see `UpperToLowerCamelCase`
* [x] <del>Is LINQ fine? Because of the GC allocations.</del> -> No
* [ ] I noticed that I have to use an explicit wrapper to assign a field/property. I cannot simply overwrite `x` with my `Testificate` object. In a wrapper (property/field assignment) it works fine. Is this by design or a bug?
* [ ] As this PR is approved I will add more tests. Just a proof of concept for now. Maybe it will be denied anyway. It works for my use case so why not share it?

### DIFF
--- a/Jint/Extensions/JavascriptExtensions.cs
+++ b/Jint/Extensions/JavascriptExtensions.cs
@@ -6,10 +6,9 @@ namespace Jint.Extensions
     {
         internal static string UpperToLowerCamelCase(this string str)
         {
-            var sb = new StringBuilder();
-            sb.Append(char.ToLowerInvariant(str[0]));
-            sb.Append(str.Substring(1));
-            return sb.ToString();
+            var arr = str.ToCharArray();
+            arr[0] = char.ToLowerInvariant(arr[0]);
+            return new string(arr);
         }
     }
 }
--- a/Jint/Extensions/JavascriptExtensions.cs
+++ b/Jint/Extensions/JavascriptExtensions.cs
@@ -2,9 +2,9 @@ using System.Text;
 
 namespace Jint.Extensions
 {
-    public static class JavascriptExtensions
+    internal static class JavascriptExtensions
     {
-        public static string UpperToLowerCamelCase(this string str)
+        internal static string UpperToLowerCamelCase(this string str)
         {
             var sb = new StringBuilder();
             sb.Append(char.ToLowerInvariant(str[0]));
--- a/Jint/Extensions/JavascriptExtensions.cs
+++ b/Jint/Extensions/JavascriptExtensions.cs
@@ -1,0 +1,15 @@
+using System.Text;
+
+namespace Jint.Extensions
+{
+    public static class JavascriptExtensions
+    {
+        public static string UpperToLowerCamelCase(this string str)
+        {
+            var sb = new StringBuilder();
+            sb.Append(char.ToLowerInvariant(str[0]));
+            sb.Append(str.Substring(1));
+            return sb.ToString();
+        }
+    }
+}
--- a/Jint/Extensions/ReflectionExtensions.cs
+++ b/Jint/Extensions/ReflectionExtensions.cs
@@ -1,0 +1,43 @@
+using System;
+using System.Reflection;
+
+namespace Jint.Extensions
+{
+    public static class ReflectionExtensions
+    {
+        public static void SetValue(this MemberInfo memberInfo, object forObject, object value)
+        {
+            switch (memberInfo.MemberType)
+            {
+                case MemberTypes.Field:
+                    var fieldInfo = (FieldInfo) memberInfo;
+                    if (fieldInfo.FieldType == value?.GetType())
+                    {
+                        fieldInfo.SetValue(forObject, value);
+                    }
+
+                    break;
+                case MemberTypes.Property:
+                    var propertyInfo = (PropertyInfo) memberInfo;
+                    if (propertyInfo.PropertyType == value?.GetType())
+                    {
+                        propertyInfo.SetValue(forObject, value);
+                    }
+                    break;
+            }
+        }
+
+        public static Type GetDefinedType(this MemberInfo memberInfo)
+        {
+            switch (memberInfo)
+            {
+                case PropertyInfo propertyInfo:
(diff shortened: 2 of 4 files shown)


Jint/Extensions/JavascriptExtensions.cs
```diff
@@ -0,0 +1,15 @@
+using System.Text;
+
+namespace Jint.Extensions
+{
+    public static class JavascriptExtensions
```
lahma: such helpers should be internal do decrease API surface

Jint/Extensions/JavascriptExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        public static string UpperToLowerCamelCase(this string str)
```
lahma: is this actually something like `ToCamelCase`?

Jint/Extensions/ReflectionExtensions.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+using System.Reflection;
+
+namespace Jint.Extensions
+{
+    public static class ReflectionExtensions
```
lahma: internal again

Jint/Runtime/Interop/DefaultTypeConverter.cs
```diff
@@ -1,8 +1,12 @@
 ﻿using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Dynamic;
+using System.Linq;
```
lahma: I dislike adding LINQ as it's always slower than using types directly, but could be tweaked later on, seems that there are no captured closures though which is good

Jint/Runtime/Interop/DefaultTypeConverter.cs
```diff
@@ -200,6 +204,29 @@ public virtual object Convert(object value, Type type, IFormatProvider formatPro
                 type = Nullable.GetUnderlyingType(type);
             }
 
+            if (value is ExpandoObject eObj)
+            {
+                // public empty constructor required
+                if (type.IsValueType && type.GetConstructors().Length > 0 ||
```
lahma: convention is to always have braces

Jint/Runtime/Interop/DefaultTypeConverter.cs
```diff
@@ -200,6 +204,29 @@ public virtual object Convert(object value, Type type, IFormatProvider formatPro
...
+                    !type.IsValueType && !type.GetConstructors().Any(x => x.GetParameters().Length == 0 && x.IsPublic)) return null;
+
+                var dict = (IDictionary<string, object>) eObj;
+                var obj = Activator.CreateInstance(type, new object[0]);
```
lahma: `ArrayExt.Empty<object>()` is better than `new object[]`

Jint/Runtime/Interop/DefaultTypeConverter.cs
```diff
@@ -200,6 +204,29 @@ public virtual object Convert(object value, Type type, IFormatProvider formatPro
...
+
+                var members = type.GetMembers().Where(x => x.MemberType == MemberTypes.Property || x.MemberType == MemberTypes.Field).ToArray();
```
lahma: no need to materialize a new array with `ToArray()`

Jint/Extensions/JavascriptExtensions.cs
MeikelLP: Yes. I explicitly named it like that to not cause confusion on when to call it.
I can rename it to `PascalCaseToCamelCase` or simply `ToCamelCase` if wanted.

Jint/Extensions/JavascriptExtensions.cs
```diff
@@ -0,0 +1,15 @@
+using System.Text;
+
+namespace Jint.Extensions
+{
+    internal static class JavascriptExtensions
+    {
+        internal static string UpperToLowerCamelCase(this string str)
+        {
+            var sb = new StringBuilder();
```
sebastienros: Why not just clone the string and update the first char? That would prevent from building a new StringBuilder on every call too.

lahma: True, when only simple concatenation of couple of strings happens, StringBuilder can be an overkill 🙂

MeikelLP: I updated my code. Is this the correct approach?

lahma: Shortest version below, I guess they don't differ much functionality/performance-wise.

```
internal static string UpperToLowerCamelCase2(this string str)
{
    return char.ToLowerInvariant(str[0]) + str.Substring(1);
}
```

MeikelLP: I think I allocate 1 string less than you. ToCharArray probably creates a copy too so we both allocate 2 objects.

Too bad we don't use .NET Core 3 D:

MeikelLP:
Thank you. I think this will help many people to write idiomatic Javascript code.

PR #641, opened 2019-07-26
Fixed: Calling Array.prototype.toString() with anything other than a …

miroslav22:
…real Array object throws a NullReferenceException

Another obscure one which prevents a polyfill library loading.

If you call Array.prototype.toString() with anything other than a real Array object then you get a NullReferenceException.

Changed method behaviour to match v8 engine and fall back to call base Object.prototype.toString() if passed something other than an array object


sebastienros:
Weird that it's not covered by the existing tests.

PR #654, opened 2019-08-29
allow overriding default clr type converter

ViceIce:
This pr fixes a regression of #625  which has made the clr type converter readonly.

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -2663,8 +2663,8 @@ function output(x) {
 };
 ";
             _engine.Execute(program);
-            var result1 = (ObjectInstance) _engine.Execute("output(x1)").GetCompletionValue();
-            var result2 = (ObjectInstance) _engine.Execute("output(x2)").GetCompletionValue();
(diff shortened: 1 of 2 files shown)


ViceIce:
My use case:

```cs
using System;
using Jint;
using Jint.Runtime.Interop;

namespace Scripting.Jint.Wrapper
{
    internal sealed class JintTypeConverter : DefaultTypeConverter
    {
        public JintTypeConverter(Engine engine) : base(engine)
        {
        }

        public override object Convert(object value, Type type, IFormatProvider formatProvider)
        {
            switch (value)
            {
                case string s when type == typeof(Guid) && Guid.TryParse(s, out var g):
                    return g;
            }
            return base.Convert(value, type, formatProvider);
        }
    }
}
```

lahma:
Thank you for fixing and creating the test case to protect from such a regression in future.

sebastienros:
Good job!

ViceIce:
@sebastienros @lahma Any chances to get a new beta release?

lahma:
Sorry for the delay, there was a release 5 days ago.

ViceIce:
I know, thanks 👍

PR #660, opened 2019-09-03
Fix array indexing with empty string

### LINKED ISSUE #659: IndexOutOfRangeException occurs when try to acces the array item by empty string
Run the following code:
```c#
var engine = new Engine();
engine.Execute("var x = [];x[\"\"]=8;");
```
Observed: Unhandled Exception: System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Jint.Native.Array.ArrayInstance.GetOwnProperty(Key& propertyName) in C:\projects\jint\Jint\Native\Array\ArrayInstance.cs:line 356

Expected:
The code should be executed without errors. And the 'x[""]' must return 8.

lahma:
fixes #659

### DIFF
--- a/Jint/Native/Array/ArrayInstance.cs
+++ b/Jint/Native/Array/ArrayInstance.cs
@@ -462,6 +462,11 @@ namespace Jint.Native.Array
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private static uint ParseArrayIndex(string p)
         {
+            if (p.Length == 0)
+            {
+                return uint.MaxValue;
+            }
+
             int d = p[0] - '0';
 
             if (d < 0 || d > 9)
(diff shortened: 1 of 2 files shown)


Jint/Native/Array/ArrayInstance.cs
```diff
@@ -462,6 +462,11 @@ private static bool IsArrayIndex(string p, out uint index)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private static uint ParseArrayIndex(string p)
         {
+            if (p.Length == 0)
+            {
+                return uint.MaxValue;
+            }
+
             int d = p[0] - '0';
```
RusKnyaz: I'm sorry to interfere, but why not use char.IsDigit(p[0]) here? I checked the performance of both approaches - they are identical.

lahma: Well we need the actual value, say 5 to return and then when we detect length being more than 1 when start addition to that value.

PR #669, opened 2019-10-05
Switch to Visual Studio 2019 and .NET Core 3 test runner

lahma:
Running on .NET Core 3 also revealed and issue with date handling. When doubles were used to represent hours, minutes etc it could cause formatting to produce -0 for zero values (format string {0:00}). Fixed this by using int where logical, this has added benefit of guiding JsNumber creation to go to fast path of Create(int) which can do fast array lookups.


lahma:
I think there's now enough band-aid on date, but in the long run it would make sense to model it with integer value always. Way faster to do calculations.

lahma:
@sebastienros thoughts?

PR #678, opened 2019-10-26
Implementing immutable bindings

sebastienros:


sebastienros:
The next step is to replace any DeclarationBindingInstantiation call (http://www.ecma-international.org/ecma-262/5.1/#sec-10.5) by all the different Declaration Instantiations  in es6:
- BlockDeclarationInstantiation (http://www.ecma-international.org/ecma-262/6.0/#sec-blockdeclarationinstantiation)
- GlobalDeclarationInstantiation (http://www.ecma-international.org/ecma-262/6.0/#sec-globaldeclarationinstantiation)
- ModuleDeclarationInstantiation (http://www.ecma-international.org/ecma-262/6.0/#sec-moduledeclarationinstantiation)
- EvalDeclarationInstantiation (http://www.ecma-international.org/ecma-262/6.0/#sec-evaldeclarationinstantiation)

Jint/Runtime/Environments/DeclarativeEnvironmentRecord.cs
```diff
@@ -41,7 +41,7 @@ internal void Reset()
             _argumentsBindingWasAccessed = false;
         }
 
-        private void SetItem(in Key key, in Binding value)
+        private ref Binding SetItem(in Key key, in Binding value)
```
lahma: Not part of this PR scope but it seems that `in` is being used for non-readonly struct. Generally performance-wise we would want `in readonly struct` or `ref struct`. Good to see PRs as it's easy to get blinded by all code otherwise.

sebastienros:
My problem right now is that there are too many things that take shortcuts from the spec, and when I try to update something then something else doesn't work anymore. 

For instance there are some SetItem, SetItemSafely that don't exist in the spec, or GetExistingItem that will create an uninitialized binding, where it shouldn't in es6.

lahma:
Maybe revert to closer to original version and we can reintroduce optimizations back if possible later on?

PR #679, opened 2019-10-27
Share common base class for exceptions

### LINKED ISSUE #676: Jint-defined exceptions should all derive from a Jint-specific Exception class
Consider having an (abstract) Exception subclass, e.g. `JintException`, that all other custom exceptions subclass. This makes it possible for client code to catch all exceptions Jint might throw, even as new features are added and exception granularity grows.

lahma:
Resolves #676


Taritsyn:
Now it will be much more convenient for me to handle errors.

PR #680, opened 2019-10-29
Removing some optimizations to prepare for refactoring

sebastienros:
- FunctionWasCalled is not used anymore but kept for future reference. It detects if a method is called to release an arguments array to the pull.
- The last binding memoization has been removed.

### DIFF
--- a/Jint/Runtime/Environments/Binding.cs
+++ b/Jint/Runtime/Environments/Binding.cs
@@ -14,5 +14,7 @@ namespace Jint.Runtime.Environments
         public JsValue Value;
         public readonly bool CanBeDeleted;
         public readonly bool Mutable;
+
+        public bool IsInitialized => !ReferenceEquals(Value, null);
     }
 }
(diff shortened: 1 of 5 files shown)


Jint/Runtime/Environments/Binding.cs
```diff
@@ -14,5 +14,7 @@ public Binding(JsValue value, bool canBeDeleted, bool mutable)
         public JsValue Value;
         public readonly bool CanBeDeleted;
         public readonly bool Mutable;
+
+        public bool IsInitialized => !ReferenceEquals(Value, null);
```
lahma: Need to add a mental note about this non-readonly-struct and methods causes defensive copies, we can revisit when things settle.

lahma:
Merge if ready for next steps, not sure if you are planning to do add onto this or smaller PRs.

lahma:
UncacheableExpressionBenchmark shows taking twice the time and allocating 8x, so we probably need to get the tweaks (or their variants) back in when the dust settles.

sebastienros:
Are you still ok with merging this one, I had forgotten about it, and the more we wait, the harder it will be. Unless your other PR would become hard to merge with these changes.

lahma:
Sure thing, please merge, we can tweak performance later on. I'm more eager to get the let/const bits 🙂

sebastienros:
Next step, before let/const work is done, is to get hoisting working based on the specs, and not based on the hackery that happens in esprima. And we can then remove it from esprima.

PR #682, opened 2019-11-06
Fix ambiguity errors when compiling with Esprima source in tree.

Commaster:
Any source file that has `using` both `Jint.Runtime` and `Esprima` fails to compile due to ambiguous class ExceptionHelper:
https://github.com/sebastienros/jint/blob/dev/Jint/Runtime/ExceptionHelper.cs
https://github.com/sebastienros/esprima-dotnet/blob/dev/src/Esprima/ExceptionHelper.cs

Only happens if you place Esprima source inside jint's source tree to build a single dll instead of several.

### DIFF
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -441,12 +441,12 @@ namespace Jint
         {
             if (_maxStatements > 0 && _statementsCount++ > _maxStatements)
             {
-                ExceptionHelper.ThrowStatementsCountOverflowException();
+                Jint.Runtime.ExceptionHelper.ThrowStatementsCountOverflowException();
             }
 
             if (_timeoutTicks > 0 && _timeoutTicks < DateTime.UtcNow.Ticks)
             {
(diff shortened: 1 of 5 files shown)


lahma:
This will probably be quite brittle and easily broken again. May I ask why aren't you just merging the two assemblies if you want to have single assembly? Is there a reason you cannot have to DLLs?

Commaster:
Merging with tools like ILMerge? They struggle to keep up with the latest .Net versions

Having a single DLL keeps my folder clean and copying one library doesn't risk creating missing assemblies errors. Not only do I wrap Esprima inside jint, but I also wrap jint in 2 further wrappers and get a single DLL out instead of 4.

Commaster:
Fixed by https://github.com/sebastienros/esprima-dotnet/pull/124

lahma:
Sorry I forgot to ping back here, but this was on the todo list 👍

Commaster:
Thank you again 👍

PR #690, opened 2019-12-30
Improve date handling and enable 262 tests

### LINKED ISSUE #689: new Date().setHours(undefined) takes too much time
bug

lahma:
fixes #689

Went through new tests, fixed issues and aligned better with spec. Also now running net452 tests for 262 suite too as there were some corner cases like handling large double.ToString() values as keys.

Time zone handling with years less than -9999 and greater than 9999 might be flaky, but probably something we can leave for the next generation..

Old ECMA tests marked as obsolete are from length property flags changing.


sebastienros:
You must have had fun. Always liked handling time, you learn a lot.

lahma:
I have this small library that has some relation to time, time zones and daylight saving changes. Always a lot of fun.

PR #693, opened 2020-01-07
Enable Symbol and RegExp tests and fix issues

lahma:
Enabling Symbol and RegExp related tests and fixing issues raised. Bringing code closer to ECMA spec (2019 version).

Mental note, we probably need some updates to Esprima in the future and can use dev feed, but before releasing shiny Jint, we need to reference a new published version from NuGet.


Jint/Runtime/Interpreter/Expressions/JintCallExpression.cs
```diff
@@ -146,7 +146,7 @@ protected override object EvaluateInternal()
             var thisObject = Undefined.Instance;
             if (r != null)
             {
-                if (r.IsPropertyReference())
+                if (r._baseValue._type < InternalTypes.ObjectEnvironmentRecord)
```
sebastienros: Isn't it any type? Or could we use another compound flag value to represent these types? Like for PrimitiveTypes. Because I have no idea what it means when I read this line, unless the spec defines it this way.

lahma: The problem here is that this is done for `NullPropagationFromArg` test case. There parameter passed as null must be considered valid and the official spec says that it needs to be primitive or object. I'm unsure if `JintCallExpression` is against the spec, it's based on original code I guess. But null propagation is a special case and not part of spec I guess.

sebastienros:
I am amazed that all the unit tests still pass ;)

lahma:
Well it took some days to get tests pass 😉 Thanks for the review!

PR #699, opened 2020-01-13
Fix string indexing detecting string starting with $ as NaN and thus 0

### LINKED ISSUE #698: Issue with property names prefixed with $
I ran into this issue when using a receiver with JSON.parse in jint. It seems when a property name if prefixed with `$`, it is evaluated to `0`, and so I get the following behavior:

```csharp
engine.SetValue("str", "Hello");
var xx1 = engine.Execute("str.$ref").GetCompletionValue().ToString(); // "H"
var xx2 = engine.Execute("str.ref").GetCompletionValue().ToString(); // "undefined"
var xx3 = engine.Execute("str.$foo").GetCompletionValue().ToString(); // "H"
var xx4 = engine.Execute("str.foo").GetCompletionValue().ToString(); // "undefined"
var xx5 = engine.Execute("str['$foo']").GetCompletionValue().ToString(); // "H"
var xx6 = engine.Execute("str['foo']").GetCompletionValue().ToString(); // "undefined"

var xx7 = engine.Execute("str.hasOwnProperty('$foo')").GetCompletionValue().ToString(); // "True"
var xx8 = engine.Execute("str.hasOwnProperty('foo')").GetCompletionValue().ToString(); // "False"
```

expected results should be
- "undefined" for xx1, xx2, xx3, xx4, xx5, xx6
- "False" for xx7, xx8

This is with jint `3.0.0-beta-1632`, on dotnet core 3.1 sdk.

lahma:
fixes #698

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -2786,6 +2786,21 @@ x.test = {
             Assert.Throws<JavaScriptException>(() => engine.Execute("c.Name"));
         }
 
+        [Fact]
+        public void ShouldAllowDollarPrefixForProperties()
+        {
+            _engine.SetValue("str", "Hello");
+            _engine.Execute("equal(undefined, str.$ref);");
+            _engine.Execute("equal(undefined, str.ref);");
+            _engine.Execute("equal(undefined, str.$foo);");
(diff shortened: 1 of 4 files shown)


lahma:
This actually gave me an idea for optimizing some basic indexing 😅

PR #702, opened 2020-01-18
typo fix

TadaWasRaped:
just fixed a small typo


sebastienros:
Thank you but that is not a typo. Expando object is a real thing in dotnet.

PR #703, opened 2020-01-19
Function call performance improvements

lahma:
I tried to return some of the performance that was lost with binding instantiation changes.

* use adapting map structures for DeclarativeEnvironmentRecord
* lazily materialize arguments when needed
* put test files to project ignore, just slows down the IDE
* add test case for running same code as arrow function, faster!

## Jint.Benchmark.UncacheableExpressionsBenchmark

``` ini

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.19041
Intel Core i7-6820HQ CPU 2.70GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.101
  [Host]     : .NET Core 3.0.0 (CoreCLR 4.700.19.46205, CoreFX 4.700.19.46214), X64 RyuJIT
  DefaultJob : .NET Core 3.0.0 (CoreCLR 4.700.19.46205, CoreFX 4.700.19.46214), X64 RyuJIT


```

## Before

|    Method |   N |     Mean |   Error |  StdDev |       Gen 0 |      Gen 1 | Gen 2 | Allocated |
|---------- |---- |---------:|--------:|--------:|------------:|-----------:|------:|----------:|
| Benchmark | 500 | 500.2 ms | 2.39 ms | 2.12 ms | 150000.0000 | 28000.0000 |     - | 598.52 MB |

## After

|    Method |   N | Arrow |     Mean |   Error |  StdDev |      Gen 0 |      Gen 1 | Gen 2 | Allocated |
|---------- |---- |------ |---------:|--------:|--------:|-----------:|-----------:|------:|----------:|
| Benchmark | 500 | False | 246.2 ms | 1.32 ms | 1.17 ms | 43000.0000 | 11000.0000 |     - | 171.78 MB |
| Benchmark | 500 |  True | 197.6 ms | 0.73 ms | 0.65 ms | 37333.3333 |  2000.0000 |     - | 149.96 MB |


Jint/Runtime/Environments/DeclarativeEnvironmentRecord.cs
```diff
@@ -78,14 +73,14 @@ public override void SetMutableBinding(in Key name, JsValue value, bool strict)
...
         {
-            ref var binding = ref GetOrCreateBinding(name);
+            _dictionary.TryGetValue(name, out var binding);
             return UnwrapBindingValue(strict, binding);
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private JsValue UnwrapBindingValue(bool strict, in Binding binding)
         {
-            if (!binding.Mutable && !binding.IsInitialized)
```
sebastienros: I'd keep IsInitialized as it shouldn't matter for perf and is easier to read the spec

Jint/Runtime/Interpreter/Expressions/JintObjectExpression.cs
```diff
@@ -107,7 +107,7 @@ private object BuildObjectFast()
             {
                 var objectProperty = _properties[i];
                 var valueExpression = _valueExpressions[i];
-                var propValue = valueExpression.GetValue();
+                var propValue = valueExpression.GetValue().Clone();
                 hasSymbols |= objectProperty.Key.IsSymbol;
```
sebastienros: I don't see the full file, is this clone method used it for Arguments instance?

Jint/Runtime/Environments/DeclarativeEnvironmentRecord.cs
lahma: Returned the old version

Jint/Runtime/Interpreter/Expressions/JintObjectExpression.cs
lahma: Yes, it's now the same logic like in ConcatenatedString, a moment to materialize things as there's an assignment that might lead to moving out of original scope (function body). Now IMO cleaner and was able to remove the FunctionWasCalled from binding context (making it better for you).

lahma:
Ready from my part, decided to use simpler HybridDictionary, about the same perf and more understandable than multi-level collection mutation that was there earlier.

sebastienros:
So arrow functions are inherently faster because they don't expose the arguments object? Do you know if that's true also in v8?

lahma:
Basically this https://www.ecma-international.org/ecma-262/10.0/index.html#sec-functiondeclarationinstantiation . We can skip object allocation and binding it to context. Especially for small arrows like array.filter this extra work adds runtime, also no need to worry about pooling etc. V8 probably has some black magic, might not show up there that much...

PR #705, opened 2020-02-18
Add current memory usage to debug information

ceifa:


sebastienros:
Looks good, but build fails

ceifa:
Thank you for merging! Do you have a prevision to publish this version in nuget?

sebastienros:
Build is running, should be ready in an hour if all goes well.

sebastienros:
Follow this: https://ci.appveyor.com/project/SebastienRos/jint/builds/31109993

ceifa:
I'm receiving this error now:
image

If i put the myget source of esprima it works, but can you create a nuget release for it too?

sebastienros:
snap :/ we must have a wrong version in the config

sebastienros:
Fixed

ceifa:
After I started using this feature, I realized the value isn't the memory usage, but the memory allocated in the lifetime, since it's not subtracted when the memory is released.

The property and some internal fields should be renamed or reimplemented to comply with the name and what the value is.

I will make a PR to fix it sometime, I'm just commenting to make sure you know it or have something against it.

PR #709, opened 2020-03-07
Execution constraints

SebastianStehle:
Hi,

I was playing around with the timeout a little bit and the problem is that the time is queried for each execution, which hurts performance. So I tried to optimize it with a simple cancellationtoken source and to test different approaches I have introduced an interface for constraints like memory usage and so on:

The performance is indeed better and it also allows to implement custom constraint, e.g. cancellation tokens.

|   Method |  N |     Mean |     Error |    StdDev |      Gen 0 |     Gen 1 | Gen 2 | Allocated |
|--------- |--- |---------:|----------:|----------:|-----------:|----------:|------:|----------:|
| Timeout1 | 10 | 195.2 ms | 0.4972 ms | 0.4651 ms | 20000.0000 | 8666.6667 |     - | 120.55 MB |
| Timeout2 | 10 | 178.3 ms | 0.5507 ms | 0.4299 ms | 20000.0000 | 8666.6667 |     - | 120.55 MB |


Jint/Constraints/MaxStatements.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Jint.Runtime;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Jint.Constraints
+{
+    class MaxStatements : IConstraint
```
lahma: nit: maybe internal?

Jint/Constraints/MemoryLimit.cs
```diff
@@ -0,0 +1,56 @@
+﻿using Jint.Runtime;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Jint.Constraints
+{
+    public sealed class MemoryLimit : IConstraint
```
lahma: making internal allows more change when there are no users depending on using directly

Jint/Constraints/TimeConstraint.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Jint.Runtime;
+using System;
+
+namespace Jint.Constraints
+{
+    public sealed class TimeConstraint : IConstraint
```
lahma: internal

Jint/Constraints/TimeConstraint2.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Jint.Runtime;
+using System;
+using System.Threading;
+
+namespace Jint.Constraints
+{
+    public sealed class TimeConstraint2 : IConstraint
```
lahma: definitely internal especially when names are "made up", should not expose

Jint/IConstraint.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Jint
+{
+    public interface IConstraint
```
lahma: could probably be internal, otherwise might need some documentation if intended for users to extend/specialize

SebastianStehle: I wanted to make a constraint that allows me to pause the time. The goal is to use scripting for async calls and only measure the time that is actually spend within the script (CPU bound) not for "waiting".

Furthermore the memory constraint needs to be enhanced to support async calls and potential thread switches.

Jint/Constraints/MaxStatements.cs
SebastianStehle: Done

Jint/Constraints/TimeConstraint.cs
SebastianStehle: Done

Jint/IConstraint.cs
lahma: One thing that came to mind also is that if this were abstract base class, dispatch would be a bit cheaper, all invocations are in hot path.

Jint/Engine.cs
```diff
@@ -452,30 +412,10 @@ public JsValue GetCompletionValue()
...
-            }
-
-            if (_timeoutTicks > 0 && _timeoutTicks < DateTime.UtcNow.Ticks)
-            {
-                ExceptionHelper.ThrowTimeoutException();
-            }
-
-            if (_memoryLimit > 0)
+            // Avoid allocating the enumerator because we run this loop very often.
+            for (var i = 0; i < _constraints.Count; i++)
```
lahma: Another idea that came to my mind to copy the constrains engine's array, then foreach is as fast as for loop and loop against array is always faster than against list. The options might even keep track of the constraints changing and invalidate the backing array and generate on request (Constrains getter) - options would be reusable and efficient. Might be future improvements.

Jint/IConstraint.cs
SebastianStehle: I have profiled it because of the DateTime.UtcNow and I think this is a micro-micro-micro-optimization with very little gain, probably the same like my for loop :D

Jint/Engine.cs
SebastianStehle: I am not sure if it makes a big difference...the list is an array under the hood and I guess the compiler will inline some of the calls to the list indexer anyway.

sebastienros:
Only one request if I may, would be to move the contraint build methods in an extension methods class, as it's using the public facing apis to register specific contraints. Then I'll merge.
If you don't want to then I'll still merge ;)

SebastianStehle:
Done

lahma:
Great work, thank you!

Issue #157, opened 2015-03-16
Interop Nullable Parameters


anyolejoe:
I've found another problem with nullable data types.

If we have a .NET class  defined - 

public class Test {

```
   public static int GetVal() { return 1; }

    public static string GetGoose(int val)
    {
        return "pink footed";
    }
```

}

we can call into this fine like so -

```
   var val = Test.GetVal();
   var goose = Test.GetGoose(val);
```

However if we change the val parameter on the GetGoose function to a nullable int (int?) then we get the following exception - No public methods with the specified arguments were found

This occus even if we change the GetVal function to return a nullable int too.

anyolejoe:
I think I've worked out a good fix for the interop nullable values issue.

Adding the following code to the bottom of the Convert function in the DefaultTypeConverter seems to resolve this - 

          //handle conversion to nullable type
            if (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Nullable<>))
            {
                Type underlyingType = Nullable.GetUnderlyingType(type);

                if (underlyingType != null)
                {
                    return (value == null) ? null : System.Convert.ChangeType(value, underlyingType);
                }              
            }

            return System.Convert.ChangeType(value, type, formatProvider);

anyolejoe:
There are a couple of other tweaks required for this to handle reading and writing nulls accurately without errors too  -

In the Invoke method on the MethodInfoFunctionInstance class - 

 if (parameters[i] != null && typeof(System.Linq.Expressions.LambdaExpression).IsAssignableFrom(parameters[i].GetType()))
                            
and on the Value setter on the PropertyInfoDescriptor class - 

  // attempt to convert the JsValue to the target type
                    obj = currentValue.ToObject();
                    if (obj == null ||  obj.GetType() != _propertyInfo.PropertyType)
                    {
                        obj = _engine.ClrTypeConverter.Convert(obj, _propertyInfo.PropertyType, CultureInfo.InvariantCulture);
                    }


Hope that is of help.

Issue #153, opened 2015-03-09
Setting Interop Nullable Int Bug


anyolejoe:
There is a bug setting nullable int values in Jint 2.4.

You can see this if we run the following code -

var p = new Person
{

```
 Name = "Mickey Mouse"
```

};

_engine.SetValue("p", p);

var age = _engine.Execute("p.Age=System.Convert.ToInt32(5); p.Age;").GetCompletionValue().ToString();

This works fine if we use this property on the Person object -

public int Age { get; set; }

but we get an exception if the property is changed to -

public int? Age { get; set; }

This is the exception produced -

Invalid cast from 'System.Double' to 'System.Nullable`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]'.

sfhaney:
Any update on this issue?

PR #715, opened 2020-03-19
Fix assigning to Nullable Enum 

KurtGokhan:
Summary:
- added test for nullable enum and struct assignment
- fix nullable enum assignment
- microoptimizations in DefaultTypeConverter (I am not sure about its benefits)

Also #153 and #157 are not reproducible anymore.


sebastienros:
I think typeof expression are resolved at compile time. Also private member should be prefixed with `_`, or in upper case for constants. Merging as these are minor.

PR #721, opened 2020-04-11
Ability to set Array.prototype as prototype for array-like object wrappers

lahma:
* when we see object implementing ICollection or read-only like interface it should be safe to access as array like
* should not assign indexers which are invalid for usage (property is string and indexer uses int)
* assign Array.prototype to array-like objects allowing filter etc, only used when object doesn't have anything suitable itself

### DIFF
--- a/Jint.Tests/Runtime/Domain/FloatIndexer.cs
+++ b/Jint.Tests/Runtime/Domain/FloatIndexer.cs
@@ -1,21 +1,7 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
-
-namespace Jint.Tests.Runtime.Domain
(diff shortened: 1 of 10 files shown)


czema:
This is great, thanks a lot.

I can't quite tell, but this doesn't seem to support IEnumerable. I realize that some array features make no sense in IEnumerable, but they don't make sense on ReadonlyCollection either.

lahma:
The key here is that the accessed object needs to be `ICollection`, there are very few things that really are only `IEnumerable`, all (that I can think of) BCL collection types implement some more functional interface than just `IEnumerable`. And yes, some of the prototypes are mutating which don't make much sense for read-only, but `filter`, `some`, `every `etc can be nice to have.

I don't like much the magic here. LINQ/JS/CLR object bridge has some conceptual problems I guess.

PR #728, opened 2020-04-22
Propagate interop exceptions from IndexDescriptor

### LINKED ISSUE #652: Exceptions in indexers are swallowed
A repro:

```C#
public class FailingObject2
        {
            public int this[int index] => throw new ArgumentException("index is bad", nameof(index));
        }

        [TestMethod]
        [ExpectedException(typeof(ArgumentException))]
        public void when_exception_is_thrown_from_indexer_during_script_invocation_then_exception_is_raised()
        {
            var engine = new Engine()
                .Execute(@"
function f2(obj)
{
    return obj[1];
}");

            var failingObject = new FailingObject2();
            engine.Invoke("f2", failingObject);
        }
```

Looking at `IndexDescriptor.cs` we see:
https://github.com/sebastienros/jint/blob/63052c2f4ab723409118011ff9705377b13b8420/Jint/Runtime/Descriptors/Specialized/IndexDescriptor.cs#L84-L91

And a test that covers the catch here
https://github.com/sebastienros/jint/blob/63052c2f4ab723409118011ff9705377b13b8420/Jint.Tests/Runtime/InteropTests.cs#L1451-L1462

If this is intended behavior--then I am confused. Why would it be desirable to suppress an exception from an indexer?

lahma:
We should let inner exception of `TargetInvocationException `to flow through. Still we can check `ArgumentOutOfRangeException` to ensure undefined is returned from invalid indexing, like our beloved JS does.

fixes #652

### DIFF
--- a/Jint/Runtime/Descriptors/Specialized/IndexDescriptor.cs
+++ b/Jint/Runtime/Descriptors/Specialized/IndexDescriptor.cs
@@ -25,7 +25,7 @@ namespace Jint.Runtime.Descriptors.Specialized
 
             Writable = engine.Options._IsClrWriteAllowed;
         }
-        
+
         public IndexDescriptor(Engine engine, string key, object item)
             : this(engine, item.GetType(), key, item)
         {
@@ -34,9 +34,9 @@ namespace Jint.Runtime.Descriptors.Specialized
         internal static bool TryFindIndexer(
             Engine engine,
             Type targetType,
-            string propertyName, 
+            string propertyName,
             out PropertyInfo indexerProperty,
-            out MethodInfo containsKeyMethod, 
+            out MethodInfo containsKeyMethod,
             out object indexerKey)
         {
             // get all instance indexers with exactly 1 argument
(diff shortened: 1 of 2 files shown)


Jint/Runtime/Descriptors/Specialized/IndexDescriptor.cs
```diff
@@ -97,12 +97,19 @@ protected internal override JsValue CustomValue
...
                 }
-                catch
+                catch (TargetInvocationException tie)
                 {
-                    return JsValue.Undefined;
+                    switch (tie.InnerException)
+                    {
+                        case null:
+                            throw;
+                        case ArgumentOutOfRangeException _:
```
sebastienros, line 106: Why is this one special? Can there be other exceptions that would need to behave the same?

lahma: I think the original idea with the exception swallowing was to behave like JavaScript, over or under-indexing should produce undefined. There could be more exceptions to "swallow", but hard to say.

lahma: I added `IndexOutOfRangeException` as that is what happens with native array, List<T> throws `ArgumentOutOfRangeException`, good catch

Issue #342, opened 2016-10-25
case of .net property and method names question


fvaneijk:
It seems that there is a fallback method to match the name of the property or method of a .net object that uses a custom case insensitive compare.  This seems wrong to me.  To me the comparison should be an exact match.  What is the reason to allow this fallback compare?  I am under the impression that JavaScript is a case sensitive language.

see GetOwnProperty() and EqualsIgnoreCasing() in /Jint/Runtime/Interop/ObjectWrapper.cs

Thanks

sebastienros:
The idea is that some users want to access .net properties with camel casing which is more widely used in JavaScript. I assume we should be specific about the camel casing thing instead of ignoring the case altogether. 

Agreed to make it an option too.

See #338 which is the opposite ask.

fvaneijk:
Thanks for the clarification.  I agree that the comparison behavior should be specified by an option.  I.E. Exact or CamelCase maybe?  CamelCase being the current implementation.  Does this also apply to native JavaScript objects?

sebastienros:
Solution is to add an option to the Options class, like `CamelCasedProperties`, maybe in `AllowClr`, then have the constructor take a `StringComparer` implementation from this options, that is used in this class.

lahma:
#952 will allow supplying a `StringComparer` instance via options to customize the matching procedure.

PR #730, opened 2020-05-05
Optionally ignore casing when resolving properties

malex984:
There is an issue with using legacy C# classes due to JINT ignoring the case of the first letter of C# object's member name when it looks them up (e.g. `public void Start();` gets confused with `public bool start;`).

This is also a very lightweight fix for #342.

### DIFF
--- a/Jint/Options.cs
+++ b/Jint/Options.cs
@@ -17,6 +17,7 @@ namespace Jint
         private bool _allowDebuggerStatement;
         private bool _allowClr;
         private bool _allowClrWrite = true;
+        private bool _ignoreCasingResolveProperty = true;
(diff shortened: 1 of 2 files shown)


lahma:
Should there option `InteropNameMatchMode` and be of type enum with members `Strict`, `CamelCase`, `IgnoreCase`, thoughts @sebastienros ?

lahma:
And another option would be to make it `IEqualityComparer<string> InteropPropertyNameComparer { get; set; }`, the it could have default value of current custom and then easily accept `StringComparer.Ordinal` or `StringComparer.OrdinalIgnoreCase`.

lahma:
I've created a new PR for this: #952 . Basically the same idea but a bit more flexibility, Thanks for you implementation and I hope the second version will suite your needs.

PR #740, opened 2020-05-27
Support passing a dictionary inside an object

cschwarz:
Currently it is only possible to pass a dictionary only directly as a method parameter, but not if it is wrapped in an object.


lahma:
Thank you, much appreciated!

PR #746, opened 2020-06-08
EnsureCapacity can throw if called against sparse mode array

### LINKED ISSUE #745: NullReferenceException for large amount of splice
**Jint version:** 3.0.0-beta-1828

**Here is my C# code:**
```
var engine = new Engine();
var parser = new JavaScriptParser(code);
var script = parser.ParseScript();
engine.Execute(script); //<--- NullReferenceException
```

**Here is the js code I used:**
```
let arr = [];
for (let i = 0; i < 10000; i++) arr.push(i);
for (let i=0;i<10000;i++) arr.splice(0, 1);
```

**What happens?**
I get a NullReferenceException in the line `engine.Execute(script);`

Here is the full stack trace from Visual Studio
```
 	Jint.dll!Jint.Native.Array.ArrayInstance.EnsureCapacity(uint capacity)	Unknown
 	Jint.dll!Jint.Native.Array.ArrayOperations.ArrayInstanceOperations.EnsureCapacity(ulong capacity)	Unknown
 	Jint.dll!Jint.Native.Array.ArrayPrototype.Splice(Jint.Native.JsValue thisObj, Jint.Native.JsValue[] arguments)	Unknown
 	Jint.dll!Jint.Runtime.Interop.ClrFunctionInstance.Call(Jint.Native.JsValue thisObject, Jint.Native.JsValue[] arguments)	Unknown
 	Jint.dll!Jint.Runtime.Interpreter.Expressions.JintCallExpression.EvaluateInternal()	Unknown
 	Jint.dll!Jint.Runtime.Interpreter.Expressions.JintExpression.Evaluate()	Unknown
 	Jint.dll!Jint.Runtime.Interpreter.Expressions.JintExpression.GetValue()	Unknown
 	Jint.dll!Jint.Runtime.Interpreter.Statements.JintExpressionStatement.ExecuteInternal()	Unknown
 	Jint.dll!Jint.Runtime.Interpreter.Statements.JintForStatement.ForBodyEvaluation()	Unknown
 	Jint.dll!Jint.Runtime.Interpreter.Statements.JintForStatement.ExecuteInternal()	Unknown
 	Jint.dll!Jint.Runtime.Interpreter.JintStatementList.Execute()	Unknown
 	Jint.dll!Jint.Engine.Execute(Esprima.Ast.Script program)	Unknown
>	jint playground.dll!jint_playground.Program.Main(string[] args) Line 40	C#
```

**Note:**
This only happens with large arrays (if I enter 1000 instead of 10000 it doesn't happen)
In case you can't reproduce the bug try adding a 0 to the 10000

lahma:
fixes #745

### DIFF
--- a/Jint.Tests/Runtime/ArrayTests.cs
+++ b/Jint.Tests/Runtime/ArrayTests.cs
@@ -52,5 +52,17 @@ namespace Jint.Tests.Runtime
 
             Assert.Equal(8, result);
         }
+
+        [Fact]
+        public void LargeArraySize()
+        {
+            const string code = @"
+            let arr = [];
+            for (let i = 0; i < 10000; i++) arr.push(i);
+            for (let i=0;i<10000;i++) arr.splice(0, 1);
+            ";
+            var engine = new Engine();
+            engine.Execute(code);
+        }
     }
 }
--- a/Jint/Native/Array/ArrayInstance.cs
+++ b/Jint/Native/Array/ArrayInstance.cs
@@ -648,13 +648,15 @@ namespace Jint.Native.Array
 
         internal void EnsureCapacity(uint capacity)
         {
-            if (capacity <= MaxDenseArrayLength && capacity > (uint) _dense.Length)
+            if (capacity > MaxDenseArrayLength || _dense is null || capacity <= (uint) _dense.Length)
             {
-                // need to grow
-                var newArray = new PropertyDescriptor[capacity];
-                System.Array.Copy(_dense, newArray, _dense.Length);
-                _dense = newArray;
+                return;
             }
+
+            // need to grow
+            var newArray = new PropertyDescriptor[capacity];
+            System.Array.Copy(_dense, newArray, _dense.Length);
+            _dense = newArray;
         }
 
         public IEnumerator<JsValue> GetEnumerator()


Jint.Tests/Runtime/ArrayTests.cs
```diff
@@ -52,5 +52,17 @@ public void EmptyStringKey()
 
             Assert.Equal(8, result);
         }
+
+        [Fact]
+        public void LargeArraySize()
+        {
+            const string code = @"
+            let arr = [];
+            for (let i = 0; i < 10000; i++) arr.push(i);
```
sebastienros, line 61: Showing off with these let keywords now ;)

Jint/Native/Array/ArrayInstance.cs
```diff
@@ -648,13 +648,15 @@ private void ConvertToSparse()
 
         internal void EnsureCapacity(uint capacity)
         {
-            if (capacity <= MaxDenseArrayLength && capacity > (uint) _dense.Length)
+            if (capacity > MaxDenseArrayLength || _dense is null || capacity <= (uint) _dense.Length)
```
sebastienros, line 651: I understand that the capacity doubles?

lahma: Capacity is decided by the caller, we can grow as asked if it hasn't flipped to be dictionary (sparse) already. Usually logic asks for double the size but for example splice knows what to expect and thus was signaling specific size when algorithm is at specific point. Same like using constructor parameter, can signal known size.

PR #758, opened 2020-07-15
Provide a way to override interop members

sebastienros:
Open to discussion. It might be better to refactor the internal type and dictionary to expose them directly.

### DIFF
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -108,40 +108,6 @@ namespace Jint
         internal readonly PropertyDescriptor _callerCalleeArgumentsThrowerConfigurable;
         internal readonly PropertyDescriptor _callerCalleeArgumentsThrowerNonConfigurable;
 
-        internal readonly struct ClrPropertyDescriptorFactoriesKey : IEquatable<ClrPropertyDescriptorFactoriesKey>
-        {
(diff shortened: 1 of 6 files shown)


sebastienros:
Thiking also that an overload that return a JsValue instead of a property descriptor, or an object, might be better suited since it's called in CLR land, for CLR instances.

Jint/Engine.cs
```diff
@@ -239,6 +239,12 @@ public Engine(Action<Engine, Options> options)
             ClrTypeConverter = new DefaultTypeConverter(this);
         }
 
+        public void SetMemberAccessor(Type type, string name, Func<Engine, object, PropertyDescriptor> accessor)
```
lahma, line 242: Is there a need to change behavior runtime or would it make sense to do these via options when initializing engine?

lahma:
If you look at how object wrapping is configured:

https://github.com/sebastienros/jint/blob/dev/Jint/Options.cs#L78-L87

And then used:

https://github.com/sebastienros/jint/blob/dev/Jint/Native/JsValue.cs#L349-L351

I think allowing to call such callback and if it returns non-null the result should be used instead the one created in normal logic? It would make the API similar.

sebastienros:
Will try to implement a callback instead.

sebastienros:
Updated

lahma:
Looks good!

Issue #753, opened 2020-07-01
FunctionInstance no longer extendable


viceice:
I seems to be a regression of #722, since then i can no longer extend the  `FunctionInstance` class, because all constructors are internal

https://github.com/lahma/jint/blob/3b8a2de4627ab1c9bbda1589f4a3cd82b63ca80c/Jint/Native/Function/FunctionInstance.cs#L31-L51

You you accept a pr to make one of them at least `protected internal` or adding a protected cxtor?

lahma:
Yes of course, I'm sorry for your trouble. I haven't seen functions extended outside of Jint so the hiding might have been a bad idea. It would be great if you could create a PR with some extending class(es) in test project to verify that your use case is covered in the future. You might need to open some other types too like `FunctionThisMode`.

I would be nice to hear about your use case too to learn more, if you would like to elaborate.

viceice:
Thanks, i'll submit a pr the next days.

My use-case is to have a native  `UUID` (`Guid`) object type in js engine which will be auto converted to .net `Guid` when calling a `Guid` indexer on a .net class instance. For that i've implemented a `UuidConstructor`, a `UuidPrototype` and a `UuidInstance`.

So the only problem for now are the internal `FunctionInstance` ctor while updating from `3.0.0-beta-1778` to `3.0.0-beta-1828`.

PR #759, opened 2020-07-16
Add protected function instance constructor

viceice:
* add protected constructor to `FunctionInstance`
* add custom uuid type as test

closes: #753

### DIFF
--- a/Jint.Tests/Runtime/UuidTests.cs
+++ b/Jint.Tests/Runtime/UuidTests.cs
@@ -1,0 +1,50 @@
+﻿using Jint.Tests.Runtime.Domain;
+using System;
+using Xunit;
+
+namespace Jint.Tests.Runtime
+{
+    public class UuidTests : IDisposable
+    {
+        private readonly Engine _engine;
+
+        public UuidTests()
+        {
+            _engine = new Engine(o => o.AddObjectConverter(new UuidConverter()))
+                .SetValue("copy", new Func<Guid, Guid>(v => new Guid(v.ToByteArray())))
+                ;
+            UuidConstructor.CreateUuidConstructor(_engine);
+        }
+
+        void IDisposable.Dispose()
+        {
+        }
+
+        private object RunTest(string source)
+        {
+            return _engine.Execute(source).GetCompletionValue().ToObject();
+        }
+
+        [Fact]
+        public void Empty()
+        {
+            Assert.Equal(Guid.Empty, RunTest($"Uuid.parse('{Guid.Empty}')"));
+            Assert.Equal(Guid.Empty, RunTest($"Uuid.Empty"));
+        }
+
+        [Fact]
+        public void Random()
+        {
+            var actual = RunTest($"new Uuid()");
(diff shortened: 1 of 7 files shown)


Jint.Tests/Runtime/UuidTests.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Jint.Tests.Runtime
+{
+    public class UuidTests : IDisposable
+    {
+        private readonly Engine _engine;
+
+        public UuidTests()
+        {
+            _engine = new Engine(o => o.AddObjectConverter(new UuidConverter()))
```
sebastienros, line 13: What do you think about adding an extension method on Options named `AddUuid()` which would register the converter and also call CreateUuidConstructor ?

viceice: Sounds good. Would need some API refactoring.

lahma: Can we get away with AddObjectConverter<T>? not sure if we can bind the constructor with that though.

sebastienros: @viceice I think we would need a way to add Action<Engine> to a list that would be executed by the engine when it has initialized the options object.

viceice: What about to add `Engine.ObjectConverters`  as ` public List<IObjectConverter> ObjectConverters => Options._ObjectConverters`?  so we can add `IObjectConverter` after `Engine` contructor finished?

Then the `AddUuid()` extension would be very easy.

sebastienros: I prefer not to since we are looking at ways to reuse engines. Having all the configuration kept in options would allow us to reuse it on any engine instance once cleared of any state. We could even add a list of scripts. Or use these options in a factory object for pooling.

Instead of calling these methods from Engine, we can also have a method Apply(engine) such that the engine code would not change every time we add options. And a Options.Configure (Action<Engine>) to register custom logic. I can add this to your PR if you prefer.

viceice: Ok, feel free to change the PR. 🙃

viceice: Maybe I need to give you write access to my fork, because it's in an org.

sebastienros: If you haven't unallowed me, then by default I can write on the branch you used to submit a PR

sebastienros:
Had not realized that all these UUid classes were only in Tests, until I start adding what I suggested.
Doing it in a different PR then.
You should write a wiki entry with this UUid example and how to get to it, would be very helpful for many.

viceice:
> Had not realized that all these UUid classes were only in Tests, until I start adding what I suggested.
> Doing it in a different PR then.
> You should write a wiki entry with this UUid example and how to get to it, would be very helpful for many.

I've started [here](https://github.com/sebastienros/jint/wiki#custom-javascript-buildin-types-for-jint-v3).

PR #762, opened 2020-07-19
Adding constructor with existing options

sebastienros:
Being able to reuse options instances, and also create custom extension methods on `Options` that can alter the engine state.


lahma:
Still looking good 😉

PR #773, opened 2020-08-05
Prefer indexing and custom properties over indexer property

### LINKED ISSUE #642: Unable to access dynamic object property when named "item" or "Item"
I came across this issue while attempting to pass nested dynamic objects to a function. Everything works as expected unless I attempt to access a property on the object named "item" or "Item". 

The call to **Invoke** will throw **System.Reflection.TargetParameterCountException**: 'Parameter count mismatch.'

For example:

``` c#
Engine engine = new Engine().Execute(@"
    function good(arg) {
        return arg.whatever
    }
    function bad(arg) {
        return arg.item
    }
");

var argument = new Dictionary<string, object>
{
    { "whatever", "This works." },
    { "item", "This does not work." }
};

JsValue result;

// Works as expected and returns a string.
result = engine.Invoke("good", argument);

// Throws an exception.
result = engine.Invoke("bad", argument);
```

I have tried this with both Dictionaries and ExpandoObjects as well as nested combinations of both. Furthermore, attempting to access a property named "item" or "Item" when it is not defined on a Dictionary/ExpandoObject also results in an error, but `arg.whatever.item` returns **undefined**, as expected.

Strangely the exception on occurs if the property is named "item" or "Item" with that capitalization specifically. I have tried other capitalization such as "ITem" and "IteM" and the code works as expected without throwing an exception.

Here is the stack trace for reference:

```
System.Reflection.TargetParameterCountException: Parameter count mismatch.
   at System.Reflection.RuntimeMethodInfo.InvokeArgumentsCheck(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimePropertyInfo.GetValue(Object obj, Object[] index)
   at Jint.Runtime.Descriptors.Specialized.PropertyInfoDescriptor.get_CustomValue() in C:\projects\jint\Jint\Runtime\Descriptors\Specialized\PropertyInfoDescriptor.cs:line 24
   at Jint.Native.Object.ObjectInstance.Get(String propertyName) in C:\projects\jint\Jint\Native\Object\ObjectInstance.cs:line 117
   at Jint.Engine.GetValue(Reference reference, Boolean returnReferenceToPool) in C:\projects\jint\Jint\Engine.cs:line 549
   at Jint.Runtime.Interpreter.Statements.JintReturnStatement.ExecuteInternal() in C:\projects\jint\Jint\Runtime\Interpreter\Statements\JintReturnStatement.cs:line 23
   at Jint.Runtime.Interpreter.JintStatementList.Execute() in C:\projects\jint\Jint\Runtime\Interpreter\JintStatementList.cs:line 70
   at Jint.Native.Function.ScriptFunctionInstance.Call(JsValue thisArg, JsValue[] arguments) in C:\projects\jint\Jint\Native\Function\ScriptFunctionInstance.cs:line 70
   at Jint.Engine.Invoke(JsValue value, Object thisObj, Object[] arguments) in C:\projects\jint\Jint\Engine.cs:line 723
   at *My Project*
```

The exception is thrown from System.Reflection which makes me think this could be an issue with corefx or perhaps some kind of inherent limitation. However I cannot find any reference to this issue anywhere, except [this stack overflow question](https://stackoverflow.com/questions/54598746/jint-errors-out-when-transform-datarow-with-column-name-item) with no answers.

lahma:
Avoid choosing property `Item` when it's an indexer until really needed.

fixes #642

### DIFF
--- a/Jint.Tests/Runtime/InteropTests.cs
+++ b/Jint.Tests/Runtime/InteropTests.cs
@@ -8,7 +8,6 @@ using System.Reflection;
 using Jint.Native;
 using Jint.Native.Array;
 using Jint.Native.Object;
-using Jint.Runtime.Descriptors;
 using Jint.Runtime.Interop;
 using Jint.Tests.Runtime.Converters;
 using Jint.Tests.Runtime.Domain;
@@ -54,6 +53,39 @@ namespace Jint.Tests.Runtime
                 assert(y === true);
                 assert(z === 'foo');
             ");
+        }
+
+        [Fact]
+        public void CanAccessMemberNamedItem()
+        {
+            _engine.Execute(@"
+                    function item2(arg) {
+                        return arg.item2
+                    }
+                    function item1(arg) {
+                        return arg.item
+                    }
+                    function item3(arg) {
+                        return arg.Item
+                    }
+            ");
+
+            var argument = new Dictionary<string, object>
+            {
+                {"item2", "item2 value"},
+                {"item", "item value"},
+                {"Item", "Item value"}
+            };
+
+            Assert.Equal("item2 value", _engine.Invoke("item2", argument));
+            Assert.Equal("item value", _engine.Invoke("item1", argument));
+            Assert.Equal("Item value", _engine.Invoke("item3", argument));
+
+            var company = new Company("Acme Ltd");
+            _engine.SetValue("c", company);
+            Assert.Equal("item thingie", _engine.Execute("c.Item").GetCompletionValue());
+            Assert.Equal("item thingie", _engine.Execute("c.item").GetCompletionValue());
+            Assert.Equal("value", _engine.Execute("c['key']").GetCompletionValue());
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


Jint.Tests/Runtime/InteropTests.cs
```diff
@@ -56,6 +55,39 @@ public void PrimitiveTypesCanBeSet()
...
+            ");
+
+            var argument = new Dictionary<string, object>
+            {
+                {"item2", "item2 value"},
+                {"item", "item value"},
+                {"Item", "Item value"}
+            };
+
+            Assert.Equal("item2 value", _engine.Invoke("item2", argument));
```
sebastienros, line 80: Is this just to ensure that it's not breaking dictionary indexers?

lahma: Yes I wanted to make sure that normal objects (non-dictionaries) still allow both indexing and using Item/item property (preferring real Item property over the indexer).

Issue #756, opened 2020-07-03
Change in behaviour when calling the Engine.EnterExecutionContext method


alexmg:
I have seen a change in behaviour related to calling `EnterExecutionContext` between `3.0.0-beta-1597` and `3.0.0-beta-1848`. It is not clear to me which behaviour is actually correct and whether I am using the modified `EnterExecutionContext` method signature correctly.

Here is a test that shows the behaviour as it was in `3.0.0-beta-1597`. The important thing to note is that after calling `LeaveExecutionContext` the `inner` variable is no longer visible and never appears on global.

```c#
var engine = new Engine();

void Assert(bool condition) => condition.Should().BeTrue();
engine.SetValue("assert", (Action<bool>)Assert);

// Set outer variable in global scope
engine.SetValue("outer", 123);
engine.Execute("assert(outer === 123)");
engine.GetValue("outer").ToObject().Should().Be(123);

// Enter new execution context
var environment = LexicalEnvironment.NewDeclarativeEnvironment(engine, engine.GlobalEnvironment);
engine.EnterExecutionContext(environment, environment, JsValue.Null);

// Can see global scope
engine.Execute("assert(outer === 123)");

// Can modify global scope
engine.Execute("outer = 321");
engine.GetValue("outer").ToObject().Should().Be(321);

// Create variable in new context
engine.Execute("var inner = 456");
engine.Execute("assert(inner === 456)");

// New variable is not in global scope
engine.GetValue("inner").ToObject().Should().BeNull();

// Leave the new execution context
engine.LeaveExecutionContext();

// The new variable is no longer visible
engine.Execute("assert(typeof inner === 'undefined')");

// The new variable is still not in the global scope
engine.GetValue("inner").ToObject().Should().BeNull();
```

The test below fails when run against `3.0.0-beta-1848`and many of the more recent beta versions I tested. I have updated the code calling `EnterExecutionContext` and cannot find a combination of environment parameters that makes the code behave as before. As soon as the `inner` variable is declared it is visible in the global scope.

```c#
var engine = new Engine();

void Assert(bool condition) => condition.Should().BeTrue();
engine.SetValue("assert", (Action<bool>)Assert);

// Set outer variable in global scope
engine.SetValue("outer", 123);
engine.Execute("assert(outer === 123)");
engine.GetValue("outer").ToObject().Should().Be(123);

// Enter new execution context
var context = engine.ExecutionContext;
[...]

Issue #695, opened 2020-01-08
Is an object pool suggested for speed performance?


rushfrisby:
With some other javascript engines I found that instantiating the engine was very expensive and to combat this I created an object pool of engines. Would you suggest this approach for Jint as well or is creating an engine lightweight enough that it isn't needed?

lahma:
There was some work done earlier to make engine instantiation faster,  https://github.com/sebastienros/jint/pull/625 . Engine will of course load features on demand so if you urilize all JS constructs the engine needs to initialize them lazily. Caching the parsed program helps too. 

Sharing engine has the problem of old global state. 
 Probably depends on your use case and scripts.

sebastienros:
@lahma what is your opinion about being able to mark an ObjectInstance immutable, such that we would not be able to change the builtin constructors, or mark pre-defined properties as not configurable. This way we could just re-initalize the lexical environments and reuse the engines. The main benefit would be that these built-in elements wouldn't have to be re-created.

It doesn't seem very hard to do, at least for the most obvious properties and constructor objects. Also it should be easy to cover everything as it would mean anything that has a property name currently in the engine code.

Forgive me if that's something I have already asked somewhere else ;)

sebastienros:
I can see these two levels of reusability:
- The first one being about what I suggested so that we can reuse the engine instances and more importantly the existing constructor instances. 
- The second one would be being able to take a snapshot of a state and restore it. This would prevent the execution of a script to reach a specific state, and internally we would clone this state of the engine. 

Maybe the second option is also the way to implement option 1. Though freezing a state (default engine state) is even faster for perf than having to clone a snapshot (allocations).

sebastienros:
But yeah, option 1 means we would disallow things like `Math.max = null`

lahma:
@rushfrisby I'm working on PR https://github.com/sebastienros/jint/pull/789 , you might want to follow that development and please chime in if it's missing something.

ChrML:
Immutables is a nice solution to this. Keep a shared Math instance until you change it, then it becomes a new instance dedicated for that engine.

Issue #491, opened 2018-03-31
Is it possible to clone or save/restore engine state?


KarolKolenda:
I've got relatively long initial code and than I use the engine to execute a single statement which can affect engine state. I would like to speed things up by not re-executing initial code all the time. Would it be possible to:
a) quickly clone the instance of engine with it entire state or
b) save and restore quickly engine state?

Many thanks,
Karol

pillepalle1:
+1
It would be great if Engine would implement ICloneable

lahma:
I'm wokring on a PR https://github.com/sebastienros/jint/issues/788 that should serve this purpose by allowing a temporary context to run commands in that can then be cleaned.

PR #789, opened 2020-10-21
Add support for isolated execution context

### LINKED ISSUE #738: Allow configuring options per Execute calls
Allow overriding defaults for max statements for example for trusted scripts, otherwise would run with engine level constraints. This would probably mean creating a context that would pass through statement calls, context would have options which is engine default or user provided.

Relates to RavenDB's [need to change max statement](https://github.com/ravendb/jint/commit/e7d511a6c81dcd36fa62ba5fa32b84342d78b549) on use case basis.

### LINKED ISSUE #805: How to reset the state of the engine between calls?
I'm trying to understand how to use the engine's EnterExecutionContext/LeaveExecutionContext methods. I would like to reset the engine after a certain operation. For example, in the following code, I don't expect to see `name` printed as `foo` but it does.

```
{
    var engine = new Engine();

    for (int i = 0; i < 2; i++)
    {
        if (i == 0)
        {
            var ctx = engine.ExecutionContext;
            var lexEnv = LexicalEnvironment.NewDeclarativeEnvironment(engine, ctx.LexicalEnvironment);
            var varEnv = LexicalEnvironment.NewDeclarativeEnvironment(engine, ctx.VariableEnvironment);
            engine.EnterExecutionContext(lexEnv, varEnv, ctx.ThisBinding);
            var x = engine.SetValue("name", "foo");
            Console.WriteLine("First " + engine.GetValue("name"));
            engine.LeaveExecutionContext();
        }
        else
        {
            Console.WriteLine("Second " + engine.GetValue("name"));
        }
    }
}
```

Is there any other way I can reset the state of the engine between calls?

lahma:
This seems to work now quite nicely, at least as a proof of concept. A disposable context can be created which allows introducing variables and whatnot and when context is disposed they disappear. Also sets everything resolved from outer context (say the real global) as read-only and throws errors if you try to do something like `Math.max = null;`.

Here's the code sample from test case demonstrating the functionality:

```csharp
var engine = new Engine(options => options.Strict());
engine.SetValue("assert", new Action<bool>(Assert.True));

// Set outer variable in global scope
engine.SetValue("outer", 123);
engine.Execute("assert(outer === 123)");
engine.GetValue("outer").ToObject().Should().Be(123);

// Enter new execution context
using (engine.EnterIsolatedContext())
{
	// Can see global scope
	engine.Execute("assert(outer === 123)");

	// Can modify global scope
	engine.Execute("outer = 321");
	engine.GetValue("outer").ToObject().Should().Be(321);

	// Create variable in new context
	engine.Execute("var inner = 456");
	var value = engine.Execute("inner").GetCompletionValue();
	engine.Execute("assert(inner === 456)");

	// cannot break anything
	Assert.Throws<NotSupportedException>(() => engine.Execute("Math.max = null;"));
}

// The new variable is no longer visible
engine.Execute("assert(typeof inner === 'undefined')");

// The new variable is not in the global scope
engine.GetValue("inner").ToObject().Should().BeNull();

var max = engine.Execute("Math.max").GetCompletionValue();
max.ToObject().Should().NotBeNull();

// but can again break things
engine.Execute("Math.max = null;");
engine.GetValue("Math.max").ToObject().Should().BeNull();
```

When coupled with default engine pool implementation (to be done) this should allow the common pattern people request, a pre-warmed engine with scripts that can then be called with temporary catching context that is cleaned after the call.

Relates to
* https://github.com/sebastienros/jint/issues/756
* https://github.com/sebastienros/jint/issues/695
* https://github.com/sebastienros/jint/issues/491

fixes #738
fixes #805

### DIFF
--- a/Jint.Benchmark/EnginePoolBenchmark.cs
+++ b/Jint.Benchmark/EnginePoolBenchmark.cs
@@ -1,0 +1,93 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using BenchmarkDotNet.Attributes;
+using Jint.Pooling;
+
+namespace Jint.Benchmark
+{
+    [MemoryDiagnoser]
+    public class EnginePoolBenchmark
+    {
+        private static readonly Options options = new Options().Strict();
+        
+        private static readonly List<string> files = new List<string>
+        {
+            "dromaeo-3d-cube",
+            "dromaeo-core-eval",
+            "dromaeo-object-array",
+            "dromaeo-object-regexp",
+            "dromaeo-object-string",
+            "dromaeo-string-base64"
+        };
+
+        private IsolatedEngineFactory factory;
+        private Engine _engine;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            factory = new IsolatedEngineFactory(options, InitializeEngine);
+
+            _engine = new Engine(options);
+            InitializeEngine(_engine);
+        }
+
+        /// <summary>
+        /// Test case where we discard the pool and data.
+        /// </summary>
+        [Benchmark]
+        public void RunUsingPooledIsolated()
+        {
+            using var engine = factory.Build();
+            RunScript(engine);
+        }
+
+        /// <summary>
+        /// Using one engine that isn't able to clean up.
+        /// </summary>
+        [Benchmark]
+        public void RunUsingDirtiedEngineInstance()
+        {
+            RunScript(_engine);
+        }
+
+        /// <summary>
+        /// Test case to do always the manual labour.
+        /// </summary>
+        [Benchmark]
+        public void RunUsingNewEngineInstance()
+        {
+            var freshEngine = new Engine(options);
+            InitializeEngine(freshEngine);
+            RunScript(freshEngine);
+        }
+
+        private static void RunScript(IScriptExecutor engine)
+        {
+            for (int i = 5; i < 20; ++i)
+            {
+                engine.Evaluate($"Init({i});");
+            }
+        }
+
+        private static void InitializeEngine(IEngine engine)
+        {
+            engine.SetValue("log", new Action<object>(Console.WriteLine));
+            engine.SetValue("assert", new Action<bool>(b => { }));
+            engine.SetValue("name", 123);
+            engine.Evaluate(@"
+                                    var startTest = function () { };
+                                    var test = function (name, fn) { fn(); };
+                                    var endTest = function () { };
+                                    var prep = function (fn) { fn(); };
+                    ");
+
+            foreach (var fileName in files)
+            {
+                var content = File.ReadAllText($"Scripts/dromaeo/{fileName}.js");
+                engine.Evaluate(content);
+            }
+        }
+    }
+}
--- a/Jint.Tests/Runtime/IsolatedContextTests.cs
+++ b/Jint.Tests/Runtime/IsolatedContextTests.cs
@@ -1,0 +1,90 @@
+using System;
+using FluentAssertions;
+using Jint.Native;
+using Jint.Runtime;
+using Xunit;
+
+namespace Jint.Tests.Runtime
+{
+    public class IsolatedContextTests
+    {
+        [Fact]
+        public void IsolatedContextCanBeUsedAndDisposed()
+        {
+            var engineFactory = new DefaultEngineFactory(new Options().Strict());
+            var engine = engineFactory.Build();
+            engine.SetValue("assert", new Action<bool>(Assert.True));
+
+            // Set outer variable in global scope
+            engine.SetValue("outer", 123);
+            engine.SetValue("outerClrArray", new[] { "a", "b", "c" });
+            engine.Evaluate("var outerJsArray = [ 'a', 'b', 'c' ];");
+            engine.Evaluate("assert(outer === 123)");
+            engine.Evaluate("outer").ToObject().Should().Be(123);
+
+            // outer scope functions and trusted, they can break things, so be aware
+            engine.Evaluate("function outerBreaker() { var m = Math.max; delete Math.max; assert(typeof Math.max === 'undefined');  Math.max = m; }");
+
+            // Enter new execution context
+            using (engine.ActivateIsolatedContext())
+            {
+                // Can see global scope
+                engine.Evaluate("assert(outer === 123)");
+
+                // Can modify global scope
+                engine.Evaluate("outer = 321");
+                engine.Evaluate("outer").ToObject().Should().Be(321);
+
+                // Create variable in new context
+                engine.Evaluate("var inner = 456");
+                engine.Evaluate("assert(inner === 456)");
+
+                engine.Evaluate("function innerBreaker() { Math.max = null; }");
+
+                engine.Evaluate("var m = Math.max; outerBreaker();");
+                engine.Evaluate("Math.max").Should().BeAssignableTo<ICallable>();
+                engine.Evaluate("m === Math.max").AsBoolean().Should().BeTrue();
+                
+                // cannot break anything
+                Assert.Throws<NotSupportedException>(() => engine.Evaluate("innerBreaker();"));
+                Assert.Throws<NotSupportedException>(() => engine.Evaluate("Math.max = null;"));
+                Assert.Throws<NotSupportedException>(() => engine.Evaluate("outerJsArray[0] = null;"));
+                Assert.Throws<NotSupportedException>(() => engine.Evaluate("outerClrArray[0] = null;"));
+                
+                // but we can redeclare the whole math thing and make it "better" in this context
+                engine.Evaluate("var Math = ({});");
+                engine.Evaluate("Math.max = () => 42;");
+
+                engine.Evaluate("outerJsArray = [ 'c' ];");
+                engine.Evaluate("outerJsArray[0]").AsString().Should().Be("c");
+
+                var maxInner = engine.Evaluate("Math.max(1, 2);").AsNumber();
+                maxInner.Should().Be(42);
+            }
+
+            // The new variable is no longer visible
+            engine.Evaluate("assert(typeof inner === 'undefined')");
+
+            // The new variable is not in the global scope
+            var ex = Assert.Throws<JavaScriptException>(()  => engine.Evaluate("inner"));
+            ex.Message.Should().Be("inner is not defined");
+
+            var max = engine.Evaluate("Math.max");
+            max.ToObject().Should().NotBeNull();
+            
+            // and we should again get reasonable results from max
+            var maxOuter = engine.Evaluate("Math.max(1, 2);").AsNumber();
+            maxOuter.Should().Be(2);
+
+            // array back to itself
+            engine.Evaluate("outerJsArray[0]").AsString().Should().Be("a");
+
+            // but can again break things
+            engine.Evaluate("Math.max = null;");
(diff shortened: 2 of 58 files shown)


sebastienros:
Should we be able to use this without the dispose pattern, such that we can also do "LeaveIsolatedContext", so the engine would store the context and dispose it when it's called.

Might be easier to make the pool implementation.

sebastienros:
I meant to keep the Dispose pattern, just that also being able to call Leave, that will invoke Dispose on the local context.

sebastienros:
How is  `engine.Execute("outer = 321");` in the scope different from updating `Math.max`  ?

Jint.Tests/Runtime/IsolatedContextTests.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            // Enter new execution context
+            using (engine.EnterIsolatedContext())
+            {
+                // Can see global scope
+                engine.Execute("assert(outer === 123)");
+
+                // Can modify global scope
+                engine.Execute("outer = 321");
+                engine.GetValue("outer").ToObject().Should().Be(321);
```
sebastienros, line 28: What is the value when the scope is released?
Why doesn't it throw an exception like `Math.max`. Not saying it should, but how do we know it won't?

Jint/Pooling/EnginePool.cs
```diff
@@ -0,0 +1,110 @@
...
+                return engine;
+            }, size);
+        }
+
+        public void ExecuteInIsolatedContext(Action<IPooledEngine> action)
+        {
+            Engine engine;
+            lock (_enginePool)
+            {
+                engine = _enginePool.Allocate();
```
sebastienros, line 31: Isn't this method already thread-safe ?

Jint/Pooling/EnginePool.cs
```diff
@@ -0,0 +1,110 @@
...
+                action(new EngineWrapper(engine));
+            }
+
+            lock (_enginePool)
+            {
+                _enginePool.Free(engine);
+            }
+        }
+
+        private sealed class EngineWrapper : IPooledEngine
```
sebastienros, line 45: Maybe create an `IEngine` instead that is implemented by `Engine` too. And the pooled version could handle the dispose pattern, and have a Release/Return() method if we don't want to use the dispose pattern.

Jint/Pooling/EnginePool.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        public void ExecuteInIsolatedContext(Action<IPooledEngine> action)
+        {
+            Engine engine;
+            lock (_enginePool)
+            {
+                engine = _enginePool.Allocate();
+            }
+
+            using (engine.EnterIsolatedContext())
```
sebastienros, line 34: With the interface, the pool could return an IEngine directly, which would be the result of this call.

lahma:
> How is `engine.Execute("outer = 321");` in the scope different from updating `Math.max` ?

This is a bit tricky and probably counter-intuitive at first. When you update anything var-like which is bound to global scope (`Math`, 'outer`) we can easily catch that as it comes as request to our `IsolatedEnvironmentRecord`. This is the key to this shadowing and isolation in my approach - environment record tracks sets and protects original record and Global object.

But when you do `Math.max = null;` it first resolves the target object and then calls `Set` logic for the `ObjectInstance`, so "get Math instance `m` and call `m.Set("max", JsNull);`.  This is why we wrap anything resolved as object instance as read-only which is resolved from outer scope (the real global environment).

So key difference is that doing `var Math = ({}); Math,max = () => 42;` Introduces new var to *our isolated context* and we can allow altering it however one wants, trying to alter real global's sub-properties will poison the engine as we cannot easily track all the object graphs and restore them later.

Jint.Tests/Runtime/IsolatedContextTests.cs
lahma: I explained this below wrt `Math.max`, TLDR; we can rewrite and modify anything that can be attached to our isolated context - this is shadowing vars. `outer = 32;` is effectively `var outer = 321;` so introducing a new variable isolated environment hiding old from global.

Jint/Pooling/EnginePool.cs
lahma: Well they used to be 😉 When I brought the pool from MS pool package, we traded thread-safety for speed as Jint is not thread-safe. Would require second impl with the origimal `Interlocked` functionality.

Jint/Pooling/EnginePool.cs
lahma: Can do, I just want to IEngine to strictly limited to execution and setting values to context. Might be role interface too. Just need to limit public access to things like `ObjectConstructor` etc which are now public and might allow doing bad things.

lahma: I think I'll stick for wrapper now so that we can implement the dispose and checks here. Or should the Engine implement it as quite no-op for now and we would just start checking `if (_disposed) { throw diposedexception}` in engine service methods?

Jint.Benchmark/EnginePoolBenchmark.cs
```diff
@@ -0,0 +1,108 @@
...
+        [GlobalSetup]
+        public void Setup()
+        {
+            _enginePool = new EnginePool(
+                10,
+                options,
+                engine =>
+                {
+
+                    engine.SetValue("log", new Action<object>(Console.WriteLine));
```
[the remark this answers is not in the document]
lahma: Fixed

lahma:
Now we have support for `ActivateTemporaryConstraints` which makes me feel that should we group these under same prefix, like `ActivateIsolatedContext` which then both follow `IDisposable` semantics. I also feel inclined to remove the `EnterExecutionContext` and `LeaveExecutionContext` for public API, I know some have been using them but I guess they are quite complex and broken for the old use cases after let/const which introuced new concepts.

The pool API also needs some love. `Allocate` and `Free` are there and now return the full Engine which allows you to do bad things... But I think those are for expert users who want to track life time and cleanup themselves..

sebastienros:
I don't see the point of IEngine if we have IScriptExecutor. These two methods in IEngine are used by the EngineWrapper, and it knows about Engine, so the interface seems useless. At least based on the fact that I don't see a user accessing them. Or I would just shift the interface names to
- IEngine (instead of IScriptExecutor)
- IActivatedEngine (instead of IEngine), if you have a reason to keep this interface.

Can you share the results of the benchmarks?

lahma:
I think this PR is coming a bit of overwhelmed with all the needs. My original thinking was to clean up `Engine` by providing `IEngine` that would only allow moderate amount of functionality, but that might be a long shot. Currently `Engine` exposes a lot of things and they make it hard to be backwards compatible, like `EnterExecutionContext` - the thing we are trying to remedy here. IEngine should probably only offer functionality somewhat described in spec (constructors, global objects).

Maybe we need to think about having the `IEngine` that exposes things JS has like `Global` and `Array.Prototype` and such that are set in stone. The pooled engine needs to have more isolation and try to hide some things to reduce possibility of "state contamination".

There are a lot of things here, like why we expose `PrototypeObject` instead of  `Prototype` which would make more sense in JS kind of way.

sebastienros:
I like it better. So anything that can be isolated should be in the interface that is returned by isolated pools, e.g. `IIsolatedEnginePool -> IsolatedEngine` and anything that can't be should be in other factories, like `DefaultEngineFactory -> IEngine`. And if we don't know which properties can't be isolated, then keep a single, minimal interface returned from pools, and let users use the DefaultEngineFactory to get `Engine` instances that can be manipulated without no assumptions about the engine being reused.

lahma:
Changed some names again, here's the results you asked:

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19041.572 (2004/?/20H1)
AMD Ryzen 7 2700X, 1 CPU, 16 logical and 8 physical cores
.NET Core SDK=5.0.100-rc.2.20479.15
  [Host]     : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT
  DefaultJob : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT


```
|                        Method |       Mean |    Error |   StdDev |       Gen 0 |      Gen 1 |      Gen 2 | Allocated |
|------------------------------ |-----------:|---------:|---------:|------------:|-----------:|-----------:|----------:|
|        RunUsingPooledIsolated |   777.3 ms |  7.50 ms |  6.65 ms |  23000.0000 |  1000.0000 |          - |  95.32 MB |
| RunUsingDirtiedEngineInstance |   777.9 ms |  4.03 ms |  3.15 ms |  23000.0000 |  1000.0000 |          - |  95.32 MB |
|     RunUsingNewEngineInstance | 2,517.1 ms | 48.53 ms | 57.77 ms | 150000.0000 | 28000.0000 | 18000.0000 | 1954.5 MB |

lahma:
I was actually reading the specs just today! I'm investigating that API as a way to upgrade AngleSharp.JS to Jint 3 as they want window as global. So far it seems that Realms is only little bit lighter than new engine instance, here this PR tries to allow existing constructs to be reused. Probably needs more investigation. Thanks for the interest, I'm open to ideas and usage scenarios you have in mind!

ejsmith:
This looks really awesome!! What are the chances of this PR landing any time soon?

lahma:
I think we should get initial version of realms/host support first in. This will allow creating a custom host for isolated context (like a browser has) that can inject custom global and participate in execution context creation.

ejsmith:
Hmm... that sounds like it's going to be a long ways off. :-(

sebastienros:
@ejsmith what scenario is really important for you that is covered in this PR?

ejsmith:
We have a multi-tenanted app that uses JS for customization and we pool engine instances but I'm worried that the global scope could be altered and leak to other tenants.

lahma:
Sharing a pool between tenants can be troublesome indeed, just needs a code like this to run `Math.max = () => 1; Math.max(2, 3);`, prints `1`. And that was just a little prank, worse things are of course possible. 

For what's it worth, [we did improve the engine creation performance](https://github.com/sebastienros/jint/pull/625) a while back, but if you have a lot of initialization going on might not suffice for your use case.

lahma:
Created https://github.com/sebastienros/jint/pull/907 for the preliminary work.

iXyles:
Hello,

Out of curiosity, are there any plans for when this might become available?

lahma:
I think we've created a Duke Nukem project here, let's see when it's ready.

lahma:
ShadowRealm was merged to main and it might contain [the functionality](https://2ality.com/2022/04/shadow-realms.html) aimed in this PR, [see Jint test case](https://github.com/sebastienros/jint/blob/main/Jint.Tests/Runtime/ShadowRealmTests.cs?rgh-link-date=2022-05-22T11%3A56%3A17Z) how it isolates a playground from the actual engine.

lahma:
I think `ShadowRealm` fulfils the purpose well enough and is a standard way to handle this. I'll close this experiment for now.

gentledepp:
thanks for the update, @lahma.
Just one question: I cannot find the jint test case anymore. Was that one removed?

lahma:
@gentledepp it was moved to public API test project: https://github.com/sebastienros/jint/blob/main/Jint.Tests.PublicInterface/ShadowRealmTests.cs

gentledepp:
Thank you so much! This is awesome!

PR #794, opened 2020-10-31
Change invalid ICallable to throw TypeError instead of ArgumentException

### LINKED ISSUE #780: Have typed exceptions for Jint errors
When I use Jint to invoke a function that does not exist, it throws an ArgumentException

I'd expect it to throw an exception inherited from JintException like NotCallableException

The code is located at https://github.com/sebastienros/jint/blob/4e86924d7e426b7bb0e7e9360b22600852afec73/Jint/Engine.cs#L665

Otherwise, in my code, I have some strange exception handling like:

            try
            {
                return engine.Invoke(functionName, arguments);
            }
            catch (ArgumentException)
            {
                throw new ProcessExecException(localizer.ExecEngineNotAFunction(functionName));
            }
            catch (JintException e)
            {
                throw new ProcessExecException(localizer.ExecEngineRuntimeError(e.Message));
            }
        
And now I am open to the fact that someone really throws an ArgumentException but not Jint itself and to have a misleading error message for the end user ...

Also, the error message could mention the NAME of the function instead of just "Can only invoke functions"

What do you think ?

lahma:
fixes #780


Jint/AssemblyInfoExtras.cs
```diff
@@ -1,3 +1,4 @@
 ﻿using System.Runtime.CompilerServices;
 
+[assembly: InternalsVisibleTo("Jint.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100bf2553c9f214cb21f1f64ed62cadad8fe4f2fa11322a5dfa1d650743145c6085aba05b145b29867af656e0bb9bfd32f5d0deb1668263a38233e7e8e5bad1a3c6edd3f2ec6c512668b4aa797283101444628650949641b4f7cb16707efba542bb754afe87ce956f3a5d43f450d14364eb9571cbf213d1061852fb9dd47a6c05c4")]
```
sebastienros, line 3: Not really blocking but do we have to pass the full name, and also sign it? I would assume that the simple name would just work. Or is it because it's signed? I don't remember I had to do this for other projects.

lahma: Without it, `Error CS1726 : Friend assembly reference 'Jint.Tests' is invalid. Strong-name signed assemblies must specify a public key in their InternalsVisibleTo declarations.`

PR #798, opened 2020-11-06
Fix indexing JArray / faster IList indexing

### LINKED ISSUE #797: Access JObject in index failed
When I try to access vaue in `items[0]`, it threw error below:
image

To reproduce:
make a json string like
```
{
  "items": [{"p1": 1}],
  "row": 1
}
```
Then use `JObject.Parse(Response.Content)` to get a `JObject`.

Below is my ugly fix, convert string `"0"` to int `0`:
image

lahma:
* protect `ClrTypeConverter` from changing during engine lifetime, should set via options
* make specialized property handlers internal, they're already sealed so let's not expose too much
* improve `TryFindIndexer` to account directly for `IList` and short-circuit there for indexer when parameter is convertible to number
* improve performance by detecting integer array indexer access

fixes #797

### DIFF
--- a/Jint/Runtime/Descriptors/Specialized/IndexDescriptor.cs
+++ b/Jint/Runtime/Descriptors/Specialized/IndexDescriptor.cs
@@ -1,11 +1,12 @@
 ﻿using System;
+using System.Collections;
 using System.Globalization;
 using System.Reflection;
 using Jint.Native;
 
 namespace Jint.Runtime.Descriptors.Specialized
 {
-    public sealed class IndexDescriptor : PropertyDescriptor
+    internal sealed class IndexDescriptor : PropertyDescriptor
     {
         private readonly Engine _engine;
         private readonly object _key;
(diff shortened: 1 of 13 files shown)


Jint/Runtime/Descriptors/Specialized/IndexDescriptor.cs
```diff
@@ -1,47 +1,64 @@
...
             var paramTypeArray = new Type[1];
+            Func<object, IndexDescriptor> ComposeIndexerFactory(PropertyInfo candidate, Type paramType)
+            {
+                if (engine.ClrTypeConverter.TryConvert(propertyName, paramType, CultureInfo.InvariantCulture, out var key))
+                {
+                    // the key can be converted for this indexer
+                    var indexerProperty = candidate;
+                    // get contains key method to avoid index exception being thrown in dictionaries
+                    paramTypeArray[0] = paramType;
+                    var containsKeyMethod = targetType.GetMethod("ContainsKey", paramTypeArray);
```
sebastienros, line 45: nameof?

No biggy since it was like it already.
Another matter for later is to cache all these reflection paths ...

sebastienros (approved):

Jint/Runtime/Descriptors/Specialized/IndexDescriptor.cs
lahma: Good improvement, changed to `nameof` and also included `IDictionary.Contains`

PR #800, opened 2020-11-07
Publish symbols on rel/3.0-beta

lahma:
The symbols didn't get published with the last release as the appveyor configuration was modified for the wrong branch 🤦

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -43,12 +43,13 @@ deploy:
     server: https://www.nuget.org/api/v2/package
     api_key:
       secure: yZBBCLlJTphpHCezRUxyDny1mBbDw7xFG/2Rwt21A8khKp6KJCxFEYx4k9IihOjO
-    skip_symbols: true
-    artifact: /.*\.nupkg/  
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -43,12 +43,13 @@ deploy:
...
-    artifact: /.*\.nupkg/  
+    artifact: /.*\.*nupkg/
   - provider: NuGet
     on:
       branch: master
     server: https://www.nuget.org/api/v2/package
     api_key:
       secure: yZBBCLlJTphpHCezRUxyDny1mBbDw7xFG/2Rwt21A8khKp6KJCxFEYx4k9IihOjO
-    artifact: /.*\.*nupkg/
+    skip_symbols: true
```
sebastienros, line 53: Why do that on master?

lahma: I don't think master has symbols didn't want to break anything as long as it's 2.x.

sebastienros: Ok. It's just that we are envisioning an new stable release, and no new 2.x one on master. So please remember this when we ship 3.0.

PR #809, opened 2020-11-30
Implementing stable sort

### LINKED ISSUE #808: Array.prototype.sort is unstable
Array.prototype.sort do not preserve the ordering of elements if the value is equal. Therefore it is not stable.
But based on this [rfc update](https://github.com/tc39/ecma262/pull/1340), Array.prototype.sort should be stable.

Is it possible to change the implementation of Array.prototype.sort to be stable? It would be great!

sebastienros:
Fixes #808

### DIFF
--- a/Jint/Native/Array/ArrayOperations.cs
+++ b/Jint/Native/Array/ArrayOperations.cs
@@ -70,24 +70,20 @@ namespace Jint.Native.Array
 
         public abstract void DeletePropertyOrThrow(ulong index);
 
-        public IEnumerator<JsValue> GetEnumerator()
-        {
-            return new ArrayOperationsIterator(this);
-        }
+        internal ArrayLikeIterator GetEnumerator() => new ArrayLikeIterator(this);
 
-        IEnumerator IEnumerable.GetEnumerator()
-        {
-            return GetEnumerator();
-        }
+        IEnumerator<JsValue> IEnumerable<JsValue>.GetEnumerator() => new ArrayLikeIterator(this);
 
-        private sealed class ArrayOperationsIterator : IEnumerator<JsValue>
+        IEnumerator IEnumerable.GetEnumerator() => new ArrayLikeIterator(this);
+
+        internal sealed class ArrayLikeIterator : IEnumerator<JsValue>
         {
             private readonly ArrayOperations _obj;
             private ulong _current;
             private bool _first;
             private readonly uint _length;
 
-            public ArrayOperationsIterator(ArrayOperations obj)
+            public ArrayLikeIterator(ArrayOperations obj)
             {
                 _obj = obj;
                 _length = obj.GetLength();
--- a/Jint/Native/Array/ArrayOperations.cs
+++ b/Jint/Native/Array/ArrayOperations.cs
@@ -1,3 +1,5 @@
+using System.Collections;
+using System.Collections.Generic;
 using Jint.Native.Number;
 using Jint.Native.Object;
 using Jint.Runtime;
@@ -5,7 +7,7 @@ using Jint.Runtime.Descriptors;
 
 namespace Jint.Native.Array
 {
-    internal abstract class ArrayOperations
+    internal abstract class ArrayOperations : IEnumerable<JsValue>
     {
         protected internal const ulong MaxArrayLength = 4294967295;
         protected internal const ulong MaxArrayLikeLength = NumberConstructor.MaxSafeInteger;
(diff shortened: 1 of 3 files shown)


lahma:
Seems to hang?

Jint/Native/Array/ArrayPrototype.cs
```diff
@@ -803,71 +804,14 @@ private JsValue Sort(JsValue thisObj, JsValue[] arguments)
...
-                    : null;
-                array[i] = value;
-            }
+            JsValue[] array;
 
             // don't eat inner exceptions
             try
             {
-                System.Array.Sort(array, Comparer);
+                array = obj.OrderBy(x => x, new ArrayComparer(compareFn)).ToArray();
```
lahma, line 814: This will have performance implications 😉

Jint/Native/Array/ArrayPrototype.cs
```diff
@@ -1290,5 +1234,67 @@ public JsValue Pop(JsValue thisObject, JsValue[] arguments)
             o.SetLength(len);
             return element;
         }
+
+        private sealed class ArrayComparer : IComparer<JsValue>
```
lahma, line 1238: I'd argue it's more common to have `sort` called without callback, so might be beneficial to have separate version of that kind of compare which has no state and could use as `ArrayComparer.Instance`

Jint/Native/Array/ArrayOperations.cs
```diff
@@ -68,6 +70,60 @@ public virtual JsValue[] GetAll(Types elementTypes)
...
+        {
+            return new ArrayOperationsIterator(this);
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
+        private sealed class ArrayOperationsIterator : IEnumerator<JsValue>
```
lahma, line 83: Maybe `ArrayLikeIterator`?

Jint/Native/Array/ArrayOperations.cs
```diff
@@ -68,6 +70,60 @@ public virtual JsValue[] GetAll(Types elementTypes)
 
         public abstract void DeletePropertyOrThrow(ulong index);
 
+        public IEnumerator<JsValue> GetEnumerator()
```
lahma, line 73: If there's an overload of concrete type `public ArrayOperationsIterator GetEnumerator()` then there would be no dispatch overhead like the interfaces have, CLR will find the version by name.

Jint/Native/Array/ArrayOperations.cs
```diff
@@ -68,6 +70,60 @@ public virtual JsValue[] GetAll(Types elementTypes)
...
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
+        private sealed class ArrayOperationsIterator : IEnumerator<JsValue>
+        {
+            private readonly ArrayOperations _obj;
+            private ulong _current;
+            private bool _first;
```
lahma, line 87: _first not really needed?

Jint/Native/Array/ArrayPrototype.cs
sebastienros: Before it was creating a copy of the array already, if this is what you mean.
We can't alter the compare function result and use the Array.Sort since it can invoke it like (a,b) or (b,a).
Any other option in mind?

Jint/Native/Array/ArrayPrototype.cs
sebastienros: I agree

Jint/Native/Array/ArrayOperations.cs
sebastienros: ok for both

Jint/Native/Array/ArrayOperations.cs
sebastienros: _current is ulong, and after the first `MoveNext()` it needs to be `0`. Because we can't use -1 I decided to use a bool instead of using a bigger signed struct.

Jint/Native/Array/ArrayOperations.cs
sebastienros: What do you suggest? To implement it in each concrete type and call `base.GetEnumerator()` ?

Jint/Native/Array/ArrayPrototype.cs
lahma: Well LINQ probably is a bit slower, but probably some benchmark would show if there's something to worry about. Unfortunately we don't seem to have a test case for `Array.sort` 🙁 So this is probably fine...

Jint/Native/Array/ArrayOperations.cs
lahma: OK

Jint/Native/Array/ArrayOperations.cs
lahma: I was referring to concrete enumerator method without interface, as seen here:

https://github.com/sebastienros/jint/blob/1d29ae72bbc87406284ef3f00d7d7bd6dabcc97b/Jint/Collections/DictionarySlim.cs#L221-L235

Explicit interface implementations for generic use case and then one of concrete type to avoid boxing (struct via interface).

sebastienros: Updated but this new member is not called, and we need to user OrderBy for a stable sort.

sebastienros:
There is a test that must be blocked in an infinite loop/ Haven't identified which one yet.

sebastienros:
@lahma new review please

lahma (approved):
Looks good and contained. I think that at some point it would make sense to make the actual `ObjectInstance` and `ArrayInstance` implement the `IEnumerable<JsValue>`, this would open resolution for extension methods too.

PR #812, opened 2020-12-03
Support extension methods

### LINKED ISSUE #807: Usage of Extenion Methods
Hi!

As mentioned on Gitter and because of the reply from Marko, i created this Issue regarding the usage of Extension Methods.
I try to explain my pain with the missing support for Extension Methods.
For demo purposes I use the Flurl Http Package as 3rd Party Library which i want to use inside Jint.
Flurl uses a few Extension Methods to create an HttpRequest and also to send the Request.

Following is a simple example:
```csharp
using System;
using System.Threading.Tasks;
using Flurl;
using Flurl.Http;
using Jint;

namespace JintExtensionMethods
{
    class Program
    {
        static async Task Main(string[] args)
        {

            var engine = new Engine(options => options.AllowClr(typeof(Url).Assembly));
            engine.SetValue("flurlHelper", new FlurlHelper());

            var script = @"

var Flurl = importNamespace('Flurl');

var url = new Flurl.Url('https://httpbin.org/anything');

url = flurlHelper
        .AppendPathSegment(url, 'person')
        .SetQueryParams({ a: 1, b: 2 });

var request = flurlHelper.WithOAuthBearerToken(url, 'my_oauth_token');
var result = flurlHelper.PostJsonAsync(request, {
                       first_name: 'Claire',
                       last_name: 'Underwood'
                   }).GetAwaiter().GetResult();

";

            engine.Execute(script);

            var result = engine.GetValue("result").ToObject();

        }

    }

    public class FlurlHelper
    {
        public Url AppendPathSegment(Url url, object segment, bool fullyEncode = false) => url.AppendPathSegment(segment, fullyEncode);
        
        public IFlurlRequest WithOAuthBearerToken(Url url, string token) => url.WithOAuthBearerToken(token);

        public Task<IFlurlResponse> PostJsonAsync(IFlurlRequest request, object data) => request.PostJsonAsync(data);
        
        public Task<string> ReceiveString(Task<IFlurlResponse> response) => response.ReceiveString();
    }
}
```

Because some of the Methods are only available as Extension Method, i created the FlurlHelper class, which only wraps the needed Extension Methods in Instance Methods.
I think it would also be possible to invoke the Static Extension Method directly, but i'm not 100% sure about this.
Another point, which i think is a pain, is that i have to set variables between the different calls, because i have to use the result from the Method before in the next Method...

Here is the same example with the usage of Extension Methods:
```csharp
using System.Threading.Tasks;
using Flurl.Http;
using Jint;

namespace FlurlTest
{
    class Program
    {
        static async Task Main(string[] args)
        {

            {
                options.AddExtensionMethod(typeof(CustomExtensionMethods));
                options.AddExtensionMethod(typeof(Flurl.GeneratedExtensions));
                options.AddExtensionMethod(typeof(Flurl.Http.GeneratedExtensions));
            });

            var script = @"
var result = 'https://httpbin.org/anything'
        .AppendPathSegment('person')
        .SetQueryParams({ a: 1, b: 2 })
        .WithOAuthBearerToken('my_oauth_token')
        .PostJsonAsync({
            first_name: 'Claire',
            last_name: 'Underwood'
         }).GetAwaiter().GetResult();
";

            engine.Execute(script);

            var result = engine.GetValue("result").ToObject();

        }

    }

    public static class CustomExtensionMethods
    {
        public static Task<IFlurlResponse> PostJsonAsync(this FlurlRequest request, object data) => ((IFlurlRequest)request).PostJsonAsync(data);
    }
}
```

For now there is an additional Extension Methods class -> `CustomExtensionMethods`, because Jint was not able to automatically cast `FlurlRequest` to `IFlurRequest` but i think this is another topic...

The same example in C#
```csharp
public static async Task<IFlurlResponse> GetResponse()
{
    return await "https://httpbin.org/anything"
        .AppendPathSegment("person")
        .SetQueryParams(new { a = 1, b = 2 })
        .WithOAuthBearerToken("my_oauth_token")
        .PostJsonAsync(new
        {
            first_name = "Claire",
            last_name = "Underwood"
        });
}
```

This is just an very simple example with Extension Methods, which already exists in the Library.
More painful is to extend existing classes in 3rd Party libraries with "selfmade" Extension Methods.
Until now i always have written a Wrapper around the class which acts like a Proxy and implement my custom extensions there.
Maybe there is a better solution i didn't think of until now?!
 
It would be great if there would be a solution for this in Jint!
If there is anything i can do, to bring this topic forward, i will be more than happy to help!

lahma:
Let's try to get all related things into this PR and work our way to master. I'm sorry if we are unable to find correct attribution for original work when/if this lands to master, but hopefully the functionality compensates for that.

* Register extension methods to JS objects via prototype, make them more lightweight as only one property needed instead of adding to each object which is getting called, extension methods and prototypes share a resemblance after all
* Object wrapper uses fast access path by caching member accessors for extension methods
* Now options if used to post-process the engine how it sees fit, options can be used between engine instances
* JS models kept clean from interop related concerns
* Support all inherited types and implemented interfaces, solving Flur usage
* Change optional parameters to pass `System.Type.Missing` which resolves the default value in method signature

closes #381
closes #616
fixes #807


lahma:
@windischb @boennhoff I think I now have the pieces together, this now also handles extension methods derived from implemented interfaces / base classes, surviving the Flurl test ( #807 ).

I did change the approach a bit not to touch ObjectInstance logic and hooking methods via prototypes when targeting JS object, this has memory benefit as it only needs to be declared per prototype instead of all object instances.

windischb:
Many thanks, that was REALLY fast!!
Can't wait to use this in my Project.
Probably i will check here every hour, until it is merged! :)

sebastienros:
The caching logic looks quite heavy for what I have in mind. I like the fact that each option instance defines which extensions types are available, so why not have a globabl (cross engines) caches per extension types, using and immutable dictionary (no locks, though about the perf differences with CD in read-only concurrent access) indexed by type, and another cache per engine that has the results for the sum of the "allowed" extension types of this engine. Could also be an Immutable dictionary (nit again).

I don't really see the need of these two new cache classes.

lahma:
I simplified the design a bit. Now caching of `type (and-subtypes) -> methods`  is done per options instance as I think that's the sweet spot. Users should cache the options anyway as they can be reused for engines. Now only one `ExtensionMethodCache` which encapsulates handling of the mapping, based on allowed targets in options.

Jint.Tests/Runtime/ExtensionMethods/FlurlExtensionTest.cs
```diff
@@ -0,0 +1,33 @@
+using Flurl.Http;
+
+namespace Jint.Tests.Runtime.ExtensionMethods
+{
+    public class FlurlExtensionTest
+    {
+        [RunnableInDebugOnlyAttribute]
+        public void CanUseFlurlExtensionMethods()
```
sebastienros, line 8: Flaky test (networking), better not have it as long as the same logic is already tested somewhere lese.

lahma: That's why iIt has the `RunnableInDebugOnlyAttribute`, it only runs under debugger.

Jint/Options.cs
```diff
@@ -80,6 +87,45 @@ public Options AddObjectConverter(IObjectConverter objectConverter)
...
+            }
+
+            foreach (var overloads in methods.GroupBy(x => x.Name))
+            {
+                var functionInstance = new MethodInfoFunctionInstance(engine, MethodDescriptor.Build(overloads.ToList()));
+                var descriptor = new PropertyDescriptor(functionInstance, PropertyFlag.None);
+
+                // make sure we register both lower case and upper case
+                prototype.SetOwnProperty(overloads.Key, descriptor);
+                if (char.IsUpper(overloads.Key[0]))
```
sebastienros, line 122: shouldn't it be a method resolution logic? Do we do that for standard methods?

lahma: Unfortunately that only applies to `ObjectWrapper` which does first-char-case-insensitive logic at the moment. We cannot bend the JS spec the same way. Maybe later this could be configurable if someone needs such option.

Jint/Runtime/Interop/Reflection/ExtensionMethodCache.cs
```diff
@@ -0,0 +1,97 @@
...
+                if (_allExtensionMethods.TryGetValue(parentType, out var parentExtensions))
+                {
+                    results.AddRange(parentExtensions);
+                }
+            }
+
+            methods = results.ToArray();
+
+            // racy, we don't care, worst case we'll catch up later
+            Interlocked.CompareExchange(ref _extensionMethods, new Dictionary<Type, MethodInfo[]>(methodLookup)
```
sebastienros, line 66: As long as you copy the _extensionMethods ref when it's iterated over, you don't need to care about this.
Or maybe an ImmutableDictionary since you can't read what you write one with it.

Jint/Options.cs
sebastienros: Ok

Jint/Runtime/Interop/Reflection/ExtensionMethodCache.cs
lahma: This only happens if it's not already there, maybe I don't quite understand,, we are mutating the dictionary if we don't yet have Person class and it's interfaces/sub-classes registered. We don't have ImmutableDictionary elsewhere, maybe some brave soul wants to change all needed places in separate PR 😉

lahma:
Is there more to do? I'd like to get feedback from end users and finishing this and maybe even NuGet release would open the can of worms (things needing to change).

PR #815, opened 2020-12-09
Prioritize Non-Generic methods

### LINKED ISSUE #814: Generic Method Invocation Problem
Hi!

While testing the new support for Extension Methods, i'll found a little bug.
I'm not sure if this only happens with Extension Methods or also with Instance Methods.

Again i'm testing with Flurl.
Flurl has an extension methods class named "ResponseExtensions"
In this class there are following Methods in the same order as below:

```csharp
public static async Task<T> ReceiveJson<T>(this Task<IFlurlResponse> response)
{
    using (IFlurlResponse resp = await response.ConfigureAwait(false))
        return resp == null ? default (T) : await resp.GetJsonAsync<T>().ConfigureAwait(false);
}

public static async Task<object> ReceiveJson(this Task<IFlurlResponse> response)
{
    using (IFlurlResponse resp = await response.ConfigureAwait(false))
        return resp == null ? (object) null : await resp.GetJsonAsync().ConfigureAwait(false);
}
```

If i want to use the second Method (Non-Generic) Jint usees the Generic one.
Seems that Jint just use the first method with matching name.
For testing i have created a custom extension method class with the same methods, but the Non-Generic Method first and registered the custom class before the "ResponseExtensions" class.

Now Jint uses the correct Method (Non-Generic).

Would it be possible the skip Generic Methods?
Or even better, find a Way to support Generic Methods?
Maybe in a Way similar to instantiate a generic object, like mentioned in the readme:

Generic types are also supported. Here is how to declare, instantiate and use a List<string>:
```js
    jint> var ListOfString = System.Collections.Generic.List(System.String);
    jint> var list = new ListOfString();
    jint> list.Add('foo');
    jint> list.Add(1); // automatically converted to String
    jint> list.Count; // 2
```
If it's a Generic Method, the first argument/s are/is the Generic Type/s, followed be the method parameters??

windischb:
As mentioned in #814 Generic Methods will now be last in order.
Prevents invoking the Generic Method instead of the Non-Generic Method of same name, if the Generic Method is declared before.

fixes #814

### DIFF
--- a/Jint/Runtime/Interop/MethodDescriptor.cs
+++ b/Jint/Runtime/Interop/MethodDescriptor.cs
@@ -61,12 +61,12 @@ namespace Jint.Runtime.Interop
             static int CreateComparison(MethodDescriptor d1, MethodDescriptor d2)
             {
                 // if its a generic method, put it on the end
-                if (d1.Method.IsGenericMethod)
+                if (d1.Method.IsGenericMethod && !d2.Method.IsGenericMethod)
                 {
                     return 1;
                 }
 
-                if (d2.Method.IsGenericMethod)
+                if (d2.Method.IsGenericMethod && !d1.Method.IsGenericMethod)
                 {
                     return -1;
                 }
--- a/Jint/Runtime/Interop/MethodDescriptor.cs
+++ b/Jint/Runtime/Interop/MethodDescriptor.cs
@@ -72,12 +72,12 @@ namespace Jint.Runtime.Interop
                 }
 
                 // put params versions to end, they can be tricky to match and can cause trouble / extra overhead
-                if (d1.HasParams)
+                if (d1.HasParams && !d2.HasParams)
                 {
                     return 1;
                 }
 
-                if (d2.HasParams)
+                if (d2.HasParams && !d1.HasParams)
                 {
                     return -1;
                 }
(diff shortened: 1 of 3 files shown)


lahma (approved):

Jint/Runtime/Interop/MethodDescriptor.cs
```diff
@@ -60,6 +60,17 @@ private static MethodDescriptor[] Prioritize(MethodDescriptor[] descriptors)
         {
             static int CreateComparison(MethodDescriptor d1, MethodDescriptor d2)
             {
+                // if its a generic method, put it on the end
+                if (d1.Method.IsGenericMethod)
```
sebastienros, line 64: This is not stable and could potentially create infinite loops with some sort algorithm. The idea being that (a, b) might then be compared again with (b, a) and both would return opposite values if a == b.

If think the test should be that if d1 is generic and not d2, then put it at the end.

Jint/Runtime/Interop/MethodDescriptor.cs
```diff
@@ -60,6 +60,17 @@ private static MethodDescriptor[] Prioritize(MethodDescriptor[] descriptors)
...
+                }
+
+                if (d2.Method.IsGenericMethod && !d1.Method.IsGenericMethod)
+                {
+                    return -1;
+                }
+
                 // put params versions to end, they can be tricky to match and can cause trouble / extra overhead
                 if (d1.HasParams)
                 {
```
windischb, line 76: Isn't then there the same problem with potentially infinite loops?

sebastienros: Probably. This is an issue I found while working on Array.Sort this week. Can you fix it too ?

windischb: off course, i just wanted to ask before

PR #816, opened 2020-12-12
Class support

lahma:


Jint.Tests.Test262/Test262Test.cs
```diff
@@ -230,6 +226,11 @@ public static IEnumerable<object[]> SourceFiles(string pathPrefix, bool skipped)
                                 skip = true;
                                 reason = "async not implemented";
                                 break;
+                            case "class-fields-private":
+                            case "class-fields-public":
+                                skip = true;
+                                reason = "private/public class fields not implemented in esprima";
```
sebastienros, line 232: A gap in esprima, or spec too recent?

Jint.Tests.Test262/Test262Test.cs
```diff
@@ -290,6 +291,48 @@ public static IEnumerable<object[]> SourceFiles(string pathPrefix, bool skipped)
                     reason = "Unicode support and its special cases need more work";
                 }
 
+                if (name.StartsWith("language/statements/class/subclass/builtin-objects/Promise"))
+                {
+                    skip = true;
+                    reason = "Promise not implemented";
```
sebastienros, line 297: WAT?

Jint/Native/RegExp/RegExpConstructor.cs
```diff
@@ -137,40 +137,13 @@ private ObjectInstance RegExpInitialize(RegExpInstance r, JsValue pattern, JsVal
...
-            };
-            return r;
-        }
-
-        public RegExpInstance Construct(string regExp, Engine engine)
-        {
-            var r = new RegExpInstance(Engine);
-            r._prototype = PrototypeObject;
-
-            var scanner = new Scanner(regExp, new ParserOptions { AdaptRegexp = true });
```
sebastienros, line 154: Where did this go?

Jint/Runtime/ExceptionHelper.cs
```diff
@@ -14,6 +15,7 @@ public static T ThrowSyntaxError<T>(Engine engine, string message = null)
             return default;
         }
 
+        [DoesNotReturn]
```
sebastienros, line 18: Saw that in your other PR, perf only? JIT?

Jint/Native/Function/FunctionPrototype.cs
```diff
@@ -65,19 +65,13 @@ private static JsValue HasInstance(JsValue thisObj, JsValue[] arguments)
...
+            if (thisObj is not ICallable)
             {
                 ExceptionHelper.ThrowTypeError(Engine, "Bind must be called on a function");
             }
 
             var thisArg = arguments.At(0);
-            var f = new BindFunctionInstance(Engine)
-            {
-                TargetFunction = thisObj,
-                BoundThis = thisObj is ArrowFunctionInstance ? Undefined : thisArg,
```
sebastienros, line 77: So no more ArrowFunctionInstance, less code repetition then?

Jint/Runtime/Interpreter/Expressions/JintExpression.cs
```diff
@@ -52,6 +55,40 @@ protected virtual void Initialize()
...
+
+                if (expression is MemberExpression memberExpression)
+                {
+                    return ToString(memberExpression.Object) + "." + ToString(memberExpression.Property);
+                }
+
+                if (expression is CallExpression callExpression)
+                {
+                    expression = callExpression.Callee;
+                    continue;
```
sebastienros, line 85: The `while(true)` is only for this case? not a recursion instead?

sebastienros (approved):

Jint.Tests.Test262/Test262Test.cs
lahma: Just a small PR and you can enable this... Just need to have the constructor logic use correct methods, like I mentioned 😉

Jint/Native/RegExp/RegExpConstructor.cs
lahma: It was never used, so removed dead code with extreme prejudice.

Jint/Runtime/ExceptionHelper.cs
lahma: Basically it helps the compiler. Can write code like:

```csharp
string? thisCanBeNull = SomeMethod();
if (thisCanBeNull is null)
   ThrowHelper.Throw();

// compiler won't complain about possibly being null as it knows the earlier branch won't return this code
thisCanBeNull.Substring(1,2);
```

Jint/Native/Function/FunctionPrototype.cs
lahma: Yes, they now follow the same logic. We can separate again later if we want to fine-tune, but spec doesn't really talk about it being a separate concept, just configured differently.

Jint/Runtime/Interpreter/Expressions/JintExpression.cs
lahma: This is honestly a hack. The correct way to handle this later is to have parser enable `options.Range = true` which should give us start and end index, then if we have the script (string instance), we can substring it. Might even make sense to always provide those out of Esprima without config switch if there's  no real performance penalty. So this is something like best-effort, should be replaced with proper solution...

PR #822, opened 2021-01-01
Allow calling Proxy revoke method without this context

KurtGokhan:
[immer](https://github.com/immerjs/immer) does not work without this fix.

As far as I understand from [the spec](https://tc39.es/ecma262/#sec-proxy.revocable), the revoke method should not depend on `this` value. 

Added a test which throws an exception without this fix.

Edit: I found another little bug which prevented defining properties on arrow functions. It was a small fix so adding it to this PR. I encountered this error while trying some packages with Webpack 5. 

For example, this works in browsers while it throws `Type Error: null` in Jint:

```
var a = () => null
Object.defineProperty(a, 'hello', { enumerable: true, get: () => 'world' })
console.assert(a.hello === 'world')
```


lahma:
Can you add a test case for the arrow function case too? I'm working on class support where I've already removed `ArrowFuctionInstance`.

lahma (approved):

lahma:
Thank you!

PR #827, opened 2021-01-14
Fix JSON parser line number initialization

### LINKED ISSUE #823: JSON.parse('') crashes the entire program
```
try {
  JSON.parse('');
} catch(err) {
  console.log('Error')
}
```

If you run this script in browser, it will print `Error` to console. However, Jint will halt the program entirely with `ArgumentOutOfRangeException` and will not go into the `catch` block. With any other invalid JSON, for example `JSON.parse('a')`, it will go into the catch block as expected.

lahma:
Initializing line number always to 1 as Esprima expects line numbers to start from 1.

fixes #823

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -1217,6 +1217,13 @@ myarr[0](0);
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


lahma:
Merged as quite trivial with little risk.

PR #843, opened 2021-02-28
Improve enum overload matching and interop

### LINKED ISSUE #203: TypeConverter.FindBestMatch do not match js number back to enum
c#==================
enum E{a,b }
class A{
   public void testFunc(string e){}
   public void testFunc(E e){}
  }
js=================
var a=new A();
a.testFunc(E.a);  // can not find testFunc(E e){}

E.a will convert to double in js  
when a.testFunc(E.a);  TypeConverter.FindBestMatch can not find a perfect match
but double is close to  c# Enum than c# string
=========modified TypeConverter.FindBestMatch seem to solve the issue========
        public static IEnumerable<MethodBase> FindBestMatch(Engine engine, MethodBase[] methods, JsValue[] arguments)
        {
            methods = methods
                .Where(m => m.GetParameters().Count() == arguments.Length)
                .ToArray();

```
        if (methods.Length == 1 && !methods[0].GetParameters().Any())
        {
            yield return methods[0];
            yield break;
        }
        MethodBase enumMethod = null;
        var objectArguments = arguments.Select(x => x.ToObject()).ToArray();
        foreach (var method in methods)
        {
            var perfectMatch = true;
            var parameters = method.GetParameters();
            for (var i = 0; i < arguments.Length; i++)
            {
                var arg = objectArguments[i];
                var paramType = parameters[i].ParameterType;

                if (arg == null)
                {
                    if (!TypeIsNullable(paramType))
                    {
                        perfectMatch = false;
                        break;
                    }
                }
                else if (arg.GetType() != paramType)
                {
                    perfectMatch = false;
                    if(arg.GetType()==typeof(double)&&paramType.IsEnum)
                    {
                        enumMethod=method;
                    }
                    else 
                    {   enumMethod=null;
                        break;
                    }
                }
            }

            if (perfectMatch)
            {
                yield return method;
                yield break;
            }
        }

        if (enumMethod != null) 
        {
            yield return enumMethod;
            yield break;
        }

        foreach (var method in methods)
        {
            yield return method;
        }
    }
```

### LINKED ISSUE #552: System.InvalidCastException for enum type
I have an user defined enum with the underlying type uint and I get a

System.InvalidCastException in JsValue.cs by

`            if (value.GetType().IsEnum)
            {
                return JsNumber.Create((int) value);
            }
`

lahma:
* handle method matching for enums as perfect match when value in integer and within range of enum
* improve object to JsNumber conversion (to accept uints for example)

fixes #203
fixes #552

### DIFF
--- a/Jint.Tests/Runtime/Domain/Enums.cs
+++ b/Jint.Tests/Runtime/Domain/Enums.cs
@@ -1,0 +1,14 @@
+namespace Jint.Tests.Runtime.Domain
+{
+    public enum IntegerEnum
+    {
+        a,
+        b
+    }   
+    
+    public enum UintEnum : uint
+    {
+        a, 
+        b
+    }
+}
--- a/Jint.Tests/Runtime/Domain/OverLoading.cs
+++ b/Jint.Tests/Runtime/Domain/OverLoading.cs
@@ -1,0 +1,20 @@
+namespace Jint.Tests.Runtime.Domain
+{
+    public class OverLoading
+    {
+        public string TestFunc(string e)
+        {
+            return "string";
+        }
+
+        public string TestFunc(IntegerEnum e)
+        {
+            return "integer-enum";
+        }
+
+        public string TestFunc(UintEnum e)
+        {
+            return "uint-enum";
+        }
+    }
+}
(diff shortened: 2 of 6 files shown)


sebastienros:
Ancient issues ;)

PR #845, opened 2021-02-28
Should not wrap Type instances as TypeReference by default

### LINKED ISSUE #533: NullReferenceException on access of Type property!
Hi folk,

I defined a class like this:

>
     public class UserClass
     {
          public Type TypeProperty { get; set; }
     }
> 

when I excute the code like this:

> 
            var userClass = new UserClass()
            {
                TypeProperty = typeof(UserClass),
            };

            var result = new Engine()
                .SetValue("userclass", userClass)
                .Execute("userclass.TypeProperty.Name;")
                .GetCompletionValue()
                .ToObject();
>

I got a  NullReferenceException  from TypeReference.cs line 204

> return new PropertyDescriptor(new MethodInfoFunctionInstance(Engine, methodInfo.ToArray()), PropertyFlag.AllForbidden);

because "methodInfo" has the value "null".

Thanks

lahma:
I couldn't find a reason why we would want to convert `Type` instances to `TypeReference` instances unless it's explicit need for type construction (`importNamespace`, explicit `TypeReference.Create`). With this change can now read values like `a.GetType().FullName`, earlier it was hidden as `TypeRefereence` which doesn't expose much. Old behavior can still be achieved by engine type mappers / object converters.

The README shows correct ways to achieve this:

```
and then to assign local namespaces the same way System does it for you, use importNamespace

jint> var Foo = importNamespace('Foo');
jint> var bar = new Foo.Bar();
jint> log(bar.ToString());

adding a specific CLR type reference can be done like this

engine.SetValue("TheType", TypeReference.CreateTypeReference(engine, typeof(TheType)))
```

But there's unit test cases that relies on auto-wrapping:

```
 _engine.SetValue("TestStruct", typeof(TestStruct));
```

So I kept the auto-wrapping when called via `engine.SetValue(name, type)`, but no longer automatic in `JsValue.FromObject`.

fixes #533

### DIFF
--- a/Jint.Tests/Runtime/Domain/Person.cs
+++ b/Jint.Tests/Runtime/Domain/Person.cs
@@ -1,9 +1,13 @@
-﻿namespace Jint.Tests.Runtime.Domain
+﻿using System;
+
+namespace Jint.Tests.Runtime.Domain
 {
     public class Person : IPerson
     {
         public string Name { get; set; }
         public int Age { get; set; }
+
+        public Type TypeProperty { get; set; } = typeof(Person);
 
         public override string ToString()
         {
--- a/Jint.Tests/Runtime/InteropTests.cs
+++ b/Jint.Tests/Runtime/InteropTests.cs
@@ -57,6 +57,20 @@ namespace Jint.Tests.Runtime
                 assert(z === 'foo');
             ");
         }
+        
+        [Fact]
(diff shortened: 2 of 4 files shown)


sebastienros:
I'll trust your judgement because this is maybe some of the oldest code that is still remaining.

PR #852, opened 2021-03-23
Debugger improvements 3

Jither:
__ETA: Summary of changes/fixes:__

* Support for stepping through return point (i.e. for inspecting return value)
* Support for inspecting scope chain in all call frames of call stack (i.e. for inspecting values in the caller(s))
* CallStack in debugger now has most information needed for devtools protocol (this, function location, location of execution, return value, scope for all call frames...)
* Scope chain changed from "Globals/Locals" into resembling Chromium/devtools protocol
* Lazy instantiation of scope chains in `DebugInformation`
* Using Jint's own `CallStack` simplifies `DebugHandler`
* All debugging-related properties, events etc. are moved from `Engine` to `DebugHandler`.
* Prevent reentrancy in `Step/Break`.
* Fix bug where `return` statements with literal argument would not call `DebugHandler.OnStep`.
* Tests and documentation

This obviously includes (entirely) breaking changes to the Debugging API.

__Original post:__
Just getting back to have another look at the debugger.

One fix I made pretty soon after the last debugger pull request was merged, was adding a guard against Pause reentrancy. It's explained in the test below, but the gist is that Step/Break event handlers may want to do evaluation through the engine (e.g. for conditional breakpoints, or evaluating getters). Doing so would cause reentrance, and since interactive debuggers will be on a separate thread, that would cause a deadlock.

The other part for now is reinstating all the skipped tests for Debugger handling of accessors. They all "automagically" pass after the call stack improvements. 😃 (after fixing a typo in one)

I've marked the pull request as a draft (but I'm fine with marking it ready if wanted), since I do intend to also have a look at proper handling of scopes. Changes in the runtime there means that consts will "disappear" from the current `Locals`/`Globals` when they're initialized - might be time to get proper division into global/local/block/catch/etc. Also going to look through the updated code for best way to add a debugger event after the last statement but before a call returns.

I've tagged this "debugger-improvements-3", in spite of there being no "2", since there were actually two debugger-improvements pull requests in the past.


Jither:
The refactor of Call and Construct really simplifies things a lot. I've removed DebugHandler's internal management of the call stack, and use Engine's instead. This does mean I've added an internal property to `JintCallStack` in order to iterate over the private stack and create a representation (with shortDescription, Location, and Arguments) for public access in `DebugInformation`.

sebastienros:
Nothing bothers me with this PR. That's a good think I guess.

lahma:
Great to hear that the call stack work reflected also to debug handler positively, looking good! If you find ways to improve the current call stack handling please feel free to do so.

Jither:
> Great to hear that the call stack work reflected also to debug handler positively, looking good! If you find ways to improve the current call stack handling please feel free to do so.

The only issue I've seen so far with the call stack handling is that it doesn't quite seem to collaborate with try/catch - the stack isn't popped appropriately. I.e., here, thrower isn't popped when catching, so after leaving `testCatch`, that also stays on the stack. That's so deep into the engine code that I don't dare try to fix it. 😄

image

Other than that, well into proper scope handling in the debugger. A few observations:

* Do we try to emulate Chromium (e.g. for future devtools protocol support)? Chromium seems to throw out all outer scopes in e.g. nested functions. My current code keeps them, filtering out bindings shadowed by inner scopes - since Chromium doesn't, that might not actually be very helpful (i.e., I've never missed being able to inspect the variables in outer functions). See dump below (and yeah, `testCatch` remaining in the call stack strikes again in that one).
* Am I right that `GlobalEnvironment` includes both block scoped and function scoped variables? (Chromium assigns let/const to a "Script" scope, separate from "Global").
* So far, Global, Block, Local, Catch, and With scopes seem to be working in early, simple tests. I've made a (hopefully) small adjustment to the `DeclarativeEnvironmentRecord` (an optional `_catchEnvironment` constructor parameter and internal field similar to `_withEnvironment` on ObjectEnvironmentRecord) in order to be able to distinguish it from other declarative records. For `with`, I just assume an ObjectEnvironmentRecord is a With environment, if it isn't a GlobalEnvironment.
* I can't seem to find anything that distinguishes Closure scopes from other declarative scopes. Haven't looked into how the context is created for them yet. Any ideas?
* Haven't looked into Eval scope yet - if it's even feasible to implement. 😄

image

lahma:
> The only issue I've seen so far with the call stack handling is that it doesn't quite seem to collaborate with try/catch - the stack isn't popped appropriately. I.e., here, thrower isn't popped when catching, so after leaving `testCatch`, that also stays on the stack.

If you can formulate a test case showing the problem it would be a good issue in the backlog to take care of 👍🏻 

> Do we try to emulate Chromium (e.g. for future devtools protocol support)? Chromium seems to throw out all outer scopes in e.g. nested functions. My current code keeps them, filtering out bindings shadowed by inner scopes - since Chromium doesn't, that might not actually be very helpful (i.e., I've never missed being able to inspect the variables in outer functions). See dump below (and yeah, testCatch remaining in the call stack strikes again in that one).

I don't have strong opinions in the overall debugger experience, but there's a PR unfortunately left behind https://github.com/sebastienros/jint/pull/354 that has the Chrome protocol that should be resurrect by any means if possible. So not an answer but a remark.

> Am I right that GlobalEnvironment includes both block scoped and function scoped variables? (Chromium assigns let/const to a "Script" scope, separate from "Global").

Maybe 😄 GlobalEnvironment is populated by global functions and `var`s , `var`s flow to global scope (and top level const/let), there's currently no script scope concept in Jint. 

> I can't seem to find anything that distinguishes Closure scopes from other declarative scopes. Haven't looked into how the context is created for them yet. Any ideas?

Not sure if this is what you're after, but if it's an arrow function then `_thisBindingStatus = ThisBindingStatus.Lexical` for `FunctionEnvironmentRecord`.

https://github.com/sebastienros/jint/blob/b27153726faf4158b8aa8001d630861076340a93/Jint/Runtime/Environments/FunctionEnvironmentRecord.cs#L29-L44

> Haven't looked into Eval scope yet - if it's even feasible to implement. 

I wouldn't keep that at the top of priorities, `eval` is a bit evil as is things like `with`. Anything in strict mode feels better than relaxed evaluation.

The debugger looks great!

Jither:
> If you can formulate a test case showing the problem it would be a good issue in the backlog to take care of 👍🏻

Added two cases in #853 

> I don't have strong opinions in the overall debugger experience, but there's a PR unfortunately left behind #354 that has the Chrome protocol that should be resurrect by any means if possible. So not an answer but a remark.

I think the DebugAgent part of that PR should be very helpful. The changes to the debugger and engine itself may have been either implemented since, or the engine has probably changed too much for a merge.
 

> Maybe 😄 GlobalEnvironment is populated by global functions and `var`s , `var`s flow to global scope (and top level const/let), there's currently no script scope concept in Jint.

Yeah - for now, I'll just assign them to Global scope.
 
> Not sure if this is what you're after, but if it's an arrow function then `_thisBindingStatus = ThisBindingStatus.Lexical` for `FunctionEnvironmentRecord`.

Will have a look later. 😃 

> I wouldn't keep that at the top of priorities, `eval` is a bit evil as is things like `with`. Anything in strict mode feels better than relaxed evaluation.

Only a "bit" evil? But then, their very evilness is what might make proper display of eval and with scope useful in some cases (forensics, e.g.). But yeah, I mainly looked at them for completion's sake.

> The debugger looks great!

Thanks 😃 It's still very finicky (doesn't reset the engine if adding a new script etc.), but it's useful enough for testing outside of the project I'm using Jint for (where the debugger is more evolved, but tied to other app internals).

Jither:
Not really sure why the latest commit fails: `Jint.Tests.Runtime.InteropTests.ShouldReturnUndefinedProperty` fails, but that test doesn't even touch any modified code as far as I can tell, and the test doesn't fail when run locally...

> I don't have strong opinions in the overall debugger experience, but there's a PR unfortunately left behind #354 that has the Chrome protocol that should be resurrect by any means if possible. So not an answer but a remark.

I'm thinking I'll refactor the DebugInformation from Jint to align more with devtools protocols, then maybe at some point have a look at whether #354 can be used to get devtools support.
 
> Not sure if this is what you're after, but if it's an arrow function then `_thisBindingStatus = ThisBindingStatus.Lexical` for `FunctionEnvironmentRecord`.

Not quite 🙂 Chromium's debugger has 9 scopes (global, local, with, closure, catch, block, script, eval, module, wasm-expression-stack). Currently, `global`, `local`, `with`, `catch`, and `block` are implemented in this PR. `script`, `module`, `wasm-expression-stack` aren't implemented in Jint (yet). That leaves `evil` and `closure`.

`closure` in Chromium lists any outer local (i.e. function) scope + its child block scope(s) as a combined list, but only displaying the bindings that are actually referenced from the inner scope. But it turns out that's due to v8 optimization - with related side effects actually being listed as `WorkingAsIntended` issues: https://bugs.chromium.org/p/v8/issues/detail?id=3491
So, I guess I might just combine outer local/block scopes (for readability) into `closure`, and not try to emulate a Chromium feature that would be undesirable in many situations. 😄 

Other than that, redesign of DebugInformation, and tests (which I've put off, because most of them would be using `DebugScopes` heavily, which hasn't really been properly designed yet), most of the intended functionality for this PR is now there in the latest commit (even if that one test inexplicably fails on appveyor).

Jither:
One step towards preparing the debug handling for devtools protocol (and make it more useful in general): Providing scope chains for all call stack frames, not just the *current* frame. To that end, when creating the `DebugInformation` version of the call frames (CallStackElements), I need the `ExecutionContext` for that frame - in order to inspect the `LexicalEnvironment` chain. So, I've done a quick fix for that - adding `CallingExecutionContext` to `CallStackElement`. If that's totally obscene, let me know. 😉 If nothing else, it's less invasive than adding the call's execution context after it's created in `Call`/`Construct`. But there are probably better ways.

The generation of the `DebugInformation` version of scope chains (and some other parts of DebugInformation) has been made lazy - the lazy way (by using `Lazy<T>`). May change it to a simple manual lazy instantiation if these experiments turn out to be solid (since we don't really need thread safety etc.). I'll probably also need to look closer at which parts can be reused, rather than re-instantiating everything for every Pause.

Other than that, the scope chains are now closer to Chromium, collapsing top-level `let`, `const` and `var` into a single "Local" scope (or "Closure" if it's an outer function).

Jither:
I think this is now in a state where it's ready for review. All functionality within scope (no pun intended) of this PR are done now, and test coverage is decent (although I may think up more).

In the latest commit, had to circumvent some optimization of `JintReturnStatement` (fast resolves) when DebugMode is true - since `OnStep` will only be called if the `return` statement actually has `Execute` called. More explanation of this in the change in `JintStatementList`. This isn't new for this PR's implementation - return statements with literal arguments are also skipped in the current version of Jint.

Jint/Runtime/Debugger/DebugCallStack.cs
```diff
@@ -0,0 +1,45 @@
...
+    public class DebugCallStack : IReadOnlyList<CallFrame>
+    {
+        private List<CallFrame> _stack;
+        public CallFrame this[int index] => _stack[index];
+
+        public int Count => _stack.Count;
+
+        internal DebugCallStack(Engine engine, Location location, JintCallStack callStack, JsValue returnValue)
+        {
+            _stack = new List<CallFrame>();
```
lahma, line 22: could initialize this to size of `callStack.Stack.Count + 1`

Jint/Runtime/Debugger/DebugCallStack.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
```
lahma, line 4: some unused usings here

Jint/Runtime/Debugger/CallFrame.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            _element = element;
+            _context = context;
+            Location = location;
+            ReturnValue = returnValue;
+
+            _scopeChain = new Lazy<DebugScopes>(() => new DebugScopes(Environment));
+        }
+
+        private JsValue GetThis()
```
lahma, line 61: looks a lot like `Engine.GetThisEnvironment`, could these share logic somehow?

Jint/Runtime/Debugger/DebugHandler.cs
```diff
@@ -1,85 +1,102 @@
...
+            // Don't reenter if we're already paused (e.g. when evaluating a getter in a Break/Step handler)
+            if (_paused)
+            {
+                return;
+            }
 
-            _debugCallStack.Push(name);
-        }
+            Location location = statement.Location;
+            BreakPoint breakpoint = _engine.BreakPoints.FirstOrDefault(breakPoint => BpTest(location, breakPoint));
```
lahma, line 38: not related to this PR but wonder why `Engine.BreakPoints` is a public list, probably would require better abstraction and offer `BpTest` as internal service instead of these context-capturing LINQ queries

Jint/Runtime/Debugger/DebugScope.cs
```diff
@@ -0,0 +1,54 @@
...
+using Jint.Native;
+
+namespace Jint.Runtime.Debugger
+{
+    /// <summary>
+    /// Dictionary of bindings for a single scope in the scope chain
+    /// </summary>
+    public class DebugScope : IReadOnlyDictionary<string, JsValue>
+    {
+        private IReadOnlyDictionary<string, JsValue> variables;
```
lahma, line 12: could probably have this as `Dictionary<string, JsValue>` as construction is internal, concrete type as field gives best access performance

Jint/Runtime/Debugger/DebugScope.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// </summary>
+        public DebugScopeType ScopeType { get; }
+
+        internal DebugScope(DebugScopeType type, IReadOnlyDictionary<string, JsValue> variables)
+        {
+            ScopeType = type;
+            this.variables = variables;
+        }
+
+        #region IReadOnlyDictionary implementation
```
lahma, line 25: ugh, regions..

Jint/Runtime/Debugger/DebugScope.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// Dictionary of bindings for a single scope in the scope chain
+    /// </summary>
+    public class DebugScope : IReadOnlyDictionary<string, JsValue>
+    {
+        private IReadOnlyDictionary<string, JsValue> variables;
+
+        /// <summary>
+        /// The type of scope. Scope types are the same as defined by Chrome devtools protocol.
+        /// </summary>
+        public DebugScopeType ScopeType { get; }
```
lahma, line 17: ideally order would be fields, constructor, properties, methods

Jint/Runtime/Debugger/CallFrame.cs
```diff
@@ -0,0 +1,77 @@
+﻿using System;
+using Esprima;
+using Esprima.Ast;
+using Jint.Native;
+using Jint.Runtime.CallStack;
+using Jint.Runtime.Environments;
+
+namespace Jint.Runtime.Debugger
+{
+    public class CallFrame
```
lahma, line 10: should be sealed by default and opened when someone has good reasoning, less to worry about how being used in the wild

Jint/Runtime/Debugger/DebugCallStack.cs
```diff
@@ -0,0 +1,45 @@
...
+using Esprima;
+using Esprima.Ast;
+using Jint.Native;
+using Jint.Runtime.CallStack;
+using Jint.Runtime.Environments;
+
+namespace Jint.Runtime.Debugger
+{
+    public class DebugCallStack : IReadOnlyList<CallFrame>
```
lahma, line 13: sealed by default

Jint/Runtime/Debugger/DebugScope.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using Jint.Native;
+
+namespace Jint.Runtime.Debugger
+{
+    /// <summary>
+    /// Dictionary of bindings for a single scope in the scope chain
+    /// </summary>
+    public class DebugScope : IReadOnlyDictionary<string, JsValue>
```
lahma, line 10: sealed by default

Jint/Runtime/Debugger/DebugInformation.cs
```diff
@@ -1,16 +1,45 @@
...
+using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
+using Esprima;
 using Esprima.Ast;
 using Jint.Native;
 
 namespace Jint.Runtime.Debugger
 {
     public class DebugInformation : EventArgs
```
lahma, line 11: we should probably seal this too

Jint/Runtime/Debugger/DebugScopes.cs
```diff
@@ -0,0 +1,162 @@
+﻿using Jint.Native;
+using Jint.Runtime.Environments;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Jint.Runtime.Debugger
+{
+    public class DebugScopes : IReadOnlyList<DebugScope>
```
lahma, line 9: sealed by default

Jint/Runtime/Debugger/DebugScopes.cs
```diff
@@ -0,0 +1,162 @@
...
+    {
+        private readonly HashSet<string> _foundBindings = new HashSet<string>();
+        private readonly List<DebugScope> _scopes = new List<DebugScope>();
+
+        /// <summary>
+        /// Shortcut to Global scope
+        /// </summary>
+        public DebugScope Global { get; private set; }
+        /// <summary>
```
lahma, line 18: maybe newline should be in between

Jint/Runtime/Debugger/DebugScopes.cs
```diff
@@ -0,0 +1,162 @@
...
+                        break;
+                    default:
+                        _foundBindings.Add(name);
+                        bindings.Add(name, jsValue);
+                        break;
+                }
+            }
+        }
+
+        #region IReadOnlyList implementation
```
lahma, line 145: maybe remove region

Jint/Runtime/Interpreter/JintStatementList.cs
```diff
@@ -68,7 +68,19 @@ public Completion Execute()
                 foreach (var pair in _jintStatements)
                 {
                     s = pair.Statement;
-                    c = pair.Value ?? s.Execute();
+
+                    // @lahma: Am I completely ruining optimization with this check? We need to Execute return statements in debug
+                    // mode, even if they're fast-resolved. Otherwise debug stepping will always step past return statements.
+                    // Could also call OnStep directly here, if value isn't null, and we're in debug mode, but that seems more prone
+                    // to introduce errors with future developments.
+                    if (!_engine._isDebugMode)
```
lahma, line 76: doesn't look too bad, could also skip fast resolve in Initialize when debug mode with comments about the need, I hope tests cover so nobody optimizes the correct behavior away 😉

lahma (commented):
Added some nits and remarks.

Jint/Runtime/Debugger/DebugCallStack.cs
Jither: Yeah - went for `callStack.Count` now (`callStack.Stack` is `IEnumerable`).

Jint/Runtime/Debugger/CallFrame.cs
Jither: Would it be OK to move `GetThisEnvironment` to `LexicalEnvironment` (since `CallFrame` needs it for different execution contexts), or keep it in `Engine` but add an optional `LexicalEnvironment` parameter?

Jint/Runtime/Debugger/DebugHandler.cs
Jither: Must admit, I'm wondering why `BreakPoints` is even a property on `Engine`, which never uses it. 🙂 Other than for ease of access in API. But I might try to move it closer to the `DebugHandler` for this PR - now that it's breaking all other debugging APIs anyway.

Jint/Runtime/Debugger/DebugScope.cs
Jither: This one is a "bit" unfinished, since we may not want a dictionary in the first place. It's somewhat left over, and probably overkill - the only real reason for its creation is to leave out shadowed bindings in outer scopes. But that might as easily be done with just a List of the names, with value being retrieved on-demand. Which would also be better for the most common use case - iteration of all of the bindings (for display in `Scopes` panel). Especially since we may want to order them in the future. For example, currently `Global` scope lists all intrinsic global properties first, with user variables - which are arguably more "important" - coming last.

Jint/Runtime/Debugger/DebugScope.cs
Jither: Oh, I never write them (or like them) either. 🤣 Except I explicitly added these two, just become *sometimes* I like easily getting boilerplate interface implementation out of the way. Sorry. 😉 They'll be gone.

Jint/Runtime/Debugger/DebugScope.cs
Jither: I'm a slob about order of members. Going through and fixing. 🙂

Jint/Runtime/Debugger/DebugInformation.cs
Jither: I've gone ahead and also gotten rid of public setters, adding an `internal` constructor to `DebugInformation`. I'm not sure what the original reasoning behind `CurrentMemoryUsage` is - it's never set in the engine. But devtools protocol does have a `getHeapUsage` method, so might be useful in future - keeping it for now, and adding an XMLDoc about it not being useful now.

Jint/Runtime/Interpreter/JintStatementList.cs
Jither: https://github.com/sebastienros/jint/blob/10851bfaca176435bac61cf0d3ed0e6bc3f92476/Jint.Tests/Runtime/Debugger/CallStackTests.cs#L99 does cover it indirectly - will fail if `OnStep` doesn't get called for the `return 'result';` line. But I'll add an explicit test for good measure. Moving debug mode check to `Initialize` - it's also more likely to be spotted there for any future initialization optimizations.

Jint/Runtime/Debugger/CallFrame.cs
lahma: Fine to move.  Only name is important as it's referred to in the spec algorithms.

Jint/Runtime/Debugger/DebugHandler.cs
lahma: Sounds good 👍🏻

Jint/Runtime/Debugger/DebugScope.cs
Jither: Thinking further on the API of DebugScope...

I may drop the flattening of `FunctionEnvironmentRecord` and its child `DeclarativeEnvironmentRecord` into "Local/Closure" scope, and let it be up to the caller to flatten them if wanted. Although V8 does this flattening, and it's nice for clarity, I think it's mainly a result of internal V8 optimization rather than a "user friendliness choice". Basically like how Jint includes `Script` top-level block scoped bindings in `Global`. And how V8 also removes closure bindings that aren't needed.

And the flattening adds some amount of unneeded complexity to the iteration of bindings - and inefficiency to random access value retrieval - in each scope:

Since each binding name may be resolved in a different record, at the very least the `DebugScope` would need to iterate over each of its records, then their associated binding names, just for enumeration. And random access retrieval of value would require a `HasBinding` call (the most efficient way I can think of, most often just involving a Dictionary lookup - rather than needing to look up in the ordered list of binding names in `DebugScope`).

So, probably better to have 1 environment record per scope. There may also be cases where the caller would *want* more granular scopes.

The only issue when the caller *wants* to flatten the scopes, is that because `DebugScopes` leaves out empty scopes, you cannot distinguish top-level block scoped variables (which should be flattened) from an empty top-level with block scoped variables in a `BlockStatement` inside. But that could be solved with a boolean `TopLevel` property on the `DebugScope`.

Might also leave the addition of `this` to the Local scope to the caller - since some callers may not actually want it included (devtools doesn't) - and it's available from `CallFrame.This` (where devtools wants it).

Jither: Clarifying the flattening vs no flattening visually:

Flattened a la Chromium:
image

Not flattened:
image

Jint/Runtime/Debugger/DebugBindings.cs
```diff
@@ -0,0 +1,97 @@
...
+        internal DebugBinding(string name, JsValue value)
+        {
+            Name = name;
+            Value = value;
+        }
+    }
+
+    internal sealed class DebugBindings : IReadOnlyList<DebugBinding>
+    {
+        private class Enumerator : IEnumerator<DebugBinding>
```
Jither, line 23: Enumerator just to avoid yield return-based enumeration

Jint/Runtime/Debugger/DebugScope.cs
```diff
@@ -1,50 +1,64 @@
...
+        /// <summary>
+        /// Names of all non-shadowed bindings in the scope.
+        /// </summary>
+        public IReadOnlyList<string> BindingNames => _bindingNames;
 
-        public bool TryGetValue(string key, out JsValue value)
-        {
-            return variables.TryGetValue(key, out value);
-        }
+        /// <summary>
```
Jither, line 52: IReadOnlyList doesn't imply that the items are lazily evaluated and `DebugBinding` instantiated on each call. However, seems "silly" that accessing via `IEnumerable` would add the overhead of IEnumerator, but actually do the iteration by using the index into `BindingNames`. That means that you'd be quite a bit better off by iterating over `BindingNames` yourself, and using `DebugScope[index]` to get the value...

Jither: Leaning towards dropping `Bindings` list entirely, and only having `BindingNames` + a `GetValue(name)` method and/or `this[name]` indexer. Getting rid of the allocations of `DebugBinding`s are another reason to do that, since most callers will likely want their own class for name + value anyway.

Jint/Runtime/Debugger/DebugHandler.cs
Jither: Moved all debugger related properties, events etc. into `DebugHandler`, and turned that into a public property. The only "issue" with that is that the `sender` of Step/Break events should conventionally be `DebugHandler` then, but that's rather useless (but then, `sender` sometimes is...) - since it only has the two public events and the `BreakPoints` collection. What's the verdict on keeping the `sender` as `Engine`? Or should the two events and their invocations remain in `Engine`?

Jint/Runtime/Debugger/CallFrame.cs
Jither: Moved to `ExecutionContext`. I've changed calls to call `(_engine).ExecutionContext.GetThisEnvironment` minimizing method calls - but let me know if you prefer a stub `Engine.GetThisEnvironment` calling `ExecutionContext.GetThisEnvironment` instead (guessing it can be inlined anyway).

That should be all in terms of addressing the remarks - although there's plenty of new stuff to remark on now. 😆

Jint.Tests/Runtime/Debugger/BreakPointTests.cs
```diff
@@ -138,11 +138,11 @@ public void BreakpointOverridesStepOut()
 
             var engine = new Engine(options => options.DebugMode());
 
-            engine.BreakPoints.Add(new BreakPoint(4, 0));
-            engine.BreakPoints.Add(new BreakPoint(6, 0));
+            engine.DebugHandler.BreakPoints.Add(new BreakPoint(4, 0));
```
lahma, line 141: `DebugHandler.AddBreakPoint` and `DebugHandler.RemoveBreakPoint` might offer cleaner encapsulation, the list could probably be private and only offer IReadOnlyList maybe?

Jint/Runtime/Debugger/BreakPoint.cs
```diff
@@ -2,31 +2,31 @@
...
+        public BreakPoint(string source, int line, int column, string condition) : this(source, line, column)
         {
             Condition = condition;
         }
 
         public string Source { get; }
         public int Line { get; }
-        public int Char { get; }
+        public int Column { get; }
         public string Condition { get; }
```
lahma, line 30: not required for this PR, but I think we should implement equality at some point, would also ensure uniqueness

Jint/Runtime/Debugger/BreakPointCollection.cs
```diff
@@ -0,0 +1,99 @@
+﻿using Esprima;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Jint.Runtime.Debugger
+{
+    public class BreakPointCollection : ICollection<BreakPoint>
```
lahma, line 7: ah now I see that it's a separte datatype, so `AddBreakPoint` not needed

lahma (approved):
All looking good to go, @sebastienros could you also please check this?

Jint/Runtime/Debugger/BreakPoint.cs
Jither: Yes, would also be needed, if `FindMatch` is turned into a dictionary lookup - which it could be, if we require that Char/Column are accurate.

Jint/Runtime/Debugger/BreakPointCollection.cs
Jither: Yeah, the reason it "needs" to be at least a readable collection is that currently the only way to remove a breakpoint again is passing a reference to it - which e.g. in my case - means nasty LINQ `SingleOrDefault` shenanigans, rather than keeping my own list of the added breakpoints. Until `BreakPoint` has been looked through (equality etc.), I didn't see much idea in adding e.g. a `Remove(int line, int column, string source = null)` method. :-)

lahma:
@sebastienros your friendly scheduled reminder here,  can we merge? 😉

sebastienros:
Looks better than before, right?

sebastienros (approved):

sebastienros:
I don't think we have that, or correct me, a list of sources and information about it. Like all the scripts an engine would be loaded with, such that we could also render where the line of code is coming from (filename for instance). I might have implemented it already or at least partially, sounds familiar ;) Just saying it could help a debugger open the sources from a running engine.
Then an component could also expose and endpoint to communicate with it, and a debugger would connect "remotely", using something like  the chrome protocol.

lahma:
I believe @Jither has something cool in the works regarding the Chrome debug protocol 😎

lahma:
Thanks @Jither for the hard work and hopefully you have time to further improve with ideas you've listed!

Jither:
@sebastienros 
> Looks better than before, right?

I hope so :wink: Thanks for approving, again 😃 

> I don't think we have that, or correct me, a list of sources and information about it. Like all the scripts an engine would be loaded with, such that we could also render where the line of code is coming from (filename for instance). I might have implemented it already or at least partially, sounds familiar ;) Just saying it could help a debugger open the sources from a running engine.

You've already implemented it, I think, in both Esprima.NET and Jint - `Esprima.Location` includes `Source` (which is a somewhat non-descript name to write other code around, but that's what the ESTree spec calls it, so it wouldn't really be an improvement to rename it now). Jint asks Esprima.NET to set `Source` to "<anonymous>" by default when not passing our own parser options, but it could really be any arbitrary string identifying the script - internally, V8 uses numeric IDs.

Breakpoints already use `Location.Source` (and did even before I threw a grenade into the debugger code :wink:) - breakpoint matching is based on line, column and source (if breakpoint source is null, any Location source will match).

The same property works fine for communication with devtools - since devtools allows any string as `ScriptId`, the application using Jint can assign filenames, paths, URLs, or incremental IDs (like V8, as far as I recall) if they want.

> Then an component could also expose and endpoint to communicate with it, and a debugger would connect "remotely", using something like the chrome protocol.

Already done :wink: - although it needs a lot of work still. And it's based on a fork that's already slightly further ahead than this PR (and which I just managed to mess up by being overeager with force push, so reconstructing now).

Jither:
@sebastienros - if interested, I've now reconstructed the branch that adds further changes to debugging for devtools protocol (and debugging in general). The very-much-work-in-progress (but working in terms of a proof of concept) repo for the devtools protocol endpoint is here: https://github.com/Jither/Jint.DevToolsProtocol

PR #853, opened 2021-03-24
Tests of CallStack unwinding after caught exception (and quick patch)

Jither:
As mentioned in #852, callstack fails to unwind after catching an exception. Not quite sure where this should be handled. Included two tests, one just showing (by way of nested calls) that it's the entire callstack that fails to unwind - it's not just off by 1.

### DIFF
--- a/Jint.Tests/Runtime/CallStackTests.cs
+++ b/Jint.Tests/Runtime/CallStackTests.cs
@@ -14,19 +14,20 @@ namespace Jint.Tests.Runtime
         {
             var engine = new Engine();
             engine.Execute(@"
-function thrower()
-{
-    throw new Error('test');
-}
+                function thrower()
+                {
+                    throw new Error('test');
+                }
 
-try
-{
-    thrower();
-}
-catch (error)
-{
-}
-");
+                try
+                {
+                    thrower();
+                }
+                catch (error)
+                {
+                }
+                "
+            );
             Assert.Equal(0, engine.CallStack.Count);
         }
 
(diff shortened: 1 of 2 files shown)


Jither:
About this issue... Dug a bit into the Engine, probably with too little understanding of it, but...

I think *ideally*, `Engine.Call` (and conversely, `Construct`)
https://github.com/sebastienros/jint/blob/5b5103633452e6704180ab694e235ebd2d17b290/Jint/Engine.cs#L1274-L1283
... should have something like:

            JsValue result;
            try
            {
                result = functionInstance.Call(thisObject, arguments);
            }
            finally
            {
                CallStack.Pop();
            }
            return result;

... to ensure that the call stack is always popped. This will yield the correct call stack when entering a `catch` (and anywhere else I can think of). However, it interferes with the stack trace in `JavascriptException` if there's no `catch` in the script:

In the case I've checked (`StackTraceCollectedOnThreeLevels`), although the stack trace is built when first throwing the Exception, when that exception is caught here:

https://github.com/sebastienros/jint/blob/5b5103633452e6704180ab694e235ebd2d17b290/Jint/Runtime/Interpreter/JintStatementList.cs#L84-L89

... it's converted to a `Completion`, which is then handled in `Engine Execute(Script script)` -  turning it *back* into a `JavascriptException` again, and replacing the original stack trace with a new one, which now no longer fits the call stack at the original site of the error:

https://github.com/sebastienros/jint/blob/5b5103633452e6704180ab694e235ebd2d17b290/Jint/Engine.cs#L379-L383

... if we unwind the stack through `finally`, that is. Currently it works only because the `CallStack` is left as is, if any Exception is thrown during execution.
 
But still, not really sure what to do about it - the back and forth of `JavascriptException` <-> `Completion` looks fishy to me. 😄 I guess an alternative to `finally` would be to do the stack unwinding specifically in JintTryStatement's catch-handling. But that seems more hack-y and obviously less general than the `finally` approach. Even if there may be no other cases where the current code would fail in practice.

lahma:
Yes the current call stack handling is quite tricky if not brittle when it comes to the exception chain and re-throw and completions. It would definitely require some love. I think ideally (JS) exceptions should be allowed to flow through until the very end where at last step the completion would be handled.

Jither:
Added a quick patch for the catch issue - but like I write in the comment, it's not actually meant as a solution - may just clarify what the issue actually is, better than my walls of text 😉 

Agree on the optimal solution being more consistent completions/throws.

lahma (approved):
This looks like an acceptable hack for time being until someone has the courage to redo some if the nested exception handling.

PR #854, opened 2021-03-25
Allow delegates to return values from .NET

derekdiamond:
Removed the void returning block in the delegate handling in function arguments.
This allows expressions like:

 **list.Sum(x => x.cost)**

To be correctly passed through to .NET. Before the change it would say that no matching function found.


lahma:
Could you also add a test case in InteropTests to ensure it doesn't break later on?

lahma (approved):

lahma:
Thank you!

PR #858, opened 2021-04-07
Allow access to declared Members on DynamicObject

### LINKED ISSUE #856: Can't access declared Methods on "DynamicObject" since 3.0.0-beta-2019
Hi!

Since Version 3.0.0-beta-2019, i can't access Methods which i have declared on DynamicObjects.
Looks like #842 has changed those behaviour

```csharp
using System;
using System.Dynamic;
using Jint;

namespace JintDynamicObject
{
    class Program
    {
        {
            var engine = new Engine();
            engine.SetValue("dyn", new SimpleDynamicObject());

            /// Works, starting with version 3.0.0-beta-2019
            var dynProp = engine.Execute("dyn.Prop = 'DynProp'; dyn.Prop").GetCompletionValue().AsString();

            /// Works, until version 3.0.0-beta-2014
            var dynName = engine.Execute("dyn.Set('Name','DynName'); dyn.ContainsKey('Name')").GetCompletionValue().AsBoolean();

        }
    }

    public class SimpleDynamicObject : DynamicObject
    {
        private readonly Dictionary<string, object> _properties = new Dictionary<string, object>();

        public override bool TryGetMember(GetMemberBinder binder, out object result)
        {
            result = binder.Name;
            if (_properties.TryGetValue(binder.Name, out var value))
            {
                result = value;
            }
            return true;
        }

        public override bool TrySetMember(SetMemberBinder binder, object value)
        {
            _properties[binder.Name] = value;
            return true;
        }

        public void Set(string key, object value)
        {
            _properties[key] = value;
        }

        public bool ContainsKey(string key)
        {
            return _properties.ContainsKey(key);
        }


    }
}

```

windischb:
Fixes #856 
Just moved the part which checks for DynamicObject after checking for Declared Members.
Extended the Test "CanAccessDynamicObject" to also access Declared Members.

Hopefully this is helpfull and not just nonsense...

### DIFF
--- a/Jint.Tests/Runtime/InteropTests.cs
+++ b/Jint.Tests/Runtime/InteropTests.cs
@@ -2344,12 +2344,17 @@ namespace Jint.Tests.Runtime
             Assert.Equal("a", engine.Execute("test.a").GetCompletionValue().AsString());
             Assert.Equal("b", engine.Execute("test.b").GetCompletionValue().AsString());
 
-            engine.Execute("test.a = 5; test.b = 10;");
+            engine.Execute("test.a = 5; test.b = 10; test.Name = 'Jint'");
 
             Assert.Equal(5, engine.Execute("test.a").GetCompletionValue().AsNumber());
             Assert.Equal(10, engine.Execute("test.b").GetCompletionValue().AsNumber());
+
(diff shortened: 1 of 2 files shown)


lahma (approved):

lahma:
Awesome, thanks!

PR #859, opened 2021-04-08
Fix block-scoped function registrations

### LINKED ISSUE #857: Problem with using functions when Strict option is enabled, since 3.0.26
Hi
The following test passes in version 3.0.25 (and older), but started failing since 3.0.26 
```c#
using Jint;
using Jint.Native;
using Xunit;

namespace Tryouts
{
    public class Program
    {
        public static void Main()
        {
            var script = @"
function execute(doc, args){
    var i = doc;
    {
        function doSomething() {
            return 'ayende';
        }

        i.Name = doSomething();
    }
}
";

            {
                options.Strict();
            });
            engine.Execute(script);

            var user = new User();
            var args = new []
            {
                JsValue.FromObject(engine, user)
            };

            var call = engine.GetValue("execute").TryCast<ICallable>();
            call.Call(Undefined.Instance, args); // throws JavaScriptException: 'doSomething is not defined', since version 3.0.26


            Assert.Equal("ayende", user.Name);

        }

        private class User
        {
        }
    }
}


```

Note that if we remove from the script the curly brackets that start after `var i = doc;` then the test passes. 
Same if `Strict` option is disabled.

lahma:
fixes #857

### DIFF
--- a/Jint/Runtime/Interpreter/JintStatementList.cs
+++ b/Jint/Runtime/Interpreter/JintStatementList.cs
@@ -113,19 +113,19 @@ namespace Jint.Runtime.Interpreter
         /// </summary>
         internal static void BlockDeclarationInstantiation(
             LexicalEnvironment env,
-            List<VariableDeclaration> lexicalDeclarations)
+            List<Declaration> declarations)
         {
             var envRec = env._record;
             var boundNames = new List<string>();
-            for (var i = 0; i < lexicalDeclarations.Count; i++)
+            for (var i = 0; i < declarations.Count; i++)
             {
-                var variableDeclaration = lexicalDeclarations[i];
+                var d = declarations[i];
                 boundNames.Clear();
-                variableDeclaration.GetBoundNames(boundNames);
+                d.GetBoundNames(boundNames);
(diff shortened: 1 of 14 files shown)


Jint/Runtime/Interpreter/JintStatementList.cs
```diff
@@ -113,19 +113,19 @@ public Completion Execute()
...
-                var variableDeclaration = lexicalDeclarations[i];
+                var d = declarations[i];
                 boundNames.Clear();
-                variableDeclaration.GetBoundNames(boundNames);
+                d.GetBoundNames(boundNames);
                 for (var j = 0; j < boundNames.Count; j++)
                 {
                     var dn = boundNames[j];
-                    if (variableDeclaration.Kind == VariableDeclarationKind.Const)
+                    if (d is VariableDeclaration { Kind: VariableDeclarationKind.Const })
```
sebastienros, line 128: you couldn't resist

lahma: Rider/R# made me do it 😇

PR #864, opened 2021-04-10
Support converting js function into any delegate.

Dogwei:
1.Compared with the previous scheme, more delegate types are now supported.
2.Shorter implementation.
3.There are still some unforeseen errors unhandled, such as ChangeType may not supported, or ref struct type.


lahma (approved):
Code did indeed get a lot cleaner and old test cases still work. We'll see if there are any problems with some cases we don't currently cover. Thank you!

PR #866, opened 2021-04-11
Fix detecting extension method when there are none

### LINKED ISSUE #865: Trying to get an object's property modifies that property of object
When a property is tried to be accessed in an `ObjectWrapper`, that property is modified. Suppose something like this:

```
// Suppose that 'value' is an ObjectWrapper and it does not have 'prop' property
console.log('prop' in value)  // Always true
console.log(value.prop) // Prints  System.Func`3[Jint.Native.JsValue,Jint.Native.JsValue[],Jint.Native.JsValue]
```

I tracked down the problem to this line:
https://github.com/sebastienros/jint/blob/dev/Jint/Runtime/Interop/ObjectWrapper.cs#L233

Is it supposed to be like this or is there a mistake?

KurtGokhan:
Fixes #865


lahma (approved):

lahma:
Great, thank you!

PR #872, opened 2021-04-16
Override Protoype Functions

### LINKED ISSUE #871: Override Protoype Functions
Hi!

Thanks for solving my problem with invoking Prototype Functions if an ExtensionMethod with same Name is registered -> #851 
But unfortunately that brings up another "problem".
Now it's not possible to use any ExtensionMethod with the same name, even if the arguments are different.
I have take a look and found a possible Solution, but i'm not sure if its ok from your side to modify the MethodInfoFunctionInstance class.
It's a little bit extended to hold a ClrFunctionInstance as fallback and if no ExtensionMethod with the provided arguments is found, it invokes the Prototype function.
With that it's also possibly to override prototype functions if an extension method with suitable arguments is registered.
Also on that point i don't know if this is a desired feature, or if it's a no-go to override prototype functions...

I'll create a Pull Request, so you can take a look if this is a reasonable way...

windischb:
fixes #871 
I dont' know why there a few newlines and removed spaces in the pull request, sorry...
I'll try to find the reason and remove that.

### DIFF
--- a/Jint.Tests/Runtime/ExtensionMethods/CustomStringExtensions.cs
+++ b/Jint.Tests/Runtime/ExtensionMethods/CustomStringExtensions.cs
@@ -26,5 +26,12 @@ namespace Jint.Tests.Runtime.ExtensionMethods
         {
             return Array.Empty<string>();
         }
+
+        public static string[] Split(this string value, int position)
+        {
+            var first = value.Substring(0, position);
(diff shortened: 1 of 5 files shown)


lahma (approved):
Looking good, this interop case seems to be a bit of can of worms 😉

PR #875, opened 2021-04-21
Fix missing NewTarget handling for basic types

### LINKED ISSUE #873: "constructor" property is not set correctly for class instances
Here is the minimal repro

```js
class MyArr extends Array {
    constructor(...args) {
        super(...args);
    } 
}
const a = new MyArr(1,2);
a.constructor === MyArr // should be true but false in Jint
```

It seems that the same applies for Errors and other places.

I notices that while working on promises, that has tests on it as a part of 262 suit.

Note: I'm not sure if the same applies for function ctors, it is possible that the issue is not unique to classes.

lahma:
* fix `OrdinaryCreateFromConstructor` cases mentioned in spec
* fix error constructor names

fixes #873

### DIFF
--- a/Jint.Tests/Runtime/ErrorTests.cs
+++ b/Jint.Tests/Runtime/ErrorTests.cs
@@ -155,10 +155,10 @@ var x = b(7);";
    at a (v) <anonymous>:2:18
    at b (v) <anonymous>:6:12
    at <anonymous>:9:9";
-            
+
             EqualIgnoringNewLineDifferences(expected, ex.ToString());
         }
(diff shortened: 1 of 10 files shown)


sebastienros (commented):
Are there some new tests we can enable?

sebastienros (approved):

lahma:
> Are there some new tests we can enable?

Unfortunately I was not able to find any specific ones. They pop up with the new features. Might be a different case with the new test suite (Jint's is years old).

PR #881, opened 2021-05-04
Add support for Optional Chaining

### LINKED ISSUE #863: Any plans to implement ECMA 2020 Optional Chaining?
Have very compelling need to support optional chaining/existential dereferencing (`myObject.someValue?.childValue` and `myObject.someMethod?.()`).  If this isn't on the roadmap yet, I'm definitely willing to devote my (or even some of my development team's) time to building it out, but this is a hefty codebase that I don't have experience with and any rough guidance on approach would be super helpful.

lahma:
* align some code more closer to spec like eval and records
 * reuse identifier/member resolution and function calls and add support for optional chain, we already have this optimized a bit different than in spec
 * added test cases from 262 suite

fixes #863


Jint.Repl/Program.cs
```diff
@@ -57,18 +57,22 @@ private static void Main(string[] args)
                 Console.ForegroundColor = defaultColor;
                 Console.Write("jint> ");
                 var input = Console.ReadLine();
-                if (input == "exit")
+                if (input is "exit" or ".exit")
```
lahma, line 60: I made the command line a bit more node like, wanted to also see statement execution results like in node console

sebastienros (approved):

PR #884, opened 2021-05-04
Fix instanceof logic to comply with specification

### LINKED ISSUE #874: "instanceof" operator doesn't work according to the latest spec
https://tc39.es/ecma262/#sec-instanceofoperator

```js
 class MyArr extends Array {
    constructor(...args) {
        super(...args);
    } 
}
const a = new MyArr(1,2);
a instanceof MyArr // should be true but false in jint
```

Not sure if this is unique to classes or a a more general problem

From spec:
>1. If Type(target) is not Object, throw a TypeError exception.
>2. Let instOfHandler be ? GetMethod(target, @@hasInstance).
>3. If instOfHandler is not undefined, then
>a. Return ! ToBoolean(? Call(instOfHandler, target, « V »)).
>4. If IsCallable(target) is false, throw a TypeError exception.
>5. Return ? OrdinaryHasInstance(target, V).

My understanding that jint does just the step 5 and step 3 doesn't work properly.

lahma:
Added tests cases, went trough logic and aligned with spec. The original test case didn't fail in dev either though...

fixes #874

### DIFF
--- a/Jint/Native/JsValue.cs
+++ b/Jint/Native/JsValue.cs
@@ -404,6 +404,30 @@ namespace Jint.Native
             return ExceptionHelper.ThrowNotSupportedException<bool>();
         }
 
+        /// <summary>
+        /// https://tc39.es/ecma262/#sec-instanceofoperator
+        /// </summary>
+        internal bool InstanceofOperator(JsValue target)
+        {
+            if (target is not ObjectInstance oi)
+            {
+                return ExceptionHelper.ThrowTypeErrorNoEngine<bool>("not an object");
+            }
+
+            var instOfHandler = oi.GetMethod(GlobalSymbolRegistry.HasInstance);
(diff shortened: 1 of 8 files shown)


Jint/Native/JsValue.cs
```diff
@@ -566,6 +590,43 @@ internal virtual JsValue DoClone()
 
         internal virtual bool IsCallable => this is ICallable;
 
+        /// <summary>
+        /// https://tc39.es/ecma262/#sec-ordinaryhasinstance
```
sebastienros, line 594: I was wondering such a weird name ... well it's the specs

sebastienros (approved):

Jint/Native/JsValue.cs
lahma: Indeed, I had trouble accepting it too 😄

PR #643, opened 2019-07-31
Complete Promise implementation

miroslav22:
Branch that provides a complete implementation of a Promise as per the specs.

Only missing method is Promise.allSettled() which it seems isn't widely supported in browsers at the moment.

The Promises are implemented, built upon and wrap .net Tasks

If the return value of a CLR method returns a Task or Task<T> then this is automatically converted to a Promise in the run time.  This allows you to now use fully asynchronous javascript code in Jint.

GetCompletionValueAsync() has been added to Engine which awaits and unwraps the execution completion value if it's a Promise type.  If the completion value isn't a Promise then it's just returned as per GetCompletionValue().

Due to the nature of the Promise continuations I've had to add thread locking to the engine execution, as otherwise it'd be possible for multiple threads to execute code concurrently within the same engine.

I've added a fairly comprehensive set of tests which covers all the specs as far as I can tell.

This opens the door for async/await keywords to be supported (as they essentially use promises 'under the hood').  Is it possible for the Esprima library to be updated to support parsing of async/await keywords and methods?  If so I'm more than happy to add support for async/await in Jint too.

PR #889, opened 2021-05-07
Promise lets go again

twop:
> This is a draft PR.

Upd: not draft anymore

reapplied initial changes from https://github.com/sebastienros/jint/pull/643. But I  only kept scheduling logic in Engine and test s, thus this is a complete rewrite internally.

Done:
- Updated 262 tests from the spec main branch
- Ability to run tests on Mac (there were file path issues and time zones handling platform differences)
- All sync Promise 262 tests are passing 
- Old tests are reworked according to spec and fixed.

Left:
- Interop/public API for creating Promises. Right now you can't create a Promise instance in C# as a library user.
- Potentially "opt-in" into Promise support. There is a lock right now in `Execute` method wether you use Promises or not. I think that this should be optional and opt-in (you don't pay the price by default).

All feedback is welcomed.

Note: I believe that `Promise` should not interop with `Task` smoothly, at least not initially. I think a better version will be something along these lines
```csharp
var  (resolve, reject) = PromiseInstance.CreateManual(engine);

// then resolve it later like so

resolve(new JsString("resolved!")); // note that type of resolve is Action<JsValue>
```

Upd: all what is left is done now and PR is "marked ready for review"


Jint/Extensions/IteratorExtensions.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+using System.Collections.Generic;
+using Jint.Native;
+using Jint.Native.Iterator;
+using Jint.Native.Object;
+using Jint.Runtime;
+
+namespace Jint.Extensions
```
twop, line 8: This file is not used anymore 

- [ ] delete it

lahma:
Initial thought is that maybe it would make sense to bring the updated test suite as a separate PR so this one would be easier to concentrate on, now it's hiding the actual thing be having so many files.

twop:
> Initial thought is that maybe it would make sense to bring the updated test suite as a separate PR so this one would be easier to concentrate on, now it's hiding the actual thing be having so many files.

There is a way to filter by file extensions, I had the same concern but just reviewed myself by using the `.js` filter. Was manageable

lahma:
> There is a way to filter by file extensions, I had the same concern but just reviewed myself by using the `.js` filter. Was manageable. Please let me know if it works for you.
> 
> I couldn't keep the old tests and enable them at the same time because some of them were not working according to the spec

Ah, TIL, thanks for the tip. Let's try out this route.

Jint/Runtime/PromiseRejectedException.cs
```diff
@@ -0,0 +1,15 @@
+using Jint.Native;
+
+namespace Jint.Runtime
+{
+    public class PromiseRejectedException : JintException
```
lahma, line 5: maybe sealed?

lahma (changes requested):
Overall looking great, pedantic follow of the spec and good comments. Things noted are mostly about hiding as much as possible (internal) as it gives us more wiggle room. I think a thing to figure out is the started task, how do we make sure it's not forgotten and as it's per engine there can be a lot of them created.

Jint/Engine.cs
```diff
@@ -381,6 +467,28 @@ public JsValue GetCompletionValue()
         {
             return _completionValue;
         }
+        
+        /// <summary>
+        /// Gets the last evaluated statement completion value.  If the completion value is a promise then the method asynchronously waits for the promise to resolve and returns the resolve result
+        /// </summary>
+        /// <returns></returns>
+        public async System.Threading.Tasks.Task<JsValue> GetCompletionValueAsync()
```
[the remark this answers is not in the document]
twop: I like that idea will see if feasible

Jint/Engine.cs
```diff
@@ -381,6 +467,28 @@ public JsValue GetCompletionValue()
...
+                _completionValue = await promise.ToTask();
+
+            return _completionValue;
+        }
+
+        /// <summary>
+        /// Asynchronously waits for a completion promise to resolve.  Returns immediately if the completion value is not a promise.
+        /// </summary>
+        /// <returns></returns>
+        public async System.Threading.Tasks.Task WaitForCompletionAsync()
```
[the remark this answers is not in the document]
sebastienros: Agree with adding Evaluate and EvaluateAsync.

Jint/Jint.csproj
```diff
@@ -9,5 +9,6 @@
   <ItemGroup>
     <PackageReference Include="Esprima" Version="2.0.0-beta-1338" />
     <PackageReference Include="Nullable" Version="1.3.0" PrivateAssets="all" />
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
```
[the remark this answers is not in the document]
twop: that's a pity :(

Jint/Native/JsValue.cs
```diff
@@ -196,7 +208,8 @@ internal bool TryGetIterator(Engine engine, out IIterator iterator)
             }
 
             var obj = callable.Call(this, Arguments.Empty) as ObjectInstance
-                      ?? ExceptionHelper.ThrowTypeError<ObjectInstance>(engine, "Result of the Symbol.iterator method is not an object");
```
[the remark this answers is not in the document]
twop: I have 13 inch air, lines longer than 130 chars don't fit :( .
line breaks are coming from Rider formatting

Jint/Native/Promise/PromiseConstructor.cs
```diff
@@ -0,0 +1,526 @@
...
+using Jint.Native.Object;
+using Jint.Native.Symbol;
+using Jint.Runtime;
+using Jint.Runtime.Descriptors;
+using Jint.Runtime.Descriptors.Specialized;
+using Jint.Runtime.Interop;
+
+namespace Jint.Native.Promise
+{
+    public struct PromiseCapability
```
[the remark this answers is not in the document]
twop: Why make then heap allocated?

Jint/Native/Promise/PromiseConstructor.cs
```diff
@@ -0,0 +1,526 @@
...
+            return obj;
+        }
+
+        protected override void Initialize()
+        {
+            const PropertyFlag propertyFlags = PropertyFlag.Configurable | PropertyFlag.Writable;
+            const PropertyFlag lengthFlags = PropertyFlag.Configurable;
+            var properties = new PropertyDictionary(5, checkExistingKeys: false)
+            {
+                ["resolve"] =
```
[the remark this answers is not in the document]
twop: autoformatting

Jint/Jint.csproj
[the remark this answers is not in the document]
twop: Pls let me know if I should remove it

Jint/Native/Promise/PromiseConstructor.cs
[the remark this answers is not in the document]
lahma: Well I didn't mean to make it a class, less class-likey, so readonly struct with constructor and read-only fields, reduces likelihood of defensive copies.

Jint/Jint.csproj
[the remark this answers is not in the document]
lahma: I think custom (readonly) structs is they way to go as long as we support old framework like net461, if I remember right net471 is first to correctly handle binding redirects, other versions had some trouble that stirred quite the discussion for libraries.

Jint/Collections/Tuple.cs
```diff
@@ -0,0 +1,20 @@
+namespace Jint.Collections
+{
+    public readonly struct Tuple<A,B>
```
lahma, line 3: internal probably

twop: I wanted to use it in public API (which I haven't flushed out yet)

Jint/Engine.cs
```diff
@@ -319,56 +329,132 @@ public void ResetCallStack()
...
+                    _continuationsTaskActive = true;
                 }
-
-                _completionValue = result.GetValueOrDefault();
             }
 
-            return this;
+            if (startContinuationsTask)
+            {
+                System.Threading.Tasks.Task.Factory.StartNew(async () =>
```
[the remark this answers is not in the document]
twop: removed `Task` dependency all-together

Jint/Runtime/EventLoop.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+using System.Collections.Generic;
+
+namespace Jint.Runtime
+{
+    public sealed class EventLoop
```
lahma, line 6: internal?

twop: Likely, as of right now it is a return value of `ExecuteWithEventLoop` method, I'm debating to not expose that at all. Tentatively it will be moved to internal if I come up with ergonomic API 

If I can get away with the following I will make it internal
```csharp 
public Engine ExecuteWithEventLoop(
   Script script,
   Action onFinished
);
```

lahma: `Execute` methods are there to allow chaining, I think promise execution would get away with the `JsValue EvaluateXXX` kind of method as it's all evaluation without (hopefully) need to get engine instance in chain call

twop:
@lahma do you understand what is not working now in CI?

sebastienros:
All tests pas but exit code is 1, weird. Will check tomorrow. Could be an issue in the runner that doesn't accept skipped tests, or an issue that we don't see in the logs, or a breaking change in the sdk.

lahma:
At least for me locally Jint.Tests hangs, might be that xUnit killed the run.

twop:
@sebastienros @lahma I added draft version of public API for creating Promises. There is a test that demonstrates it: `RegisterPromise_CalledWithinExecuteWithEventLoop_ResolvesCorrectly`

twop:
thoughts about public API are more than welcome :)

lahma:
I might be way off base here, but some thoughts, feel free to shoot them down (I'm just writing from end-user perspective). `EventLoop` is implementation details so I don't think it's good thing to expose in API. This is partly pseudo-code, using things that might not be there (await is shorter example).

```csharp
Engine engine = new Engine(options => options.EnablePromiseEvaluation()); // if really needed
JsValue result1 = engine.Evaluate("1 + 1");

// translates to event loop if enabled
JsValue result2 = engine.Evaluate("await 1;"); , 

// above translates to something like, ideally enables event loop when first seen that required
-> engine.InternalEventLoop(() => manualResetEvent.Set(true)); manualResetEvent.Wait();

// then there's possibility maybe later to configure per call, not part of this PR
engine.Evaluate("await 1", options => options.EnablePromiseEvaluation = true);
```

So trying to hide as much about the internals for now, I'm leaving postfix etc out of scope now. The evaluate allow configuring the execution context parameter we've talked before, which would flow through all calls etc, call level configuration.

Having to manually state EnablePromiseEvaluation (enable event loop) seems odd to me, we could just automatically enable when first seeing need to do so and then if started, always wait to end? Probably with timeout.

twop:
@lahma I think usage of EventLoop is fundamentally different than just `Execute`, specifically **it assumes long running tasks**. I think because it is fundamentally different from synchronous "input -> output" of normal scripting that it justifies explicit API for it. 

I'm open to any suggestions and ideas but I think I would love to land the current public API with the caveat that it is experimental and subject to change

Jint/Runtime/EventLoop.cs
twop: Marked EventLoop as internal. It is no longer exposed

Jint/Engine.cs
```diff
@@ -382,6 +534,28 @@ public JsValue GetCompletionValue()
             return _completionValue;
         }
 
+        /// <summary>
+        /// Gets the last evaluated statement completion value, if it is a Promise unwraps it
+        /// </summary>
+        public JsValue GetPromiseCompletionValue()
```
sebastienros, line 540: Why isn't it part of `GetCompletionValue()` directly? (unless I find the reason in one of your older comments)

twop: Then it will seem "funny" relative to other places.
e.g.
```csharp
engine.Execute("Promise.resolve(1);").GetCompletionValue()
```
Will return a `PromiseInstance` or `1`?

So I decided that it makes more sense to have it as explicit unwrapping. It is possible that it should only work with Promises. like so

```csharp
 public JsValue UnwrapPromise(PromiseInstance promise)
 ```

What do you think?

**Review by lahma — APPROVED** [review]

File: Jint/Engine.cs (hunk 12 above)
Comment (sebastienros) [reply Jint/Engine.cs:544]: In this case the promise, and if you do "await Promise.resolve(1)" then it returns the value.

File: Jint/Engine.cs [hunk 13]
```diff
@@ -369,6 +400,131 @@ public Engine Execute(Script script)
...
+        /// EXPERIMENTAL! Subject to change.
+        /// 
+        /// Executes a script and then tries to process all schedules events.
+        /// It triggers onFinished callback when there are no scheduled events in the EventLoop
+        /// and when there are no user registered promises left unresolved.
+        /// </summary>
+        /// <param name="script"></param>
+        /// <param name="onFinished">called when EventLoop had no more events neither pending user promises</param>
+        /// <returns></returns>
+        public Engine ExecuteWithEventLoop(Script script, Action onFinished)
```
sebastienros, line 413: What is a use case for `onFinished`?

Jint/Engine.cs
twop: `await Promise.resolve(1)` is semantically equivalent to just `Promise.resolve(1)`. So I would personally expect them to behave the same way

sebastienros: Try it in the chrome dev tools (or Safari?), and you'll see the difference there too. Even MDN says so: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve#return_value
and so does the spec: https://262.ecma-international.org/10.0/#sec-promise-resolve

Can't beat that, right?

sebastienros: Just to say that it won't be a surprise if the Promise is not unwrapped automatically. Only when `await` is used on it.

twop: My understand is `await Promise.resolve(1)` is using top-level await feature. Without it `await Promise.resolve(1)` is not even valid syntax, at least that is my current understanding

Jint/Engine.cs
twop: `onFinished` is a way to determine if everything is settled, but it can probably be tracked in user space, because the only reason why it can be not settled are user created promises

twop: killed it

Jint/Engine.cs
twop: settled with "public JsValue UnwrapIfPromise(this JsValue value)" extension method on `JsValue`

ChrML:
I don't think opt-in Promise support is necessary. It should be enabled by default.

A lock in a single-threaded environment is very cheap as they will complete immediately with spinlocks, without involving the OS. Locks only have significant negative impact in multithreaded environments in areas with concurrency.

PR #891, opened 2021-05-08
NUKE build

lahma:
Idea is to use the awesome NUKE build to create build system configured via C# and then use auto-generated workflow definitions to run on GitHub Actions and AppVeyor. Ideally we get the Linux and MacOS builds covered too.


sebastienros:
I'd rather just copy the two files from .github in my other repos that work very well and are only a few lines of code. Change my mind.

lahma:
I'll resurrect this when we need some more logic...

PR #892, opened 2021-05-08
Run engine tests

sebastienros:


lahma:
Windows, Linux and MacOS will make a lovely combo 👌🏻

sebastienros:
Focusing on Linux only for now, will add the matrix when it works. Though not sure it should run everything, will depend on the time it takes. Or maybe just for the Publish step.

lahma:
Some build logic probably helps, maybe in c# 🤔

sebastienros:
Everything seems to work, even the publish task using the tags (from any branch). I just used a fake API key to test.

Jint.Tests.CommonScripts/Jint.Tests.CommonScripts.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net461;net5.0</TargetFrameworks>
+    <TargetFrameworks>net5.0</TargetFrameworks>
```
lahma, line 3: would be nice to have windows included and also net461 tested

lahma (commented):
Looking good. Would be nice to have the build matrix setup for windows and macos too. I guess with this we won't easily get the release notes generation based on commits and contributors document generation, but maybe just basic yaml then instead of new era build automation 😛

sebastienros:
Added windows builds, running met461 only on it since 5.0 is tested on linux. That keeps the PR builds under 5 minutes.
Also not running on windows for publication.

sebastienros:
Re-added net5.0 on windows

PR #893, opened 2021-05-11
Implement Operator Overloading

KurtGokhan:
Normally, Ecmascript does not allow operator overloading and [the specs](https://tc39.es/ecma262/#sec-addition-operator-plus) are pretty clear on how to handle operators. But since we are running on C#, we may as well use operators that are defined on the C# side. So I added an option for that. The option is off by default.

Binary operators will try to find a C# side overloaded operator first and use it. If it cannot find, will fall back to Ecmascript behavior. (I forgot to do it for unary so the PR is draft for now. Will do asap.)
Also, when this option is turned on, the default type converter will try to find implicit/explicit cast operator and use it as a last resort.

### DIFF
--- a/Jint/Runtime/Interpreter/Expressions/JintBinaryExpression.cs
+++ b/Jint/Runtime/Interpreter/Expressions/JintBinaryExpression.cs
@@ -1,7 +1,5 @@
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
-using System.Globalization;
 using System.Linq;
 using System.Reflection;
 using Esprima.Ast;
@@ -14,8 +12,8 @@ namespace Jint.Runtime.Interpreter.Expressions
     internal abstract class JintBinaryExpression : JintExpression
     {
 #if NETSTANDARD
-        private static readonly ConcurrentDictionary<(string OperatorName, Type Source, Type Target), MethodDescriptor> _knownOperators = 
-            new ConcurrentDictionary<(string OperatorName, Type Source, Type Target), MethodDescriptor>();
+        private static readonly ConcurrentDictionary<(string OperatorName, System.Type Left, System.Type Right), MethodDescriptor> _knownOperators = 
+            new ConcurrentDictionary<(string OperatorName, System.Type Left, System.Type Right), MethodDescriptor>();
 #else
         private static readonly ConcurrentDictionary<string, MethodDescriptor> _knownOperators = new ConcurrentDictionary<string, MethodDescriptor>();
 #endif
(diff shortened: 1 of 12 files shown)


Jint/Runtime/Interpreter/Expressions/JintBinaryExpression.cs
```diff
@@ -17,6 +30,85 @@ private JintBinaryExpression(Engine engine, BinaryExpression expression) : base(
             _right = Build(engine, expression.Right);
         }
 
+        protected abstract object EvaluateBinaryExpression();
+
+        protected override object EvaluateInternal()
```
lahma, line 35: this part is a bit iffy for me, this is on performance hot path and now introduces another layer of indirection

lahma: might require sealing, also minimize code size for non-default case, i.e. put if `if (!_engine.Options._IsOperatorOverloadingAllowed)` first and then call separate method for doing the operator overloading version?

lahma (commented):
A great addition with a perfect use case demonstrated by the test, thanks for preparing this! I had just one comment of possible performance impact.

Jint/Runtime/Interpreter/Expressions/JintBinaryExpression.cs
KurtGokhan: Removed the calls to virtual methods. Please check now.

KurtGokhan:
Done all (most) binary and unary methods.

I did not implement operator overloading for [true and false operators](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/true-false-operators). I guess that operator can be used in a context where Ecmascript spec asks to convert the value to boolean. For example `if(here)` or `here && ...`. I would implement it but I don't know where to start yet. Also seems like it may affect performance if not done right. I can do it in the future as another PR.

lahma (approved):

lahma:
Looking good, @sebastienros please take a look.

Jint/Runtime/TypeConverter.cs
```diff
@@ -634,24 +638,47 @@ public static void CheckObjectCoercible(Engine engine, JsValue o)
...
-                            // no can do
-                            perfectMatch = false;
-                            break;
+                            if (paramType.IsAssignableFrom(argType))
+                            {
+                                score -= 1;
+                            }
+                            else
+                            {
+                                if (argType.GetMethods(BindingFlags.Public | BindingFlags.Static)
```
sebastienros, line 647: Is there a test for that and the `scrore` thing to understand what is happening?

KurtGokhan: Basically this method finds the best match when there are multiple methods with the same name and number of arguments. Before my changes, this code was checking for a method that was perfectly matching to the passed arguments. If it could not find a perfect match, it would choose the first method it encountered (and this would cause an error in most cases).

My code gives each method a penalty based on how close each argument is to the parameter it is going to be casted to. Score is divided into different levels and assigned with a number in base 10. This is kind of a heuristic algorithm.

This is demonstrated in `MethodAmbiguityTests.cs`.

To be honest, this became deeper than I initially thought. It deserves its own PR. I did not want to go overkill here. C# compiler can resolve the best method even better. But that is probably only possible with static analysis. Also when there are multiple matching methods, C# calls it an ambiguity and does not let you compile.

Also this can become a breaking change for people who use interop a lot and have methods with same names. Perfect match will still be same but when there are multiple non-matching methods, people may experience different results with this change. But in my opinion, before this change, the result would be undefined behavior.

Issue #304, opened 2016-05-20
Jint does not interperet System.Linq expressions


greenjeffreym:
for example, this code does not work:

var newArray = someArray.Select(function (x) {
    return X*X;
});

In my code I am using .Net assemblies and so I assume the arrays are declared as c# arrays. However, Jint throws an exception with the message: Object has no method 'Select'. I assume that since I am using the CLR that it would pick up the Linq methods.

sebastienros:
Probably because they are extension methods and they need to be discovered differently.

greenjeffreym:
How might that be done?

sebastienros:
If you look where the reflection is used to find namespaces and members, and nothing is found, you could fallback to extension methods.
What I would suggest is to add an option on the engine, like for assemblies, to specify a set of classes that contain extension methods to lookup for.

Like `option.AddExtensionMethods<System.Linq.Enumerable>` for your example

sebastienros:
I don't see any reason to close this issue so far, it's legitimate. Someone might want to implement it.

lahma:
Thanks to @KurtGokhan 's work, this is now supported in dev branch (MyGet release as of writing):

```csharp
var engine = new Engine(cfg => cfg
    .AllowClr()
    .AddExtensionMethods(typeof(System.Linq.Enumerable))
);
```

```javascript
let someArray = [ 2, 4 ];
let newArray = someArray.Select(x => x * x).ToArray();
// newArray now has [ 4, 16]
```

I'm closing this issue and we can investigate more specific use cases if/when they rise. [Like described in the PR](https://github.com/sebastienros/jint/pull/896), some cases are hard to handle.

PR #896, opened 2021-05-14
Improve generic extension methods and Linq interop

KurtGokhan:
I tried to implement this in an effort to improve interop capabilities. I thought it would be easy since Jint supports extension methods now. But the problem is, Linq extension methods are generic, and they work with type inference. It is hard (often impossible) to infer types in Javascript, and there are many issues it creates with covariance, contravariance, generic type constraints, and some other stuff. So, a true Linq compatibility is hard. But I did some improvements that make Linq and other generic extension methods partially work.

Related issue #304


lahma (approved):

lahma:
Nice improvement to resolution logic, thank you!

PR #899, opened 2021-05-18
Introduce Engine.Evaluate

lahma:
Code gets more succinct as chaining is no longer needed to get result.  Keeping the GetCompletionValue with Obsolete attribute, maybe someone will yell.


sebastienros (approved):
Needs update on readme

lahma:
Removed references to `Execute` in favor of `Evalute`

PR #902, opened 2021-05-20
Fix ToInt32 conversion

jfd16:
The current ToInt32 (and ToUint32) conversions are incorrect, according to the ECMA spec they have to be done using modulo 2<sup>32</sup> arithmetic (https://262.ecma-international.org/11.0/#sec-toint32)

|Expression|Current value|Correct value|
|----------|---:|--:|
| `((2 ** 53 - 1) * (2 ** 15)) \| 0` | 0 | -32768|
| `((2 ** 53 - 1) * (2 ** 15)) >>> 0` | 0 | 4294934528|
|`((2 ** 53 - 1) * (2 ** 31)) \| 0` |0|-2147483648|
|`((2 ** 53 - 1) * (2 ** 31)) >>> 0` |0|2147483648|


Jint/Runtime/TypeConverter.cs
```diff
@@ -375,14 +375,36 @@ internal static double ToInteger(string o)
             return (long) number;
         }
 
+        internal static int ToInt32(double o)
+        {
+            // Computes the integral value of the number mod 2^32.
```
sebastienros, line 380: Can there be a fast path when the `double` is in a specific range, in which case it could be a simple cast?

jfd16: Casting from double to int/uint gives the correct result when the double value is within the range of whatever integer type you are casting to. Outside of it, the result is CPU architecture dependent.

Adding checks for this range makes these conversions faster, at the cost of making conversions *outside* the range *slower*.

This can be seen from the benchmark:

```csharp
using System;
using System.Linq;
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Diagnosers;
using BenchmarkDotNet.Jobs;

[SimpleJob]
[IterationTime(250)]
public class BenchmarkJob {

    public IEnumerable<object> args =>
        new double[]
        {
            0, 1.0, 0.0000004, 2147483647, -2147483648, 485739298832.45432, Double.PositiveInfinity, Double.NaN
        }
        .Select(x => new object[] { x });

    [Benchmark]
    [ArgumentsSource(nameof(args))]
    public int DoubleToInt32_WithCast(double d) =>
        (d >= -2147483648.0 && d <= 2147483647.0) ? (int)d : DoubleToInt32Slow(d);

    [Benchmark]
    [ArgumentsSource(nameof(args))]
    public int DoubleToInt32_NoCast(double d) => DoubleToInt32Slow(d);

    public int DoubleToInt32Slow(double d)
    {
        // Computes the integral value of the number mod 2^32.

        long doubleBits = BitConverter.DoubleToInt64Bits(d);
        int sign = (int)(doubleBits >> 63);   // 0 if positive, -1 if negative
        int exponent = (int)((doubleBits >> 52) & 0x7FF) - 1023;

        if ((uint)exponent >= 84)
        {
            // Anything with an exponent that is negative or >= 84 will convert to zero.
            // This includes infinities and NaNs, which have exponent = 1024
            // The 84 comes from 52 (bits in double mantissa) + 32 (bits in integer)
            return 0;
        }

        long mantissa = (doubleBits & 0xFFFFFFFFFFFFFL) | 0x10000000000000L;
        int int32Value = (exponent >= 52) ? (int)(mantissa << (exponent - 52)) : (int)(mantissa >> (52 - exponent));

        return (int32Value + sign) ^ sign;
    }

}
```

```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
Intel Core i7-8550U CPU 1.80GHz (Kaby Lake R), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=5.0.102
  [Host]     : .NET Core 5.0.2 (CoreCLR 5.0.220.61120, CoreFX 5.0.220.61120), X64 RyuJIT
  Job-XURBHB : .NET Core 5.0.2 (CoreCLR 5.0.220.61120, CoreFX 5.0.220.61120), X64 RyuJIT

IterationTime=250.0000 ms
```
|                 Method |                  d |      Mean |     Error |    StdDev | 
|----------------------- |-------------------: |----------:|----------:|--------:|
| DoubleToInt32_WithCast |                  0 | 0.3614 ns | 0.0317 ns | 0.0297 ns | 
| DoubleToInt32_WithCast |              4E-07 | 0.3813 ns | 0.0205 ns | 0.0191 ns |  
| DoubleToInt32_WithCast |                  1 | 0.3863 ns | 0.0171 ns | 0.0160 ns |  
| DoubleToInt32_WithCast |        -2147483648 | 0.4082 ns | 0.0359 ns | 0.0318 ns |  
| DoubleToInt32_WithCast |         2147483647 | 0.3975 ns | 0.0327 ns | 0.0306 ns |  
| DoubleToInt32_WithCast | 485739298832.45435 | 2.8664 ns | 0.0665 ns | 0.0622 ns |  
| DoubleToInt32_WithCast |           Infinity | 1.3476 ns | 0.0351 ns | 0.0328 ns |  
| DoubleToInt32_WithCast |                NaN | 1.6198 ns | 0.0399 ns | 0.0373 ns | 
|   DoubleToInt32_NoCast |                  0 | 0.9683 ns | 0.0016 ns | 0.0013 ns |  
|   DoubleToInt32_NoCast |              4E-07 | 0.9155 ns | 0.0304 ns | 0.0285 ns |  
|   DoubleToInt32_NoCast |                  1 | 1.9896 ns | 0.0544 ns | 0.0509 ns |  
|   DoubleToInt32_NoCast |        -2147483648 | 2.0823 ns | 0.0726 ns | 0.1041 ns |  
|   DoubleToInt32_NoCast |         2147483647 | 2.0123 ns | 0.0356 ns | 0.0333 ns | 
|   DoubleToInt32_NoCast | 485739298832.45435 | 2.0095 ns | 0.0241 ns | 0.0226 ns |   
|   DoubleToInt32_NoCast |           Infinity | 0.9410 ns | 0.0200 ns | 0.0167 ns |  
|   DoubleToInt32_NoCast |                NaN | 1.0096 ns | 0.0501 ns | 0.0492 ns |

lahma: I would think that this code path is usually hit with the expectation of value being in the range?

lahma:
Could we have a test case case to cover so it doesn't get broken later?

Jint/Runtime/TypeConverter.cs
sebastienros: I agree with that too. Proof being that all current tests and scripts that are evaluated in the test suite didn't trigger any issue.

sebastienros, line 380:
```suggestion
            // Computes the integral value of the number mod 2^32.
            if (o >= -2147483648.0 && o <= 2147483647.0)
            {
                return (int)o;
            }
```

lahma (approved):
Thanks for making the tweaks and adding test test cases, looking good!

PR #904, opened 2021-05-26
Implement Error stack property

KurtGokhan:
Some libraries can do things like `Error().stack` to get the current callstack. I did an implementation that is working but somewhat hacky maybe (using `engine.GetLastSyntaxNode()`). 

It does not give the exactly same string as in V8 or other engines but afaik there is no spec that says what string to return. Only WIP spec is this: https://github.com/tc39/proposal-error-stacks


lahma (approved):
As this is not part of the official spec and browser implementations vary, I think this is a fine and useful implementation. Thanks for this.

PR #907, opened 2021-05-27
Refactor internals to support realm concept

lahma:
This PR will hopefully produce first version of [Realm](https://tc39.es/ecma262/#sec-code-realms) and [Host](https://tc39.es/ecma262/#sec-hosts-and-implementations) concepts.

 This will cater way for to additional features:

* [Isolated execution context,](https://github.com/sebastienros/jint/pull/789) we can have custom global built and some callbacks
* BrowserHost (or something similar) can be implemented for scenarios like [AngleSharp.JS](https://github.com/AngleSharp/AngleSharp.Js/issues/66), where default environment is not what is wanted
* Having separate initialization could also mean that we could reset the global environment without ditching the entire engine

This PR will also unify and simplify initial engine by introducing separate `Intrinsics` object instead having it all in `Engine`.


sebastienros:
What's your feeling about how much closer it is to the specs with this PR?

lahma:
Current status is that it should now follow spec and constructs quite closely. I *think* that the constructs are now easier to understand and one can follow the spec with regards to references to realms (current realm etc). Tests are passing and now it's time for some polish and checking if AngleSharp custom host (Window object) can be implemented and what kind of hosting API it requires, close to finish I would say.

lahma:
Because this is already quite massive change I think it would be nice to get this into main before continuing the integration story via Host concepts. I've tested this API-wise against RavenDB and AngleSharp.JS, engine.Obsolete.cs helps to transition to realm and their code compiles as usual.

lahma:
Another nice outcome of making engine initialization lazier is that now one gets to run minimal scripts faster:

```csharp
var engine = new Engine();
return engine.Evaluate("return [].length + ''.length").AsNumber();
```

*Before*

|      Method |     Mean |     Error |    StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------ |---------:|----------:|----------:|-------:|-------:|------:|----------:|
| BuildEngine | 5.587 us | 0.0265 us | 0.0207 us | 1.2894 | 0.0687 |     - |  21.09 KB |

*After*

|      Method |     Mean |     Error |    StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------ |---------:|----------:|----------:|-------:|-------:|------:|----------:|
| BuildEngine | 4.312 us | 0.0159 us | 0.0149 us | 0.8545 | 0.0229 |     - |  14.04 KB |

The change is more visible when the script is parsed beforehand and only the Script instance is executed against new engine:

*Before*

|      Method |     Mean |     Error |    StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------ |---------:|----------:|----------:|-------:|-------:|------:|----------:|
| BuildEngine | 3.323 us | 0.0163 us | 0.0136 us | 1.0071 | 0.0381 |     - |  16.48 KB |

*After*

|      Method |     Mean |     Error |    StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------ |---------:|----------:|----------:|-------:|-------:|------:|----------:|
| BuildEngine | 2.366 us | 0.0109 us | 0.0102 us | 0.5760 | 0.0153 |     - |   9.43 KB |

lahma:
[Example how spec can now be followed using the new constructs](https://github.com/sebastienros/jint/pull/907/files#diff-1b5cc28bf240558bbe821785127badf825812d58b96abfa497ea270066523cf1L426).

PR #910, opened 2021-05-29
Parse scripts only once while testing

lahma:
We are parsing some scripts over and over again through thousands of test cases, let's parse them only once. From realms branch originally.


sebastienros:
It's better for the environment

PR #911, opened 2021-05-29
Fix Promise.all behavior under interop

### LINKED ISSUE #908: Unexpected Promise.all behavior
First of all, thank you very much guys for the new release and the support of promises! Great work!

I am testing promises right now, and run into an issue.

On some occasions, when I do

`var evaluationResults = engine.Evaluate("return Promise.all(someArray)").UnwrapIfPromise().ToObject();`,

and `someArray` does not contain any promises at all, but only contains _one_ object, namely a .Net `Dictionary<string, object>`,

instead of an array with a single element in `evaluationResults` I get array of `KeyValuePair<string, object>` of the dictionary within an array.

If instead of an instance of `Dictionary<string, object>` I put into that array a single simple JS object, it works just as expected.

If I simply add to `someArray` any additional element, say, `true`, it works like a charm.

lahma:
Using direct array create to bypass possible CLR conversion logic, fixes #908 .

### DIFF
--- a/Jint.Tests/Runtime/InteropTests.cs
+++ b/Jint.Tests/Runtime/InteropTests.cs
@@ -2556,5 +2556,12 @@ namespace Jint.Tests.Runtime
             Assert.Equal(1, _engine.Evaluate("arr.indexOf(b)").AsNumber());
             Assert.True(_engine.Evaluate("arr.includes(b)").AsBoolean());
         }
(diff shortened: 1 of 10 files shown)


lahma:
@sebastienros ping 😉

sebastienros (approved):

PR #921, opened 2021-06-27
Store and display original stack frames to the user if the same object is re-thrown later

whizzter:
Frameworks like React catches error objects and re-throws later, debugging these errors can be hard if the stack is lost.

This is a small patch that store the first stack trace and appends the last throw (there can be intermediate traces but showing al of them can create really long and unreadable printouts).

### DIFF
--- a/Jint/Runtime/CommonProperties.cs
+++ b/Jint/Runtime/CommonProperties.cs
@@ -25,5 +25,6 @@ namespace Jint.Runtime
         internal static readonly JsString Configurable = new JsString("configurable");
         internal static readonly JsString Stack = new JsString("stack");
         internal static readonly JsString Message = new JsString("message");
+        internal static readonly JsString OriginalStack = new JsString("original_stack");
     }
 }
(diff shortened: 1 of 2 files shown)


lahma:
Would it make sense to model this per the [new Error Cause proposal](https://github.com/tc39/proposal-error-cause)?

whizzter:
It'd be an ideal end-state for user-code to use the proposal, current practice by user code however is another thing. Tried in the Chrome Dev console and at a quick try it seems like V8 just creates the call-stack upon constructing Error objects rather than doing anything at throw-time.

KurtGokhan:
If this PR is about defining Error stack upon constructing, it was already done at #904. I also had problems with React before implementing that PR but not anymore. So this problem may already been solved.

whizzter:
Looked at the patch and it did seem to be more in line what was needed. Gotta pull in latest and test everything (I haven't had time to work on this project since the summer).

lahma:
@whizzter should we close this or is there still something missing in the error.cause?

PR #923, opened 2021-07-08
Refactor options and add MaxArraySize constraint

### LINKED ISSUE #683: Memory limit and timeout bypassed with new Array and Join
For my use case we need to evaluate Javascript expressions for validation. I wanted to test Jint security features to ensure that the there would be minimal impact to the application if a bad expression was run. I was testing to see if it was possible to exceed the run limits and the following test case allows the 0.5s timeout to be exceeded and the memory limit of 500 to also be exceeded.

Eventually the application fails with a .NET OutOfMemoryException.

```
var javascript = "new Array(2147483647).join('*')";
var result = new Engine(options =>
{
    options.LimitRecursion(1);
    options.Strict();
    options.DiscardGlobal();
    options.TimeoutInterval(TimeSpan.FromSeconds(0.5));
    options.MaxStatements(1);
    options.LimitMemory(500);
} )
.Execute(javascript)
.GetCompletionValue() 
.ToObject();
```

From browsing the source code it looks like the timeout and memory limits are only evaluated before each statement statement. Therefore the statement can exceed these limits and the execution only stops on the next statement. If it is not possible to resolve this case it is worth documenting this as a limitation of the timeout and memory limits.

### LINKED ISSUE #887: Memory limits bypass
```
var arr = Array(1000000000).fill(new Array(1000000000));
```

I think it's fine to not check the actual memory usage of each array or string allocations though. These could use an option with a max length instead, even very big. Then throw a specific exception. And concatenations would check upon this limit instead of the thread memory.

The `fill` function however might need to check the memory limits. There might be other functions too that loop over the same value and need a check. Looping over a function evaluation is safe though since thread memory is ensured.

### LINKED ISSUE #917: Trying to crash server running jint
I'm offering users of my website to run scripts, but I don't want to let them be able to crash my server (or make it unresponsive).
So I've been actively trying to make jint unresponsive.
I'm running the following code to attempt to use up all the memory:

`new Engine(o => o.MaxStatements(1000)).Evaluate("for (let a = [1, 2, 3];; a = a.concat(a)) ;");`

Interestingly, it doesn't use up all the memory, but it freezes (even though I set a maximum number of statements). No exceptions are thrown.

3.0.0-beta-2032

lahma:
* introduce public options object that has sub-categories for interop, debugger, constraints - and API documented 
* move fluent configuration to extension methods to keep model clean
* add array size constraint that allows limiting the size of created arrays
* don't switch to dictionary based arrays too eagerly, there's a large performance penalty
* move constraint related tests to own class

fixes #917
fixes #887
fixes #683


lahma:
While implementing came up with couple questions, should this be `MaxCollectionSize` and enforce to other collection types too later (typed array, map, set?) and whether `Options` should be renamed to `JintOptions`, maybe easier story for disambiguating but a breaking change.

lahma:
@sebastienros please take a look.

ByronAP:
Bump requesting this get resolved

lahma:
Let's refine in later pull requests if needed, the groundwork is now done.

sebastienros:
Can I still approve?

lahma:
Thanks for the gesture 😄 You can still also list improvement ideas though!

PR #927, opened 2021-07-10
Fix Array.from, entries, spread and other iterator related methods

KurtGokhan:
Summary of fixes:
- Added `Symbol.iterator` to iterator objects
- Added String `trimLeft` and `trimRight`. They are the same as `trimStart` and `trimEnd` [according to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trimEnd)
- Fixed `Array.from`, `entries`, spread syntax
  - Removed `ExtractValueFromIteratorInstance` method. It was causing trouble for nested arrays. I think it was added mistakenly because of the bug in `entries`
- Added Babel test which helped me catch all these errors

I was surprised that no tests fail because of `entries` change. Maybe there was no test coverage for that?


lahma (approved):
Great fixes, thank you! I tried with newer test262 test suite and some untested paths but none did show these problems so I guess the official test suite is just missing the coverage (or I can't find the correct place). Thanks for adding the missing coverage 👍🏻

Issue #506, opened 2018-06-04
Javascript Regex parse error


dejanberic:
Some javascript libraries have this as regex:
F=/[^]*?(:[rp][el]a[\w-]+)[^]*/;
The issue is with [^], which means any character including line breaks in javascript regex but in C# it is not a valid regex.
I have fixed this issues by doing .Replace("[^]", "(?:.|\r|\n)").

PeterWone:
@dejanberic This is exactly the problem I have. Where did you put the expression fixup?

lahma:
The latest beta should have this working now: https://github.com/sebastienros/jint/releases/tag/v3.0.0-beta-2050

PR #931, opened 2021-07-17
Do not throw error immediately when receiving null Regex from Esprima 

KurtGokhan:
Sometimes Esprima will return null Regex. That means parsing did not fail, but the Regex is incompatible with C#.

With this PR, Jint will not fail during the parsing when such a Regex is received. It will only fail when that Regex is actually used. This way the user can handle the error, and also error can be avoided if that Regex is not used at all. 

See also: sebastienros/esprima-dotnet#173

Keeping this as draft until Esprima PR is merged and released. (Test will fail without the new Esprima)

### DIFF
--- a/Jint.Tests.Test262/Test262Test.cs
+++ b/Jint.Tests.Test262/Test262Test.cs
@@ -321,12 +321,6 @@ namespace Jint.Tests.Test262
                     reason = "Unicode support and its special cases need more work";
                 }
 
-                if (name.StartsWith("built-ins/RegExp/CharacterClassEscapes/"))
-                {
-                    skip = true;
-                    reason = "for-of not implemented";
-                }
-
                 // Promises
                 if (name.StartsWith("built-ins/Promise/allSettled") ||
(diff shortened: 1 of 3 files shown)


lahma:
@KurtGokhan 2.0.2 now available on NuGet.

lahma:
Will this resolve #506 ?

lahma:
I've updated main to have 2.0.2 as part of small bug fixing, new lib was required to get some test suite JS parsing working.

KurtGokhan:
@lahma  that Regex is from [stylis](https://github.com/thysultan/stylis.js), which is the library I also noticed this issue. 

This PR will not fix the issue. But at least the parsing will pass. The regex will still fail on runtime if it is executed. In my case the regex is never executed so it works for me.

`code.Replace("[^]", "(?:.|\r|\n)")` is a legit fix for this case. But as I said, if we apply this hack blindly, we may break already valid regex like `[[^]`.

lahma:
Cool, thanks for the clarification. We can merge?

PR #944, opened 2021-08-02
Allow overloading assignment operators

### LINKED ISSUE #941: Overloading Assignment Operators
Hello!

In C#, the operator += is implicitly overloaded when the operator + is defined. However, the current version of Jint (3.0.0.beta-2033) doesn't behave correctly when the operator += is encountered. This is the test I used:

```
class A
{
    public int Value { get; set; } = 0;
    public A(int value) { Value = value; }
    public static A operator +(A lhs, A rhs) { return new A(lhs.Value + rhs.Value); }
    public override string ToString() { return Value.ToString(); }
}
```
```
var engine = new Jint.Engine(cfg => cfg.AllowOperatorOverloading());
engine.SetValue("a", new A(1));
engine.SetValue("b", new A(2));
engine.Execute("a += b");
var a = engine.GetValue("a"); // Result is a string with value "12", while it should be an object A with Value = 3
```
When the method "ToString" is not provided, the result will be a string containing the type name of the class.

Is it something that would be hard to implement or even desirable in the context of Jint?

Regards

KurtGokhan:
Fixes #941


lahma (approved):
Well that didn't take long. Thanks for looking into this and helping out!

PR #945, opened 2021-08-03
Fix double evaluation when Operator Overloading is enabled

n1xx1:
When you have operator overloading enabled the operands of almost all binary and unary expression are evaluated multiple times. This PR should fix it but it's marked as draft because it was suggested to add some tests to test against it.

The problem is easily verifiable like this:

```
var engine = new Engine(cfg => { cfg.AllowOperatorOverloading(); });
engine.SetValue("Log", (Action<string>) (s => Console.Out.WriteLine(s)));
engine.Execute("const test = () => { Log('Example'); return 1; }; test() + 1; ");
```

This should only print a single `Example` line, but instead prints two.

### DIFF
--- a/Jint/Runtime/Interpreter/Expressions/JintUnaryExpression.cs
+++ b/Jint/Runtime/Interpreter/Expressions/JintUnaryExpression.cs
@@ -53,6 +53,7 @@ namespace Jint.Runtime.Interpreter.Expressions
 
         protected override object EvaluateInternal()
         {
+            var v = _argument.GetValue();
             if (_engine.Options._IsOperatorOverloadingAllowed)
             {
                 string operatorClrName = null;
@@ -75,7 +76,7 @@ namespace Jint.Runtime.Interpreter.Expressions
                 }
 
                 if (operatorClrName != null &&
-                    TryOperatorOverloading(_engine, _argument.GetValue(), operatorClrName, out var result))
+                    TryOperatorOverloading(_engine, v, operatorClrName, out var result))
                 {
                     return result;
                 }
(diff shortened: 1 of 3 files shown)


lahma:
Change seems logical, it would be great if you can add test case along the other operator overloading tests.

Jint/Runtime/Interpreter/Expressions/JintUnaryExpression.cs
```diff
@@ -53,6 +53,7 @@ public override JsValue GetValue()
 
         protected override object EvaluateInternal()
         {
+            var v = _argument.GetValue();
```
lahma, line 56: This might have broken the test suite as it seems that this could have behavioral change.

n1xx1: I found some issues with the UnaryExpression fixes, I will add another commit to this soon that should also avoid this behavioral change.

Jint/Runtime/Interpreter/Expressions/JintUnaryExpression.cs
```diff
@@ -165,7 +166,8 @@ protected override object EvaluateInternal()
                         }
                     }
 
-                    var v = _engine.GetValue(value, true);
+                    // TODO: double evaluation problem
```
n1xx1, line 169: This is a problem that need more work to fix. GetValue is overloaded by different expressions, but we need to first check if it's an undefined reference so that we can return `"undefined"` without an error. At the moment this should pass all the tests, but it not the correct behaviour as you can try with this example:

```
const fn = () => Log('test');
typeof fn();
```

lahma: So what's the correct behavior, this is what node gives:

```
Welcome to Node.js v14.16.1.
Type ".help" for more information.
> const fn = () => Log('test');
undefined
> typeof fn();
Uncaught ReferenceError: Log is not defined
    at fn (REPL1:1:18)
```

And Jint follows the logic:

```
jint> const fn = () => Log('test');
undefined
jint> typeof fn();
Jint.Runtime.JavaScriptException: Log is not defined
   at fn () repl:1:1
   at repl:1:8
```

n1xx1: ye, of course, Log is not defined 😅, you have to use the corresponding print function.

With node:
```
Type ".help" for more information.
> const fn = () => { console.log('test'); }
undefined
> typeof fn();
test
'undefined'
```

With jint:
```
jint> const fn = () => { print('test'); }
undefined
jint> typeof fn();
test
test
"undefined"
jint>
```

lahma: OK thanks, can you open a separate issue so that can be looked into, unless you want to solve all in this PR 😉

lahma (approved):

lahma:
Thanks for going through the process and helping out, always appreciated!

Issue #598, opened 2019-02-25
Duplicate __proto__ fields are not allowed in object literals


windischb:
Hi!

I use "typescript.js" to transpile typescript code to javascript code.
Currently working with version 2.11.58
If i do exactly the same, with version 3.0.0-beta-1476 i'll get the following error:
```
Esprima.ParserException: Line 62: Duplicate __proto__ fields are not allowed in object literals
```
Is this a bug in Esprima, so duplicate `__proto__` fields sholud be allowed??
Or is this the correct behavior and the typescript.js library is not ES6 compliant?

wizzard0:
I believe this is a known bug in Esprima, discussed in js version at https://github.com/jquery/esprima/issues/1485

wizzard0:
ugly hack is set `"importHelpers": true,` in tsconfig and patch `tslib`, commenting out the line 
`({ __proto__: [] } instanceof Array && function (d, b) { d.__proto__ = b; }) ||`

even uglier would be regex-replace this line in emitted helpers :)

windischb:
If i understand it correct, Esprima .Net should behave exactly as the Esprima js version.
So if it's a gug in the js version, it's also a bug in the .Net version!

Sounds like there is a good chance that @sebastienros wil fix it also in the .Net version...

For now i can stay at Jint 2.11.58.
But in this version i'll get an Stackoverflow expection whit circular references, and i hoped that this is maybe fixed in Jint 3.x

lahma:
I believe you just haven't confiured the JavaScript parser to be tolerant:

```
var engine = new Engine();
engine.Execute(
    "({ __proto__: [] } instanceof Array && function (d, b) { d.__proto__ = b; })",
    new ParserOptions
    {
        Tolerant = true
    });
```

@sebastienros can chime in if the being non-tolerant is the wrong default..

sebastienros:
About tolerant mode: https://esprima.readthedocs.io/en/latest/syntactic-analysis.html#tolerant-mode

But this doesn't excuse a stackoverflow, it should at least return a parse error.
Then does this duplicate proto definition follows the spec? https://tc39.github.io/ecma262/#sec-__proto__-property-names-in-object-initializers

If you think this instance is compliant with the spec then it should work in non-tolerant mode. If it doesn't follow the spec then you will need to enable the Tolerant mode.

windischb:
I have tested a few things!

```csharp
var tsCode1 = "import {test} from 'test'";
[...]

PR #948, opened 2021-08-23
Fix eval and function constructor failing for object literals with __proto__ prop

KurtGokhan:
Related to #598

Fixes weren't applied to Eval and Function Ctor before. 

Edit after tests: `Tolerant = true` breaks other tests. Do you know what may be a better solution?

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -64,7 +64,7 @@ namespace Jint.Tests.Runtime
         public EngineTests(ITestOutputHelper output)
         {
             _engine = new Engine()
-                .SetValue("log", new Action<object>( o => output.WriteLine(o.ToString())))
+                .SetValue("log", new Action<object>(o => output.WriteLine(o.ToString())))
                 .SetValue("assert", new Action<bool>(Assert.True))
                 .SetValue("equal", new Action<object, object>(Assert.Equal))
                 ;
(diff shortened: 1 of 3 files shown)


lahma:
I quickly checked just one test case and its contents:

```javascript
function testcase()
{
  try 
  {
    eval('"use strict"; var x = "\\17";');
    return false;
  }
  catch (e) {
    return (e instanceof SyntaxError);
  }
}
runTestCase(testcase);
```
So my hunch is that it's a strict mode issue. At least `PerformEval` has `strictCaller` parameter which you could use to choose between two different option setups (tolerant true/false), so maybe another static field for choosing between. Didn't dig deep into the issue so not sure.

KurtGokhan:
I think this should be solved in Esprima.NET. Original esprima  also seems to have this bug.

KurtGokhan:
Created sebastienros/esprima-dotnet#192

There is no need to change any code in Jint. Only tests.

But I noticed that by making `Tolerant = true` in the engine [in this PR](https://github.com/sebastienros/jint/pull/605/files#diff-28d7cd25ff148366bda545a8d3c4e400be429c5d2aa5be771d993d0a88c2d49dR40), some behaviors are not compliant with spec. Should we revert it? But it can be a breaking change.

lahma:
Main branch has been updated with latest Esprima. Is there still an issue that we need to fix?

KurtGokhan:
Let me test real quick.

KurtGokhan:
Good to go. It is only tests.

lahma (approved):

lahma:
Thanks for verifying!

PR #959, opened 2021-09-18
Fix extension and class methods being enumerated

KurtGokhan:
Gitter discussion here: https://gitter.im/sebastienros/jint?at=6143e1ecfd7409696e1612ad

Additionally, I run into a case where if you use a class method once, it registers as a enumerable property onto that object. Included fix for that as well.


lahma (approved):
AllForbidden seems like a safe default. Let's see if someone needs runtime mutation. Thanks for fixing this!

PR #965, opened 2021-09-22
Capture original exception stacktrace in ThrowMeaningfulException

### LINKED ISSUE #963: ThrowMeaningfulException does not capture inner stacktrace
In [ExceptionHelper ThrowMeaningfulException](https://github.com/sebastienros/jint/blob/87ee8e2c4b1be511900947e22cc254f5740c2e6c/Jint/Runtime/ExceptionHelper.cs#L135), the inner exception gets thrown but the stacktrace gets rewritten to the stacktrace of ThrowMeaningfulException

eg
```
System.InvalidOperationException: <some exception>.
 at Jint.Runtime.ExceptionHelper.ThrowMeaningfulException(Engine engine, TargetInvocationException exception)
 at Jint.Runtime.Interop.MethodInfoFunctionInstance.Call(JsValue thisObject, JsValue[] jsArguments)
 at Jint.Engine.Call(FunctionInstance functionInstance, JsValue thisObject, JsValue[] arguments, JintExpression expression)
 at Jint.Runtime.Interpreter.Expressions.JintCallExpression.EvaluateCall(JsValue func, Object reference, NodeList`1& arguments, Boolean tailPosition)
etc
```

It would be helpful if we could instead throw with the original stacktrace using ExceptionDispatchInfo, as in this stackoverflow post:
https://stackoverflow.com/questions/57383/how-to-rethrow-innerexception-without-losing-stack-trace-in-c

What does everyone think about this proposed change?

caseyzhang123:
Use ExceptionDispatchInfo to capture the original stacktrace in ThrowMeaningfulException

fixes #963

### DIFF
--- a/Jint.Tests/Runtime/ErrorTests.cs
+++ b/Jint.Tests/Runtime/ErrorTests.cs
@@ -1,5 +1,6 @@
 ﻿using Esprima;
 using Jint.Runtime;
+using Jint.Tests.Runtime.TestClasses;
 using System;
 using System.Collections.Generic;
 using Xunit;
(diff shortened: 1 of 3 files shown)


Jint.Tests/Runtime/ErrorTests.cs
```diff
@@ -282,6 +283,21 @@ public void StackTraceCollectedForImmediatelyInvokedFunctionExpression()
...
+        [Fact]
+        public void StackTraceIsForOriginalException()
+        {
+            var engine = new Engine();
+            engine.SetValue("HelloWorld", new HelloWorld());
+            const string script = @"HelloWorld.ThrowException();";
+
+            var ex = Assert.Throws<DivideByZeroException>(() => engine.Execute(script));
+
+            const string expected = @"System.DivideByZeroException: Attempted to divide by zero.
```
sebastienros, line 295: I would just check for the type of exception. Checking on the string is too brittle.

sebastienros: Isn't that a stupid comment ;) Just look for some part of the string that you think should be there, but not the exact one, the filenames or lines might change

caseyzhang123: I agree the string is brittle, but if we don't somehow string check the stacktrace then it won't properly test my PR is effective.  Should we just remove it regardless?

caseyzhang123: fixed

lahma (approved):

lahma:
Thank you for your contribution!

PR #967, opened 2021-09-25
Fix iteration in interop

lahma:
* array-like when list or array should iterate values (like Set)
* dictionary-likes should iterate keys and values (like Map)
* IEnumerable behaves like custom Iterable added to regular JS object (enumerates values as-is)
* any other object type is unsupported like in normal JS (like basic objects { "A": 1 })


sebastienros:
Not the easiest of changes.

sebastienros (approved):

PR #971, opened 2021-10-01
Add typed array helpers for IsArray and AsArray

lahma:
* moved `Is` and `As` helpers to extensions methods
* added `Is` and `As` helpers for each supported typed array type

### DIFF
--- a/Jint/Native/TypedArray/TypedArrayInstance.cs
+++ b/Jint/Native/TypedArray/TypedArrayInstance.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using Jint.Native.ArrayBuffer;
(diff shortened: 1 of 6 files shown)


Jint/Native/TypedArray/TypedArrayInstance.cs
```diff
@@ -330,5 +331,23 @@ private static bool IsIntegralNumber(double value)
                    && !double.IsInfinity(value)
                    && System.Math.Floor(System.Math.Abs(value)) == System.Math.Abs(value);
         }
+
+        internal T[] ToNativeArray<T>()
+        {
+            var conversionType = typeof(T);
+            var elementSize = _arrayElementType.GetElementSize();
```
sebastienros, line 338: I believe this could be a static field?

sebastienros: super nit right?

sebastienros (approved):

Jint/Native/TypedArray/TypedArrayInstance.cs
lahma: Size is calculated via switch as there's only one TypeArrayInstance type and no specific for each one, it should be quite cheap...

sebastienros: I agree. It just reminded me this trick that things on the generic type can be done statically without interfering between types.

PR #973, opened 2021-10-11
Configurable value coercion under interop

lahma:
Noticed this when investigating AngleSharp.Js code and its expectations. Currently `o.StringProperty = true;` leads to having `"True"` as value, JS engine should inject values based on JS ToString conventions, thus having the value `"true"`.

Supporting three modes `Boolean`, `Number` and `String`, and any combination of them. This allows things like:

* `o.StringProperty = [ 1, 2, 3 ]; o.StringProperty = true;` 
* `o.BooleanProperty = "OK"; o.BooleanProperty = array.length; o.BooleanProperty = {}` 
* `"class MyClass { valueOf() { return 42; } }"; o.IntProperty = new MyClass();` 

The default is just `String` which should yield quite predictable behavior, `Number`and `Boolean` might surprise as they are different from c# semantics.


lahma:
After this I think next up is `o.BooleanProperty = someObject;` or `o.BooleanProperty = array.length` (when array has elements) should evaluate to `true`. Using `TypeConverter` logic.

Jint/Extensions/ReflectionExtensions.cs
```diff
@@ -54,5 +58,108 @@ private static bool IsExtensionMethod(this MethodBase methodInfo)
...
+            if (type == null || type.IsEnum)
+            {
+                return false;
+            }
+
+            switch (Type.GetTypeCode(type))
+            {
+                case TypeCode.Decimal:
+                case TypeCode.Double:
+                case TypeCode.Int32:
```
sebastienros, line 104: Not the smaller ones?

sebastienros:
Next level interoperability

Jint/Extensions/ReflectionExtensions.cs
lahma: My thinking was less surprises, like overflows etc. `int` should be quite common and we could open up later on if someone has the use case.

lahma:
Are we good to merge?

sebastienros (approved):

Issue #534, opened 2018-08-27
Access to an indexer of a userdefined csharp class failed.


AlPecin:
Hi 

I have  this class defined

> 
        public class UserClass
        {
            public Type TypeProperty { get; set; }
            public object Value { get; set; }

            public string this[int index]
            {
                get
                {
                    return "int";
                }
            }
            public string this[string index]
            {
                get
                {
                    return "string";
                }
            }
        }

> 

and I execute the following javascript snippets:

>  userclass['Whistler'] + ':' +  userclass[10];
    userclass[10] + ':' + userclass['Whistler'];
> 

so I got an error on passing the unit test in the attachment
[AccessIndexer.zip](https://github.com/sebastienros/jint/files/2323784/AccessIndexer.zip)

if I change the order of the snippets in the unit test, it passed without an error! 

What's going wrong?

AlPecin:
IMHO it depends on the "static readonly Dictionary<string, bool> _knownConversions"!

KurtGokhan:
Tested this. It seems to work now. Issue can be closed?

PR #992, opened 2021-10-30
Implement Error cause and other Error specs

KurtGokhan:
- Added Error Cause
  - Stage 4
  - Specs: https://github.com/tc39/proposal-error-cause
  - MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error
- Refreshed the test262 for Error specs. Some of them were failing, so fixed them.
- Also added test for #534


sebastienros:
Was looking at all the proposals and what stage 4 means, FYI:
https://github.com/tc39/proposals/blob/master/finished-proposals.md

Not sure which ones haven't been implemented in Jint. 
And there there are all the features that are in stages 0/1: https://github.com/tc39/proposals

sebastienros (approved):

KurtGokhan:
I implemented this one specifically because it was mentioned at #921

Maybe we can do a compat-table for Jint to detect all missing features. https://github.com/kangax/compat-table

lahma (approved):
Looking good.

PR #994, opened 2021-10-31
Prevent GetType and System.Reflection namespace under interop by default

### LINKED ISSUE #993: Security leak
Security leak:
even I do not expose Assembly type it is accessible via any type like below:
```javascript
log(model.GetType().Module.GetType().Module.GetType('System.DateTime'));
```

```c#
var jintEngine = new Engine();
jintEngine.SetValue("log", new Action<object>(Console.WriteLine));
jintEngine.SetValue("model", new List<int>());
jintEngine.Execute(Code);
```
With this leak a script can access any .NET class in the system and can execute any arbitrary code using server privileges.
Is there any mechanism to prevent such leaks?

lahma:
Seems that this trips people often enough that we shouldn't allow `GetType` by default. Created separate property to control so that overwritten member filter won't cause us trouble, GetType will always by filtered based on configuration.

fixes #993

### DIFF
--- a/Jint/Runtime/Interop/TypeResolver.cs
+++ b/Jint/Runtime/Interop/TypeResolver.cs
@@ -12,14 +12,27 @@ namespace Jint.Runtime.Interop
     /// </summary>
     public sealed class TypeResolver
     {
-        public static readonly TypeResolver Default = new TypeResolver();
+        public static readonly TypeResolver Default = new();
 
         private Dictionary<ClrPropertyDescriptorFactoriesKey, ReflectionAccessor> _reflectionAccessors = new();
 
         /// <summary>
-        /// Registers a filter that determines whether given member is wrapped to interop or returned as undefined.
+        /// Whether to expose <see cref="object.GetType"></see> which can allow bypassing allow lists.
+        /// Defaults to false.
         /// </summary>
+        public bool ExposeGetType { get; set; }
+
+        /// <summary>
+        /// Registers a filter that determines whether given member is wrapped to interop or returned as undefined.
+        /// By default allows all but will also by limited by <see cref="ExposeGetType"/> configuration.
+        /// </summary>
+        /// <seealso cref="ExposeGetType"/>
         public Predicate<MemberInfo> MemberFilter { get; set; } = _ => true;
+
+        internal bool Filter(MemberInfo m)
+        {
+            return (ExposeGetType || m.Name != nameof(GetType)) && MemberFilter(m);
+        }
 
         /// <summary>
         /// Gives the exposed names for a member. Allows to expose C# convention following member like IsSelected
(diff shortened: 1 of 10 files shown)


sebastienros:
Fine with disabling it by default.
What about just reflection as a whole? Like checking on the `System.Reflection` namespace.

What about having a bool in `AllowClr` to also allow reflection. Just to give more ideas, not sure it's better, just maybe easier. No sure if with a custom filter we could actually remove the GetType limitation as a side effect. This bool would be taken into account and block the whole namespace whatever the filter says.

lahma:
Yeah makes sense to consider separate reflection namespace restriction, but `GetType` is a gateway to all the magic (`GetType().Assembly.GetType("HardDriveFormatter).GetMethod("Format").Invoke()`). `GetType` is defined on object and the `Type` is under System namespace. So we would then allow access to Type information at least, which might leak something - dunno.

So I guess we could do something extra, maybe another filter to check for Reflection namespace before allowing wrapping, like the `AllowClr(bool allowReflectionNamespace)` you suggested.

sebastienros:
Didn't realize that GetType() was in System, so obvious.

koculu:
I have done following for Topaz:
https://github.com/koculu/topaz/commit/00bf8f1936e2f88c204f17c26c1630c6699f28d3
Type is under system and it is safe. So any of its members are callable including Assembly.
Since Assembly is under System.Reflection namespace, it's members not callable or accessible through script.
I believe blocking System.Reflection namespace is enough for safety.
Any other dangerous use case for Type?

koculu:
ok I found my own answer after throwing the question.
`someValue.GetType().GetProperty("xxx").GetValue(someValue);`
With above statement script can access private members of a value which are not supposed to be.
And tons of other functions are evil.
Instead of disabling GetType() completely, it is best to write a white list for Type Function and allow them.

lahma:
I've added separate `AllowSystemReflection` which controls the namespace wrapping, in addition to separate `AllowGetType`. I've harmonized the naming a bit and moved options under `Options.Interop`.

Jint/Runtime/Interop/TypeResolver.cs
```diff
@@ -12,15 +12,22 @@ namespace Jint.Runtime.Interop
...
     public sealed class TypeResolver
     {
-        public static readonly TypeResolver Default = new TypeResolver();
+        public static readonly TypeResolver Default = new();
 
         private Dictionary<ClrPropertyDescriptorFactoriesKey, ReflectionAccessor> _reflectionAccessors = new();
 
         /// <summary>
         /// Registers a filter that determines whether given member is wrapped to interop or returned as undefined.
+        /// By default allows all but will also by limited by <see cref="AllowGetType"/> configuration.
```
sebastienros, line 21:
```suggestion
        /// By default allows all but will also be limited by <see cref="AllowGetType"/> configuration.
```

sebastienros (approved):

PR #996, opened 2021-11-03
Fix ExpandoObject serialization

### LINKED ISSUE #995: JSON.stringify incorrectly serializes ExpandoObject when it contains a `Values` field
## TLDR;

- ExpandoObjects containing a `Values` key don't get stringified properly.

**Expected result**
```json
{"Values":["Foo","Bar","Baz"]}
```

**Actual result**
```json
{"Values":{"Count":1,"IsReadOnly":true}}
```

## Steps to reproduce:
Take the following JavaScript script:

```javascript
({ Values: ["Foo", "Bar", "Baz"] })
```

When evaluated, it will return an `ExpandoObject` containing a `Values` key and an array value.
So far, so good.

> We could also just use the following code to create an expando object of course:
> ```csharp
> dynamic dynamicObject = new ExpandoObject();
> dynamicObject.Values = new[] { "Foo", "Bar", "Baz" };
> ```

When we now `JSON.stringify` this expando object, the result will be:

```json
{"Values":{"Count":1,"IsReadOnly":true}}
```

Instead of the expected:

```json
{"Values":["Foo","Bar","Baz"]}
```

Interestingly enough, when I change the `Values` key to e.g. `MyValues`, the stringify operation does return the expected JSON:

```json
{"MyValues":["Foo","Bar","Baz"]}
```

I attached a small console app demonstrating the issue. It also shows that using `Values` as a key does not cause any issues when using Newtonsoft.Json's `JsonConvert` to serialize the expando object:

image

Sample project:
[JintExpandoTest.zip](https://github.com/sebastienros/jint/files/7460251/JintExpandoTest.zip)

sebastienros:
Fixes #995

### DIFF
--- a/Jint/Runtime/Interop/ObjectWrapper.cs
+++ b/Jint/Runtime/Interop/ObjectWrapper.cs
@@ -98,19 +98,19 @@ namespace Jint.Runtime.Interop
                 return Undefined;
             }
 
-            if (Target is IDictionary<string, object> stringKeyedDictionary && property is JsString dicKey) // expando object for instance
-            {
-                var member = dicKey.ToString();
-
-                if (stringKeyedDictionary.TryGetValue(member, out var value))
-                {
-                    return FromObject(_engine, value);
-                }
-            }
-
             if (property is JsString stringKey)
             {
                 var member = stringKey.ToString();
+
+                // expando object for instance
+                if (Target is IDictionary<string, object> stringKeyedDictionary)
+                {
+                    if (stringKeyedDictionary.TryGetValue(member, out var value))
+                    {
+                        return FromObject(_engine, value);
+                    }
+                }
+
                 var result = Engine.Options.Interop.MemberAccessor?.Invoke(Engine, Target, member);
                 if (result is not null)
                 {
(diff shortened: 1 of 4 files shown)


Jint/Runtime/Interop/ObjectWrapper.cs
```diff
@@ -98,6 +98,16 @@ public override JsValue Get(JsValue property, JsValue receiver)
                 return Undefined;
             }
 
+            if (Target is IDictionary<string, object> stringKeyedDictionary && property is JsString dicKey) // expando object for instance
```
lahma, line 101: this could also be part of next if also where we know it's a JsString and also `ToString` will be done? No sure if we could be preventing registerin custom ExpandObject accessor now in engine, but probably this is anyway what end user is after.

lahma (approved):

Jint/Runtime/Interop/ObjectWrapper.cs
sebastienros: Done

hyzx86:
996 what an amazing number 😂
Hoping to solve this problem, I went to try the following

sebastienros:
Do you refer to the 996 work schedule? I had to google that, seems very well-known in China though.

hyzx86:
No, we're 10 7 5 ！  996 is all work, no life.

back to the bug，the problem in my code probably didn't come from Jint
https://github.com/OrchardCMS/OrchardCore/discussions/10628

PR #1002, opened 2021-11-05
Generic string-keyed dictionary indexing under interop

### LINKED ISSUE #1001: Can't access JValue properties
Given this `JObject` payload:

```json
"Properties": {
"x": {
"Type": "Content",
"ContentId": "4m1pcbaa6t6gzr87qebcw0ysc1"
},
"expirationDate": {
"Value": "2021-10-09T00:00:00Z"
}
}
```

Trying to access the `Value` fails when the JValue contains a DateTime

lahma:
fixes #1001


sebastienros:
Only improvement would be to create a delegate instead of invoking the PropertyInfo.

sebastienros (approved):

lahma:
I think Delegate would require binding to target, here we have static type information and resolved Property which needs to be called with provided target.

Issue #1005, opened 2021-11-08
JObjectArray  Stringify error


hyzx86:
https://github.com/OrchardCMS/OrchardCore/issues/10648

```
      [Fact]
       public void EngineShouldStringifyJObjectArrayWithValuesCorrectly()
        {
           //https://github.com/OrchardCMS/OrchardCore/issues/10648
            var engine = new Engine();
            var source = new JObject[] {
                 JObject.FromObject(new { Text = "Text1", Value = 1 }),
                 JObject.FromObject(new { Text = "Text2", Value = 2 }) }; 

            engine.SetValue("testSubject", source);
            var fromEngine = engine.Evaluate("return JSON.stringify(testSubject);");
            var result = fromEngine.ToString();
            Assert.Equal(result, "[{\"Text\":\"Text1\",\"Value\":1},{\"Text\":\"Text2\",\"Value\":2}]");
        }
```
out put ：

```
[{"Type":null,"HasValues":null,"First":null,"Last":null,"Count":null,"Parent":null,"Root":null,"Next":null,"Previous":null,"Path":null},{"Type":null,"HasValues":null,"First":null,"Last":null,"Count":null,"Parent":null,"Root":null,"Next":null,"Previous":null,"Path":null}]
```

hyzx86:
This test can be run and passed,

```
        [Fact]
        public void EngineShouldStringifyDynamicObjectArrayWithValuesCorrectly()
        {
            var engine = new Engine();
            var source = new dynamic[] { new { Text = "Text1", Value = 1 }, new { Text = "Text2", Value = 2 } };

            engine.SetValue("testSubject", source.AsEnumerable());
            var fromEngine = engine.Evaluate("return JSON.stringify(testSubject);");
            var result = fromEngine.ToString();
            Assert.Equal(result, "[{\"Text\":\"Text1\",\"Value\":1},{\"Text\":\"Text2\",\"Value\":2}]");

        }
```
but not work in OrchardCore ，Even though I commented out these lines of code
https://github.com/OrchardCMS/OrchardCore/blob/afcc979897f0d1a0e26581066bce53f2c319e166/src/OrchardCore.Modules/OrchardCore.Queries/Sql/SqlQuerySource.cs#L91

```
IEnumerable<dynamic> queryResults;

                using (connection)
                {
                    await connection.OpenAsync();

                    using (var transaction = connection.BeginTransaction(_session.Store.Configuration.IsolationLevel))
                    {
[...]

PR #1009, opened 2021-11-16
Improve stringify for JObject

lahma:
* https://github.com/sebastienros/jint/issues/1005#issuecomment-968517060
* improves property traversal of JObject fixing serialization
* still requires custom JTokenConverter in order to map Newtonsoft primitives to correct JS primitives, we just don't know how to handle them otherwise

### DIFF
--- a/Jint/Runtime/Interop/ObjectWrapper.cs
+++ b/Jint/Runtime/Interop/ObjectWrapper.cs
@@ -98,37 +98,6 @@ namespace Jint.Runtime.Interop
                 return Undefined;
             }
 
-            if (property is JsString stringKey)
-            {
-                var member = stringKey.ToString();
-
-                // expando object for instance
-                if (_typeDescriptor.IsStringKeyedGenericDictionary)
-                {
-                    if (_typeDescriptor.TryGetValue(Target, member, out var value))
-                    {
-                        return FromObject(_engine, value);
-                    }
-                }
-
-                var result = Engine.Options.Interop.MemberAccessor?.Invoke(Engine, Target, member);
-                if (result is not null)
-                {
-                    return result;
-                }
-
-                if (_properties is null || !_properties.ContainsKey(member))
-                {
-                    // can try utilize fast path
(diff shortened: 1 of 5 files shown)


Jint/Runtime/Interop/ObjectWrapper.cs
```diff
@@ -166,7 +136,9 @@ private IEnumerable<JsValue> EnumerateOwnPropertyKeys(Types types)
                 // we take values exposed as dictionary keys only
                 foreach (var key in dictionary.Keys)
                 {
-                    if (_engine.ClrTypeConverter.TryConvert(key, typeof(string), CultureInfo.InvariantCulture, out var stringKey))
+                    object stringKey = key as string;
+                    if (stringKey is not null
+                        || _engine.ClrTypeConverter.TryConvert(key, typeof(string), CultureInfo.InvariantCulture, out stringKey))
```
sebastienros, line 141: Would probably work without the custom IObjectConverter by checking for `IConvertible` and ultimately for `IFormattable` if we know we want a string.

sebastienros: That's how Fluid can handle `JToken` without configuration.

lahma: Here I just wanted to create a fast path when we already have a string, but might be worth some PR to improve string coercion

lahma:
@sebastienros OK to merge?

sebastienros (approved):

sebastienros:
> we just don't know how to handle them otherwise

`IFormattable` and `IConvertible`

lahma:
Did I get it right? Cool stuff, nonetheless.

sebastienros:
It's better than what I do in FLuid ... I might want to do it this way too. Thanks!

sebastienros:
Apparently `Convert.ChangeType(Object, Type)` will invoke these methods if the object implements `IConvertible`.

lahma:
This version doesn't (probably) cause boxing so I guess this is a bit more optimal for these primary types. We don't need to go finding a converter on a second pass when we do this already in our own code. Can I merge or something to improve?

PR #1011, opened 2021-11-19
Improve operator overloading logic

### LINKED ISSUE #1004: Operator overloading and string concatenation
Jint version: 1426408817.0.0

Usually when using the `+` operator on a class/struct and a string it concatenates the result of the class/struct `ToString()` method and the string. However, when enabling operator overloading and implementing an override for `+` this automatic conversion stops working.

Have a look at this example:

```cs
using System;
using Jint;
using Jint.Runtime.Interop;

namespace jinttestx
{
	struct Vector2D
	{
		public double X { get; set; }
		public double Y { get; set; }

		public Vector2D(double x, double y)
		{
			X = x;
			Y = y;
		}

		public static Vector2D operator +(Vector2D lhs, Vector2D rhs)
		{
			return new Vector2D(lhs.X + rhs.X, lhs.Y + rhs.Y);
		}

		public override string ToString()
		{
			return $"({X}, {Y})";
		}
	}

	class Program
	{
		{
			Engine engine = new Engine(cfg => cfg.AllowOperatorOverloading());
			engine.SetValue("Vector2D", TypeReference.CreateTypeReference(engine, typeof(Vector2D)));
			engine.SetValue("log", new Action<object>(Console.WriteLine));

			engine.Execute(@"

let v1 = new Vector2D(1, 2);
log(v1); // Works fine, prints '(1, 2)'
log('### ' + v1 + ' ###'); // Results in InvalidCastException

				");

			Console.WriteLine("End of progream reached");
			Console.ReadLine();
		}
	}
}
```
The example shows that the `+` operator is overloaded to add two `Vector2D`. In the JS portion I use the `+` operator with the intention of string concatenation, but doing so throws an `InvalidCastException`. It works fine when commenting out the overloading method.

This can be prevented by adding more overloading method, i.e. ones that accept strings for the left hand and right hand side:

```cs
public static string operator +(Vector2D lhs, string rhs)
{
	return lhs.ToString() + rhs;
}

public static string operator +(string lhs, Vector2D rhs)
{
	return lhs + rhs.ToString();
}
```

Is it intentional that the automatic string concatenation fails and has to be implemented manually?

### LINKED ISSUE #1008: Wrong overloaded method is called under certain circumstances
Jint version: 1439970634.0.0

When using overloaded method Jint usually choses the correct method to call depending on the parameters given in the JS code. However, this stops working when one of the overloaded methods accepts `object`. In that case the method accepting `object` is always called, except for - as far as I can tell - a version that accepts `ExpandoObject`.

Example:

```cs
using System;
using System.Dynamic;
using Jint;
using Jint.Runtime.Interop;

namespace jinttestx
{
	struct Class1
	{
		public static void Print(ExpandoObject eo)
		{
			Console.WriteLine("Class1: method Print(ExpandoObject eo)");
		}

		public static void Print(double[] a)
		{
			Console.WriteLine("Class1: method Print(double[] a)");
		}

		public static void Print(int i)
		{
			Console.WriteLine("Class1: method Print(int i)");
		}
	}

	struct Class2
	{
		public static void Print(ExpandoObject o)
		{
			Console.WriteLine("Class2: method Print(ExpandoObject eo)");
		}

		public static void Print(double[] a)
		{
			Console.WriteLine("Class2: method Print(double[] a)");
		}

		public static void Print(int i)
		{
			Console.WriteLine("Class2: method Print(int i)");
		}

		public static void Print(object o)
		{
			Console.WriteLine("Class2: method Print(object o)");
		}
	}

	class Program
	{
		{
			Engine engine = new Engine(cfg => cfg.AllowOperatorOverloading());
			engine.SetValue("Class1", TypeReference.CreateTypeReference(engine, typeof(Class1)));
			engine.SetValue("Class2", TypeReference.CreateTypeReference(engine, typeof(Class2)));
			engine.SetValue("log", new Action<object>(Console.WriteLine));

			engine.Execute(@"

Class1.Print([ 1, 2 ]); // Calls Print(double[] a)
Class1.Print({ x: 1, y: 2 }); // Calls Print(ExpandoObject eo)
Class1.Print(5); // Calls Print(int i)
log('-----');
Class2.Print([ 1, 2 ]); // Calls Print(object o), but should call Print(double[] a)
Class2.Print({ x: 1, y: 2 }); // Calls Print(ExpandoObject eo) as expected
Class2.Print(5); // Calls Print(object o), but should call Print(int i)

				");

			Console.WriteLine("End of progream reached");
			Console.ReadLine();
		}
	}
}
```

This example has two classes, `Class1` and `Class2`. Both have a method called `Print` that's overloaded with several different parameter types. Unlike `Class1`, `Class2` also has a method that accepts `object` as a parameter. When called the methods just print out their name and parameter.

The JS code calls all those methods of both classes, and while it calls the appropriate method in `Class1` as expected, for `Class2` it calls the method accepting `object` for the first and third call, while I'd expect it to call the methods accepting `double[]` and `int`, just as for `Class1`.

lahma:
Hopefully the overload selection logic now a bit easier to read and maintain.

* cleanup and refactor code
* use `struct record` instead of value tuples and custom code
* prefer more specific types over object
* prefer CLR array against JS array match over changing types
* as `TryOperatorOverloading` is `Try` pattern, it should not throw on conversion error, just return false

fixes #1004
fixes #1008


Jint/Runtime/Interpreter/Expressions/JintBinaryExpression.cs
```diff
@@ -11,12 +11,8 @@ namespace Jint.Runtime.Interpreter.Expressions
 {
     internal abstract class JintBinaryExpression : JintExpression
     {
-#if NETSTANDARD
-        private static readonly ConcurrentDictionary<(string OperatorName, System.Type Left, System.Type Right), MethodDescriptor> _knownOperators =
-            new ConcurrentDictionary<(string OperatorName, System.Type Left, System.Type Right), MethodDescriptor>();
-#else
-        private static readonly ConcurrentDictionary<string, MethodDescriptor> _knownOperators = new ConcurrentDictionary<string, MethodDescriptor>();
-#endif
+        private readonly record struct OperatorKey(string OperatorName, Type Left, Type Right);
```
sebastienros, line 14: https://nietras.com/2021/06/14/csharp-10-record-struct/
TL;DR;
Amen

sebastienros (approved):

Jint/Runtime/TypeConverter.cs
```diff
@@ -845,84 +858,156 @@ public static void CheckObjectCoercible(Engine engine, JsValue o)
...
+        /// Method's match score tells how far away it's from ideal candidate. 0 = ideal, bigger the the number,
+        /// the farther away the candidate is from ideal match. Negative signals impossible match.
+        /// </summary>
+        private static int CalculateMethodScore(Engine engine, MethodDescriptor method, JsValue[] arguments)
+        {
+            if (method.Parameters.Length == 0 && arguments.Length == 0)
+            {
+                // perfect
+                return 0;
+            }
```
KurtGokhan, line 882: Does this mean method will match perfectly if argument lengths are the same, regardless of the argument types? Isn't this wrong, or am I missing something?

lahma: Method has no parameters and JS call has no arguments? What is there to match?

KurtGokhan: I had a little brain fart. I saw another thing when reviewing.

Sorry for the trouble. Thank you.

lahma: No worries, I always appreciate reviews and and extra pair of eyes 👍🏻

Issue #1016, opened 2021-12-02
Write some documentation for debugger support


PeterWone:
@Jither You've done so much work on this, it would be a tragedy for it to go unused because there's no documentation whatsoever. Or if there is, it's not turning up in searches.

Is the debugger interface based on some standard for which docs are available?

Perhaps a worked comprehensive demonstration by using it from a well known debug client implementation like Visual Studio Code. This would not only show people how to use it, it would be a worthwhile piece of software in its own right.

Jither:
There's a bit of indirect documentation in this semi-duplicate: https://github.com/sebastienros/jint/issues/674 - and the source documentation very much follows Chromium concepts and terminology.

Unfortunately, I have way too many projects going on at the moment. Started some work on implementing the DevTools protocol around the debugger (and adding support to Jint itself - mostly rewriting the Breakpoint implementation to be more intuitive and less error prone). But that's on hold too - and quite a bit has happened to Jint since I last looked at it: https://github.com/Jither/Jint.DevToolsProtocol

I do intend to look at it again at some point, but can't really say when that point will be. 😐 But until then (or until someone else jumps at it):

This pretty much demonstrates use of the main (breakpoint/stepping) API: https://github.com/Jither/Jint.DevToolsProtocol/blob/main/Jint.DevToolsProtocol/Protocol/Debugger.cs

The rest of the API really consists of the properties in https://github.com/sebastienros/jint/blob/main/Jint/Runtime/Debugger/DebugInformation.cs - which are all documented in the source.

PeterWone:
I can see that it shouldn't be all on you. I'll look at this myself though I doubt work and my other project will allow a something this big.

Is the design of what you're building wholly original or based on some convention or another work?

Jither:
It's really based on the API that was already in Jint before I started - just elaborated on it. The only change I've made to the public API itself is moving breakpoints into a separate class and adding more information to DebugInformation. All of that information is based on the DevTools protocol (i.e. https://chromedevtools.github.io/devtools-protocol/)

Jither:
I've (manually) rebased the changes in the `devtools-protocol` branch in my fork onto the latest Jint changes and placed it in a new `debugger-improvements-4`, which I may make a pull request for soon. The `devtools-protocol` branch is now abandoned, and it's likely that I'll focus on implementing a DebugAdapter for VS Code as an example, rather than continue on the devtools protocol server.

The DebugAdapter approach has several advantages:

- Better documentation and stability
[...]

PR #1018, opened 2021-12-05
Debugger improvements 4

Jither:
As discussed, and referenced in https://github.com/sebastienros/jint/issues/1016

Minor compared to previous debugger-improvements PR's, but a relatively important step for Chrome devtools protocol and/or Debug Adapter Protocol (VSCode).

## Summary of changes
 __bold__ areas = breaking changes (breaking for debugger implementations, that is)

- Event (`Parsed`) on engine for notification of when scripts have been through Esprima. This avoids the need for double parsing, because debugger implementations are likely to need the AST too (for e.g. determining proper locations for breakpoints).
- Closer to Chromium devtools scopes: __Splitting global scope__ into Global (Object Environment Record) and Script (Declarative Environment Record).
- Debug scopes now hold a reference to the BindingObject for OER's - this eases inspection a lot in devtools, and probably (not checked yet) in Debug Adapter Protocol (VSCode) - because the BindingObject can simply be serialized for the protocol, rather than copying every single property to a new object and serializing that.
- Breakpoints:
    - Only allows a single breakpoint at a given location (source/line/column) - hence, __API is now "Set" instead of "Add"__.
    - __Requires the breakpoint to be set at *exactly* the correct line/column__ (this allows faster matching, column breakpoints, and more).
    - Additional BreakPoint tests
    - Breakpoint class unsealed - it's useful for a debugger to be able to add additional data to the class. As such, the data important to the Jint side (other than condition - source, line, column) has moved into a sealed `BreakLocation` class.
- Fixed tests broken by addition of Script scope and new breakpoint semantics.
- Fixed inspection of uninitialized block scoped bindings (const/let)

Breakpoints still need work on the API (and probably implementation) side.


Jint/Runtime/Debugger/BreakLocation.cs
```diff
@@ -0,0 +1,117 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Jint.Runtime.Debugger
+{
+    /// <summary>
+    /// BreakLocation is a combination of an Esprima position (line and column) and a source (path or identifier of script).
+    /// Like Esprima, first column is 0 and first line is 1.
+    /// </summary>
+    public sealed class BreakLocation : IEquatable<BreakLocation>
```
lahma, line 10: you could also start using records here, would remove a lot of boiler-plate code

Jither: Done 🙂

Jint/Runtime/Debugger/BreakLocation.cs
```diff
@@ -0,0 +1,77 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Jint.Runtime.Debugger
+{
+    /// <summary>
+    /// BreakLocation is a combination of an Esprima position (line and column) and a source (path or identifier of script).
+    /// Like Esprima, first column is 0 and first line is 1.
+    /// </summary>
+    public sealed record BreakLocation
```
lahma, line 10: btw, if you want really make the code short, this should also work:

```csharp
public sealed record BreakLocation(string Source, int Line, int Column)
{
    public BreakLocation(int line, int column) : this(null, line, column)
    {
    }

    public BreakLocation(string source, Esprima.Position position) : this(source, position.Line, position.Column)
    {
    }
}
```
 These are of course personal preferences and not important for this PR, I've just really liked the expressiveness they have 🙂

Jint/Runtime/Debugger/BreakPoint.cs
```diff
@@ -1,32 +1,28 @@
 ﻿namespace Jint.Runtime.Debugger
 {
-    public sealed class BreakPoint
+    // BreakPoint is not sealed. It's useful to be able to add additional properties on a derived BreakPoint class (e.g. a breakpoint ID
+    // or breakpoint type) but still let it be managed by Jint's breakpoint collection.
+    public class BreakPoint
```
lahma, line 5: record could also work here

Jint/Runtime/Debugger/BreakPointCollection.cs
```diff
@@ -1,98 +1,100 @@
...
-    public class BreakPointCollection : ICollection<BreakPoint>
+    /// <summary>
+    /// Collection of breakpoints.
+    /// </summary>
+    /// <remarks>
+    /// Only allows a single breakpoint at the same location (source, column and line).
+    /// Adding a new breakpoint at the same location <i>replaces</i> the old one - this allows replacing e.g. a 
+    /// conditional breakpoint with a new condition (or remove the condition).
+    /// </remarks>
+    public class BreakPointCollection : IEnumerable<BreakPoint>
```
lahma, line 14: should be sealed unless this is a extension point

Jint/SourceParsedEventArgs.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Esprima.Ast;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Jint
+{
+    public class SourceParsedEventArgs
```
lahma, line 10: should be sealed if not an extension point, maybe even a record 😉

Jint/Runtime/Debugger/BreakLocation.cs
Jither: Not a fan (yet) of positional records, when the properties aren't self-evidently positional (emphasized here by the multiple constructors). But that's also just philosophy. May change it anyway. 😉

Jint/Runtime/Debugger/BreakPoint.cs
Jither: For the same reason as above, not sure `record` adds something to `BreakPoint` - value equality, yes, but not entirely sure if value equality is even appropriate for a breakpoint. 🙂 (But that's just my impaired brain speaking)

Jint/SourceParsedEventArgs.cs
Jither: Sealed, absolutely. But philosophy-wise, here I'm quite sure record doesn't add much, though - can't remember ever needing to compare two eventargs. 🙂 (and since some EventArgs may oftentimes need to be mutable, I'll generalize and say no EventArgs should be records, even when they're immutable 😉).

Jint/Runtime/Environments/DeclarativeEnvironmentRecord.cs
```diff
@@ -122,6 +122,18 @@ public sealed override JsValue GetBindingValue(string name, bool strict)
             return null;
         }
 
+        internal override bool TryGetBindingValue(string name, bool strict, out JsValue value)
```
Jither, line 125: Keeping the implementations entirely separate from `GetBindingValue`, simply because code reuse would either require catching exceptions from `GetBindingValue` (somewhat defeating the point of `Try...` in the first place). Or `GetBindingValue` would have to make a call to `TryGetBindingValue`, which is what I'd do in just about any other project. Not sure if a call could be hinted as inlined, but if any method in Jint doesn't need additional overhead, it's probably `GetBindingValue` - is there anything in Jint that's called more often during interpretation?

lahma: Yes this is quite performance critical and inlining could get hard due to virtuals. I think this is the right approach.

Jint/SourceParsedEventArgs.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Jint
+{
+    public sealed class SourceParsedEventArgs
+    {
+        public string SourceId { get; }
```
lahma, line 12: nit: I think properties should come after constructors (fields before constructors), sorry for dropping notes but I seem to spot things in iterations, lazy brain or something..

Jither: Nah, that's _my_ lazy brain - we've been through that convention already. 🙂

Jither:
Hmmm... Judging from that list of commits, I clearly messed up when fetching the upstream. I'll probably manually rebase the changes onto the latest dev, when they're done - since so far, they don't interfere at all with the rest of the engine.

Jither:
Messed up history - will open a new clean slate, based on latest main

PR #1027, opened 2021-12-12
ES2020 BigInt

lahma:
* can ignore .js file changes with filter - only updating test suite under `BigInt` folder
* new `BigInt` primitive type
* moved logic towards `JsValue.Equals` (===, strict) and `JsValue.LooseEquals` (==)
* about 800 new test cases passing

There might be things to improve, but I think some refactoring can be left later. I'm probably going to try separate operator overloading via decorators or something to make core logic more concise.


lahma:
I think I still need to validate against latest test suite and bigint related tests.

lahma:
All bigint related tests pass in my custom workbench after two latest commits, still requires new Esprima release for full compatibility.

sebastienros (approved):

PR #1032, opened 2021-12-22
#278309: Bugfix/fix stack overflow error on parseint

rafael-santiago-take:
Fixed stack overflow error when call parseint method with big strings

### DIFF
--- a/Common.props
+++ b/Common.props
@@ -1,8 +1,7 @@
 ﻿<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
-    <Copyright>Sebastien Ros</Copyright>
-    <Authors>Sebastien Ros</Authors>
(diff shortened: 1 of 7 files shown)


lahma:
wrong repo?

mhnakashima (approved):

rafael-santiago-take:
yes, wrong rpo. sorry.

sebastienros:
Or maybe not really, if there are things worth fixing please file issues, and if you are fixes worth contributing please file the PRs.

lahma:
Well this particular fix seems to make no sense, but a failing test case might educate me more about the number parsing problem. We also have BigInts now!

sebastienros:
It's just that I checked, and these packages (Esprima and Jint) actually exist on NuGet ... not sure it helps the ecosystem to have a separate set of packages with very similar names.

sebastienros:
Though the "unofficial" in the name makes it clearer.

PR #1033, opened 2021-12-26
Fix some unicode tests

sebastienros:


sebastienros:
I believe all the ones that fail are the ones that are now fixed.

Jint.Tests.Test262/Test262Test.cs
```diff
@@ -323,18 +323,86 @@ public static IEnumerable<object[]> SourceFiles(string pathPrefix, bool skipped)
                     reason = "Esprima problem, Unexpected token *";
                 }
 
+                if (name.StartsWith("language/white-space/mongolian-vowel-separator-eval.js"))
```
lahma, line 326: if it's a full file name it can also be in skipped.json , there's own section for esprima problems

README.md
```diff
@@ -42,9 +42,9 @@ The entire execution engine was rebuild with performance in mind, in many cases
 - ✔ Proxies
 - ✔ Reflect
 - ✔ Symbols
-- ❌ Tail calls
+- ✔ Tail calls (_deprecated_)
```
lahma, line 45: I don't think we can make this claim. This would special logic in runtime and we won't currently pass the test cases. Maybe we could just state `Won't be implemented`.

lahma (approved):
Looking good, one comment about tail calls support claim.

Issue #1035, opened 2021-12-30
int vs float in function overloads


CreepGin:
In working with math-heavy libraries and/or game engine APIs, it'd be helpful for C# function overloads to be able to distinguish between ints and floats when called from JS. Jint works with doubles behind the scene for obvious reasons. But essentially, I'd like the following to work.

```C#
// C#
public class Foo {
    public static void Ace(float f) {
        print("FLOAT!");
    }
    public static void Ace(int i) {
        print("INT!");
    }
}
```

```C#
// C#
jsEngine.SetValue("f", new Func<Double, float>((obj) => {
    return (float)obj;
}));
jsEngine.SetValue("i", new Func<Double, int>((obj) => {
    return (int)obj;
}));
```

```JS
// JS
Foo.Ace(f(12.3))  // calls the float version
Foo.Ace(i(123))   // also calls the float version, but ideally should be calling the int version
```

Right now with Jint, only the float version will be called both times (because it's defined first). To fix that, do I need to modify TypeConverter.CalculateMethodScore? Or are there easier alternatives?

Thanks

lahma:
> To fix that, do I need to modify TypeConverter.CalculateMethodScore? Or are there easier alternatives?

I think that is the preferred approach. You can check for `JsNumber` and then `if (jsNumber.IsInteger() || TypeConverter.IsIntegralNumber(jsNumber._value)) { give better score than float/double }`. There's of course other integral ones like uint etc, but in should be quite safe default to check.

CreepGin:
Thanks! Yes this route is probably easier. No longer need to rely on some intermediate functions like `f` and `i` above. 

I made the changes and a test. Everything seem to work fine on my side so far. Made a pull request in case this is desirable for Jint by default.

#1036

PR #1036, opened 2021-12-30
added int vs float diff for function overloads

CreepGin:
#1035 

First time contributing here. It felt appropriate to follow and reuse the existing Enum resolution/overloading code and tests.

Also, I added an extra check in the existing Enum resolution code `paramType.GetEnumUnderlyingType() == typeof(int)` because `Enum.IsDefined` would throw if `paramType.GetEnumUnderlyingType()` is not an int (as in the case of an overloaded function with an uint enum param).


lahma:
Thank you!

PR #1037, opened 2021-12-30
Fix JSON-serialization issues (function-valued properties)

resnickj:
Issues:
* function-valued properties should be excluded entirely from JSON, not serialized as objects
* proxies incorrectly treated as callable

Changes:
* Remove extraneous early exit case (JsonSerializer line 175) which diverges from the algorithm in the spec (https://tc39.es/ecma262/#sec-serializejsonproperty). This appears to correct the serialization of function-valued properties, but unmasks another issue where ProxyInstance is mistakenly treated as callable even when target is not
* Change "value is ICallbable" test to "value.IsCallable" so that serialization of ProxyInstance is handled correctly (according to the spec proxies are callable only if the target is callable - https://tc39.es/ecma262/#sec-proxycreate)
* Add test to verify serialization of function-valued properties

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -2059,6 +2059,20 @@ var prep = function (fn) { fn(); };
             Assert.True(res == "Cyclic reference detected.");
         }
 
+        [Fact]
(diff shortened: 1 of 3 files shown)


lahma:
Thank you!

lahma (approved):

PR #1038, opened 2022-01-06
Fix JsonParser error behavior

resnickj:
Issues:
* JsonParser sometimes throws JavaScriptException (SyntaxError) and sometimes throws ParserException; should be consistent and always throw one or the other.
* A script that calls JSON.parse(input) should always see a SyntaxError in the case of syntactically invalid input.
* SyntaxError messages should clearly indicate that the error is in JSON data, and indicate the position within the input string.
* SyntaxError messages should not include the entire input string in the message, as it may be arbitrarily long.

Changes:
* Route all exceptions through a common ThrowError(...) method that always throws a SyntaxError
* Clean-up content of error messages and include position of the error. Unfortunately the parser does not track line number accurately, so position is rendered as a single number as opposed to line:column format.
* Add tests to verify the expected error output for each call site of ThrowError(...)

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -1006,8 +1006,8 @@ namespace Jint.Tests.Runtime
         [Fact]
         public void JsonParserShouldHandleEmptyString()
         {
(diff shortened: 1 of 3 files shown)


lahma (approved):

lahma:
Great work, thank you!

PR #1039, opened 2022-01-07
Use Test262Harness to run tests

lahma:
* tests loaded from test262 repository
* test cases generated from test files
* settings file used to know what to ignore/include during generation
* deleted EcmaTest.csproj so they won't be run anymore, separate *BIG* delete commit required for deleting test files in VCS


NuGet.config
```diff
@@ -4,6 +4,7 @@
     <clear />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
     <add key="Esprima" value="https://www.myget.org/F/esprimadotnet/api/v3/index.json" />
+    <add key="Local" value="libs" />
```
sebastienros, line 7: left-over?

Jint/Native/Date/DateConstructor.cs
```diff
@@ -87,9 +87,26 @@ protected override void Initialize()
             SetProperties(properties);
         }
 
+        /// <summary>
+        /// https://tc39.es/ecma262/#sec-date.parse
+        /// </summary>
         private JsValue Parse(JsValue thisObj, JsValue[] arguments)
         {
             var date = TypeConverter.ToString(arguments.At(0));
+            var negative = date.StartsWith("-");
```
sebastienros, line 96: wtf

Jint.Tests.Test262/TestHarness.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        // NOTE: The Date tests in test262 assume the local timezone is Pacific Standard Time
+        try
+        {
+            State.TimeZone = TimeZoneInfo.FindSystemTimeZoneById("Pacific Standard Time");
+        }
+        catch (TimeZoneNotFoundException)
+        {
+            // https://stackoverflow.com/questions/47848111/how-should-i-fetch-timezoneinfo-in-a-platform-agnostic-way
+            // should be natively supported soon https://github.com/dotnet/runtime/issues/18644
```
sebastienros, line 24: It's a lie. It depends on how Windows is configured, so for instance it won't work on Azure App Service.

sebastienros (approved):

Jint.Tests.Test262/TestHarness.cs
lahma: Good point, I think this is "old code".

Jint/Native/Date/DateConstructor.cs
lahma: Trial error horror I guess, I will be having only things related to test in this PR when merged, will exclude Jint changes.

NuGet.config
lahma: Still needed for NUnit local lib

lahma:
I'll try to port fixes to main to reduce failures, I'd just hate to create a large todo list by having skipped test cases which fail in new tests, but the amount of failures is a bit overwhelming to fix..

NuGet.config
lahma: No longer needed, using nuget.org package now

lahma:
/cc recent code contributors @christianrondeau @Jither @KurtGokhan @source-transformer @shlomo-ildar 

You now need to generate Test262 test suite in order for the solution to compile. See [Jint.Tests.Test262](https://github.com/sebastienros/jint/tree/main/Jint.Tests.Test262) project's README.

PR #1042, opened 2022-01-09
Allowing to use nullable propagation on callables again. Fixed #1041

### LINKED ISSUE #1041: RavenDB: referenceResolver.TryGetCallable is not called after switching to new Jint version
Hi! 
We have switched to new Jint version in branch RavenDB-16912 (is to get into release 6.0):
old jint: https://github.com/ravendb/jint/commit/4b9c78e3fc3e1c96a8e54838a6181c3b519352f1
5:33
new jint: https://github.com/ravendb/jint/commit/d84af69871637c15e5db8a42fe5edb09cb397604

The test CanProjectFromCollectionNotInJson passes in v5.3 branch and fails in RavenDB-16912 on executing x.Tags.filter(...) in case x.Tags === undefined. Previously engine._referenceResolver.TryGetCallable() was called, but now it doesn't get called in line 184, as method EvaluateCall returns in line 170 after TryUnresolvableReference got called twice (first time in GetValue method):
https://github.com/sebastienros/jint/blob/main/Jint/Runtime/Interpreter/Expressions/JintCallExpression.cs#L166:171

ayende:
In `EvaluateCall`, we shouldn't short circuit if we detected that the value is null and the nullable reference returned true, we still need to call to the callable

### DIFF
--- a/Jint.Tests/Runtime/NullPropagation.cs
+++ b/Jint.Tests/Runtime/NullPropagation.cs
@@ -1,4 +1,5 @@
-﻿using Esprima;
+﻿using System;
+using Esprima;
 using Jint.Native;
 using Jint.Runtime;
 using Jint.Runtime.Interop;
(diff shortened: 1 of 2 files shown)


Jint.Tests/Runtime/NullPropagation.cs
```diff
@@ -22,8 +23,18 @@ public bool TryPropertyReference(Engine engine, Reference reference, ref JsValue
...
 
-            public bool TryGetCallable(Engine engine, object reference, out JsValue value)
+            public bool TryGetCallable(Engine engine, object callee, out JsValue value)
             {
+                if (callee is Reference reference)
+                {
+                    var name = reference.GetReferencedName().AsString();
+                    if (name == "filter")
+                    {
+                        value = new ClrFunctionInstance(engine, "map", (thisObj, values) => engine.Array.Construct(Array.Empty<JsValue>()));
```
sebastienros, line 33:
```suggestion
                        value = new ClrFunctionInstance(engine, "map", (thisObj, values) => engine.Realm.Intrinsics.Array.ConstructFast(0));
```

sebastienros (approved):

lahma (approved):

lahma:
GitHub requires merge sync from main before merge, and we cannot do it as source branch is main.

PR #1047, opened 2022-01-11
Add Engine.Construct method

### LINKED ISSUE #1046: Expose a canonical way to construct a user-defined class
Assume I have a user-defined class, Test, defined as follows:

```
engine.Execute(
          "Test = /** @class */ (function() {\n" +
          "  function Test(a) {\n" +
          "    this.a = a;\n" +
          "  }\n" +
          "  return Test;\n" +
          "}());");
```

What is the canonical way to construct an instance of this class?

The closest answer seems to be:

```
var jsValue = engine.GetValue("Test");
var args = ...;
var instance = ((IConstructor)jsValue).Construct(args, jsValue);
```

According to @sebastienros, from our gitter chat https://gitter.im/sebastienros/jint?at=61dcd02446529077f57ac218

> the "issue is that we can't call Construct like this on the engine directly
> it expected an environment context to be created, which we don't have when calling it directly
> only if invoked from an Execute method
> I had to hard-code it in the Construct method
> I believe this method should be internal knowing this. And expose a new one specifically to simulate a new() and create the necessary context like Execute is doing."

@lahma, what are your thoughts? I'm happy to code up a PR if given some direction.

lahma:
Now possible:

```csharp
engine.Evaluate("class TestClass { constructor(a, b) { this.a = a; this.b = b; }}");

ObjectInstance instanceFromClass = engine.Construct("TestClass", "abc", 123);
Assert.Equal("abc", instanceFromClass.Get("a"));
Assert.Equal(123, instanceFromClass.Get("b"));
```

fixes #1046

### DIFF
--- a/Jint.Tests/Runtime/FunctionTests.cs
+++ b/Jint.Tests/Runtime/FunctionTests.cs
@@ -1,5 +1,6 @@
 using System;
 using Jint.Native;
+using Jint.Native.Array;
 using Jint.Runtime;
 using Jint.Runtime.Interop;
 using Xunit;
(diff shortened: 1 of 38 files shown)


sebastienros (approved):

markroth8:
This is amazing. Thank you, @lahma!

PR #1049, opened 2022-01-11
Allow JS class to extend CLR type

### LINKED ISSUE #1048: Allow extending ES6 class from a base CLR type; currently throws with JavascriptException (message: "null")
I'd like to implement a base class in C# that users will be able to inherit from, but it looks like it's not supported. Because the error message throws with "null", I'm not sure if it should be supported, if I'm doing something wrong, or if at least there should be a better error message.

To reproduce:

```c#
public class MyBaseType
{
    public string Value { get; set; } = "Hello World!";
}

[Test]
public void CanInheritFromClrType()
{
    var engine = new Engine();
    engine.SetValue("MyBaseType", TypeReference.CreateTypeReference<MyBaseType>(engine));
    // This line fails with JavascriptException, message = null
    engine.Execute("class MyExtendedType extends MyBaseType {}");
    engine.Invoke("let obj = new MyExtendedType();");
    var value = engine.Invoke("obj.Value");

    Assert.That(value.AsString(), Is.EqualTo("Hello World!"));
}
```

I think this is related to #165

Note that I can achieve a similar result by implementing the base class using `engine.Invoke("class MyBaseType { ... }")` but it's more practical to make a base class in C# in this case.

lahma:
I'm probably opening a can of worms here.

fixes #1048

### DIFF
--- a/Jint.Tests/Runtime/InteropTests.cs
+++ b/Jint.Tests/Runtime/InteropTests.cs
@@ -787,6 +787,23 @@ namespace Jint.Tests.Runtime
             ");
         }
 
+        [Fact]
+        public void JavaScriptClassCanExtendClrType()
+        {
+            var engine = new Engine();
+            engine.SetValue("TestClass", TypeReference.CreateTypeReference<TestClass>(engine));
+
+            engine.Execute("class ExtendedType extends TestClass { constructor() { super(); this.a = 1; } }");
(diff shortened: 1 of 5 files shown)


sebastienros:
You'd better define all the limitations beforehand or bugs/missing features will keep flowing.

lahma:
I was able to get it work with my test case which should cover most obvious things, deep inheritance chain with accessing properties from both JS and base CLR type.

Issue #636, opened 2019-07-19
Support Esprima.Ast.ExportNamedDeclartion


MeikelLP:
I want to write a Javascript file like

```js
export var name = 'Mod1'
```

But this gives me: 
```
InvalidCastException: Unable to cast object of type 'Esprima.Ast.ExportNamedDeclaration' to type 'Esprima.Ast.Statement'.
  at Jint.Runtime.Interpreter.JintStatementList.Initialize() in C:\projects\jint\Jint\Runtime\Interpreter\JintStatementList.cs:34
```

This is an ES6 feature.

lahma:
Import/export should be covered in https://github.com/sebastienros/jint/issues/430 , I'll close this and we should continue tracking on the specific feature issue.

christianrondeau:
It seems #430 was closed, but as far as I can tell it's still impossible to use something like `engine.Execute("export const x = 12;");`.

I guess it would also be useful to have a way to "get" those exported values. Right now the alternative is to do something like `x = 12;` which creates a global variable. Using `const x = 12` makes `x` innaccessible from outside the JS code.

Should this be re-opened, did I miss something or should I open a new issue instead?

Thanks for the amazing project by the way :)

lahma:
The module runtime behavior is still lacking, the static constructs and services are there, but runtime evaluation is missing some bits.

christianrondeau:
I was considering investigating at least two things, so I'm not creating a monster :) My questions:

1. What do you think of the following? Would that be aligned with how Jint is supposed to work, or am I stretching?
2. do you have any pointers or ideas on how it should work, or should I start with a PR outlining the idea first?

## Ability to use the export keyword and make exported objects visible in a dictionary on the engine.

Given the script

```javascript
export const x = 5;
```

```c#
engine.Execute(script);

Assert.That(engine.Module.Exports["x"].AsInt(), Is.EqualTo(5));
```

(I'll work on the details such as default exports if that makes sense to you)

## Ability to resolve C#-provided imports

```javascript
import { MyType } from '@mycompany/whatever';
```

```c#
var engine = new Engine(opts => opts.AddModuleResolver<MyModuleResolver>());
```

```c#
public class MyModuleResolver : IModuleResolver {
    // This has to be fleshed out once I better understand how Esprima implements this
}
```

lahma:
[...]

PR #1054, opened 2022-01-17
Support ECMAScript modules (export/import statements, modules definition)

christianrondeau:
This attempts to implement ECMAScript modules, to resolve https://github.com/sebastienros/jint/issues/636. This is a draft and will definitely change, but as of now, I think the basic "export named" flow is mostly correct.

What I want to implement:

- [x] Full export specification
  - [x] Export named
  - [x] Export default
  - [x] Export all
- [x] Create modules from C# (engine.DefineModule)
- [x] Import modules from C# (engine.ImportModule)
- [x] Create modules from a C# class
  - [x] Create modules that import multiple C# classes in a single module, e.g. `import { Class1, Class2 } from 'my-module';`
- [x] Partial import specification (exclude promise import and focus on C#-defined modules)
- [x] File-based module loading (from @lahma branch)

I do not intend on implementing module file resolution (only modules specified in C#) for this PR, but I feel like there may not be much left to do once this works.

Note that I did (and will) use code from @lahma https://github.com/sebastienros/jint/pull/1051 - the main difference is that I started with the assumption that I would only implement `export`, however, to make sure my implementation is correct, I need to implement at least one simple `import` statement.


Jint/Engine.Modules.cs
```diff
@@ -28,6 +42,60 @@ internal JsModule LoadModule(string referencingModuleLocation, string specifier)
...
+
+            return module;
+        }
+
+        public ObjectInstance ImportModule(string specifier)
+        {
+            var key = new ModuleCacheKey(string.Empty, specifier);
+
+            if (!_modules.TryGetValue(key, out var module))
+                throw new ArgumentOutOfRangeException(nameof(specifier), "No module was found for this specified");
```
sebastienros, line 68: Typo on 'specified'. I assume wither 'specifier' or the actual name used.

sebastienros:
Was wondering why Christian's name was so familiar. It's because he's been active on Jint since 2015, way before Marko.

Jint/Engine.Modules.cs
christianrondeau: Fixed :)

christianrondeau:
@sebastienros yeah it's been a while, I was glad to see that Jint was still alive and well!

So, I went much further than I initially planned to, but I felt that without all those pieces interacting together, I wouldn't be able to trust that my approach made sense (and indeed, by adding pieces I realized a few times I approached it wrong!)

There are a lot of things in this PR, so while I think a good review would be important, I'll try and raise some specific uncertainties I had myself so we can get them out of the way. I'll use the GitHub code comments to do so, I'll write a message when I reviewed it all myself at least once. After those are resolved, we can proceed with an actual review.

I don't think I completely mapped to the ECMAScript specifications, but to be honest I have trouble fully understanding it. I'll try and review it the best I can and raise questions if I'm not sure.

There are also a few things that are Jint-specific:

1. The ability to create a module in C#, e.g. `engine.DefineModule("export const x = 1;", "my-module")`
2. The ability to  create a module from values directly, e.g. `engine.DefineModule<MyClass>()` or `engine.DefineModule("exportName", JsString.Create("Hello"), "my-module");
3. The ability to import a module from C#, e.g. `var ns = engine.ImportModule("specified")` which allows reading the exports in C#
4. I added a new `DefaultModuleResolver`, which is responsible for resolving paths and/or bare specifiers. This is a small breaking changes since now the `DefaultModuleLoader` is not responsible for doing the resolution, and receives a `ResolvedSpecifier`. This was necessary in order to have modules `./my-module` and `../folder/my-module` reference the same module instance.

I hope this PR will see the light of day and that I didn't do anything dumb :)

christianrondeau:
The tests fails because `System.IO.DirectoryNotFoundException : Could not find a part of the path 'C:\Users\runneradmin\AppData\Local\Temp\2a69ed27-1218-4ccb-b5d7-45e8321e1e0a\2a69ed27-1218-4ccb-b5d7-45e8321e1e0a\assembly\Runtime\Scripts\modules\format-name.js'.` - it seems the way I find the scripts folder is incorrect, I'll get back to this too later. I'm not sure why the Linux tests fail though.

christianrondeau:
I'm done with my first review; I added comments and questions on things that I'm not sure about. It's a pretty big PR for someone with limited knowledge of ASTs, language specifications etc but I feel like the vast majority of the code required was already written. I'm looking forward to hearing your thoughts on this!

lahma (commented):
Looking good! As you can see most of my comments were stylistic nit-picking. The import/export API via c# engine still doesn't seem intuitive *to me*, but I'm just one person. You really went the extra mile with this.

I don't have good guidance whether this is spec-wise "perfect", when I read the relevant parts it seemed that there were some places where "magic just happens", so I'm happy we have an interpretation of behavior that passes tests!

Jint.Tests/Runtime/ModuleTests.cs
```diff
@@ -0,0 +1,195 @@
...
+        public ModuleTests()
+        {
+            _engine = new Engine();
+        }
+
+        [Fact]
+        public void ShouldExportNamed()
+        {
+            _engine.DefineModule(@"export const value = 'exported value';", "my-module");
+            var ns = _engine.ImportModule("my-module");
```
sebastienros, line 23: Does `ImportModule` actually import a module in the environment, or does it just return it to the caller, and doing an `Execute('return value')` would not work.

Jint.Tests/Runtime/ModuleTests.cs
```diff
@@ -0,0 +1,195 @@
...
+
+            Assert.Equal("exported value", ns.Get("value").AsString());
+        }
+
+        [Fact]
+        public void ShouldExportNamedListRenamed()
+        {
+            _engine.DefineModule(@"const value1 = 1; const value2 = 2; export { value1 as renamed1, value2 as renamed2 }", "my-module");
```
sebastienros, line 31: Modules don't seem to be imported when "defined", looks good, could be used to setup an engine with external optional dependencies. Is the module parsed at this point of on import? I would prefer `Set/Add/Create` module.

Jint.Tests/Runtime/ModuleTests.cs
```diff
@@ -0,0 +1,195 @@
...
+            Assert.Equal("John Doe", ns["user"].AsObject()["name"].AsString());
+        }
+
+        [Fact]
+        public void CanImportFromFile()
+        {
+            var engine = new Engine(options => options.EnableModules(GetBasePath()));
+            var ns = engine.ImportModule("./modules/format-name.js");
+
+            Assert.Equal("John Doe", ns["formatName"].AsFunctionInstance().Call(JsString.Create("John"), JsString.Create("Doe")).AsString());
```
sebastienros, line 182: isn't there an `Invoke()` method?
Indexer? Not great for consistency with the way the engine uses `GetValue()`

Jint/Engine.Modules.cs
```diff
@@ -7,27 +16,119 @@ public partial class Engine
...
+            module = new JsModule(this, _host.CreateRealm(), loadedModule, moduleResolution.Uri?.LocalPath, false);
+
+            _modules[moduleResolution.Key] = module;
+
+            return module;
+        }
+
+        public JsModule DefineModule(string source, string specifier)
+        {
+            return DefineModule(new JavaScriptParser(source).ParseModule(), specifier);
```
sebastienros, line 50: Maybe we need a kind of `Module` that does the parsing lazily.

Jint/Native/Function/ScriptFunctionInstance.cs
```diff
@@ -138,7 +138,7 @@ ObjectInstance IConstructor.Construct(JsValue[] arguments, JsValue newTarget)
             {
                 try
                 {
-                    var result = OrdinaryCallEvaluateBody(_engine._activeEvaluationContext, arguments, calleeContext);
+                    var result = OrdinaryCallEvaluateBody(_engine._activeEvaluationContext ?? new EvaluationContext(_engine), arguments, calleeContext);
```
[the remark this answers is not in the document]
sebastienros: I agree, this needs to be found. Please share the stacktrace when this happens.

Jint/Native/Object/ObjectInstance.cs
```diff
@@ -82,6 +82,8 @@ internal ObjectClass Class
             get => _class;
         }
 
+        public JsValue this[JsValue property] => Get(property);
```
sebastienros, line 85: @lahma Should this be generalized? Like on Engine (get and set) ?

Jint.Tests/Runtime/ModuleTests.cs
```diff
@@ -0,0 +1,195 @@
...
+#if(NET6_0_OR_GREATER)
+
+        [Fact]
+        public void CanLoadModuleImportsFromFiles()
+        {
+            var engine = new Engine(options => options.EnableModules(GetBasePath()));
+            engine.DefineModule("import { User } from './modules/user.js'; export const user = new User('John', 'Doe');", "my-module");
+            var ns = engine.ImportModule("my-module");
+
+            Assert.Equal("John Doe", ns["user"].AsObject()["name"].AsString());
```
sebastienros, line 173: Isn't it sufficient to let users access the module properties from the engine since the module was loaded? `engine.GetValue("user")` just to reduce the amounts of ways to do something, which could lead to inconsistencies, or discrepancies in behavior. So `ImportModule` would return nothing. (or Engine)

sebastienros (commented):
Some of my comments my collide with @lahma 's around naming.

Jint/Engine.Modules.cs
```diff
@@ -1,4 +1,6 @@
-﻿using System;
+﻿#nullable enable
```
sebastienros, line 1: I need a drink.

sebastienros: Obviously this is the kind of comment that adds nothing to the conversation ;) I know

christianrondeau:
I will now add the `ModuleBuilder` idea that came up a few times in the reviews, I think it will slightly reduce the impact on Engine (remove DefineModule variants) and refactor the tests accordingly. The idea is to do something like this:

```c#
var module = new ModuleBuilder();
module.ExportClass<T>("name");
module.ExportValue("name", JsValue);
engine.LoadModule(module, "module-name");
```

If you disagree it's now or never :D Technically that should also allow me to do things like:

```
module.CreateImmutableBinding("varName", JsValue);
module.Add("const test = varName + 1");
module.ExportBinding("test", "exportName");
```

It won't be part of the PR but the idea behind the ModuleBuilder is that it'll allow extending the process of building a module in the future.

christianrondeau:
All right I think I have something I'm happy with that solves some of the comments. Here's how you can define modules now:

Method 1, using source code. Nothing new here, except the rename:

```c#
engine.AddModule("module-name", "export const x = 1;");
```

Method 2, using the builder. This allows you to build a module with code, objects, types, etc.

```c#
engine.AddModule("module-name", builder =>
  .AddSource("export const value1 = 'hello';")
  .ExportType<MyType>()
  .ExportValue("value2", "world")
  .ExportObject("value3", new MyType())
  .ExportFunction("fn", args => Console.WriteLine(args.Length))
);
```

This is, I think, fairly discoverable (there's only AddModule there, and only two overloads for it), it allows building modules from multiple pieces of data, it can be extended later, and doesn't need to interact with engine twice.

Also, bonus, it really lazy loads modules now, so even custom-built modules won't be linked until they are loaded, making it much better if you wanted to offer a large set of modules to optionally import.

I feel like actually supporting node_modules wouldn't be a stretch, but I'm already way out of my initial need so shush me :D

christianrondeau:
Two quick questions.

1. Do you want me to rebase or should we wait for all reviews/feedbacks to be completed?
2. I was thinking of @sebastienros feedback on wanting to use "Execute" to run modules. While I still stand by my opinion that "modules should be separated from scripts", I thought of a few improvements to AddModule that I'd like to share. They won't change any of the existing code so it shouldn't invalidate any of the remaining reviews.

Right now to "execute" a module you need to "import" it in the C# context:

```c#
engine.AddModule("module-name", "export const x = 1;");
var ns = engine.ImportModule("module-name");
Console.WriteLine(ns.Get("x").AsString());
```

This is the equivalent of:

```
echo "export const x = 1;" > module-name.js
echo "import { x } from 'module-name.js'; console.log(x);" > run.js
node run.js
```

This is why it's confusing, we're actually creating AND running modules from the same context. However to create and run modules will probably be a pretty common pattern. So to simplify it, I had two new "sugar" APIs in mind.

1) Fluent API to "Add and Import"

This keeps the same proposed semantics.

```c#
var ns = engine.AddModule("module-name", "export const x = 1;").ImportModule();
Console.WriteLine(ns.Get("x").AsString());
```

2) Sugar API to "AddAndImport":

This adds a new word, "Execute" which here means "Add And Import". It also hides the module name, making it more in line with what you expect from Execute. _However_ it does not add any of the lexical context into the "normal" Execute" method. The module name would be automatically generated (e.g. a GUID or something like "*dynamic*")

```c#
var ns = engine.ExecuteModule("export const x = 1;");
Console.WriteLine(ns.Get("x").AsString());
```

Let me know what you think :)

christianrondeau:
I may have greatly reduced availability soon; while I won't push for a merge (I'm completely aware of what maintaining an open-source project means) I'd like to know if you believe this will eventually get merged so that I can start building on top of it. If you're not sure yet, fair enough :) If it's just a matter of taking the time to fully review but you feel that the approach is sound, I'll temporarily work on this branch for my downstream projects.

Thanks :)

Jint/Runtime/Modules/DefaultModuleLoader.cs
```diff
@@ -52,20 +140,20 @@ protected virtual ModuleLoaderResult LoadModule(Engine engine, Uri location)
...
+            ExceptionHelper.ThrowSyntaxError(engine.Realm, $"Error while loading module: error in module '{resolved.Uri.LocalPath}': {ex.Error}");
             module = null;
         }
 
-        return new ModuleLoaderResult(module, location);
+        return module;
     }
 
-    protected virtual string LoadModuleSourceCode(Uri location)
+    protected virtual string ReadAllText(Uri uri)
```
sebastienros, line 150: This is not reusable anymore in a sub-class. The fact that the checks are done outside will prevent custom implementation from loading urls.

sebastienros:
I added only one comment, but would like to see @lahma 's answered too.
When this is done I will merge it.

lahma:
@christianrondeau can we help out here somehow? would be nice to get this finalized and merged 🚀

lahma:
I checked the PR and there are some comments, can you mark resolved the ones are non-actionable etc?

Jint.Tests/Runtime/Modules/DefaultModuleResolverTests.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
+using Jint.Runtime.Modules;
+using Xunit;
+
+namespace Jint.Tests.Runtime.Modules
```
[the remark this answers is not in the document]
sebastienros: How is that resolved?

Jint/Runtime/Modules/DefaultModuleLoader.cs
sebastienros: We'll need this one fixed before merging. I will do it otherwise.

christianrondeau:
@lahma I'm not allowed to mark comments as resolved :) I did go through all comments last time, most of them are still unanswered. I added a comment on as many as I could to mark the ones I think are not actionable, and deleted my own unanswered questions to reduce the scope. I don't think I can do anything more at this point (but I'm willing to!).

christianrondeau:
(in case it wasn't clear, the list I posted shows everything that I can see but I have no todos on my end as far as I can tell, I'm on hold)

lahma:
@christianrondeau awesome work! thank you for pushing this through and being patient with us!

christianrondeau:
Woot! Thanks @lahma and @sebastienros ! Since I hope to contribute again, do you have any feedback on this PR? I know it was too large (I didn't find a way to reliably test the import without export and vice versa) but otherwise, is there something I could have done to make it easier for you guys? I feel like throwing tons of questions directly in the file changes might not have been the best approach :)

lahma:
I think the process was mostly long due to the feature being both large and had no specific spec to rely on at times (implementation specific). Hard to predict all the use cases and how people would like to interact with it "outside of JS".

Feedback loop is always a bit of a problem when people try to find some free time (which is taken away from something else) and are also on different time zones 🙂

lahma:
I'll try to do a quick round with more seamless test suite integration and allowing people to run scripts with imports without jumping many hoops. I'll create a separate PR for feedback.

christianrondeau:
@lahma I didn't mean that it wasn't fast enough, no worries!!! I just wanted to make sure my next PR is painless for you guys :) Thanks for helping me through!

PR #1055, opened 2022-01-18
Added context null check for OrdinaryCallEvaluateBody

CreepGin:
Without the null check and fallback, ScriptFunctionInstance.Construct would throw for usages like the following:

```C#
engine.Evaluate(@"
    (function () {
        function Foo(){}
        let f = new Foo()
    })
").As<FunctionInstance>().Call();
```


sebastienros:
Can you add a unit test please?
Is it only happening since this recent change ? https://github.com/sebastienros/jint/pull/1047

CreepGin:
I don't think it was affected by #1047 because I observed the bug before merging in past 20 days' commits. 

So the reason I didn't add a test was because in most places `_activeEvaluationContext` always have a fallback: `_engine._activeEvaluationContext ?? new EvaluationContext(_engine)`. I just assumed it was an oversight that `ScriptFunctionInstance.Construct` didn't have it. 

Also, I don't know whether it's better to just handle null contexts in `OrdinaryCallEvaluateBody()`. If you want me to add a unit test for the specific use case this patch fixes for me (shown in OP), I can certainly do that.

CreepGin (commented):
Sorry, typo in the new commit msg. Should be **unit test** instead of **unity test**

lahma:
Thanks for reporting this. This one is a bit problematic as I thin we should not support calling "bare function instances" without any context (this also bypasses any execution constraints, though not a big deal here - I guess).

We just took the approach with constructors, you need to go through the engine to avoid problems like this. 

Current workaround is to use just eval without the IIFE:

```csharp
var engine = new Engine();

var result = engine.Evaluate(@"
    class Foo { test() { return 123 } }
    let f = new Foo()
    return f.test()
");

Assert.True(result.IsInteger());
Assert.Equal(123, result.AsInteger());
```

Is is the IIFE case real use case for you that you need to invoke outside of engine as anonymous instance that hasn't proper handle via `engine.Invoke`?

 I'd rather hide that whole possibility to issue `FunctionInstance.Call` directly via function's API, should we just expose `engine.Call()` method like we have the construct, e.g.:

```csharp
var engine = new Engine();

var function = engine.Evaluate(@"
    (function () {
        class Foo { test() { return 123 } }
        let f = new Foo()
        return f.test()
    })
");

// we don't currently have this
// Call(JsValue value, params JsValue[] arguments) + Call(string name, params JsValue[] arguments)
var result = engine.Call(function, Arguments.Empty);

Assert.True(result.IsInteger());
Assert.Equal(123, result.AsInteger());
```

This would bring parity with `Construct` API and shift usage again towards using `Engine` as entry-point.

CreepGin:
Thanks for the details. The use case that I used was actually from [Jint.CommonJS](https://github.com/tylerjwatson/Jint.CommonJS). It wraps every module in a `function(module, exports, __dirname, require)` closure. I can see it as a convenient use case for other similar interop scenarios.

I'm too inexperienced with Jint to comment on the execution constraints and other nuances associated with it, but I should mention that right now the IIFE works fine if there's no constructor involved inside the closure (e.g. no new'ing). So from an user's perspective, they could be using the IIFE and `FunctionInstance.Call` without any problem for a long time. Then one day they add in some code in the closure that contains a `new` operator. The resulting crash will lead to some head scratching as it did to me. 

Design and usability wise, I think both `As<FunctionInstance>().Call()` and `engine.Call()` feel natural to use.

lahma:
The idea would be to hide `FunctionInstance.Call` as protected method like we did for constructors and explicitly implement `IConstructor` interface so the `Call` method would not be visible to outside, would require to use `engine.Call()` to invoke This would reduce risk of calling the function directly without proper setup.

We could optionally still have extension method to allow such call that would dig function's engine instance and route via it's `Call` if needed.

@sebastienros thoughts?

CreepGin:
Calling functions with default parameters will also throw. But I'm not sure how to properly fix this one.

```C#
// Throws NullReferenceException
engine.Evaluate(@"
    (function () {
        function foo(a = false) { }
        foo()
    })
").As<FunctionInstance>().Call();
```

```txt
NullReferenceException: Object reference not set to an instance of an object
Jint.Runtime.Interpreter.Expressions.JintConstantExpression.GetValue (Jint.Runtime.Interpreter.EvaluationContext context) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Runtime.Environments.FunctionEnvironmentRecord.HandleAssignmentPatternOrExpression (Jint.Runtime.Interpreter.EvaluationContext context, Esprima.Ast.Node left, Esprima.Ast.Node right, Jint.Native.JsValue argument, System.Boolean initiallyEmpty) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Runtime.Environments.FunctionEnvironmentRecord.SetFunctionParameterUnlikely (Jint.Runtime.Interpreter.EvaluationContext context, Esprima.Ast.Node parameter, Jint.Native.JsValue[] arguments, System.Int32 index, System.Boolean initiallyEmpty) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Runtime.Environments.FunctionEnvironmentRecord.SetFunctionParameter (Jint.Runtime.Interpreter.EvaluationContext context, Esprima.Ast.Node parameter, Jint.Native.JsValue[] arguments, System.Int32 index, System.Boolean initiallyEmpty) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Runtime.Environments.FunctionEnvironmentRecord.AddFunctionParameters (Jint.Runtime.Interpreter.EvaluationContext context, Esprima.Ast.IFunction functionDeclaration, Jint.Native.JsValue[] arguments) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Engine.FunctionDeclarationInstantiation (Jint.Native.Function.FunctionInstance functionInstance, Jint.Native.JsValue[] argumentsList) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Native.Function.FunctionInstance.OrdinaryCallEvaluateBody (Jint.Runtime.Interpreter.EvaluationContext context, Jint.Native.JsValue[] arguments, Jint.Runtime.Environments.ExecutionContext calleeContext) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Native.Function.ScriptFunctionInstance.Call (Jint.Native.JsValue thisArgument, Jint.Native.JsValue[] arguments) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Engine.Call (Jint.Native.Function.FunctionInstance functionInstance, Jint.Native.JsValue thisObject, Jint.Native.JsValue[] arguments, Jint.Runtime.Interpreter.Expressions.JintExpression expression) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Engine.Call (Jint.Native.ICallable callable, Jint.Native.JsValue thisObject, Jint.Native.JsValue[] arguments, Jint.Runtime.Interpreter.Expressions.JintExpression expression) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Runtime.Interpreter.Expressions.JintCallExpression.EvaluateCall (Jint.Runtime.Interpreter.EvaluationContext context, Jint.Native.JsValue func, System.Object reference, Esprima.Ast.NodeList`1[Esprima.Ast.Expression]& arguments, System.Boolean tailPosition) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Runtime.Interpreter.Expressions.JintCallExpression.Call (Jint.Runtime.Interpreter.EvaluationContext context) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Runtime.Interpreter.Expressions.JintCallExpression.EvaluateInternal (Jint.Runtime.Interpreter.EvaluationContext context) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Runtime.Interpreter.Expressions.JintExpression.Evaluate (Jint.Runtime.Interpreter.EvaluationContext context) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Runtime.Interpreter.Statements.JintExpressionStatement.ExecuteInternal (Jint.Runtime.Interpreter.EvaluationContext context) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Runtime.Interpreter.Statements.JintStatement.Execute (Jint.Runtime.Interpreter.EvaluationContext context) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Runtime.Interpreter.JintStatementList.Execute (Jint.Runtime.Interpreter.EvaluationContext context) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Runtime.Interpreter.JintFunctionDefinition.Execute (Jint.Runtime.Interpreter.EvaluationContext context) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Native.Function.FunctionInstance.OrdinaryCallEvaluateBody (Jint.Runtime.Interpreter.EvaluationContext context, Jint.Native.JsValue[] arguments, Jint.Runtime.Environments.ExecutionContext calleeContext) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.Native.Function.ScriptFunctionInstance.Call (Jint.Native.JsValue thisArgument, Jint.Native.JsValue[] arguments) (at <47e46eb871964f00ac051ccd1a997686>:0)
Jint.JsValueExtensions.Call (Jint.Native.JsValue value, Jint.Native.JsValue[] arguments) (at <47e46eb871964f00ac051ccd1a997686>:0)
```

lahma:
@CreepGin I've created https://github.com/sebastienros/jint/pull/1062 to show a version that both hides `Call` from public API and still allows old code to compile - also your both test cases pass. Would this fit the bill?

CreepGin:
Great! I'll close this pull and comment in #1062

PR #1062, opened 2022-01-22
Hide Call method from public API

lahma:
Hiding `Call` like we hid `Construct`, it's now `protected internal` and `ICallable` is explicit implementation, but can still be accessed via extension method helper that gets the Engine instance and then invokes the call via it. 

`Engine.Call` allows to get via name, use existing `JsValue` and pass parameters, overload to also supply thisObj that requires using proper array as arguments parameter to disambiguate and follow JS style.

### DIFF
--- a/Jint.Tests/Runtime/Domain/UuidConstructor.cs
+++ b/Jint.Tests/Runtime/Domain/UuidConstructor.cs
@@ -55,7 +55,7 @@ namespace Jint.Tests.Runtime.Domain
             return obj;
         }
 
-        public override JsValue Call(JsValue thisObject, JsValue[] arguments) => Construct(arguments, null);
+        protected internal override JsValue Call(JsValue thisObject, JsValue[] arguments) => Construct(arguments, null);
 
         public void Configure()
         {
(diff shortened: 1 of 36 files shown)


CreepGin:
Just tested this on my side. Everything seems to work well. But the inability to set thisObj breaks my old code that depends on it. Here's one of my use cases:

I'm interop'ing C# with Preact (light weight React). So all the dom APIs that Preact depends on, I implemented them in C#. 

```js
// Preact code
function eventProxy(e) {
	this._listeners[e.type + false](options.event ? options.event(e) : e);
}
dom.addEventListener(name, eventProxy, useCapture);
```

```C#
// Example dom api that I have to implement. I need to pass the dom obj back as 'this' for eventProxy(e)
public class dom {
    public Dictionary<string, object> _listeners;

    ...

    public void addEventListener(string name, JsValue jsval, bool useCapture = false) {
        var func = jsval.As<FunctionInstance>();
        var engineThis = JsValue.FromObject(UIEngine.Instance.Engine, this);
        var callback = (EventCallback<EventBase>)((e) => {
            func.Call(engineThis, JsValue.FromObject(UIEngine.Instance.Engine, e));  // <-- passing dom as 'this'
        });
        switch (name) {
            case "MouseUp":
                _ve.RegisterCallback<MouseUpEvent>(callback);
                break;
            ...
        }
        _registeredCallbacks.Add(name, callback);
    }
}
```

lahma:
@CreepGin thanks for the use case, I've updated the PR to include overloads to allow passing in the wanted `thisObject`.

CreepGin:
Awesome! Looks and works well on my side.

PR #1063, opened 2022-01-23
Various Interop fixes

KurtGokhan:
I am not sure about the Dictionary accessor. It now uses `TryGetValue` instead of using `Item` and expecting it to throw error when a key is missing. That makes more sense to me logically. But there is this case:

https://github.com/sebastienros/jint/compare/main...KurtGokhan:interop-fixes?expand=1#diff-6f0fbc1958b9b5e571672e8b8c8795a77ac611e89ee8bf7fbe31d2d1eb476942R145

I am not sure if the result should be `undefined` or `0` here.


lahma:
Thank you for this! Couldn't spot anything bad logic-wise and think the returning undefined is natural for missing member like JS does, we can revisit when someone finds a problem with the logic 😉

PR #1074, opened 2022-02-03
Improve function interop

KurtGokhan:
The last fixes I did were not ideal. Same function would resolve to different reference, and bound functions would not work. Also `Call` utility is a more proper way to solve this now I think.

### DIFF
--- a/Jint/Native/Function/FunctionInstance.cs
+++ b/Jint/Native/Function/FunctionInstance.cs
@@ -307,7 +307,7 @@ namespace Jint.Native.Function
             }
             else
             {
-                if (thisArgument == null || thisArgument.IsNullOrUndefined())
+                if (thisArgument is null || thisArgument.IsNullOrUndefined())
                 {
                     var globalEnv = calleeRealm.GlobalEnv;
                     thisValue = globalEnv.GlobalThisValue;
--- a/Jint/Native/Function/FunctionInstance.cs
+++ b/Jint/Native/Function/FunctionInstance.cs
@@ -307,7 +307,7 @@ namespace Jint.Native.Function
             }
             else
             {
-                if (thisArgument.IsNullOrUndefined())
+                if (thisArgument == null || thisArgument.IsNullOrUndefined())
                 {
                     var globalEnv = calleeRealm.GlobalEnv;
                     thisValue = globalEnv.GlobalThisValue;
--- a/Jint/Native/Function/ScriptFunctionInstance.cs
+++ b/Jint/Native/Function/ScriptFunctionInstance.cs
@@ -138,7 +138,8 @@ namespace Jint.Native.Function
             {
                 try
                 {
-                    var result = OrdinaryCallEvaluateBody(_engine._activeEvaluationContext, arguments, calleeContext);
+                    var context = _engine._activeEvaluationContext ?? new EvaluationContext(_engine);
+                    var result = OrdinaryCallEvaluateBody(context, arguments, calleeContext);
 
                     // The DebugHandler needs the current execution context before the return for stepping through the return point
                     if (_engine._isDebugMode && result.Type != CompletionType.Throw)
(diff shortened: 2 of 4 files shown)


Jint/Native/Function/FunctionInstance.cs
```diff
@@ -307,7 +307,7 @@ internal void OrdinaryCallBindThis(ExecutionContext calleeContext, JsValue thisA
             }
             else
             {
-                if (thisArgument.IsNullOrUndefined())
+                if (thisArgument == null || thisArgument.IsNullOrUndefined())
```
KurtGokhan, line 310: Added this for convenience of calling a function from C# side with `null` thisArgument. Maybe this check can be moved inside `IsNullOrUndefined`, but it is used in a lot of places so I did not touch it.

KurtGokhan:
I think I may have introduced the old problem again. Will test this more before it is ready for merging.

Jint/Native/Function/FunctionInstance.cs
lahma: Nit: should use `is null` and `is not null` syntax against JsValues as equality goes to operator overloading unnecessarily.

Jint/Native/Object/ObjectInstance.cs
```diff
@@ -959,10 +959,9 @@ private object ToObject(ObjectTraverseStack stack)
                     break;
 
                 case ObjectClass.Function:
-                    if (this is FunctionInstance function)
+                    if (this is ICallable function)
```
lahma, line 962: There's a PR trying to hide ICallable and possibly trying to remove it at some point altogether if possible. In spec there are function objects (in Jint FunctionInstance) that may be callable (`IsCallable`)

KurtGokhan: Are you suggesting I should check `IsCallable`? As I see it does `this is ICallable` for everything except `Proxy`. 

Also this function is for converting script objects to their CLR counterpart. I don't think anything other than a function should be converted to a CLR function. So, proxies will not be hitting this line anyway.

As for removing `ICallable`, I can't imagine how it is going to work. If we agree on test cases, let's keep this for now and change when `ICallable` removed?

lahma: Yeah this was just a note. Proxy is a good example that function object might not be callable, even though it has to implement `ICallable` if the target would happen to be callable. My thinking is that we should be able to get away by having `FunctionInstance` (write your own implementation), `ScriptFunctionInstance` (anything from script) and `ClrFunctionInstance` (basically native function). All these have `Call` method from `FunctionInstance`. IConstructor and ICallable are a bit artificial constructs, is all.

The more test coverage the better. Always comes back biting for advanced usage that people like you have when doing interop. All coverage is much appreciated!

KurtGokhan: Ah, ok. It was `FunctionInstance` before but I noticed bound functions did not derive from `FunctionInstance` and `ICallable` seemed like the only thing that covers all of these cases.

lahma: Ah the BoundFunction... OK that was the pain point we discussed earlier. Well it's a start to make `ICallable` internal at least..

Jint/Native/Function/ScriptFunctionInstance.cs
```diff
@@ -138,7 +138,8 @@ ObjectInstance IConstructor.Construct(JsValue[] arguments, JsValue newTarget)
             {
                 try
                 {
-                    var result = OrdinaryCallEvaluateBody(_engine._activeEvaluationContext, arguments, calleeContext);
+                    var context = _engine._activeEvaluationContext ?? new EvaluationContext(_engine);
+                    var result = OrdinaryCallEvaluateBody(context, arguments, calleeContext);
```
KurtGokhan, line 142: I know this change is required (because `_activeEvaluationContext` is used this way everywhere else). My app was crashing because of this, but I could not write a test which reproduces the issue in isolation.

KurtGokhan:
Ready for review

lahma (approved):
Looks good, especially new test cases. We need to stay vigilant with that execution context that it doesn't get broken later on.

PR #1078, opened 2022-02-10
External serialisation of set fails

Ghostbird:
This adds a test that shows how the external serialisation of a set fails.


Ghostbird:
Issue closed since it's probably not a bug.

PR #1082, opened 2022-02-11
WIP: Scheduler

SebastianStehle:
WIP for https://github.com/sebastienros/jint/discussions/1081

### DIFF
--- a/Jint/Scheduling/Scheduler.cs
+++ b/Jint/Scheduling/Scheduler.cs
@@ -1,0 +1,101 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Jint.Scheduling
+{
+    internal sealed class Scheduler : IDisposable
+    {
+        private readonly TaskCompletionSource<object> _taskCompletionSource = new TaskCompletionSource<object>();
+        private readonly HashSet<DeferredTask> _pendingTasks = new HashSet<DeferredTask>();
+        private readonly Queue<Action> _inlinedTasks = new Queue<Action>();
+        private bool _isRunning;
+        private bool _isDisposed;
+
+        public Task Completion
+        {
+            get => _taskCompletionSource.Task;
+        }
+
+        public void Dispose()
+        {
+            _isDisposed = true;
+        }
+
+        public IDeferredTask CreateTask()
+        {
+            var task = new DeferredTask(this);
+
+            _pendingTasks.Add(task);
+
+            return task;
+        }
+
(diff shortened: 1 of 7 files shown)


SebastianStehle:
@twop This is a WIP of POC for the discussion I opened in the discussions.

Please ignore the changes to the *.csproj files. I will revert them if necessary, I was just too lazy to install .NET Framework.

Jint/Scheduling/Scheduler.cs
```diff
@@ -0,0 +1,101 @@
...
+{
+    internal sealed class Scheduler : IDisposable
+    {
+        private readonly TaskCompletionSource<object> _taskCompletionSource = new TaskCompletionSource<object>();
+        private readonly HashSet<DeferredTask> _pendingTasks = new HashSet<DeferredTask>();
+        private readonly Queue<Action> _inlinedTasks = new Queue<Action>();
+        private bool _isRunning;
+        private bool _isDisposed;
+
+        public Task Completion
```
twop, line 15: I have a strong opinion that Jint should not have `Task` as a part of it's API surface area, otherwise there is an implicit promise (no pun intended) of taking care of thread safety, which is not something Jint is interested in (I might be wrong here).

twop: Maybe as a small piece of "evidence", no JS engines I know provide comprehensive async support. You basically have to do all the steps I described in the issue in a host environment

SebastianStehle: The Scheduler is internal.

SebastianStehle: What do you mean with async support? 

For me, in this context "async" is everything that is not invoked immedately. SetTimeout, setInterval, Promises, callbacks and so on. Even module loading is actually a promise in javascript.

twop: I mean the actual async runtime is usually outside of JS engines, e.g. they have a low level primitive that a hosting environment needs to implement. Specifically: `HostEnqueuePromiseJob` from the language spec https://www.ecma-international.org/wp-content/uploads/ECMA-262_12th_edition_june_2021.pdf

lahma: Here's a bit easier link: https://tc39.es/ecma262/#sec-hostenqueuepromisejob , I haven't really gone into this but basically custom hosts should be possible in Jint (we should make it so and improve when features are missing), there's already the Host concept and ideally Jint would have all expected behavior and allow adding missing more controversial bits.

PR #1083, opened 2022-02-13
Date.parse should default to local time

### LINKED ISSUE #1077: Date.parse treats date-time string without time zone specified as UTC time instead of local one
"When the time zone offset is absent, date-only forms are interpreted as a UTC time and date-time forms are interpreted as local time."
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse

Thus, this is to work like this (in the time zone GMT+2):
console.log(new Date(Date.parse("2022-02-06T23:07:13.7796076Z")).getUTCHours());
console.log(new Date(Date.parse("2022-02-06T23:07:13.7796076")).getUTCHours());
> 23
> 21

However, the last code line seems to incorrectly produce 23 through Jint, treating date-time string without time zone specified (here, without 'Z' at the end) as UTC time instead of local one.

shlomo-ildar:
Moved the following formats from DefaultFormats to SecondaryFormats:
            "yyyy-MM-ddTHH:mm:ss.FFF",
            "yyyy-MM-ddTHH:mm:ss",
            "yyyy-MM-ddTHH:mm",

Test updates:
ShouldParseAsUtc:
        [InlineData("1970-01-01T00:00Z")]
        [InlineData("1970-01-01T00:00:00Z")]
        [InlineData("1970-01-01T00:00:00.000Z")]

ShouldParseAsLocalTime:
        [InlineData("1970-01-01T00:00")]
        [InlineData("1970-01-01T00:00:00")]
        [InlineData("1970-01-01T00:00:00.000")]

fixes #1077

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -1109,9 +1109,9 @@ namespace Jint.Tests.Runtime
         [InlineData("1970")]
         [InlineData("1970-01")]
         [InlineData("1970-01-01")]
-        [InlineData("1970-01-01T00:00")]
-        [InlineData("1970-01-01T00:00:00")]
-        [InlineData("1970-01-01T00:00:00.000")]
+        [InlineData("1970-01-01T00:00Z")]
+        [InlineData("1970-01-01T00:00:00Z")]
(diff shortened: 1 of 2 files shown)


lahma (approved):

lahma:
Thank you!

PR #1091, opened 2022-02-16
Fix integer addition/subtraction overflow

### LINKED ISSUE #1088: Integer addition/subtraction leading to an overflow is not converted to Number
Hello,

The following code doesn't behave as expected:

```
var result = new Jint.Engine().Evaluate("2147483647 + 1");
// Actual: 'result' is of type Integer with value -2147483648
// Expected: 'result' would be of type Number with value 2147483648
```
Subtraction has a similar behaviour.

Thanks for your time.

conicgames:
Closes #1088


lahma:
Awesome, thank you!

Issue #1094, opened 2022-02-18
Exception when trying to pass object as method parameter


KubaZ2:
Cannot pass [DiscordId](https://github.com/KubaZ2/NetCord/blob/alpha/NetCord/DiscordId.cs) as method parameter. It throws `System.ArgumentException: 'Object of type 'System.String' cannot be converted to type 'System.UInt64'.'`

KubaZ2:
I'm new to jint and I don't really know how it works. I don't think I can fix it 😕

lahma:
@KubaZ2 there was a bug in resolution algorithm that caused Jint to choose wrong overload in this case. You also need to set `AllowOperatorOverloading` to true in order for implicit/explicit conversions to kick in. 

https://github.com/sebastienros/jint/blob/0231800a0261f2f09a3d1e953c62ad219b73d8a5/Jint.Tests/Runtime/InteropTests.cs#L2770-L2780

There should be a build on MyGet feed soon (on NuGet later), see README on repo root how to access it.

lahma:
And with #1099 merged you shouldn't even need to allow operator overloading, it's allowed by default with latest soon to be on MyGet version.

PR #1099, opened 2022-03-06
Always allow operator overloaded conversion

lahma:
After fixing #1094 I felt that we should actually always allow implicit/explicit operators for conversion. I've adapted code to prefer `ChangeType` and if it doesn't succeed we can still always try operator overloading, which has semantically the same goal as supporting `IConvertible`. So removing checks whether operator overloading is allowed when doing interop, it's already behind allow interop flag anyway.

Thoughts @sebastienros @KurtGokhan ?


KurtGokhan:
Looks good

PR #1103, opened 2022-03-08
Add support for generic methods under interop

source-transformer:
add support for generic methods

### DIFF
--- a/Jint.Tests/Runtime/GenericMethodTests.cs
+++ b/Jint.Tests/Runtime/GenericMethodTests.cs
@@ -12,18 +12,9 @@ namespace Jint.Tests.Runtime
 {
     public class GenericMethodTests : IDisposable
     {
-        public GenericMethodTests()
-        {
-        }
-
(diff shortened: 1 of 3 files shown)


lahma (changes requested):
Mostly stylistic issue, tabs seems to running wild again. Also could you add description to PR what is to be expected?

lahma:
If you rebase against main you should get `.editorconfig` that should guide your editor to use spaces over tabs, maybe that helps.

source-transformer:
no problem for switching tabs to spaces - each time I do it (use tabs instead of spaces) - it is accidental. The convention just differs from the other projects I work in - so my editor defaults to tabs - I've grabbed the editor config - will try updating those individual spots you mentioned - see if the editor does the right thing.  In case it is of interest I'm using Visual Studio Preview/2022 on a mac.

lahma (approved):
I did small cleanup, there were still some tabs left and made new TypeConverter method internal.

PR #1106, opened 2022-03-12
Generic Methods: better generic method support.  

source-transformer:
covariance + contravariance - where generic arguments are derived from base generic argument type

### DIFF
--- a/Jint/Runtime/TypeConverter.cs
+++ b/Jint/Runtime/TypeConverter.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Numerics;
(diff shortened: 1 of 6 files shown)


source-transformer:
no rush - but just bumping this PR @lahma + @sebastienros in case you didn't see it.

added new tests to show how the existing support for generic extensions is broken.  So if you just grab my tests and put them into mainline - you can see how they fail.

I have another PR incoming to notify programmers with a "thread safety" exception when they try to use JINT across multiple threads.  It's a small update - but I think the wording of the exception and probably the type of exception will require your input.

Jint/Runtime/TypeConverter.cs
```diff
@@ -1124,6 +1124,19 @@ private static int CalculateMethodScore(Engine engine, MethodDescriptor method,
             return score;
         }
 
+        internal class AssignableResult
```
lahma, line 1127: readonly struct?

PR #1107, opened 2022-03-12
Date related fixes and improvements

lahma:
Already reviewed in test branch, bringing as only containing date-related fixes.


lahma:
Comment from @Jither 

> By the way, a Date.toString() quick-fix, if anyone looks at its code in the future:
return ToDateTime().ToString("ddd MMM dd yyyy HH:mm:ss 'GMT'K", CultureInfo.InvariantCulture);
... should probably be:
return ToDateTime().ToString("ddd MMM dd yyyy HH:mm:ss 'GMT'zzz", CultureInfo.InvariantCulture);
K outputs Z for 0 offset from UTC - meaning the string will be GMTZ (incorrect). zzz outputs +00:00 as it should be - GMT+00:00.

PR #1110, opened 2022-03-18
Thread Safety Check

source-transformer:
JINT is not designed with thread-safety.

JINT will now throw exception if callbacks are attempted on different thread than thread that created Engine.

in some scenarios - javascript callback parameters end up as not defined (exp: "parameter is not defined") - so instead of failing with a seemingly unrelated error - fail when different thread is used in callback.

### DIFF
--- a/Jint.Tests/Runtime/AsyncAndDisposeTests.cs
+++ b/Jint.Tests/Runtime/AsyncAndDisposeTests.cs
@@ -1,0 +1,172 @@
+﻿using Jint.Native;
+using System;
+using System.Dynamic;
+using Jint.Runtime.Interop;
+using Xunit;
+
+using Xunit.Abstractions;
+using System.IO;
+using System.Text;
+
+namespace Jint.Tests.Runtime
+{
+    public class AsyncAndDisposeTests : IDisposable
+    {
+        private readonly Engine _engine;
+
+        public AsyncAndDisposeTests(ITestOutputHelper output)
+        {
+            // TPC: remove the following 2 statements - just for enabling Console.WriteLine(..)
+            var converter = new Converter(output);
+            System.Console.SetOut(converter);
+
+
+            _engine = new Engine(cfg => cfg
+                .AllowOperatorOverloading())
+                .SetValue("log", new Action<object>(Console.WriteLine))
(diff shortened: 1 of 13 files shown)


sebastienros:
Async code can jump threads, and it will still be thread-safe.

Honestly, I think it's easy enough to understand what "not thread-safe" means that we don't need to try and check it's incorrectly used. And the current solution would not work for what's not using interop.

source-transformer:
> Async code can jump threads, and it will still be thread-safe.
> 
> Honestly, I think it's easy enough to understand what "not thread-safe" means that we don't need to try and check it's incorrectly used. And the current solution would not work for what's not using interop.

Makes sense - I've put my "thread safety check" behind an option (which defaults to disabled) to see if that works

ChrML:
I don't think this check adds any value in it's current state. It's perfectly valid to invoke a single Jint engine from multiple threads, and it's very common to do so, e.g. a single method with an 'await' in it backed by a thread-pool would. And it's perfectly safe. So I can't see why this check would be opt-ed in by anyone in such environment. My production code does this everywhere.

 What's important is to not invoke Jint multiple times simultaneously. At the moment doing so will cause weird exceptions and races at bizzare locations, and screw up the state of the engine. So I believe we could forget about threads a little, and instead add checks for improper concurrent use. This could protect the Engine, while at the same time not add much additional CPU cost, not restrict correct usage from multiple threads, and not give the user an option that requires usability trade-offs.

Example code:

```c#
public Execute()
{
    if (Interlocked.Exchange(ref this.engineBusy, 1) != 0)
    {
          throw InvalidOperationException("Concurrent usage detected.");
    }

    try
    {
          // Engine processing goes here.
     }
     finally
     {
          Volatile.Write(ref this.engineBusy, 0);
     }
}
```

NOTE: Jint will have to set engineBusy to 0 each time it calls into C# code, because we could call Javascript callbacks from it (re-entry). And back to 1 once the C# code is done with the interlocked guard

This way would allow the same usability as today, but throw an exception the second your code runs into concurrency issues (calling into Jint while Jint is still executing stuff on another thread) that could corrupt the engine.

PR #1111, opened 2022-03-18
fix Jint.Tests.csproj to properly reference files in Runtime

source-transformer:
fix Jint.Tests.csproj to properly reference files in Runtime directory and subdirectories so that build is properly triggered when you update test files (under that directory)

### DIFF
--- a/Jint.Tests/Jint.Tests.csproj
+++ b/Jint.Tests/Jint.Tests.csproj
@@ -9,7 +9,7 @@
     <NoWarn>612</NoWarn>
   </PropertyGroup>
   <ItemGroup>
-    <EmbeddedResource Include="Runtime\Scripts\*.*;Parser\Scripts\*.*" />
+    <EmbeddedResource Include="Runtime\**\*.cs;Parser\Scripts\*.*;" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\Jint\Jint.csproj" />


lahma:
I don't understand the need for this change.  Not everything should be embedded.

source-transformer:
> I don't understand the need for this change. Not everything should be embedded.

I was looking for a way to get my test code to be rebuilt when I click build.  Currently I have to rebuild the entire test csproj every time I update a test file in the test project - because the project doesn't see any updates when I update my test file.  So when I click "build" - it just immediately returns saying "everything is already built" (i.e. "build successful").

With this change the following four tests are failing:

Jint.Tests.Runtime.EngineTests.ShouldExecuteDromaeoBase64
Jint.Tests.Runtime.EngineTests.ShouldExecuteLodash
Jint.Tests.Runtime.EngineTests.ShouldExecuteKnockoutWithoutErrorWhetherTolerantOrIntolerant
Jint.Tests.Runtime.EngineTests.ShouldExecuteHandlebars

Is the correct solution in this case to relocated my test files into this directory:

Runtime\Scripts\

So that visual studio correctly rebuilds the tests when I update my test code?

lahma:
If it's and IDE issue, hard to help you there. For me tests and rebuilds work just fine. I'm using Rider though, but occasionally VS 2022.

source-transformer:
> If it's an IDE issue, hard to help you there. For me tests and rebuilds work just fine. I'm using Rider though, but occasionally VS 2022.

I just retested this with the latest update from visual studio preview/2022 (on mac) and it appears this issue is fixed without updating the csproj.  I'll close this MR.

PR #1135, opened 2022-04-18
Replace JsValue[] arguments with Arguments struct

lahma:
Just pushing this into queue so it won't get lost. Concept is based on how [YantraJS handles arguments](https://github.com/yantrajs/yantra/blob/main/YantraJS.Core/Core/Arguments.cs).

Idea is to replace `JsValue[] arguments` in used method signatures for call receiver and instead use a custom struct that hides what's the backing way of handling things. `Arguments` gives a better abstraction and offers the same indexing and `At()` capabilties. 

For four or less parameters they are just fields and don't require array handling, this way `JsValueArrayPool` pretty much becomes obsolete. For larger parameter sets (which should be unusual, 4 is largest handled by EcmaScript built-in functions).

The code gets simpler, no need for array pooling and performance doesn't change that much - I think it's in margin of error. API is now also cleaner and passing parameters always goes via `Arguments()` constructor and no need to think about array pooling and releasing of items.

Things to think, should we also cover `thisObject` in Arguments and make all calls take just `Call(in Arguments arguments)`. This would also allow passing more date in the future if needed. Many methods require engine reference which is a reason from them not be `static`.

BREAKING CHANGE.


victoriaquasar:
Why not implement an overload instead of replacing? You can deprecate with time

lahma:
This might not be that good idea after all. This will probably make stack grow more and we've been recently battling with making stack small enough to allow deep nesting and recursions.

lahma:
Coming back to this after four years, with measurements rather than intuition. Short version: the idea
was sound, most of its value has since been captured by other means, and the remaining slice does not
justify the break. I'm closing it.

## Both halves of the idea already shipped, in effective form

The engine now has, for built-ins:

* `Function.CallFast(thisObject, arg0, arg1)` — arguments passed in registers, no pooled array. That is
  this PR's payload without the struct.
* `Function.CallFastVariadic(thisObject, ReadOnlySpan<JsValue>)`, backed by an `[InlineArray(2)]` stack
  buffer — the `ReadOnlySpan` variant, with stack backing where the runtime allows it.

Those were gated in July: −8.2% on `string-base64`, −7.2% on `3d-morph`, and **~0% on Babel, TypeScript,
Prettier and Terser**. About half of even that came from eliding the call-stack frame rather than from
argument passing.

What that gate never covered is the **user-function** path — Babel's transform phase runs roughly four
interpreted script-function calls for every built-in call. So I bounded that separately.

## The bounding experiment

A throwaway branch adds a register-args lane for interpreted callees whose FDI takes the fixed-slot fast
path (`state.CanUseFastFDI`, which already excludes rest parameters, `arguments` objects and duplicate
parameter names). Four registers, matching the four inline fields this PR proposes. The call-stack frame,
the recursion check and the environment reuse all stay exactly as they are, so the delta is the argument
array and nothing else.

Deliberately the *optimistic* bound: raw register parameters, no struct, no frame elision. Green on
`Jint.Tests` (4660/0 net10, 4579/0 net472) and Test262 (99484/0/157), so it is the same program.

First, engagement counters — what fraction of calls could even take the lane:

| script | eligible and actually renting an array |
|---|---|
| `controlflow-recursive` | 100.00% |
| `3d-raytrace` | 96.16% |
| `crypto-aes` | 78.24% |
| `babel-standalone` | 56.28% |
| `access-binary-trees` | 50.00% |
| `access-nbody` | 0.01% |

Worth noting the all-literal path in `ExpressionCache` — where `Unsafe.As<JsValue[]>(_expressions)` makes
an argument list cost literally nothing — covers only 7.34% of Babel's calls. I had assumed argument
arrays were mostly free already. They are not.

## Result

Same binary both sides, mode switched by a `static readonly bool`, rep loop outer / mode loop inner,
best-of-5, on an idle machine. `access-nbody` is the control: 49,620 argument-building dispatches, 4 of
them eligible, so it exercises the call path hard while being immune to the change.

| script | off ms | on ms | Δ |
|---|---|---|---|
| `access-nbody` **(control)** | 69.527 | 70.196 | +0.96% |
| `controlflow-recursive` | 61.359 | 56.219 | **−8.38%** |
| `3d-raytrace` | 68.953 | 67.167 | −2.59% |
| `crypto-aes` | 54.428 | 53.498 | −1.71% *(inside noise)* |
| `access-binary-trees` | 42.940 | 42.609 | −0.77% *(inside noise)* |
| `babel-standalone` | 795.511 | 794.114 | **~0%** |

The control moved the wrong way, which means mode 1 running second carried a systematic ~1% handicap —
the real deltas are if anything slightly more negative than printed.

`controlflow-recursive` is the trustworthy row: five reps within 0.9 points of each other
(−8.49/−8.38/−8.47/−7.64/−8.55). Babel needed 20 reps and several estimators before it could be called
at all — its per-rep spread is ±3% against the control's ±1.3%, and its raw mean is tail-dominated and
useless. Min −0.16%, median +1.71%, trimmed mean of the lowest 10 of 15 +0.84%, against a control reading
+0.66% / +0.36% on the same estimators. It does not move.

## The number, and why Babel is flat

From `controlflow-recursive`: 245,489 eligible calls saving 5.14 ms is **~21 ns per call**. Its whole
per-call cost is 61.4 ms ÷ 245,489 = 248 ns, so the argument array is **~8% of an interpreted call** —
and 21/248 is exactly the −8.38% measured.

That makes Babel a prediction rather than an absence of evidence:

> 231,603 eligible calls × 22 ns ≈ 5.1 ms on a 790 ms run = **−0.64% expected**

which is what was observed, inside a ±1.5% floor. Babel is flat *not* because the lane misses — it
engages on 56% of its calls — but because 232k eligible calls in 790 ms is too few for 21 ns each to
matter. Reaching 3% there would take ~1.1M eligible calls, five times what it does. Real applications
spend their time in parsing, strings and allocation, not in call plumbing.

## Why the struct would land below that bound

The measurement is a ceiling. The struct as designed here would additionally pay:

* a `_first is JsValue` type test — against a non-sealed class — on **every** argument read, replacing
  today's single bounds-checked load;
* construction cost at all-literal call sites, which today cost nothing at all;
* the stack growth I raised in [my 2022 comment](#issuecomment-1259068916), which has only got more
  relevant now that `StackGuard` exists — ~40 bytes across several .NET frames per JS frame.

`JsValueArrayPool` also no longer allocates for 1–4 arguments and takes no interlock on its fast path, so
there is no GC win left to collect either.

On the API side: 120 public/protected signatures in `Jint/` mention `JsCallArguments`, and
`JsCallDelegate = Func<JsValue, JsValue[], JsValue>` is the most-used embedder entry point there is. A
lambda written `(thisObj, args) => …` does not bind to a delegate with an `in` parameter — which is why
the diff in this PR spells `(JsValue _, in Arguments args) =>` at every call site. Every host would
rewrite every function registration. `ReadOnlySpan` is a non-starter separately from perf: it cannot
appear in `Func<>` on `net462`/`netstandard2.0`, and arguments legitimately escape into `arguments`
objects, rest parameters, `BindFunction.BoundArguments` and generator/async suspension.

To answer @scarletquasar's question from 2022 — an additive overload would have avoided the break, but it
would also have to be threaded through every internal consumer to be worth anything, and the measurement
above says there is nothing at the end of that to collect.

## If anyone wants the 21 ns

`ScriptFunction` is sealed, so this is capturable entirely **internally**, with
`JsCallArguments = JsValue[]` untouched at the public boundary. The experimental branch is already that
shape and passes both suites. It is worth ~8% on recursion-heavy scripts and nothing on real code, so
it's on the shelf rather than in a PR — happy to dust it off if someone turns up with a workload shaped
like `controlflow-recursive`.

Closing this one. Thanks for the idea, past me.

PR #1140, opened 2022-04-21
Support custom CLR object creator for ObjectInstance.ToObject

### LINKED ISSUE #1138: Performance improvement for interop of ObjectInstance
Currently in `ObjectInstance.ToObject` method objects of `ObjectClass.Object` class are transformed to `ExpandoObject`
https://github.com/sebastienros/jint/blob/de1d2de7665f251584d6156785fecaace3a53023/Jint/Native/Object/ObjectInstance.cs#L985

If we have an `ObjectInstance` with many (10-50k) properties that conversion causes significant slowness due to O(n) for `ExpandObject.Add` method and the fact `ToObject` is called twice (in `TypeConverter.FindBestMatch` and in `MethodInfoFunctionInstance.Call`)

Sample Js code that causes issue:
```javascript
var obj={}; 
for(var i=0;i<20000;i++){obj['i'+i]=i;} 
worker.getObject(obj);
```
`worker` object is set to the following C# class:

```c#
private class CustomWorker {
  public void GetObject(IDictionary<string, object> o) {}
}
```

The slowness of `ExpandoObject` can be observed with:
```c#
IDictionary<string, object> eo = new ExpandoObject();
for (var i = 0; i < 20000; i++) {
  eo.Add($"i{i}", i);
}
```

Is it possible to use `Dictionary<string, object>` instead of `ExpandoObject` for cases when we know for sure that CLR method expects `IDictionary<string, object>` and not just generic `object`?

lahma:
fixes #1138

### DIFF
--- a/Jint.Tests/Runtime/InteropTests.cs
+++ b/Jint.Tests/Runtime/InteropTests.cs
@@ -919,8 +919,8 @@ namespace Jint.Tests.Runtime
             var name = e.Evaluate("o.values.filter(x => x.age == 12)[0].name").ToString();
             Assert.Equal("John", name);
         }
-        
-                
+
+
         [Fact]
         public void CanSetIsConcatSpreadableForArrays()
         {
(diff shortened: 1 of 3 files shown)


sebastienros:
Next question might be to return something that is not even IDictionary, but that would also mean that it would contain the logic to iterate the properties. I think that's fine for now and solves the problem.

sebastienros (approved):

PR #1143, opened 2022-04-22
Improve method scoring logic between double and float

### LINKED ISSUE #1142: An incorrect double value is returned from a static method of the .NET type via interop
Hello!

In Jint version 3.0.0 Beta 2038 an incorrect double value is returned from a static method of the .NET type via interop. Moreover, this error occurs only in version for .NET Framework (everything works correctly in .NET Core). If you run the following console application:

```csharp
using System;

using Jint;
using Jint.Runtime.Interop;

namespace TestJint
{
    class Program
    {
        {
            var engine = new Engine();
            TypeReference mathTypeReference = TypeReference.CreateTypeReference(engine, typeof(Math));
            engine.SetValue("Math2", mathTypeReference);
            double result = engine.Evaluate("Math2.Max(5.37, 5.56)").AsNumber();

            Console.WriteLine("Result: {0}", result);
        }
    }
}
```

Previous version (version 3.0.0 Beta 2037) and current version for .NET Core output a correct result:

```
Result: 5,56
```

Current version for .NET Framework outputs an incorrect result:

```
Result: 5,55999994277954
```

lahma:
fixes #1142

### DIFF
--- a/Jint.Tests/Runtime/InteropTests.cs
+++ b/Jint.Tests/Runtime/InteropTests.cs
@@ -2888,6 +2888,18 @@ namespace Jint.Tests.Runtime
             Assert.Equal("b", dictionary["a"]);
         }
 
+        [Fact]
+        public void ShouldScoreDoubleToDoubleParameterMatchHigherThanDoubleToFloat()
+        {
+            var engine = new Engine();
+            var mathTypeReference = TypeReference.CreateTypeReference(engine, typeof(Math));
+            
+            engine.SetValue("Math2", mathTypeReference);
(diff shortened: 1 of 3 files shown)


sebastienros (approved):

sebastienros:
You should try the magic button if you don't already
image

lahma:
But if I enable auto merge it's harder to get some feedback 😉

sebastienros:
Ah, forgot it doesn't require an approval (aspnet core does it, and it's not the best sometimes)

lahma:
It's a though process to determine which changes are dead-simple and which are ones that require review, and on top of that steal your time to get the review 😉

PR #1146, opened 2022-04-23
Cache created dynamic delegates

### LINKED ISSUE #1139: Removing Event Handler doesn't seem to work
`remove_EventName(handler)` doesn't really remove the handler. Take an existing unit test for example:

```C#
public void ShouldExecuteActionCallbackOnEventChanged()
{
    var collection = new System.Collections.ObjectModel.ObservableCollection<string>();
    Assert.True(collection.Count == 0);

    _engine.SetValue("collection", collection);

    RunTest(@"
        var eventAction = 1;
        collection.add_CollectionChanged(handler);
        collection.remove_CollectionChanged(handler);    // <- does not remove the original handler
        collection.Add('test');

        function handler(sender, eventArgs) { eventAction = 0 }
    ");

    var eventAction = _engine.GetValue("eventAction").AsNumber();
    Assert.Equal(1, eventAction);    // Throws. Handler is still called
}
```

lahma:
Because dynamic delegates are built again and again when function is passed around it's impossible to add and remove *same* function instance from collection that has delegate listener. ~~I'm introducing delegate cache that is based on target delegate type and the Func instance being called with it.~~

Created delegate is now cached inside `FunctionInstance`'s hidden custom field. They should now share the same lifecycle.

@Dogwei  this is your original code, do you see a problem with caching the generation result per function instance?

fixes #1139

### DIFF
--- a/Jint/Runtime/Interop/DefaultTypeConverter.cs
+++ b/Jint/Runtime/Interop/DefaultTypeConverter.cs
@@ -18,6 +18,11 @@ namespace Jint.Runtime.Interop
         private static readonly ConcurrentDictionary<TypeConversionKey, bool> _knownConversions = new();
         private static readonly ConcurrentDictionary<TypeConversionKey, MethodInfo> _knownCastOperators = new();
 
+        // cache of delegates built against function functions, cannot be static because engine specific
+        // this way we get same delegate instance for each func being used
+        private readonly record struct DelegateCacheKey(Func<JsValue, JsValue[], JsValue> Call, Type DelegateType);
+        private readonly ConcurrentDictionary<DelegateCacheKey,Delegate> _delegateCompilations = new();
+
         private static readonly Type intType = typeof(int);
         private static readonly Type iCallableType = typeof(Func<JsValue, JsValue[], JsValue>);
         private static readonly Type jsValueType = typeof(JsValue);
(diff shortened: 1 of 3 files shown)


Jint/Runtime/Interop/DefaultTypeConverter.cs
```diff
@@ -18,6 +18,11 @@ public class DefaultTypeConverter : ITypeConverter
         private static readonly ConcurrentDictionary<TypeConversionKey, bool> _knownConversions = new();
         private static readonly ConcurrentDictionary<TypeConversionKey, MethodInfo> _knownCastOperators = new();
 
+        // cache of delegates built against function functions, cannot be static because engine specific
+        // this way we get same delegate instance for each func being used
+        private readonly record struct DelegateCacheKey(Func<JsValue, JsValue[], JsValue> Call, Type DelegateType);
+        private readonly ConcurrentDictionary<DelegateCacheKey,Delegate> _delegateCompilations = new();
```
KurtGokhan, line 24: I had a quick look at this and I don't know the answer. But wouldn't this prevent GC? Maybe `ConditionalWeakTable` is more suitable.

lahma: That's a very good point. I've changed the code to cache the delegate inside the function instance as hidden property. A rare use case so not introducing actual CLR field for this. Now the target and generated delegate should share the same lifetime.

KurtGokhan: The new changes will not work with multiple target types. I think you should revert to the old code and use `ConditionalWeakTable` (which is basically WeakMap) instead of `ConcurrentDictionary`.

I think creating a JS property to keep a C# data feels hacky. It could be a field in `FunctionInstance`. It must be `Dictionary<Type, Delegate>` then. But `ConditionalWeakTable` feels better for this. According to C# docs: 

> The [ConditionalWeakTable<TKey,TValue>](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2?view=net-6.0) class enables language compilers to attach arbitrary properties to managed objects at run time.

This is a weird and confusing way to say it is a WeakMap. But summarizes the current use case.

lahma: I think adding the type to key should be enough to work around the uniqueness problem. ConditionalWeaktable probably has more overhead as functions can be short-lived (during single script) whereas the weak table would have engine's lifetime and then the liveness-maintaining.

lahma: I made the get and set extension method to make it easier to change later. The `ConditionalWeakTable` is a solid choice but I think I'm leaning towards more instance-lifetime for now.  Fixed the key to contain Type's GUID.

sebastienros:
Did you mess up with merged? Can't see a change anymore

lahma:
I think I can blame GitHub with this one. It's in main already but PR still active. Oh well.

PR #1149, opened 2022-04-24
Update README.md

lahma:
Adding ECMAScript 2022 status and Stage 3 implementations


lahma:
@sebastienros now that I have your attention, can't merge this because no builds are being run as it's only .md change...

sebastienros:
Can you remove the md files skip as part of this PR? It's not the first time we are having the issue.

lahma:
Done, and green!

PR #1160, opened 2022-05-07
Fix JsonParser not able to parse "\u0000"

Genteure:
Currently `JSON.parse` does not handle `\u0000` correctly:

```cs
var engine = new Engine();
engine.SetValue("log", new Action<object>(Console.WriteLine));
engine.Execute(@"js code below");
```

```js
const null_text = '{"a":"\\u0000"}';
const null_obj = JSON.parse(null_text);
log(null_text);
log(null_obj.a.length);
log(null_obj.a);

const a_text = '{"a":"\\u0061"}';
const a_obj = JSON.parse(a_text);
log(a_text);
log(a_obj.a.length);
log(a_obj.a);
```

Expected output:

```
{"a":"\u0000"}
1

{"a":"\u0061"}
1
a
```

Current output:

```
{"a":"\u0000"}
5
u0000
{"a":"\u0061"}
1
a
```

This PR removed unneeded check for escaped characters, and added tests to verify escaped characters are correctly parsed.

### DIFF
--- a/Jint.Tests/Runtime/JsonTests.cs
+++ b/Jint.Tests/Runtime/JsonTests.cs
@@ -16,6 +16,28 @@ namespace Jint.Tests.Runtime
         }
 
         [Theory]
+        [InlineData("{\"a\":\"\\\"\"}", "\"")] // " quotation mark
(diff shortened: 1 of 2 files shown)


lahma (approved):
Looks good, thank you!

PR #1161, opened 2022-05-07
Make JsonParser escape character parsing more strict

Genteure:
This is a continuation of #1160 

This PR removes support for non-standard `\v`, `\x00`, `\0123` and `\(any char here)`, and will throw an error instead.

This behavior is matching other common JSON parser like `JSON.parse` found in nodejs.

```js
// node v16.13.0
JSON.parse('{"\\v":1}')
// Uncaught SyntaxError: Unexpected token v in JSON at position 3
```

```cs
// Newtonsoft.Json 13.0.1
const string json = @"{""\v"":1}";
var obj1 = Newtonsoft.Json.JsonConvert.DeserializeObject(json);
// Newtonsoft.Json.JsonReaderException: Bad JSON escape sequence: \v. Path '', line 1, position 4.
```


lahma (approved):
I like when things get simpler!

PR #1171, opened 2022-05-20
Keep C# stack trace in exceptions thrown in JS, provide JS stack in inner exception instead

christianrondeau:
This is a draft to highlight what this would look like, and some of the challenges.

Not done:

- Format JS exceptions in C# style (with localization using StackFrame I guess) in the mixed stack
- Deal with exceptions thrown using object instances (e.g. calling CLR from JS)
- Test is fragile
- Didn't review what debugging 262 tests look like now

Open questions:

- Do you agree with the property names:
  - `.JavaScriptStackTrace` for the JS-only stack
  - `.OriginalStackTrace`
  - `.StackTrace` for the mixed stack
- Do you agree with keeping `.ToString()` behaving normally and adding `.JavaScriptToString()`? Any idea for a better name?
- I'm wondering if the mixed stack is useful, and/or whether we should format it differently. The problem is that it's _already_ a string once in the Exception class, so without a bigger refactoring I'm wondering if the common formatting is worth it? Opinions?

Here's an example:

```
   at run () mixstacks.js:1:18
   at mixstacks.js:2:1
   at Jint.Engine.ScriptEvaluation(ScriptRecord scriptRecord) in D:\Dev\jint\Jint\Engine.cs:line 308
   at Jint.Engine.<>c__DisplayClass63_0.<Execute>b__0() in D:\Dev\jint\Jint\Engine.cs:line 265
   at Jint.Engine.ExecuteWithConstraints[T](Boolean strict, Func`1 callback) in D:\Dev\jint\Jint\Engine.cs:line 651
   at Jint.Engine.Execute(Script script) in D:\Dev\jint\Jint\Engine.cs:line 265
   at Jint.Engine.Execute(String source, ParserOptions parserOptions) in D:\Dev\jint\Jint\Engine.cs:line 259
   at Jint.Tests.Runtime.ErrorTests.<>c__DisplayClass12_0.<MixStackTracesFromJsError>b__0() in D:\Dev\jint\Jint.Tests\Runtime\ErrorTests.cs:line 295
   at Xunit.Assert.RecordException(Func`1 testCode) in C:\Dev\xunit\xunit\src\xunit.assert\Asserts\Record.cs:line 50
```


lahma:
Thanks @christianrondeau for creating the PR with the details, on top of our usual suspect @sebastienros I'd like to ask feedback from @Jither who has been improving the debugger and I think stacks and debugging are somewhat related...

I personally don't have opinions (as usual) that would push this to better direction, but if stakeholders could chime in with ideas and use cases.

Jither:
> Thanks @christianrondeau for creating the PR with the details, on top of our usual suspect @sebastienros I'd like to ask feedback from @Jither who has been improving the debugger and I think stacks and debugging are somewhat related...

The debugger doesn't care much about stack traces currently - it creates its own debuggable list of stack frames from `_engine.CallStack` - so in my mind, the two aren't directly related. My mind isn't in Jint mode currently, though - but I'd expect the stack trace in general becomes somewhat less useful due to ye olde "Completions vs. Exceptions" issue? 🙂

christianrondeau:
I guess it depends how you use Jint. If you use it to run user code, you want the C# stack trace as debugging information, and the js stack trace as user data. Then there is no need to mix the stack traces. This is why I'm wondering if the mixed stack is worth it. On the other hand if Jint calls C# code, it'd be nice to see the full c#->js->c# stack all at once. I'm thinking saying in the JavascriptException "see the X property for the js stack" could work too. I'll try and see what's both natural and doesn't require huge amounts of code.

lahma:
I'm all in with improving error stack (with security in mind). I've had my share of "Error: null" messages during creating features, and I usually can blame the person between chair and keyboard not taking the time to replicate scenario in Node and check what kind of error this reports.

Like Jither mentioned, in order to implement generators "by-the-book", I'm afraid we need to drop throwing exceptions to control execution paths and start using `Completion { Type: Throw }`. Because... who doesn't want to yield from inside `try` or `catch` or `finally` or... Great to implement a feature that probably 0.0001% of users need. But hey...

christianrondeau:
@lahma if I hear that right, JavaScriptException will eventually _not_ be thrown, therefore any effort there would be wasted, right? That actually sounds right by me. So, for now:

- `JavaScriptException` will have the native C# `.StackTrace`, `.Message` and `.ToString()`
- `JavaScriptException` will have the new properties `.JavaScriptStackTrace`, which will be used in the unit tests (big try catch) and eventually by developers who want to inform users of errors in their code;

And if we get rid of throwing exceptions eventually, I think it would be "nice" to offer in Engine Options the ability to `ThrowExceptionOnJavaScriptErrors`, which would simply be a check for the Completion at the end of Evaluate, Execute, and ImportModule.

Does that sound right? If so I'll just review the code here and move the PR out of Draft.

lahma:
I think exceptions would be thrown at the edge, like when interpreter exits back from engine and sees Throw completion, like engine.Evaluate. Something probably that needs iterations (at least I don't have clear picture). 

All work with exception reportingis most welcome (test cases showing ideal UX), we just might need to reduce expectations along the way if we cannot meet those reporting features with interpretation - nothing done yet.

christianrondeau:
So, I did find one gotcha going down that route. See https://github.com/sebastienros/jint/pull/1093

If we don't throw `JavaScriptException` anymore in the future, that means the idea of subclassing JavaScriptException doesn't really make sense anymore, does it? It's not such a big deal, I can add a method to construct an exception subclass from an error by allowing access to that error through the `JintScriptExecutionException` so another exception can be built, but I'm just wondering if the goal of that other PR was not to allow implementers to do what I'm doing right now anyway? @source-transformer did that make sense?

Said differently, if we have an exception with the C# stack trace that contains the "JavaScript error" with the JS stack trace, does that achieve the same goal you had in your PR? If so I'd delete the copy constructor tests since they wouldn't be needed anymore.

christianrondeau:
Just pushed what the result would be, I think it makes sense to have an edge exception when running a script fail, so eventually, the `JavaScriptException` can be replaced by anything that provides the Error and a Location.

Here's what shows up in Rider when I fail a test because of an error in JavaScript. There are both stacks but no fancy formatting nor disclosed information from JavaScript code.

```
Jint.Runtime.JintScriptExecutionException: My error thrown in js

Jint.Runtime.JintScriptExecutionException
My error thrown in js
   at Jint.Runtime.ExceptionHelper.ThrowJintExecutionException(Engine engine, JsValue value, Completion& result) in D:\Dev\jint\Jint\Runtime\ExceptionHelper.cs:line 139
   at Jint.Engine.EvaluateModule(String specifier, ModuleRecord cyclicModule) in D:\Dev\jint\Jint\Engine.Modules.cs:line 165
   at Jint.Engine.<>c__DisplayClass119_0.<ImportModule>b__0() in D:\Dev\jint\Jint\Engine.Modules.cs:line 122
   at Jint.Engine.ExecuteWithConstraints[T](Boolean strict, Func`1 callback) in D:\Dev\jint\Jint\Engine.cs:line 653
   at Jint.Engine.ImportModule(String specifier, String referencingModuleLocation) in D:\Dev\jint\Jint\Engine.Modules.cs:line 122
   at Jint.Engine.ImportModule(String specifier) in D:\Dev\jint\Jint\Engine.Modules.cs:line 90
   at Jint.Tests.Runtime.ModuleTests.ShouldExportDefault() in D:\Dev\jint\Jint.Tests\Runtime\ModuleTests.cs:line 46

Jint.Runtime.JavaScriptException
My error thrown in js
   at my-module:0:1
```

I think it's ok, formatting the js stack trace using .NET's StackFrame would be a nice addition eventually but I don't think it's a deal-breaker for this PR.

One thing that also shows is "what's the edge". I have identified a few places that affect the tests: Evaluate, Execute and ImportModule as said before, but there are other places like JsValue.Invoke(), Realm.GlobalObject.ToObject() for example (there might be others). So if you move to Completion, that question will be pop up again :)

Unless there's something conceptually wrong, I'll do a review and move out of draft.

christianrondeau:
Mmmh I found another gotcha; it seems the "edge" is pretty unclear, since Engine.Execute and Engine.Evaluate seems to be called by JintStatementList... which then turns it back into a JavaScript error :)

To avoid having this PR actually becoming that "edge-defining" task, I'll cheat a little bit and catch both internal and external exceptions and treat them the same internally (in other words, a JavaScriptException base class from which will inherit the JavaScriptInternalException (that contains the JS stack trace) and the JintScriptExecutionException (that contains the C# stack trace). It's not as big a change as it sounds, but it's still a little bit more than I thought I'd need. So hold for reviewing the code (though the previous concepts still apply)

lahma:
Thanks for sharing the progress. I'm afraid I don't have much to give as help, but I'm glad that your solution is aimed at a need that you can describe and even implement 😉 

Not sure about `JintScriptExecutionException`, Jint prefix is a bit redundant as it's a `JintException`, `Script` refers to `Script` type, but we also support `Module`. Is it just `EvaluationException`? 🤷🏻‍♂️

christianrondeau:
Yeah I don't like the name either. And I also don't like having a stack trace that makes it seem like the JS stack is starting from wherever the evaluation exception is thrown... but it's also very practical and simple, so I'm hesitating (hence too many words).

I'm thinking, if every method returns Completion instead of throwing, I'd see something like `EnsureSuccessStatusCode` but for Completion so instead of throwing "at the edge", we let users decide whether to throw:

```c#
var result = engine
  .Evaluate("throw new Error('test');")
  .EnsureNormalCompletion();
```

This would bite people who wouldn't know when their code failed silently and didn't think of calling this, but it would also solve the problem of the "edge" completely.

I just wanted to get exceptions I could work with but as always I ended up in a rabbit hole haha! I'll probably move forward with the PR anyway since it's (I think) better but I'm pretty sure this'll need to be reworked once the Completion thing is done.

christianrondeau:
All right so I found an "elegant" but "ugly" solution to this mess (can it be both?). Instead of throwing special exceptions at the edge, I just wrap a JavaScriptErrorWrapperException inside the JavaScriptException. So whenever you throw JavaScriptException, the inner exception is automatically created. This creates usable stack traces like this:

```c#
Jint.Runtime.JavaScriptException: Date.parse(zeroDate.toString()) Expected SameValue(«0», «NaN») to be true
 ---> Error: Date.parse(zeroDate.toString()) Expected SameValue(«0», «NaN») to be true
   at assert.sameValue (actual, expected, message) harness/assert.js:42:3
   at built-ins/Date/parse/zero.js:5:1
   --- End of inner exception stack trace ---
   at Jint.Engine.ScriptEvaluation(ScriptRecord scriptRecord) in D:\Dev\jint\Jint\Engine.cs:line 308
   at Jint.Engine.<>c__DisplayClass63_0.<Execute>b__0() in D:\Dev\jint\Jint\Engine.cs:line 265
  etc...
```

This is still not "exactly true" because depending where the exception is thrown, the stack trace may contain parts of Jint, but this is going to become a non-issue when Completion replaces throw. The only "but" is that it breaks @source-transformer's PR, hopefully that commit will solve the problem it tried to fix.

I also checked for formatting the stack trace correctly (localized, etc.) and it wouldn't be very hard, but unless someone has a need for it I don't think it's worth the effort. My objective here was pretty much to be able to know when it fails, what did fail, so we can debug it.

lahma (approved):
Looks good to me. Now is good time as any to break things and gather feedback about the changes. Thanks for yet another improvement!

Issue #1144, opened 2022-04-22
internal interface `IConstructor`


lahma:
### Discussed in https://github.com/sebastienros/jint/discussions/1141

<div type='discussions-op-text'>

<sup>Originally posted by **viceice** April 22, 2022</sup>
Was this intentional to no longer make that interface public? I use that for a custom type, which implements this interface.

https://github.com/sebastienros/jint/blob/fe4ee89b85a8ea2bd7d3406b55e6597df368b58d/Jint/Native/IConstructor.cs#L5

I'm basicall doing something like this:
https://github.com/sebastienros/jint/blob/fe4ee89b85a8ea2bd7d3406b55e6597df368b58d/Jint.Tests/Runtime/Domain/UuidConstructor.cs#L11

changed in [`v3.0.0-beta-2038`](https://github.com/sebastienros/jint/releases/tag/v3.0.0-beta-2038) via
- #1047</div>

viceice:
Can we make those interfaces public again until a better solution is found?

dan3988:
Could `FunctionInstance` not implement `IConstructor`, with a virtual implementation that throws a type error by default? That way `IConstructor` can be made internal and we can still create custom constructors by overriding that method.

lahma:
@dan3988 what's your use case, I'd favor something simpler like [the model I'm proposing in this PR](https://github.com/sebastienros/jint/pull/1183). Building against `IConstructor` usually means that you are implementing the ECMA spec.

dan3988:
@lahma I have some custom classes I've implemented similarly to the ECMA classes inside Jint/Native, where I have a separate C# class for the constructor, prototype and instance.

lahma:
Sorry to sound like a broken record,  but can you give a more specific example. Doing the constructor-prototype-instance route seems quite involved compared to what it would be if we could make the interop do the leg work.

dan3988:
I've extracted some of the code from the project into a [repo](https://github.com/dan3988/JintExample) with some examples of javascript classes:
-  [WmDate](https://github.com/dan3988/JintExample/blob/master/WmDate.cs)
-  [WmTime](https://github.com/dan3988/JintExample/blob/master/WmTime.cs)
-  [Uuid](https://github.com/dan3988/JintExample/blob/master/Uuid.cs)

There are 30-40 classes implemented this way in our project. It was written this way as we wanted the following
- Classes in the script that inherit other custom classes, so that `instanceof` would work in our scripts
[...]

PR #1179, opened 2022-05-24
fix: make `IConstructor` public again

viceice:
Make `IConstructor` public again until a better solution is found

- #1141
- #1144

### DIFF
--- a/Jint.Tests/Runtime/Domain/JsUuid.cs
+++ b/Jint.Tests/Runtime/Domain/JsUuid.cs
@@ -3,6 +3,9 @@ using Jint.Native;
 
 namespace Jint.Tests.Runtime.Domain
 {
+    /// <summary>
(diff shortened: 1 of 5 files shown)


viceice:
@lahma @sebastienros  Are you ok with that solution?

lahma:
I think I pondering with this one was left in the state of "Why not use a TypeReference" for this one... Probably need to check again with some thought at some point...

viceice:
> I think I pondering with this one was left in the state of "Why not use a TypeReference" for this one... Probably need to check again with some thought at some point...

Ok, but how to with the uuid scenario? I'm happy to use another compatible solution if you can guide me.

viceice:
I'll give it a try, need a couple of days to fine some time.

viceice:
I no longer need this

PR #1183, opened 2022-05-25
Use interop capabilities for JsUuid

lahma:
@viceice here's an example how using TypeReference would work. There's slight change in behavior as value returned from Engine is now `JsUuid` instead of `Guid`. I think it's a bit hard to model primitive types the way behavior was earlier as Jint doesn't know about it and might fail to account for it at some point (even though it was considered a string).


viceice:
sorry, not yet. had some days off with family

lahma:
Well that's a great reason 👍🏻, no hurry

Jint.Tests/Runtime/Domain/JsUuid.cs
```diff
@@ -3,25 +3,38 @@
...
+
+    public static readonly JsUuid Empty = new JsUuid(Guid.Empty);
+
+    public static JsUuid Parse(JsValue value)
     {
-        internal readonly Guid _value;
-        public static readonly JsUuid Empty = new JsUuid(Guid.Empty);
+        return value switch
+        {
+            { } js when Guid.TryParse(js.AsString(), out var res) => new JsUuid(res),
```
viceice, line 16:
```suggestion
            IObjectWrapper ow when ow.Target is JsUuid res => res,
            { } js when Guid.TryParse(js.AsString(), out var res) => new JsUuid(res),
```

can already be a JsUuid. `Uuid.parse(Uuid.Empty)`

Jint.Tests/Runtime/UuidTests.cs
```diff
@@ -28,16 +28,16 @@ private object RunTest(string source)
         [Fact]
         public void Empty()
         {
-            Assert.Equal(Guid.Empty, RunTest($"Uuid.parse('{Guid.Empty}')"));
-            Assert.Equal(Guid.Empty, RunTest($"Uuid.Empty"));
+            Assert.Equal(JsUuid.Empty, RunTest($"Uuid.parse('{Guid.Empty}')"));
+            Assert.Equal(JsUuid.Empty, RunTest($"Uuid.Empty"));
```
viceice, line 32:
```suggestion
            Assert.Equal(JsUuid.Empty, RunTest($"Uuid.Empty"));
            Assert.Equal(JsUuid.Empty, RunTest($"Uuid.parse(Uuid.Empty)"));
```

Jint.Tests/Runtime/UuidTests.cs
```diff
@@ -9,11 +10,10 @@ public class UuidTests : IDisposable
         private readonly Engine _engine;
 
         public UuidTests()
-        {
-            _engine = new Engine(o => o.AddObjectConverter(new UuidConverter()))
-                .SetValue("copy", new Func<Guid, Guid>(v => new Guid(v.ToByteArray())))
-                ;
-            UuidConstructor.CreateUuidConstructor(_engine);
+        {;
+            _engine = new Engine(options => options.AllowOperatorOverloading());
```
viceice, line 14: Whats this option for?

Jint.Tests/Runtime/UuidTests.cs
```diff
@@ -9,11 +10,10 @@ public class UuidTests : IDisposable
...
+            _engine.SetValue("copy", new Func<Guid, Guid>(v => new Guid(v.ToByteArray())));
```
viceice, line 15: is it used?

viceice (commented):
Works in general, needed some additional changes for our internal code

Issue #1186, opened 2022-06-02
Jint doesn't throw exception from arithmetic operation overload


AndrewLityagin:
Hello👋
In my work project I found  a strange problem: When I try to execute wrong arithmetic operation ( + ) with my type I don't catch exception and I have wrong result.
I have forked your branch and added such the type (Dimensional).  Also I have added two tests  (ShouldBeDimensionalWithValue100kg, ShouldBeInvalidOperationException). Could you explain why don't  I catch  InvalidOperationException in test ShouldBeInvalidOperationException and Jint returns wrong value ?🙏
Thanks in advance! 

[Link to my fork ](https://github.com/AndrewLityagin/jint)
[Link to Dimensional.cs ](https://github.com/AndrewLityagin/jint/blob/main/Jint.Tests/Runtime/Domain/Dimensional.cs)
[Link to ShouldBeDimensionalWithValue100kg](https://github.com/AndrewLityagin/jint/blob/main/Jint.Tests/Runtime/InteropTests.cs#L2958)
[Link to ShouldBeInvalidOperationException](https://github.com/AndrewLityagin/jint/blob/main/Jint.Tests/Runtime/InteropTests.cs#L2979)

lahma:
Thanks for the detailed report! I see that you are using `ArrayList` which can be confusing as API takes a `params array`, this means that you parameter will be considered as single element of an array, being of type object wrapper (for ArrayList) .

So if you change your code to:

```c#
var evapParamsList = new object[]
{
    new Dimensional("kg", 30),
    new Dimensional("piece", 70)
};

var result = engine.Invoke("Eval", evapParamsList).ToObject();
```

Jint will consider this as a list of arguments that each need to be converted, instead of the earlier ArrayList. After this, do things look better then?

AndrewLityagin:
Thanks for your advice. I implemented it. Now I get NaN value from wrong expression - it is good. But I get NaN value from right expression. Test [ShouldBeDimensionalWithValue100kg ](https://github.com/AndrewLityagin/jint/blob/main/Jint.Tests/Runtime/InteropTests.cs#L2958) did not pass.
I did some changes to tests and added [ShouldBeArgumentException](https://github.com/AndrewLityagin/jint/blob/main/Jint.Tests/Runtime/InteropTests.cs#L2985).
I have two questions:
1. Is it possible to catch exception from overrided arithmetic operation?
[...]

PR #1191, opened 2022-06-07
Added the ability to catch the CLR exception from an overridden arithmetic operator.

AndrewLityagin:
In my work project, I need catch exception from wrong arithmetic operation with my types. See [Issue#1866](https://github.com/sebastienros/jint/issues/1186)


lahma (approved):
Great, thanks for the PR. We'll merge soon if no objections are being raised.

PR #1203, opened 2022-06-25
Implement __defineGetter__ and __defineSetter__

lahma:
Not endorsed by spec, but easy addition and allows legacy to work.

/cc @AudriusButkevicius


EternalQuasar0206:
Good, but there will be an option to enable/disable legacy features?

lahma:
No, it's standard runtime feature. You should be able to overwrite it though. And again, closest prototype wins.

AudriusButkevicius:
Nice! I can rid of my hacks.

PR #1206, opened 2022-06-27
Decode import URI before using it as a file name

fluffware:
This fixes imports from paths that contains spaces (and other characters that are escaped in URIs).


lahma (changes requested):
Can you also add a test case that shows what's being fixed here (and also protects from future regressions)?

lahma (approved):
Great work, thank you!

PR #1212, opened 2022-06-30
Fix missing evaluation context in constructor

KurtGokhan:
I had fixed this in #1074 before but it was overwritten by some other changes. Unfortunately my code fails without this change. I can't reproduce it in isolation though.


lahma:
Thanks!

PR #1213, opened 2022-07-01
Fix interop stack trace unwind (#1202)

### LINKED ISSUE #1202: Weird stack trace behaviour
I'm having an issue where the JS stack trace appears to accumulate with each error, when previous errors are handled by catching a `JavaScriptException` in CLR code. Here's a test case that reproduces the issue (apologies if it's convoluted, I couldn't find a way to make it any simpler). If you remove the calls to `foo` in the script, the test passes. Each call to `foo` causes the stack trace to accumulate more frames from the previous (handled) errors, causing the test to fail. I'd be willing to submit a PR to fix this, if I could figure out where to start... 

    public class MyTests
    {
        private static void Foo(Func<int> callback)
        {
            try
            {
                var i = callback();
            }
            catch(JavaScriptException)
            {
                // suppress JS errors
            }
        }

        [Fact]
        public void StackTraceBehaviour()
        {
            var engine = new Engine()
                .SetValue("foo", new Action<Func<int>>(Foo));

            const string script = @"
    function throwIt(message) {
        throw new Error(message);
    }

    foo(() => throwIt('e1'));
    foo(() => throwIt('e2'));
    foo(() => throwIt('e3'));
    
    try {
        throwIt('e4');
    } catch(x){
        // return x to host environment
        x;
    }
    ";
            var result = engine.Evaluate(script);
            var stack = result.Get("stack").AsString();
            Assert.Equal("   at throwIt (message) <anonymous>:3:5\r\n   at <anonymous>:11:5", stack);
        }
    }

With calls to `foo` removed, the stack trace is as expected:

    at throwIt (message) <anonymous>:3:5
    at <anonymous>:11:5

With calls to `foo` left in, the observed stack trace ends up looking like this:

    at throwIt (message) <anonymous>:3:5
    at delegate <anonymous>:11:5
    at delegate <anonymous>:8:1
    at delegate <anonymous>:7:1
    at <anonymous>:6:1

resnickj:
* remove quick patch from JintTryStatement introduced in #853
* implement "ideal" solution described in #853 (try/finally in Engine.Call and Engine.Construct)
* change behaviour of JavaScriptException.SetJavaScriptCallstack() so that it will not overwrite the existing stack trace unless explicitly told to
* add test

fixes #1202

### DIFF
--- a/Jint.Tests/Runtime/ErrorTests.cs
+++ b/Jint.Tests/Runtime/ErrorTests.cs
@@ -284,6 +284,46 @@ var x = b(7);";
             EqualIgnoringNewLineDifferences(expected, ex.GetJavaScriptErrorString());
         }
 
+        // Verify #1202
+        [Fact]
+        public void StackIsUnwoundWhenExceptionHandledByInteropCode()
+        {
+            var engine = new Engine()
+                .SetValue("handle", new Action<Action>(Handler));
+
+            const string Script = @"
+function throwIt(message) {
+    throw new Error(message);
+}
+
+handle(() => throwIt('e1'));
+handle(() => throwIt('e2'));
+handle(() => throwIt('e3'));
+    
+try {
+    throwIt('e4');
+} catch(x){
+    x.stack; // return stack trace string
+}
+";
+            var stack = engine.Evaluate(Script).AsString();
+            EqualIgnoringNewLineDifferences(@"   at Error <anonymous>:3:21
+   at throwIt (message) <anonymous>:3:15
+   at <anonymous>:11:5", stack);
+
+            static void Handler(Action callback)
+            {
+                try
+                {
+                    callback();
+                }
+                catch (JavaScriptException)
+                {
+                    // handle JS error
+                }
+            }
+        }
+
         [Fact]
         public void StackTraceIsForOriginalException()
         {
--- a/Jint/Runtime/TypeConverter.cs
+++ b/Jint/Runtime/TypeConverter.cs
@@ -1030,7 +1030,8 @@ namespace Jint.Runtime
         {
             referencedName ??= "unknown";
             var message = $"Cannot read property '{referencedName}' of {o}";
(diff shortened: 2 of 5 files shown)


Jint.Tests/Runtime/ErrorTests.cs
```diff
@@ -284,6 +284,46 @@ public void StackTraceCollectedForImmediatelyInvokedFunctionExpression()
...
+try {
+    throwIt('e4');
+} catch(x){
+    x.stack; // return stack trace string
+}
+";
+            var stack = engine.Evaluate(Script).AsString();
+            EqualIgnoringNewLineDifferences(@"   at Error <anonymous>:3:21
+   at throwIt (message) <anonymous>:3:15
+   at <anonymous>:11:5", stack);
```
resnickj, line 312: Note: this stack trace still isn't quite right, as the Error ctor should not be included. However, that is a separate issue (there are other tests that expect it to be there), which I may try to address in a subsequent PR.

Jint/Runtime/TypeConverter.cs
```diff
@@ -1030,7 +1030,8 @@ private static ObjectInstance ToObjectNonObject(Realm realm, JsValue value)
         {
             referencedName ??= "unknown";
             var message = $"Cannot read property '{referencedName}' of {o}";
-            throw new JavaScriptException(engine.Realm.Intrinsics.TypeError, message).SetJavaScriptCallstack(engine, sourceNode.Location);
+            throw new JavaScriptException(engine.Realm.Intrinsics.TypeError, message)
+                .SetJavaScriptCallstack(engine, sourceNode.Location, overwriteExisting: true);
```
resnickj, line 1034: In this case we need to overwrite the existing stack trace (the one that is created within the Error ctor) in order to use the correct location (i.e. the passed `source.Location`).

lahma (approved):
Great work here, thanks for the contribution!

PR #1217, opened 2022-07-02
Exclude Error constructor from stack trace

resnickj:
When an error is explicitly thrown from a script, it does make sense to include the Error constructor itself in the captured stack trace. The trace should begin at the point where the Error object is constructed.

Changes:
* If the current function on the call stack is the Error constructor function, exclude it from the captured stack trace.
* Fix issues with reported location so that JavaScriptException.Location is consistent with the location recorded in the stack trace.
* Add new test, adjust existing tests


lahma (approved):
Nice improvement, the change makes sense 👍🏻

PR #1222, opened 2022-07-11
Fix Stackoverflow error when parseInt is executed

rafael-santiago-take:
Fix Stackoverflow error when parseInt is executed using a big word as the parameter


lahma (changes requested):
I think it would make more sense to improve the `Parse` method to cope with larger strings, basically removing the recursion.

lahma:
I've fixed the issue with #1229 which tackles the problem a bit more thoroughly.

PR #1246, opened 2022-07-26
Rethrow CLR function exceptions as JS errors

christianrondeau:
As discussed with @lahma on [gitter](https://gitter.im/sebastienros/jint?at=62de95cbb408e830dc50fdc5). This allows using the `CatchClrExceptions` handler on ClrFunctionInstance. Works for modules and scripts.

### DIFF
--- a/Jint/Runtime/Intrinsics.cs
+++ b/Jint/Runtime/Intrinsics.cs
@@ -38,6 +38,7 @@ namespace Jint.Runtime
         private static readonly JsString _syntaxErrorFunctionName = new("SyntaxError");
         private static readonly JsString _typeErrorFunctionName = new("TypeError");
         private static readonly JsString _uriErrorFunctionName = new("URIError");
+        private static readonly JsString _nativeErrorFunctionName = new("NativeError");
 
         private readonly Engine _engine;
         private readonly Realm _realm;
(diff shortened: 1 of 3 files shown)


Jint/Runtime/Intrinsics.cs
```diff
@@ -255,5 +257,8 @@ internal Intrinsics(Engine engine, Realm realm)
 
         public ThrowTypeError ThrowTypeError =>
             _throwTypeError ??= new ThrowTypeError(_engine, _engine.Realm) { _prototype = _engine.Realm.Intrinsics.Function.PrototypeObject };
+        
+        public ErrorConstructor NativeError =>
```
lahma, line 261: I'm not sure if this concept of `NativeError` exists as JS primitive. `TypeError` and friends are `NativeError`s by nature, it's a grouping name for error and their behavior AFAIK.

christianrondeau: I think you're right, re-reading seems like it's not just a placeholder for "anything-non-js" like I understood it on my first read: https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-nativeerror-object-structure. It's not super clear though, even more confusing when reading TypeError definition: `TypeError is used to indicate an unsuccessful operation when none of the other NativeError objects are an appropriate indication of the failure cause.`

I'll just use Error and move on then ;)

lahma: Yes, the spec should be always enjoyed with a grain of salt. I generally re-learn things by reading it again after some time. There are some edge-cases where some imagination is required.. 😃

lahma (approved):
Looks good!

PR #1250, opened 2022-08-08
Removed ProxyInstance ToString override so calls the property 'toString'

### LINKED ISSUE #1249: Proxy with template literals does not use toString property
```javascript
const proxy = new Proxy({}, {})
proxy.toString = () => 'Test'
const test = `${proxy}`

test === 'function () { [native code] }'
```

I would expect test === 'Test'. This works with an online tester I tried. 

I see that this was changed with https://github.com/sebastienros/jint/pull/736, should the ToString override look for toString properties first, then fall back to 'function () { [native code] }'?

omegatmm:
This reverts a change in ProxyInstance.ToString from #736. All the proxy string conversions match V8 10.4.132.20. I added some tests, not sure if you want to test to that level.

fixes #1249

### DIFF
--- a/Jint.Tests/Runtime/ProxyTests.cs
+++ b/Jint.Tests/Runtime/ProxyTests.cs
@@ -1,0 +1,40 @@
+﻿namespace Jint.Tests.Runtime
+{
+    public class ProxyTests
+    {
+        [Fact]
+        public void ProxyCanBeRevokedWithoutContext()
+        {
+            new Engine()
+                .Execute(@"
+                    var revocable = Proxy.revocable({}, {});
+                    var revoke = revocable.revoke;
+                    revoke.call(null);
+                ");
+        }
+
+        [Fact]
+        public void ProxyToStringUseTarget()
+        {
+            var engine = new Engine().Execute(@"
+                const targetWithToString = {toString: () => 'target'}
+            ");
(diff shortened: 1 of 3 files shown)


lahma:
Seems that the test suite Jint.Tests.Test262 no longer passes: `The active test run was aborted. Reason: Test host process crashed : Stack overflow.`

Jint.Tests/Runtime/ProxyTests.cs
```diff
@@ -0,0 +1,46 @@
+﻿using Jint.Native;
+using Jint.Native.Array;
+using Jint.Native.Function;
+using Jint.Runtime;
+using Jint.Runtime.Interop;
+
+namespace Jint.Tests.Runtime
```
lahma, line 7: nit: the usings are redundant and can use file-scoped namespace if you wish

omegatmm:
Not sure why my visual studio did not indicate that failure...

I hope I'm not over simplifying this, but the test it fails on explicitly states when the Proxy target "IsCallable" it should return the function string.

proxy-arrow-function.js
  If Type(func) is Object and IsCallable(func) is true, then return an
  implementation-dependent String source code representation of func.
  The representation must have the syntax of a NativeFunction.

lahma (approved):
Looks valid, thanks for your contribution!

PR #1259, opened 2022-08-15
Restore public access to JavaScriptException ctor

resnickj:
* This may not be the optimal solution for allowing interop code to raise JS errors, but for the time being it is the most straightfoward and simple thing to do.


lahma:
Handled via second PR containing test coverage,  thanks for pursuing this.

resnickj:
Sorry I didn't have a chance to get to this. Thank you @Genteure for doing it.

PR #1260, opened 2022-08-21
Upgrade to Esprima 3.0.0-beta-5

lahma:


lahma:
There's an unrelated problem in GH actions after some OS upgrade where full framework is now stricter in type check or something similar.

PR #1264, opened 2022-08-24
Fix Regex Iterator

### LINKED ISSUE #1263: RegExp String Iterator always returns false
Spreading an iterator after using `matchAll()` does not work properly:

```js
let string = "one two three";
let matches = [...string.matchAll(/t/g)];
// matches.length should be 2, but Jint returns 0
```

After doing some digging, the source of the problem seems to be that `RegExpStringIterator` may be returning the wrong thing:
https://github.com/sebastienros/jint/blob/8b89a992415aa2c598e01b5b3ad666dc48696493/Jint/Native/Iterator/IteratorInstance.cs#L273
Should this line be `return true;`?

vicr123:
Fixes #1263

### DIFF
--- a/Jint.Tests/Runtime/RegExpTests.cs
+++ b/Jint.Tests/Runtime/RegExpTests.cs
@@ -59,6 +59,14 @@ namespace Jint.Tests.Runtime
             Assert.Equal("/a/b", result);
         }
 
+        [Fact]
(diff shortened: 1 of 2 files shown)


lahma:
Thank you!

vicr123:
Now just hoping it doesn't break anything anywhere else 😉

PR #1276, opened 2022-08-30
Upgrade benchmark project and add YantraJS

lahma:
Somewhat noise, but basically adding YantraJS as comparison target and some cleanup.


lahma:
[Benchmark results](https://gist.github.com/lahma/d74d69521be0b47e9896b31506129629)

PR #1284, opened 2022-09-04
Fix array performance regression

lahma:
* `ObjectChangeFlags.NonDefaultDataDescriptorUsage `should only be raised for array index
* inline some unnecessary extra methods
* reuse property descriptor if possible on `Set`

### DIFF
--- a/Jint.Benchmark/SingleScriptBenchmark.cs
+++ b/Jint.Benchmark/SingleScriptBenchmark.cs
@@ -23,7 +23,7 @@ public abstract class SingleScriptBenchmark
     public void Execute()
     {
         var engine = new Engine(static options => options.Strict());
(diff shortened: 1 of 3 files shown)


lahma:
## Jint.Benchmark.ArrayStressBenchmark

| **Diff**|Method|Mean|Error|Allocated|
|------- |-------|-------:|-------|-------:|
| Old |Execute|13.40 ms|0.085 ms|8.34 MB|
| **New** |	| **11.18 ms (-17%)** | **0.128 ms** | **8.14 MB (-2%)** |
| Old |Execute_ParsedScript|13.47 ms|0.037 ms|8.32 MB|
| **New** |	| **10.87 ms (-19%)** | **0.056 ms** | **8.12 MB (-2%)** |

PR #1286, opened 2022-09-06
Call Dispose on enumerator (#1285)

jm-merle:
See discussion #1285 : call dispose on enumerator (or execute finally clause on enumerator methods/properties)
- Added a call to (_enumerator as IDisposable)?.Dispose() in EnumerableIterator.Close
- Added 3 test cases : normal termination, on break, on exception


lahma (approved):
Thanks for the contribution!

PR #1288, opened 2022-09-07
Make BindFunctionInstance public

### LINKED ISSUE #1287: Make BindFunctionInstance public
In #1044, the BindFunctionInstance class was moved to internal.  This interferes with some valid use cases for the JS .bind() function.  See discussion thread under the referenced issue.

paynesworld:
Make the BindFunctionInstance class public.  Includes a new unit-test project that contains a valid use case for the bind() function.  Note that a new test project was required since the Jint internals are visible to the existing Jint.Tests project.

fixes #1287

### DIFF
--- a/Jint.Tests.PublicInterface/Jint.Tests.PublicInterface.csproj
+++ b/Jint.Tests.PublicInterface/Jint.Tests.PublicInterface.csproj
@@ -1,0 +1,28 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.1.0" />
+    <PackageReference Include="xunit" Version="2.4.1" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <PackageReference Include="coverlet.collector" Version="3.1.2">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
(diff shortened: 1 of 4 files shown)


Jint.Tests.PublicInterface/Jint.Tests.PublicInterface.csproj
```diff
@@ -0,0 +1,28 @@
...
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.1.0" />
```
lahma, line 12: please use the same dependencies as in Jint.Tests.csproj:

```xml
  <ItemGroup>
    <PackageReference Include="GitHubActionsTestLogger" Version="2.0.1" PrivateAssets="all" />
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
    <PackageReference Include="xunit" Version="2.4.1" />
    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.5" />
  </ItemGroup>
```

Jint.Tests.PublicInterface/PublicInterfaceTests.cs
```diff
@@ -0,0 +1,100 @@
+using System.Collections.Concurrent;
+using Jint.Native;
+using Jint.Native.Function;
+using Xunit;
+
+namespace Jint.Tests.PublicInterface
```
lahma, line 6: can use file-scoped namespace

lahma (changes requested):
Excellent initiative, just what is needed to ensure we don't go breaking things. The csproj needs some tuning though I think.

Jint.Tests.PublicInterface/Jint.Tests.PublicInterface.csproj
paynesworld: Resolved, thanks.  I cloned the Jint.Tests.csproj file instead of using the default content that VS generated.

Jint.Tests.PublicInterface/PublicInterfaceTests.cs
paynesworld: Removed the now-redundant using statement for Xunit.

lahma (approved):
Excellent, thank you!

PR #1295, opened 2022-09-11
Restore public access to JavaScriptException ctor

Genteure:
This is the same as pull request #1259 with requested test added.


lahma (approved):
Great, thank you!

PR #1312, opened 2022-10-13
Implement Promise.any

KurtGokhan:
This implements `Promise.any`
Spec: https://tc39.es/ecma262/#sec-promise.any

I am planning to implement `Promise.allSettled` too and I will try to reduce duplicated codes when implementing that.

### DIFF
--- a/Jint/Native/Promise/PromiseConstructor.cs
+++ b/Jint/Native/Promise/PromiseConstructor.cs
@@ -13,7 +13,8 @@ namespace Jint.Native.Promise
         JsValue PromiseInstance,
         ICallable Resolve,
         ICallable Reject,
-        JsValue RejectObj
+        JsValue RejectObj,
+        JsValue ResolveObj
     );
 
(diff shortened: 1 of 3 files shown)


Jint/Native/Promise/PromiseConstructor.cs
```diff
@@ -295,16 +298,143 @@ void ResolveIfFinished()
...
             {
                 iterator.Close(CompletionType.Throw);
-                reject.Call(Undefined, new[] {e.Error});
+                reject.Call(Undefined, new[] { e.Error });
                 return resultingPromise;
             }
 
-            // if there were not items but the iteration was successful
-            // e.g. "[]" empty array as an example
-            // resolve the promise sync
```
KurtGokhan, line 304: Removed this because `ResolveIfFinished` already does that

lahma (approved):
Great work, eagerly waiting for the `Promise.allSettled` now 😉

PR #1316, opened 2022-10-14
Implement Promise.allSettled

KurtGokhan:
Implements Promise.allSettled
Spec: https://tc39.es/ecma262/#sec-promise.allsettled

Also reduces code duplication in Promise static methods like `all`, `any`, `allSettled` and `race` since the first 6 steps of those are same in the spec.


lahma (approved):
Awesome, that didn't take long 👍🏻 I made minor tweak to ensure properties dictionary is initialized with correct size after new member was introduced.

Issue #206, opened 2015-07-03
interop of objects not working two-ways


doronguttman:
I'm trying to save an object generating in a javascript code (obj below) into a value of a CLS object (nameValue.Value below) and later get it back from nameValue and get use a property of obj.
It looks like when you export the javascript object to the CLR it is converted to an ExpandoObject (which I have verified has all the properties set correctly), but when importing the ExpandoObject back to javascript, it is not usable in the same way as the original object (can't access it's properties).
Below is a way to verify the behavior.
Any suggestions?

Use this C# class:

``` c#
namespace jintTester.TestClasses
{
    public class NameValue
    {
        private string _name;
        private object _value;

        public string Name
        {
            get { return _name; }
            set { _name = value; }
        }

        public object Value
        {
            get { return _value; }
            set { _value = value; }
        }

        public override string ToString()
        {
            return string.Format("{{ Name: \"{0}\", Value: {1} }}", Name, Value);
        }
    }
}

```

Use this script:

``` javascript
(function () {
    log("** 2.interop.js: interop test");
    log("--> creating object");
    var obj = {
        s: "string",
        i: 10,
        d: 10.01,
        b: true,
        a: ["string1", "string2"],
        o: {
            s: "object"
        },
        f: function(a) { return a; }
    };
    assert(typeof obj !== "undefined" && obj !== null);             //True
    log("obj is: " + obj);                                          //obj is: [object Object]

    log("--> creating NameValue instance");
    var TestClasses = importNamespace('jintTester.TestClasses');
    var nameValue = new TestClasses.NameValue();
    assert(typeof nameValue !== "undefined" && nameValue !== null); //True
    log("nameValue is: " + nameValue.ToString());                   //nameValue is: { Name: "", Value:  }

    log("--> saving obj into nameValue");
    nameValue.Name = "Object Test";
    nameValue.Value = obj;
    log("nameValue is: " + nameValue.ToString());

    log("--> getting obj back out of nameValue");
    var rsltObj = nameValue.Value;
    //log("rslt is: " + rsltObj);                                   //TypeError
    log("rslt is: " + rsltObj.ToString());                          //rslt is: System.Dynamic.ExpandoObject
    assert(rsltObj.s === obj.s);                                    //False
    assert(rsltObj.i === obj.i);                                    //False
    assert(rsltObj.d === obj.d);                                    //False
    assert(rsltObj.b === obj.b);                                    //False
    assert(rsltObj.a === obj.a);                                    //False
[...]

PR #1318, opened 2022-10-16
Fix some Object static methods returning wrong value

KurtGokhan:
[As seen in specs](https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-object.isextensible), these methods should return a boolean. Somehow this is not covered by test262.

Also added test for #206 so it can be closed.

### DIFF
--- a/Jint/Native/Object/ObjectConstructor.cs
+++ b/Jint/Native/Object/ObjectConstructor.cs
@@ -496,7 +496,7 @@ namespace Jint.Native.Object
         {
             if (arguments.At(0) is not ObjectInstance o)
             {
-                return arguments.At(0);
+                return true;
             }
 
             return TestIntegrityLevel(o, IntegrityLevel.Sealed);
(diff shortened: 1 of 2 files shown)


Jint/Native/Object/ObjectConstructor.cs
```diff
@@ -496,7 +496,7 @@ private static JsValue IsSealed(JsValue thisObject, JsValue[] arguments)
         {
             if (arguments.At(0) is not ObjectInstance o)
             {
-                return arguments.At(0);
+                return true;
```
lahma, line 499: nit: could return `JsBoolen.True` and `JsBoolean.False`

lahma (approved):
Checked the spec and indeed, boolean is the correct end result for these. Thanks!

lahma:
I think that was your 3rd `hactoberfest-accepted` label in this repo! 🥳

PR #1319, opened 2022-10-16
Fix toPrimitive evaluating multiple times during "in" expression

KurtGokhan:
Found another bug. This should fix it.

> @lahma: That was your 3rd hactoberfest-accepted label in this repo

Let's make it 4 😆


lahma (approved):
LGTM

PR #1323, opened 2022-10-20
Implement async/await

lahma:
Contains a feature that some people have asked for, some kind of `async` and `await`. Fixed small issues in earlier Promise implementations too. 

**Before**
`Passed: 47738, Skipped: 13720, Total: 61458`

**After**
`Passed: 49859, Skipped: 12480, Total: 62339`


sebastienros:
That was easy ;)

lahma:
I actually thought that we would need to have generators for this like being state machine kind of feature (support for suspending state). But so far it seems that all snapshotting etc works on function level correctly already, until proven otherwise.

PR #1326, opened 2022-10-23
Make object wrapper cache hold ObjectInstance type

libaowei:


Jint/Engine.cs
```diff
@@ -53,7 +53,7 @@ public sealed partial class Engine : IDisposable
         internal readonly Dictionary<string, Type?> TypeCache = new();
 
         // cache for already wrapped CLR objects to keep object identity
-        internal readonly ConditionalWeakTable<object, ObjectWrapper> _objectWrapperCache = new();
+        internal readonly ConditionalWeakTable<object, ObjectInstance> _objectWrapperCache = new();
```
lahma, line 56: Like I've said before, the type should probably be `IObjectWrapper` as this is cache for *object wrappers*.

lahma (changes requested):

Jint/Engine.cs
lahma: Ah, but the current contract is for object instance.. Fine..

lahma:
Thanks!

PR #1327, opened 2022-10-23
Cache wrapped object returned by WrapObjectHandler

Genteure:
As discussed with @libaowei on Gitter related to #1321 and #1326, turns out they has some script that rely on some non-standard behavior of a older version of Jint. Because they want to maintain compatibility for those scripts they decided to use a custom object wrapper.

This pull request make the object wrapper cache available to `WrapObjectHandler` through argument, which was their original intention if I understand correctly.

### DIFF
--- a/Jint/Options.cs
+++ b/Jint/Options.cs
@@ -254,18 +254,7 @@ namespace Jint
         /// </summary>
         public WrapObjectDelegate WrapObjectHandler { get; set; } = static (engine, target) =>
         {
-            // check global cache, have we already wrapped the value?
-            if (engine._objectWrapperCache.TryGetValue(target, out var wrapped))
-            {
-                return wrapped;
-            }
-
-            wrapped = new ObjectWrapper(engine, target);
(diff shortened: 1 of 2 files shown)


lahma:
Because there's already configuration property `TrackObjectWrapperIdentity` which should be respected. Would it make more sense to have another layer for retrieving the wrapper. Basically one like (pseudo):

```
GetWrapper {
   if cached => return cached;
   cached = TheDelegate(original params)
   return cached
}
```

There should be no need to expose the cache itself. So could use 90% of the original wrap handler and call the customization and store into cache.

Genteure:
Yes it does make sense. Changed as suggested.

Genteure:
Not sure why it's failing on linux, I can't reproduce on my Windows machine, I'll take a look tomorrow.

lahma (approved):
Looks good, the Linux build seems to have trouble at times. Shouldn't be related to this.

lahma:
Thank you!

Issue #1345, opened 2022-11-03
CancellationConstraint constructor cannot be called anymore


EnCey:
In PR #1290 the `CancellationConstrait` ctor was made `internal` and thus can't be called anymore.

The [README](https://github.com/sebastienros/jint#execution-constraints) advertizes to create the constraint and `Reset()` it each time `Execute()` is called:

```
var constraint = new CancellationConstraint();  // !!! internal now !!!!

var engine = new Engine(options =>
{
    options.Constraint(constraint);
});

for (var i = 0; i < 10; i++) 
{
    using (var tcs = new CancellationTokenSource(TimeSpan.FromSeconds(10)))
    {
        constraint.Reset(tcs.Token);

        engine.SetValue("a", 1);
        engine.Execute("a++");
    }
}
```

Since the `TimeConstraint` is just a wrapper around a `CancellationTokenSource` I was planning to just use my own CTS to support both  timeouts and manual cancellation.

Was this change accidental (the `CancellationConstraint` itself is still `public`) or is this approach no longer recommended?

lahma:
Do you need the public constructor? The API can be used via options:

```c#
using var tcs = new CancellationTokenSource();
using var waitHandle = new ManualResetEvent(false);
var engine = new Engine(cfg => cfg.CancellationToken(tcs.Token));
```

This is part of my trial of internalizing everything possible and then opening only when need (there's `Jint.Tests.PublicInterface` that tries to catch usage patterns to protected expected/needed public API).

lahma:
You can reset constraints manually if needed:

https://github.com/sebastienros/jint/blob/943ac858bb92f5c61e656b90104188f3c9257df6/Jint/Engine.cs#L226-L232

But engine already does this for you between calls:

https://github.com/sebastienros/jint/blob/943ac858bb92f5c61e656b90104188f3c9257df6/Jint/Engine.cs#L659-L685

EnCey:
Alright, then I'll go with the `FindConstraint` approach, thank you for providing these insights!

Should I open a PR to add the `Reset(CT)` use case to `Jint.Tests.PublicInterface` and maybe update the README to use `FindConstraint` instead of `new CancellationConstraint()` ?

lahma:
Both would be great as a PR!

PR #1350, opened 2022-11-08
PublicInterface test for CancellationConstraint.Reset

EnCey:
Adds a PublicInterface test that fetches a `CancellationConstraint` with the `FindConstraint<>` method and calls its overloaded `Reset()` method in order to pass in a new `CancellationToken`.

---
### Update README

The example usage of CancellationConstraint demonstrated in the README doesn't work anymore because the constructor was made internal.

A dummy CancellationToken must be provided when creating the constraint if one intends to pass the actual one via Reset(). `default` doesn't work though, as the constraint won't be created then.

see #1345


lahma (approved):
Looks good, thank you!

PR #1373, opened 2022-12-08
Add Loaded event to DefaultModuleLoader

Jither:
As discussed on Gitter.

Simple addition of event to the default module loader, which is triggered when the loader loads a module. This is helpful for e.g. debuggers and other dev tools, which may need to load the same sources as the engine.

Since the module's AST is readily available, and is also currently needed by debuggers (e.g. for determining possible breakpoint positions), the event passes the AST, along with the source ID generated by the loader.


lahma (approved):
Nice addition, thank you!

PR #1374, opened 2022-12-08
Module-related debugger improvements

Jither:
Fixed undefined binding values in module scope
Implemented DebugScopeType.Module - that's one more alignment with Chromium dev-tools.

Not too happy about creating BindingName instances to call `TryGetBinding` - but of course, in terms of performance, this only happens when inspecting - and the change gets rid of the (I think) now redundant `TryGetBindingValue`: `DeclarativeEnvironmentRecord.TryGetBinding` has identical semantics; `ObjectEnvironmentRecord.TryGetBinding` is more up-to-date development-wise (handles `with` and other edge cases); I *think* the same goes for `GlobalEnvironmentRecord`, but that one is harder for me to wrap my head around. Of course, this also circumvents the need for writing a `TryGetBindingValue` for `ModuleEnvironmentRecord`. 😁

The pull request is draft for now, just to have time for spotting edge cases where `TryGetBinding` fails.

ETA: I've done a lot of manual test cases and, as expected, don't really see anywhere it breaks. I don't think it's appropriate to add lots of test cases for `DebugHandler.GetBindingValue` - it's merely a wrapper around `TryGetBinding` which is already thoroughly tested (much more than the old `TryGetBindingValue`). And we don't actually have a spec of how the results should behave - in general, they should simply give the same access to binding values as the script has.

As a "bonus", I threw in the small change to avoid the implicit class constructor triggering a return point step.

### DIFF
--- a/Jint.Tests/Runtime/Debugger/ScopeTests.cs
+++ b/Jint.Tests/Runtime/Debugger/ScopeTests.cs
@@ -333,6 +333,38 @@ namespace Jint.Tests.Runtime.Debugger
         }
 
         [Fact]
+        public void HasCorrectScopeChainForModule()
+        {
+            string imported = @"
+            function add(a, b)
+            {
+                debugger;
+                return a + b;
+            }
+            
+            export { add };";
+
+            string main = @"
+            import { add } from 'imported-module';
+            const x = 1;
+            const y = 2;
+            add(x, y);";
+            TestHelpers.TestAtBreak(engine =>
+            {
+                engine.AddModule("imported-module", imported);
+                engine.AddModule("main", main);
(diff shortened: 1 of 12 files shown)


Jither:
Also improved StepFlowTests a bit while adding tests for the new fixes: Checking "dummy" nodes with `IsLiteral` is more telling and strict than checking if the node is a directive - or, worse (outside functions), checking if it's an expression statement. Since that meant changing lines throughout the source file anyway, I went ahead and changed `string` to `var` in the file, per the Jint style.

lahma (approved):
Looks good, thanks for improving the debug experience!

PR #1382, opened 2022-12-17
Debugger API improvements

Jither:
* Added `Skip` event to `DebugHandler` (see https://github.com/Jither/Jint.DebuggerExample/blob/main/NOTES.md#Events for rationale)
* Added `Parsed` event to `Engine` for giving e.g. debuggers access to the AST as parsed by the engine, when passing code rather than AST to `Execute`/`Evaluate`.
* `BindingNames` on scopes are now lazily evaluated. Because we still want to leave out scopes without any bindings from scope chains (for now), a `HasBindings()` method has been added to environment records.

### Breaking changes:
* `Global` and `Local` properties on `CallFrame` have been removed. Because Jint's scope system is now more complex (and mostly aligned with other Javascript debuggers), `Global` and `Local` are very arbitrary scopes to make special cases for. Either can, of course, still be retrieved through the `ScopeChain` property on `CallFrame`.
* Shadowed bindings are no longer removed from outer scopes - they're inspectable like any other binding.
* When a breakpoint placed at a `debugger` statement is hit, only one `Break` event will be triggered (with `PauseType = Break`), rather than two (one with `PauseType = Break`, the other with `PauseType = DebuggerStatement`). This aligns the behavior with other debuggers - e.g. Chromium dev tools and Visual Studio's debugger - generally, only one event is triggered per execution point, with the priority being: `Step` > `Break` > `DebuggerStatement` > `Skip`.

These changes would bring https://github.com/Jither/Jint.DebugAdapter back to a place where it can be built against official Jint (although it still has some major issues, and module support in particular hasn't been tested at all yet).

### DIFF
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -292,7 +292,7 @@ namespace Jint
         /// </summary>
         private Engine ScriptEvaluation(ScriptRecord scriptRecord)
         {
-            DebugHandler.OnParsed(scriptRecord.EcmaScriptCode);
+            DebugHandler.OnBeforeEvaluate(scriptRecord.EcmaScriptCode);
 
             var globalEnv = Realm.GlobalEnv;
 
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -61,6 +61,13 @@ namespace Jint
         internal Intrinsics _originalIntrinsics = null!;
         internal Host _host = null!;
 
+        public delegate void ParsedHandler(object sender, string source, Script script);
+
+        /// <summary>
+        /// The Parsed event is triggered whenever a script has been passed as a result of execution or evaluation.
+        /// </summary>
+        public event ParsedHandler? Parsed;
+
         /// <summary>
         /// Constructs a new engine instance.
         /// </summary>
@@ -255,6 +262,7 @@ namespace Jint
                 : new JavaScriptParser(parserOptions);
 
             var script = parser.ParseScript(code, source, _isStrict);
+            Parsed?.Invoke(this, source, script);
 
             return Evaluate(script);
         }
@@ -275,6 +283,7 @@ namespace Jint
                 : new JavaScriptParser(parserOptions);
 
             var script = parser.ParseScript(code, source, _isStrict);
+            Parsed?.Invoke(this, source, script);
 
             return Execute(script);
         }
@@ -363,7 +372,7 @@ namespace Jint
 
             Action<JsValue> SettleWith(FunctionInstance settle) => value =>
             {
-                settle.Call(JsValue.Undefined, new[] {value});
+                settle.Call(JsValue.Undefined, new[] { value });
                 RunAvailableContinuations();
             };
 
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -64,7 +64,7 @@ namespace Jint
         public delegate void ParsedHandler(object sender, string source, Script script);
 
         /// <summary>
-        /// The Parsed event is triggered whenever a script has been passed as a result of execution or evaluation.
+        /// The Parsed event is triggered whenever a script has been parsed as a result of execution or evaluation.
         /// </summary>
         public event ParsedHandler? Parsed;
 
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -61,13 +61,6 @@ namespace Jint
         internal Intrinsics _originalIntrinsics = null!;
         internal Host _host = null!;
 
-        public delegate void ParsedHandler(object sender, string source, Script script);
-
-        /// <summary>
-        /// The Parsed event is triggered whenever a script has been parsed as a result of execution or evaluation.
-        /// </summary>
-        public event ParsedHandler? Parsed;
-
         /// <summary>
         /// Constructs a new engine instance.
         /// </summary>
@@ -262,7 +255,6 @@ namespace Jint
                 : new JavaScriptParser(parserOptions);
 
             var script = parser.ParseScript(code, source, _isStrict);
-            Parsed?.Invoke(this, source, script);
 
             return Evaluate(script);
         }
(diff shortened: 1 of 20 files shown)


Jint/Engine.cs
```diff
@@ -61,6 +61,13 @@ public sealed partial class Engine : IDisposable
         internal Intrinsics _originalIntrinsics = null!;
         internal Host _host = null!;
 
+        public delegate void ParsedHandler(object sender, string source, Script script);
+
+        /// <summary>
+        /// The Parsed event is triggered whenever a script has been parsed as a result of execution or evaluation.
+        /// </summary>
+        public event ParsedHandler? Parsed;
```
lahma, line 69: nit: kind of more like `ScriptParsed` as doesn't support module code

lahma: should these be in options? usually anything configurable is found there

Jither: > nit: kind of more like `ScriptParsed` as doesn't support module code

I was actually writing this while you replied:

Rethinking the `Parsed` event - and the recent `Loaded` event on `DefaultModuleLoader`. Ideally (especially for the VS Code DebugAdapter), those two should be combined into just `Parsed` on `Engine`. A debugger would need at least the `source` (ID) of the parsed script - to use as for breakpoints and as a general key - as well as the AST. But of course, after parsing, the `source` is already available on the AST's root node. DebugAdapter needs the `Parsed` event (or possibly more accurately a "script ready to execute" event) in order to wait for the debugging client (e.g. VS Code) to set e.g. breakpoint positions (after which VS Code sends `configurationDone` and the script can start evaluation).

So, a more general approach would be to remove `Loaded` from `DefaultModuleLoader` again, and instead let the engine handle it for *all* module loaders, by triggering the same `Parsed` event as a "classic" script `Execute` (with arguments `object sender` and `Program ast`) here:

https://github.com/sebastienros/jint/blob/main/Jint/Engine.Modules.cs#L44

(or does it make more sense to add it in `ImportModule`?)

... and, for "classic" scripts, somewhere... inhere:

https://github.com/sebastienros/jint/blob/4b6f44596bac1eb98a9492f7a8e60de5929c820d/Jint/Engine.cs#L295

... which would "unify" `Parsed` for the two methods of loading scripts. For the case of modules, e.g. a DebugAdapter would only listen for the first time the event is triggered.

Any idea at which points the event would make most sense to add? Considering the `Parsed` event will, in the case of e.g. a DebugAdapter, be used to pause execution of the `Engine` thread - e.g. (in the VS Code example) until the debugging client is done setting up evaluation.

Jither: > should these be in options? usually anything configurable is found there

In the original draft from last year (which I ended up deleting), they were actually calls to `DebugHandler.OnScriptReady`, and the `ScriptReady` event was on `DebugHandler` - along with `Step/Break/Skip`. Not sure about that, since I guess they may be useful outside of debugging too. How would we handle events with the `Options` fluent API?

lahma: I'm good if you're good with the solution 😉

 I wouldn't mind if fluent API isn't there for more special options. It's Options patterns more now with fluent available for common cases IMO. There are already many switches not visible on fluent side of I remember right.

Jither: Problem is, if we start introducing C# events to `Options`, that would require `Options` to also include logic for dispatching those events, or at the very least have an internal `OnParsed` on `Options`. Not just from a pattern perspective - that's enforced by the compiler.

Jither: Moved `Parsed` to `DebugHandler` for now, pending this discussion - just to implement the event for modules too - and test it with the DebugAdapter as shown above.

Jint/Engine.cs
```diff
@@ -292,6 +292,8 @@ public Engine Execute(Script script)
         /// </summary>
         private Engine ScriptEvaluation(ScriptRecord scriptRecord)
         {
+            DebugHandler.OnParsed(scriptRecord.EcmaScriptCode);
```
lahma, line 295: nit: script might be existing `Script` instance and thus not parsed in literal sense

lahma (approved):
LGTM, we can merge when you say so!

Jint/Engine.cs
lahma: Well it's internal so it's all about splitting hairs here. `Before`/`After` is good pairing, maybe even `OnEvaluating`/`OnEvaluated`, but all good as it's again, internal 🙂

Jither: Well, the event that it ends up triggering - `Parsed` - isn't internal. 🙂

lahma: Ah true. We're still in beta, we can break things even after this 😉

Jither: But the Debugger API has been so constantly breaking that I'd like for once to make a nice choice. 😐 Ended up with `BeforeEvaluate` (never liked `-ing` in event names, because it can both indicate "before" and "during").

And, more importantly, moved it into `EvaluateModule` rather than `LoadModule`, which is also used for loading modules for other purposes than evaluation. If that looks OK, it's good to go, I think.

Jither: Yeah, because now it doesn't take `LinkModule` into account, only `EvaluateModule` - and `LinkModule` is actually the main reason it's needed (scripts you don't pass to the engine yourself). 😋 But `LinkModule` doesn't do evaluation, so the event name, once again, becomes meaningless... 🤨

Jither: So there... The name stays for now, although it's now moved back to `LoadModule` 😋 Added a better test case (for two levels of import), and also tested again with Jint.ExampleDebugger and Jint.DebugAdapter.

Jither:
> Thanks once again!

My pleasure - I've updated `Jint.DebugAdapter` and `Jint.DebuggerExample` (as well as NOTES.md in the latter) to reflect the changes. 🙂

PR #1403, opened 2023-01-12
Fix nullable annotation of Engine.Invoke arguments

### LINKED ISSUE #1402: Wrong nullable annotation on Engine.Invoke
The `Engine.Invoke` methods only accept `object[]` as arguments, but should allow `object?[]` . We are already passing `null` here in our code and it works fine. We only noticed this when we migrated our own code to Nullable Reference Types.

mus65:
Fixes #1402


lahma (approved):

lahma:
Excellent, thank you!

PR #1425, opened 2023-02-07
Change JsDate.ToString to follow Date prototype logic

### LINKED ISSUE #1424: cfg.LocalTimeZone not work
Jint Version: 3.0.0-beta-2043

this is my code 

```C#
            var siteService = serviceProvider.GetRequiredService<ISiteService>();
            var siteSettings = siteService.GetSiteSettingsAsync().Result;
            var siteTimeZone = TimeZoneInfo.FindSystemTimeZoneById(siteSettings.TimeZoneId);
            {
                //TODO: 更新 Jint 版本后可移除
                options.Interop.TrackObjectWrapperIdentity = false;
                options.AddExtensionMethods(typeof(FreeSQLExtensions));
                options.TimeZone= siteTimeZone;
                options.LocalTimeZone(siteTimeZone);//"Asia/Shanghai"
            });
```
input : new Date("2022-1-1")

result : "2021-12-31T16:00:00Z"

Is it because my configuration method is wrong?

Thanks !

lahma:
fixes #1424

### DIFF
--- a/Jint/Native/JsDate.cs
+++ b/Jint/Native/JsDate.cs
@@ -1,4 +1,3 @@
-using System.Globalization;
 using Jint.Native.Date;
 using Jint.Native.Object;
 using Jint.Runtime;
@@ -56,6 +55,6 @@ public sealed class JsDate : ObjectInstance
             return "Infinity";
         }
 
-        return _engine.Realm.Intrinsics.Date.PrototypeObject.ToDateString(DateValue).ToString();
+        return TypeConverter.ToString(this);
     }
 }
--- a/Jint/Native/JsDate.cs
+++ b/Jint/Native/JsDate.cs
@@ -56,6 +56,6 @@ public sealed class JsDate : ObjectInstance
             return "Infinity";
         }
 
-        return ToDateTime().ToString("ddd MMM dd yyyy HH:mm:ss 'GMT'zzz", CultureInfo.InvariantCulture);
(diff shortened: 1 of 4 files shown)


Jint/Native/JsDate.cs
```diff
@@ -56,6 +56,6 @@ public override string ToString()
             return "Infinity";
         }
 
-        return ToDateTime().ToString("ddd MMM dd yyyy HH:mm:ss 'GMT'zzz", CultureInfo.InvariantCulture);
+        return _engine.Realm.Intrinsics.Date.PrototypeObject.ToDateString(DateValue).ToString();
```
sebastienros, line 59: do other JS object follow this pattern?

sebastienros: meaning JsNumber, ...

lahma: well the primitives are quite limited with toString, should we just ignore this - don't you like the behavior?

lahma: Well JsNumber does follow JS semantincs:

```
    public override string ToString()
    {
        return TypeConverter.ToString(_value);
    }
```

lahma: Changed to use `TypeConverter` here too

sebastienros (approved):

sebastienros:
Not sure we need the new dependency. I love the library, use it in Fluid also, but in this case we just need a culture info, we can just use a string that is known from both linux and windows in dotnet core.

PR #1434, opened 2023-02-15
Improve date parsing by utilizing MimeKit logic

### LINKED ISSUE #766: NaN Date created from what seems to be a valid Date string
As a followup to the Date issue I mentioned yesterday, I noticed that my Date should not have been getting constructed with a NaN argument in the first place.

The value passed to the Date constructor is the string `Thu, 30 Jan 2020 08:00:00 PST`.

I have inspected, and that is indeed the value of `var date` at [DateConstructor L97](https://github.com/sebastienros/jint/blob/dev/Jint/Native/Date/DateConstructor.cs#L97).

However all the checks fail and the statement `return JsNumber.DoubleNaN;` is executed.

Is that expected? Constructing a Date from this string succeeds in the browser console:

```js
new Date("Thu, 30 Jan 2020 08:00:00 PST");
```

Outputs:

```bash
> Date Thu Jan 30 2020 16:00:00 GMT+0000 (Coordinated Universal Time)
```

I can reproduce it with this unit test:

```cs
[Fact]
public void CanParseLocaleString()
{
    var result = _engine.Execute("new Date(\"Thu, 30 Jan 2020 08:00:00 PST\");").GetCompletionValue();

    Assert.True(result.IsDate() && !double.IsNaN(result.AsDate().PrimitiveValue));
}
```

lahma:
Utilizing extracted code from MimeKit, removed all things not needed. No dependency on library, but attribution given in both source and CREDITS.txt.

A fast turnaround for soon three year old issue.

fixes #766

### DIFF
--- a/CREDITS.txt
+++ b/CREDITS.txt
@@ -21,3 +21,11 @@ Jint number serialization is based on the Java port of the FastDtoa algorithm fr
 Website:      http://mozilla.org/
 Copyright:    Copyright (c) Mozilla
 License:      MPL 2.0 - http://mozilla.org/MPL/2.0/
+
+MimeKit
+-----
+Jint date parsing utilizes code extracted from the excellent MimeKit library.
+
+Website:      https://github.com/jstedfast/MimeKit
+Copyright:    Copyright (C) 2012-2022 .NET Foundation and Contributors
+License:      MIT - https://github.com/jstedfast/MimeKit/blob/master/LICENSE
--- a/Jint.Tests/Runtime/DateTests.cs
+++ b/Jint.Tests/Runtime/DateTests.cs
@@ -83,4 +83,18 @@ public class DateTests
         Assert.Equal("Tue Feb 01 2022 00:00:00 GMT+0800 (China Standard Time)", engine.Evaluate("new Date(2022,1,1).toString()"));
         Assert.Equal("Tue Feb 01 2022 00:00:00 GMT+0800 (China Standard Time)", engine.Evaluate("new Date(2022,1,1)").ToString());
     }
+
+    [Theory]
+    [InlineData("Thu, 30 Jan 2020 08:00:00 PST", 1580400000000)]
+    [InlineData("Thursday January 01 1970 00:00:25 UTC", 25000)]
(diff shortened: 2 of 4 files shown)


sebastienros:
Saw this and looks good, was wondering how better it is from the logic in Fluid which I improved a lot recently for different reasons, and maybe there should be a separate library ;)

lahma:
I tried the same test cases with Fluid using following test code, 7/8 test cases failed so I think I'll merge this now and we can revisit when yet another library emerges 😉 

```c#
[Theory]
[InlineData("Thu, 30 Jan 2020 08:00:00 PST", 1580400000000)]
[InlineData("Thursday January 01 1970 00:00:25 UTC", 25000)]
[InlineData("Wednesday 31 December 1969 18:01:26 MDT", 86000)]
[InlineData("Wednesday 31 December 1969 19:00:08 EST", 8000)]
[InlineData("Wednesday 31 December 1969 17:01:59 PDT", 119000)]
[InlineData("December 31 1969 17:01:14 MST", 74000)]
[InlineData("January 01 1970 01:46:06 +0145", 66000)]
[InlineData("December 31 1969 17:00:50 PDT", 50000)]
public void CanParseLocaleString(string dateTime, long expected)
{
    var options = new TemplateOptions() { CultureInfo = new CultureInfo("en-US", useUserOverride: false), TimeZone = TimeZoneInfo.Utc };
    new StringValue(dateTime).TryGetDateTimeInput(new TemplateContext(), out var customDateTime);
    Assert.Equal(expected, FromDateTimeOffset(customDateTime));
}

internal static readonly DateTime Epoch = new(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);

private static long FromDateTimeOffset(DateTimeOffset dt, bool negative = false)
{
    var dateAsUtc = dt.ToUniversalTime();

    double result;
    if (negative)
    {
        var zero = (Epoch - DateTime.MinValue).TotalMilliseconds;
        result = zero - TimeSpan.FromTicks(dateAsUtc.Ticks).TotalMilliseconds;
        result *= -1;
    }
    else
    {
        result = (dateAsUtc - Epoch).TotalMilliseconds;
    }

    return (long) System.Math.Floor(result);
}
```

PR #1443, opened 2023-02-19
jint provides the ability to hold additional data

libaowei:
I want jint enable hold additional data, and used by customize ModuleLoader
Currently I implemente it through SetValue, but SetValue always Wrapper clr object to ObjectWrapper


lahma:
I don't think there's need for separate holder. You could go with `SetValue` route and user either object wrapper - you can access data via `(MyClass) ((IObjectWrapper) GetValue(myKey)).Target` or inherit from `ObjectInstance` if you really need to.

lahma:
Another option to have `ConditionalWeakTable<Engine, IDictionary>` to hold your data. Closing this as too much use-case specific.

PR #1458, opened 2023-02-22
Engine.SetValue accepts a nullable object.

czema:
When Jint was modified to use nullable reference types, it lost the ability to pass a null value into Engine.SetValue(string, object).  This pull request restores that functionality.


lahma (approved):
Great, thank you!

PR #1511, opened 2023-03-25
Support inheritance chain of CLR objects with instanceof operator

mainlyer:
Hi! I implemented the functionality to make instanceof operator behave following the ineritance chain of CLR objects. But, because this not should be the default behavior, I suggest an explicit activation via Options.Runtime or a class Attribute. I made a test for net 6.0 too.

Regards.

### DIFF
--- a/Jint.Tests/Runtime/InstanceOfTests.cs
+++ b/Jint.Tests/Runtime/InstanceOfTests.cs
@@ -1,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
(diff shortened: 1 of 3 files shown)


lahma:
Thanks for the PR, I did a small tweak and moved the logic to be inside of `TypeReference` and bound via `hasInstance` symbol.

lahma:
Hoping to see more PRs and next time I'd suggest creating a separate branch for the PR, will make things usually easier 👍🏻

mainlyer:
I can relate the tweak, makes more sense the place 👍🏻

PR #1515, opened 2023-03-27
Add net60 target

lahma:


Jint/SkipLocalsInitAttribute.cs
```diff
@@ -1,3 +1,5 @@
+#if !NET6_0_OR_GREATER
```
sebastienros, line 1: a new TFM target just because of this?

NB: I have no idea what it's doing ;)

lahma: Performance optimization attribute that "does less" when preparing function call. NET 6 has this natively so conflicts. Adding new target to open more optimization paths via some APIs like `JsonEncodedText` - see gitter chat 😉

lahma: Here are some perf benefits from this attribute: https://github.com/sebastienros/esprima-dotnet/pull/339

PR #1517, opened 2023-03-28
Support optional constructor parameters under interop

mainlyer:
Optional constructor parameters working when creating CLR objects from JS.

### DIFF
--- a/Jint.Tests/Runtime/ConstructorSignature.cs
+++ b/Jint.Tests/Runtime/ConstructorSignature.cs
@@ -11,11 +11,22 @@ namespace Jint.Tests.Runtime
 
             engine.SetValue("A", TypeReference.CreateTypeReference(engine, typeof(A)));
 
-            Assert.Equal("6", engine.Evaluate("new A(1).Result").AsString());
+            // ParamArray tests
             Assert.Equal("3", engine.Evaluate("new A(1, 2).Result").AsString());
-            Assert.Equal("10", engine.Evaluate("new A(5, undefined).Result").AsString());
+            Assert.Equal("3", engine.Evaluate("new A(1, 2, null).Result").AsString());
+            Assert.Equal("3", engine.Evaluate("new A(1, 2, undefined).Result").AsString());
+            Assert.Equal("5", engine.Evaluate("new A(1, 2, null, undefined).Result").AsString());
+            Assert.Equal("9", engine.Evaluate("new A(1, 2, ...'packed').Result").AsString());
+            Assert.Equal("3", engine.Evaluate("new A(1, 2, []).Result").AsString());
+            Assert.Equal("7", engine.Evaluate("new A(1, 2, [...'abcd']).Result").AsString());
+
+            // Optional parameter tests
+            Assert.Equal("3", engine.Evaluate("new A(1, 2).Result").AsString());
+            Assert.Equal("6", engine.Evaluate("new A(1).Result").AsString());
+            Assert.Equal("7", engine.Evaluate("new A(2, undefined).Result").AsString());
+            Assert.Equal("8", engine.Evaluate("new A(3, undefined).Result").AsString());
             Assert.Equal("ab", engine.Evaluate("new A('a').Result").AsString());
-            Assert.Equal("a", engine.Evaluate("new A('a', null).Result").AsString());
+            Assert.Equal("ab", engine.Evaluate("new A('a', undefined).Result").AsString());
             Assert.Equal("ac", engine.Evaluate("new A('a', 'c').Result").AsString());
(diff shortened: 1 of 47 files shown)


Jint.Tests/Runtime/ConstructorSignature.cs
```diff
@@ -0,0 +1,33 @@
+﻿using Jint.Runtime.Interop;
+
+namespace Jint.Tests.Runtime
```
lahma, line 3: nit: you can use file-scoped namespaces for new code (no need for braces)

Jint/Runtime/Interop/TypeReference.cs
```diff
@@ -1,4 +1,5 @@
 using System.Collections.Concurrent;
+using System.Linq;
```
lahma, line 2: we try to avoid LINQ usage in Jint (excluding tests)

Jint/Runtime/Interop/TypeReference.cs
```diff
@@ -73,9 +74,31 @@ static ObjectInstance ObjectCreator(Engine engine, Realm realm, ObjectCreateStat
                         referenceType,
                         t => MethodDescriptor.Build(t.GetConstructors(BindingFlags.Public | BindingFlags.Instance)));
 
-                    foreach (var (method, _, _) in TypeConverter.FindBestMatch(engine, constructors, _ => arguments))
+                    var argumentProvider = new Func<MethodDescriptor, JsValue[]>(method =>
                     {
-                        var retVal = method.Call(engine, null, arguments);
+                        var parameters = method.Parameters;
+                        var list = new List<JsValue>(arguments);
```
lahma, line 80: can you replace this with less-LINQ solution, something like (untested):

```csharp
var extraParameters = Math.Max(0, parameters.Length - arguments.Length));
var array = JsValue[arguments.Count + extraParameters];
Array.Copy(arguments to array)
for (var i = start; i < array.Length; ++i)
{
 array[i] = // check if optional, make object
}
```
If you get the gist, so we could get away by using just single array without extra allocations as this is searching, also can do array init in same pass I guess?

lahma (changes requested):

mainlyer:
Hi! Are there any issues with the tests suites? I tried the same changes twice and the first time fail on Linux and the second time on Windows, always in the same test **[Jint.Tests.PublicInterface.PublicInterfaceTests.BindFunctionInstancesArePublic](https://github.com/sebastienros/jint/pull/1517/files#annotation_10070700790)**. The most weird thing is that failed test is not affected by the change.

Jint/Runtime/Interop/TypeReference.cs
mainlyer: I can relate! Investigating a little more, I found the issue deeper as I expected, because of the use of the **params** keyword.  For example:

A class with a constructor in C#:

```c#
class A
{
    public A(int param1, int param2, params object[] param3) {}
}
```

The constructor expects exactly 3 parameters:

```c#
var c = typeof(A).GetConstructors().First();

/* case 1: throws a System.Reflection.TargetParameterCountException */
c.Invoke(new object[] { 1, 2 });

/* case 2: works, but param3 is null, when expected an empty array */
c.Invoke(new object[] { 1, 2, null });

/* case 3: works */
c.Invoke(new object[] { 1, 2, new string[] { "1", null, "2" } });

/* case 4: works */
c.Invoke(new object[] { 1, 2, new object[] { "1", 2, ConsoleKey.A } });
```
I'm mixing this with optional parameters because from the point of view of JS, this should work:

```js
const a = new A(1, 2);
```
In fact, this must work transparently because the remaining parameters are _optional_ in this context but not in the background inside the runtime, this is the case 1. A simple workarround is add a third parameter as an empty array (not elegant). Case 2 is explicity set to null, then is the programmer's responsability, not the library. Cases 3 and 4 works when:

```js
const a = new A(1, 2, ['1', null, '2']);
```
The reality is that the following code is expected to work from a JS programmer:

```js
const a = new A(1, 2, 'a', 'b', 'c');
const a = new A(1, 2, ...'abc');
```
But to make it work you should need:

```js
const a = new A(1, 2, ['a', 'b', 'c']);
const a = new A(1, 2, [...'abc']);
```
Thats because the runtime receive 5 parameters as expected, but the CLR needs 3 and the last must be an array. Is for that I attach this subject to the initial optional parameters issue in CLR (with default value), because the concepts in CLR and JS are the same but in the background differs.

What do yo think about this @lahma?

lahma (approved):
Quite a lot of logic build for seems to cover edge cases quite nicely. Thanks for implementing this!

PR #1518, opened 2023-03-30
Support CLR object inheritance chain with prototypes

mainlyer:
Hi! Added a case for instanceof operator comparing the prototype against the type reference. Now the following code works when B is derived from A:

```js
A.prototype instanceof A // true
B.prototype instanceof A // true
````

Regards.


lahma:
Thank you! I made small adjustments for clarity.

PR #1520, opened 2023-03-31
Support params arrays in delegate invocation

### LINKED ISSUE #1218: Script delegates with params are called with object[]
Here to report more quirky behaviour with `params`, however I'm not sure if this one is as easy to fix as the last one...

I'm trying to implement [setInterval](https://developer.mozilla.org/en-US/docs/Web/API/setInterval) which accepts a variable amount of arguments after the delay:
```js
setInterval(func, delay, arg0, /* ... ,*/ argN)
```
However in my code, when invoking the func and providing the parameters, those parameters are passed as a single object array (`object[]`) instead of being "unpacked" to a series of variable parameters for use in the function.

## Demonstration
This demonstrates how invoking a delegate with params incorrectly servers all params in a single object[]
```csharp
using Jint;
using Jint.Runtime.Interop;
using System;

namespace TestParams
{
    internal class Program
    {
        public delegate void RegisterCallbackDelegate(CallbackAction callback, params object[] arguments);
        
        public delegate void CallbackAction(params object[] arguments);

        {
            var engine = new Engine();
            engine.SetValue("registerCallback", new DelegateWrapper(engine, new RegisterCallbackDelegate(RegisterCallback)));

            engine.Execute(@"
                var argsConcat = '';
                registerCallback((valTest, valOther, valNumber) => {
                    argsConcat+=typeof valTest;
                    argsConcat+=' ' + typeof valOther;
                    argsConcat+=' ' + typeof valNumber;
                }, 'test', 'other', 1337);
            ");

            if (engine.Evaluate("argsConcat === 'string string number'") == true)
                Console.WriteLine("Params in delegates work as expected");
            else
            {
                Console.WriteLine(engine.Evaluate("argsConcat"));
                Console.WriteLine("Params for delegates are incorrectly served as a single object[]");
            }

            Console.ReadKey();
        }

        static void RegisterCallback(CallbackAction callback, params object[] arguments)
        {
            callback.Invoke(arguments);
        }
    }
}
```

## Problematic code
I've spent quite sometime trying to figure out what is wrong and I suspect it's around here: https://github.com/sebastienros/jint/blob/110ea56bce734fad575becc1bd475283947c440d/Jint/Runtime/Interop/DefaultTypeConverter.cs#L210

It seems to me like at some point in this BuildDelegate function, the information that the argument has a [ParamsArrayAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.paramarrayattribute?view=net-6.0) is lost/ignored.
That or it's being passed incorrectly to the lambda here:
https://github.com/sebastienros/jint/blob/110ea56bce734fad575becc1bd475283947c440d/Jint/Runtime/Interop/DefaultTypeConverter.cs#L256

Sadly I have no experience with Linq.Expressions. Neither trial and error nor looking online has yielded positive results or anymore understanding of where the problem lies. Perhaps it's not even here, but where the Expression.Lambda is eventually called?

Hope this helps in some way! 

Have a nice Sunday.

PS: I'm on the latest 3.x version

mainlyer:
fixes #1218


Jint/Native/Function/ScriptFunctionInstance.cs
```diff
@@ -58,6 +59,22 @@ protected internal override JsValue Call(JsValue thisArgument, JsValue[] argumen
             var strict = _functionDefinition.Strict || _thisMode == FunctionThisMode.Strict;
             using (new StrictModeScope(strict, true))
             {
+                // check the stack for the presence of a DelegateWrapper with SpreadParameters set to true
+                if (_engine.CallStack.TryPeek(out var stackItem) &&
```
lahma, line 63: this does not seem the correct place to check for interop, JS spec and interop are generally different things and interop should prepare correct parameters for interop

mainlyer: Yeah! I wondered that too... :) I changed it to Interop. Due to ScriptFunctionInstance is created early to bind a JS function, is not posible to achieve this kind of interop, because happens inside the call.

lahma:
I still think the problem is in `DefaultTypeConverter.BuildDelegate`  which basically should inject far more complicated code that would also check for `params`. This thing makes my head hurt due to mu shortcomings in dynamic code generation. 

I wonder if original author of the current logic @Dogwei could chime in whether such `params` handling would be easy to add to current logic.

mainlyer:
I can relate. I had not seen the ```DefaultTypeConverter.BuildDelegate``` method. I can create an ```Attribute``` in order to decorate the delegate what you want to spread the ```ParamArray``` and do the logic there. This makes more sense and now I can understand clearly the logic for creating and calling a delegate from JS function so transparently.

lahma:
Not sure what you mean with the attribute, but generally consumer code shouldn't need to use attributes or special configuration. Jint should be able to pass parameters in correct format by using information from reflection.

mainlyer:
I mean, for example:

```c#
private static void RegisterCallback(CallbackAction callback, params object[] arguments)
{
    callback.Invoke(arguments);
}

private delegate void RegisterCallbackDelegate(CallbackAction callback, params object[] arguments);

private delegate void CallbackAction([SpreadParameters] params object[] arguments);
```

lahma:
Such extra attribute shouldn't be needed, compiler already decorates the parameter with `ParamsArrayAttribute`.

mainlyer:
I known, but the attribute should be used only when you want that behavior. Or is preferible to make it behaves always in this way?

mainlyer:
Hi! I did the stuff. If you want to remove the SpreadParametersAttribute, just do it. I don't know how many people is using this library and a change of the default behavior maybe break the logic in case of a library update. Feel free!

Regards.

Jint/Runtime/Interop/DefaultTypeConverter.cs
```diff
@@ -256,18 +256,43 @@ private Delegate BuildDelegate(Type type, Func<JsValue, JsValue[], JsValue> func
                 parameters[i] = Expression.Parameter(arguments[i].ParameterType, arguments[i].Name);
             }
 
-            var initializers = new MethodCallExpression[parameters.Length];
-            for (var i = 0; i < parameters.Length; i++)
+            MethodCallExpression[] initializers;
+
+            if (arguments.Length == 1 &&
```
lahma, line 261: now this limits to having single parameter in function declaration, if you look how `MethodInfoFunctionInstance.Call` handles this it checks for each parameter. I'm unsure if this can easily handle same logic or should this somehow use the same logic.

Jint/Runtime/Interop/SpreadParametersAttribute.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Jint.Runtime.Interop
+{
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public sealed class SpreadParametersAttribute : Attribute
```
lahma, line 4: we don't use attributes for any other interop scenario at the moment in Jint so probably shouldn't use for this either. if changing logic would be breaking I think it's find as long as the behavior changes for the better (expected).

mainlyer: I agree!

Jint/Runtime/Interop/DefaultTypeConverter.cs
mainlyer: Changed. There are no limits now. If the use of ```List``` is not allowed, I can change it to an array querying before the number of parameters the ```FunctionInstance``` expects.

lahma (approved):
Thanks for hanging in there and iterating the solution, this one was quite complicated just like the original issue noted!

mainlyer:
Thanks you @lahma for guiding me accross the file/logic organization. Now I have clearer the spec/(interpreter-runtime)/interop organization. Good job.

lahma:
@mainlyer your much appreciated work for interop enhancements has been released as part of the latest NuGet beta release.

mainlyer:
A pleasure @lahma !

Issue #514, opened 2018-06-18
async/await and callback support


christianrondeau:
I was surprised to find no issues nor pull requests to support C#/JS async/await and Task support.

In a perfect world, I could make that transparent and "halt" the calling method (like #192), or use a callback mechanism (which would make scripts much more complex however).

I'd prefer to avoid using `return task.Result` and friends, so that I don't get into a deadlock eventually.

Before investigating this further, did anyone actually attempt any async/await support in Jint? @sebastienros is this something you'd be interested in supporting, e.g. through the native ECMA-262 async/await (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function)? Any failed or partial attempt?

To avoid any confusion, given this C# code:

```
public async Task<string> loadSomething(string path) {
  return await File.ReadAllTextAsync(path);
}
```

We could consider async/await support in one or many of these expressions in JavaScript:

```
// Pausing (easiest for script writers):
var x = loadSomething('test.txt');
// Callbacks
var x;
loadSomething('test.txt', function(result) { x = result });;
// Promises
var x;
loadSomething('test.txt').then(function(result) { x = result });;
// Async Function
var x = await loadSomething('test.txt');
```

Any ideas or feedback on that is welcome; I'd like to avoid spending any time on this if it would not be merged, or if the effort is too large for what I can afford.

christianrondeau:
@ackava thanks for sharing your ideas. It looks like you are using TypeScript, which would explain why using the `Promise` constructor, the `=>` syntax and eventually using `await` would work for you :) Using a transpiler is an idea worth investigating though.

I realize that supporting async/await would pretty much be supporting ES2017, which has it's own open issue: https://github.com/sebastienros/jint/issues/343 - this would probably (I can only guess) solve my issue, depending on how @sebastienros decided to implement "await" for C#-backed functions. For now, the `Promise` constructor looks like it's not implemented in the `3.0` branch, so I guess we'll have to wait and see. If I had more time, I'd try contribute this back in Jint, but I don't think I'll be able to take the lead on this for some time.

If someone else wonders how to do transpiling in Jint: https://github.com/sebastienros/jint/issues/274 (did not try it yet)

ta10212:
Hi everyone, I echo @christianrondeau in the desire for async/await support.  

While promises would work, it would be even simpler if the Engine could just "await" a function delegate passed to "SetValue" - if that delegate wrapped an asynchronous method.

[...]

PR #1567, opened 2023-06-28
Experimental - Task to Promise conversion

Xicy:
#514

### DIFF
--- a/Jint/Runtime/Interop/DelegateWrapper.cs
+++ b/Jint/Runtime/Interop/DelegateWrapper.cs
@@ -130,7 +130,47 @@ namespace Jint.Runtime.Interop
 
             try
             {
-                return FromObject(Engine, _d.DynamicInvoke(parameters));
+                var result = _d.DynamicInvoke(parameters);
+                if (result is Task task)
+                {
+                    var (promise, resolve, reject) = _engine.RegisterPromise();
+                    task = task.ContinueWith(continuationAction =>
+                    {
+                        if (continuationAction.IsFaulted)
+                        {
+                            reject(JsValue.FromObject(_engine, continuationAction.Exception));
+                        }
+                        else if (continuationAction.IsCanceled)
+                        {
+                            reject(JsValue.Undefined);
+                        }
+                        else
+                        {
+                            var result = continuationAction.GetType().GetProperty(nameof(Task<object>.Result));
+                            if (result is not null)
+                            {
+                                resolve(JsValue.FromObject(_engine, result.GetValue(continuationAction)));
+                            }
+                            else
+                            {
+                                resolve(JsValue.FromObject(_engine, JsValue.Undefined));
+                            }
+                        }
+                    });
+                    _engine.AddToEventLoop(() =>
+                    {
+                        if (!task.IsCompleted)
+                        {
+                            // Task.Wait has the potential of inlining the task's execution on the current thread; avoid this.
+                            ((IAsyncResult) task).AsyncWaitHandle.WaitOne();
+                        }
+                    });
+                    return promise;
+                }
+                else
+                {
+                    return FromObject(Engine, result);
+                }
             }
             catch (TargetInvocationException exception)
             {
--- a/Jint/Runtime/Interop/DelegateWrapper.cs
+++ b/Jint/Runtime/Interop/DelegateWrapper.cs
@@ -13,6 +13,7 @@ namespace Jint.Runtime.Interop
     public sealed class DelegateWrapper : FunctionInstance
     {
         private static readonly JsString _name = new JsString("delegate");
+        private static readonly Type _voidTaskResultType = Task.CompletedTask.GetType.GetGenericArguments()[0];
         private readonly Delegate _d;
         private readonly bool _delegateContainsParamsArgument;
 
@@ -159,7 +160,17 @@ namespace Jint.Runtime.Interop
                 }
                 else
                 {
-                    var result = continuationAction.GetType().GetProperty(nameof(Task<object>.Result));
+                    var taskType = continuationAction.GetType();
+
+                    // Special case: Marshal `async Task` as undefined, as this is `Task<VoidTaskResult>` at runtime
+                    // See https://github.com/sebastienros/jint/pull/1567#issuecomment-1681987702
+                    if (taskType.IsGenericType && taskType.GetGenericArguments()[0] == _voidTaskResult)
+                    {
+                        resolve(FromObject(Engine, JsValue.Undefined));
+                        return;
+                    }
+
+                    var result = taskType.GetProperty(nameof(Task<object>.Result));
                     if (result is not null)
                     {
                         resolve(FromObject(Engine, result.GetValue(continuationAction)));
--- a/Jint/Runtime/Interop/DelegateWrapper.cs
+++ b/Jint/Runtime/Interop/DelegateWrapper.cs
@@ -131,52 +131,56 @@ namespace Jint.Runtime.Interop
             try
             {
                 var result = _d.DynamicInvoke(parameters);
-                if (result is Task task)
-                {
-                    var (promise, resolve, reject) = _engine.RegisterPromise();
-                    task = task.ContinueWith(continuationAction =>
-                    {
-                        if (continuationAction.IsFaulted)
-                        {
-                            reject(JsValue.FromObject(_engine, continuationAction.Exception));
-                        }
-                        else if (continuationAction.IsCanceled)
-                        {
-                            reject(JsValue.Undefined);
-                        }
-                        else
-                        {
-                            var result = continuationAction.GetType().GetProperty(nameof(Task<object>.Result));
-                            if (result is not null)
-                            {
-                                resolve(JsValue.FromObject(_engine, result.GetValue(continuationAction)));
-                            }
-                            else
-                            {
-                                resolve(JsValue.FromObject(_engine, JsValue.Undefined));
-                            }
-                        }
-                    });
-                    _engine.AddToEventLoop(() =>
-                    {
-                        if (!task.IsCompleted)
-                        {
-                            // Task.Wait has the potential of inlining the task's execution on the current thread; avoid this.
-                            ((IAsyncResult) task).AsyncWaitHandle.WaitOne();
-                        }
-                    });
-                    return promise;
-                }
-                else
+                if (result is not Task task)
                 {
                     return FromObject(Engine, result);
                 }
+                return ConvertTaskToPromise(task);
             }
             catch (TargetInvocationException exception)
             {
-                ExceptionHelper.ThrowMeaningfulException(_engine, exception);
+                ExceptionHelper.ThrowMeaningfulException(Engine, exception);
                 throw;
             }
+        }
+
+        private JsValue ConvertTaskToPromise(Task task)
+        {
+            var (promise, resolve, reject) = Engine.RegisterPromise();
+            task = task.ContinueWith(continuationAction =>
+            {
+                if (continuationAction.IsFaulted)
+                {
+                    reject(FromObject(Engine, continuationAction.Exception));
+                }
+                else if (continuationAction.IsCanceled)
+                {
+                    reject(FromObject(Engine, new ExecutionCanceledException()));
+                }
+                else
+                {
+                    var result = continuationAction.GetType().GetProperty(nameof(Task<object>.Result));
+                    if (result is not null)
+                    {
+                        resolve(FromObject(Engine, result.GetValue(continuationAction)));
+                    }
+                    else
+                    {
+                        resolve(FromObject(Engine, JsValue.Undefined));
+                    }
+                }
+            });
+
+            Engine.AddToEventLoop(() =>
+            {
+                if (!task.IsCompleted)
+                {
+                    // Task.Wait has the potential of inlining the task's execution on the current thread; avoid this.
+                    ((IAsyncResult) task).AsyncWaitHandle.WaitOne();
+                }
+            });
+
+            return promise;
         }
     }
 }
--- a/Jint/Runtime/Interop/DelegateWrapper.cs
+++ b/Jint/Runtime/Interop/DelegateWrapper.cs
@@ -13,7 +13,7 @@ namespace Jint.Runtime.Interop
     public sealed class DelegateWrapper : FunctionInstance
     {
         private static readonly JsString _name = new JsString("delegate");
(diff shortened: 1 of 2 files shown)


lahma:
Great initiative here too. I don't have strong opinions, but might open a can of worms when people would start submitting issues with corner cases - but of course they can be referred to create a PR to improve things. It seems that this would handle the most requested cases anyway.

/cc @twop 
/

Jint/Runtime/Interop/DelegateWrapper.cs
```diff
@@ -130,13 +130,57 @@ protected internal override JsValue Call(JsValue thisObject, JsValue[] jsArgumen
...
+                    }
+                }
+            });
+
+            Engine.AddToEventLoop(() =>
+            {
+                if (!task.IsCompleted)
+                {
+                    // Task.Wait has the potential of inlining the task's execution on the current thread; avoid this.
+                    ((IAsyncResult) task).AsyncWaitHandle.WaitOne();
```
twop, line 179: Does this mean that JS thread will be blocked until task is resolved?

Xicy: Right, when promise needs to resolve, block to js thread until resolved.

I'm not sure but Evalutate function in unit tests, only process the script. It does not block until call [RunAvailableContinuations](https://github.com/sebastienros/jint/blob/75b2d703e6ca5f5661b9107a3285ee4e19f097df/Jint/Engine.cs#L410C45-L410C45)

twop: Yeah, I think that is the main problem that I encountered with the Task mapping too.

Tasks (plural in general case) can be resolved out of order at any time on any thread. But in JS it is one thread that collects continuations in the order they came in, in other words **non blocking at any time**. So I would probably advocate to use that as a guiding principal, and to my understanding the current code is not aligned with that, or I'm missing something obv?

Xicy: >my understanding the current code is not aligned with that
As per your explanation, yes code does not align. Maybe EventLoop works on another thread that can be closer to working asynchrony.

By the way, I don't have experience with that. If you lead the way, I try to enhance the code.

lahma:
At the same time I see great value in this but do you folks see possible *bad* drawbacks (via bug reports and hard to track errors)? This would create a reasonable bridge (AFAIK) to allow common patterns and interoperability. Of course bug reports can be closed with "please send a PR to fix instead", but there's some fundamental impossibility - that might not be fair.

VelvetToroyashi:
Wanted to poke at this to see where things were in terms of progress; this is great, and I'm excited to seeing these changes be implemented, however in my attempt at testing it, I've noticed a strange quirk.

Take the following C# code:

```cs
var engine = new Engine().SetValue("asyncMethod", () => Task.Delay(2000));
```

And in JS:

```js
(
  async () => // Workaround for no top-level async/await
  {
    await asyncMethod();
  }
)();
```

When calling `engine.Execute` with that script, it runs as expected; the code is delayed for 2 seconds, and that's that. 

However, if the delegate is instead made async, and awaits the task that represents the work:

`async () => await Task.Delay(2000)`

The promise is resolved immediately which is unexpected behavior, especially considering that on the surface, [practically identical](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAIAYAEqCMBWAbgFgAoVAZmwCZMBhTAb1MxepqZNa8wDcBDMJj6YAvJgAUASlEA+bADYAdABEApgBs+AT3HpJxTtxZ9FASSgBnAA6qAxsHEAiVPIC0AJVXAIYKAEsoAHNHfWYjTDCjfkEAI1FsAA4JaRE5VABOBRUNbV19CMNw1kjuGLNLG3snAEELLShbTE0AWxi4PhCDIpKuEwBZLwALAHs4RU9vXwAVLRty6zsHZzcJn38gzp7WMoHgEbHV6dnVecql2vrGtXVVQL5gVU3C7i2WaKFhMT4pLvD3mLiYhiP1eBSKfBM5gWVWWHi8awCwVCzy4ANOixqdQamGut3uj2RXAAvqQiUA). My initial suspicion was it was something to do with how the task is handled, but I suspect this may be down to something deeper in the TPL implementation? I'll take a deeper look at this later in the day, but an interesting behavior nonetheless.

Xicy:
@VelvetToroyashi, Thank you also here :), 
Is there any chance to check failed test cases?

VelvetToroyashi:
@Xicy I can run the tests locally here in a second

Jint/Runtime/Interop/DelegateWrapper.cs
```diff
@@ -130,13 +131,67 @@ protected internal override JsValue Call(JsValue thisObject, JsValue[] jsArgumen
...
+                {
+                    reject(FromObject(Engine, new ExecutionCanceledException()));
+                }
+                else
+                {
+                    var taskType = continuationAction.GetType();
+
+                    // Special case: Marshal `async Task` as undefined, as this is `Task<VoidTaskResult>` at runtime
+                    // See https://github.com/sebastienros/jint/pull/1567#issuecomment-1681987702
+                    if (taskType.IsGenericType && taskType.GetGenericArguments()[0] == _voidTaskResult)
```
VelvetToroyashi, line 167:
```suggestion
- if (taskType.IsGenericType && taskType.GetGenericArguments()[0] == _voidTaskResult)
+ if (taskType.IsGenericType && taskType.GetGenericArguments()[0] == _voidTaskResultType)

Comment (Xicy) [reply Jint/Runtime/Interop/DelegateWrapper.cs]: I update pr thanks :)

File: Jint/Runtime/Interop/DelegateWrapper.cs [hunk 3]
```diff
@@ -13,6 +13,7 @@ namespace Jint.Runtime.Interop
     public sealed class DelegateWrapper : FunctionInstance
     {
         private static readonly JsString _name = new JsString("delegate");
+        private static readonly Type _voidTaskResultType = Task.CompletedTask.GetType.GetGenericArguments()[0];
```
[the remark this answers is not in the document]
Xicy: Not a big deal, it's fixed :)

ejsmith:
This looks awesome! Can't wait to get this ability. I was wondering if we can await a C# task from the a script... would it be possible to freeze and serialize the state of the JS engine at that point and resume it at a later time? I'm thinking of long running workflow type scenarios where there is a call out from the script to wait for some event to occur. That is implemented in the form of an await inside of the script that is calling a C# async method. That async method could then serialize engine state, wait some undefined amount of time and when that event occurs it would deserialize the engine state and resume the script execution. Thoughts?

sebastienros:
@ejsmith maybe not rely on Jint as the core but use it as a way to run an activity in the workflow. Check [Elsa Workflows](https://v2.elsaworkflows.io/) or the one in Orchard Core (Else is coming from there) which are using Jint to run customizable tasks.

ejsmith:
> @ejsmith maybe not rely on Jint as the core but use it as a way to run an activity in the workflow. Check [Elsa Workflows](https://v2.elsaworkflows.io/) or the one in Orchard Core (Else is coming from there) which are using Jint to run customizable tasks.

Yeah, I'm going to look into that as well. Was just thinking it would be really nice and simpler to just have a script that did the entire workflow.

sebastienros:
@ejsmith why not after all, currently Elsa/OrchardCore are using a Json description of it. Now I am starting to understand your vision. And being able to serialize the whole state and continue is something that would be helpful in other scenarios (it has bugged us many times). Even with some constraints like assuming the global objects are immutable so they don't have to be part of the serialization. Or that we can't even store state and it's up to the workflow engine to store what it needs and rehydrate it when it's resumed (custom C# bindings, state objects, ...).

lahma:
I think that the easiest way to handle "game state" is to have a separate state object that can be JSON serialized. JavaScript engine state is far too complex to serialize successfully as one can mutate so many things (functions etc).

sebastienros:
@lahma exactly, that's what I meant by "it's up to the workflow engine to store what it needs and rehydrate it"

So the remaining problem is to be able to stop the engine, maybe with a custom API call invoked from a C# callback, which would return some kind of serializable code pointer, and they another call to "resume" a script instead of "execute". I have no idea how it could work with the current implementation. Meaning that I believe that is technically impossible without changing how a script is currently processed.

ejsmith:
> I think that the easiest way to handle "game state" is to have a separate state object that can be JSON serialized. JavaScript engine state is far too complex to serialize successfully as one can mutate so many things (functions etc).

Yeah, I can see that. I'd think you'd have to restrict it to no global state and just serialize the call stack and variables passed to those stacks and then be able to jump back to that stack. But I'm sure that it's way more complicated than that. 🙂

ejsmith:
> @lahma exactly, that's what I meant by "it's up to the workflow engine to store what it needs and rehydrate it"
> 
> 
> 
> So the remaining problem is to be able to stop the engine, maybe with a custom API call invoked from a C# callback, which would return some kind of serializable code pointer, and they another call to "resume" a script instead of "execute". I have no idea how it could work with the current implementation. Meaning that I believe that is technically impossible without changing how a script is currently processed.

Thanks! I'm going to dive into that.

I currently have a really good scripting setup for custom logic and actions was just thinking that it would be really nice to just add something like `await waitForDays(12)` right into the script and various other kinds of awaits for different triggers. I think Elsa will work really well too. Just didn't want to get into having to do some sort of designer for building workflows. Does Elsa have a web designer?

sebastienros:
Else does have a web designer.

sfmskywalker:
Yep: https://github.com/elsa-workflows/elsa-studio

It even has [an installation guide](https://v3.elsaworkflows.io/docs/installation/elsa-studio-blazorwasm).

sfmskywalker:
> This looks awesome! Can't wait to get this ability. I was wondering if we can await a C# task from the a script... would it be possible to freeze and serialize the state of the JS engine at that point and resume it at a later time? I'm thinking of long running workflow type scenarios where there is a call out from the script to wait for some event to occur. That is implemented in the form of an await inside of the script that is calling a C# async method. That async method could then serialize engine state, wait some undefined amount of time and when that event occurs it would deserialize the engine state and resume the script execution. Thoughts?
> 
> Could potentially use this to provide the long running task support: https://github.com/Azure/durabletask

This is super interesting to me. I've been experimenting with a custom DSL that reads like JS, but one where keywords and statements translate to actual workflow activities, making it easy to describe a long running process that reads sequentially. As an example:


```javascript
variable age = GetInput("age")";

if age > 16 then
{
    print("Enjoy your driver's license! And be careful!");
    event("crash");
    print("I told you to be careful -_-");
}
else
    print("Enjoy your bicycle!");
                        
print("Goodbye.");
```

In the above pseudo code, `if/else` keywords translate to Elsa's `If` activity where the first body becomes a `Sequence` activity for the `Then` property and the `else` body becomes a `Sequence` activity for the `Else` property. `print` translates to the `WriteLine` activity,

The relevant part here is the `event` function, which translates to the `Event` activity, which is a blocking activity that puts the workflow in a suspended state, and gets resumed when the application triggers the "crash" event.

But there are two significant limitations with this DSL for the moment:

- It's not finished
- It doesn't support all of the scripting stuff that a real JavaScript interpreter does.

So, I think that the idea to actually use real JavaScript and have it yield to the workflow engine, and have that resume the script at some point, is brilliant. If we can somehow pull this off, I might not have to write a custom DSL in the first place to be able to define a workflow in a (limited) scripting language. Instead, an entire workflow, both long-and short-running, could be defined in JS 🤯 

And although this could be supported inside a JS activity running in a workflow, I think it would make just as much sense, if not more, to also have the ability to just write .js files to define long running processes - which is the point of the custom DSL.

I have no clue what it would take to make Jint support something like this (being able to yield back to the host app mid-script and then return), but let's keep marinating on it 🤔

sfmskywalker:
`yield` sounds like it could be a perfect fit. If I understand correctly, then it might be used like this from a `RunJavaScritpt` activity:

```javascript
const foo = 'bar"l
yield*; // Yields back to the RunJavaScript activity, which will now create a bookmark. In most samples I;ve seen, yield* is used to generate data (e.g. yield* 42), but in our case, we just want to yield control back. Maybe it would be nice to make that explicit, like yield* suspend or whatever.
const someInput = getInput(); // Input would be provided by the application resuming the JS bookmark.
return someInput;
```

sfmskywalker:
And I wonder, if that works, then could we also yield control from within a custom JS function? So that we could write a long-running workflow in pure JS:

```javascript
const approveSignalUrl.= generateSignalUrl("approve");
const rejectSignalUrl.= generateSignalUrl("reject");

await sendEmail("approver@acme.com", `<a href="${approveSignalUrl}">Approve</a> or <a href="${rejectSignalUrl}">Reject</a>`);

if(signalReceived("approve")) // signalReceived should yield control back to engine.
{
   await sendEmail("contributor@acme.com", "Great job!");
}

if(signalReceived("reject"))
{
   await sendEmail("contributor@acme.com", "You and your work suck!");
}
```

In C#, `signalReceived` would somehow be able to tell Jint to yield, e.g.:

```csharp
engine.SetValue("signalReceived", () => engine.YieldMagic("signalReceived"));
```

But that might be too farfetched. Perhaps the following approach might be more realistic, where the yield* keyword is returned in front of the suspending activity:

```javascript
if(yield* signalReceived("approve")) // signalReceived should yield control back to engine.
{
   await sendEmail("contributor@acme.com", "Great job!");
}

if(yield* signalReceived("reject"))
{
   await sendEmail("contributor@acme.com", "You and your work suck!");
}
```

Here, the engine would understand that the SignalReceived activity is to be scheduled next, and when that gets resumed, it knows from the bookmark from where to resume JS execution.

sfmskywalker:
Looking at the sample script, it would be important that the yield* keyword used in this way doesn't stop running the remaining code in order to also create a bookmark for the "reject" signal.

That got complicated real fast 😅

ejsmith:
BTW, all of this first requires this Task to Promise conversion PR to be accepted which we have totally hi-jacked. Sorry @Xicy! @lahma what are the chances this PR is going to be accepted?

lahma:
@Xicy thank you for the PR, looks good and kudos for the test coverage. I've rebased against `main` and it still seems to have some problems running against full .NET framework, could you please take a look?

Xicy:
> @Xicy thank you for the PR, looks good and kudos for the test coverage. I've rebased against `main` and it still seems to have some problems running against full .NET framework, could you please take a look?

First of all, I'm sorry for your loss. I'll try to take care of it next weekend.

lahma (approved):

lahma:
Thank you!

Issue #1468, opened 2023-02-25
Feature Request - Performance Logging/Profiling support


edokan:
Hello,

I want to profile function and property calls to CLR objects. I looked at the code and saw that most of the points I can hook (ObjectWrapper, TypeResolver, etc.) are either sealed or do not let me intercept calls. I tried using Debugger by stepping in, but it calls me only before the call and does not work for properties. I cannot use a regular profiler like dotTrace for various reasons, including security. 

Is there a way for me to be able to log/trace function calls/property fetches to CLR without changing the source code, which I'm open to but prefer not to in a sizeable project without a good grasp of the things?

Best regards

PR #1569, opened 2023-07-01
Enable Intercept function call

libaowei:
Hi, @lahma , I add a feature, enable intercept function call via configure Options.
My implementation is based on my understanding of jint, but I don't know if my implementation meets the requirements

### DIFF
--- a/Jint.Tests/Runtime/InteropTests.cs
+++ b/Jint.Tests/Runtime/InteropTests.cs
@@ -2,6 +2,7 @@
 using System.Globalization;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using Esprima.Ast;
 using Jint.Native;
 using Jint.Native.Object;
 using Jint.Native.Symbol;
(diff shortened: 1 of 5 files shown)


lahma:
Closing in favor of the CLR proxy participation API (`ProxyHandler` + `Engine.Advanced.CreateProxy`) in #2678. Rationale:

- The motivating scenario (#1468) is *per-object* interception — profiling function calls and property access on specific .NET objects handed to a script. ECMAScript already has the idiomatic construct for that: **Proxy**. The new API lets hosts implement traps in C# with no cost anywhere except on the proxied objects themselves, composes with `WrapObjectHandler` for automatic wrapping, and (unlike this PR's observe-only callbacks) can modify arguments, short-circuit, or substitute results.
- This PR's hook fires unconditionally inside `ScriptFunction.Call` — the hottest path in the engine, one that multiple performance campaigns have been spent on — and allocates a `Guid` per invocation even when unused. That cost profile isn't viable for a diagnostic feature, and the callbacks were also filed under `Options.Interop` although they intercept script functions rather than interop.
- Whole-program tracing genuinely can't be done with proxies, but for that need `DebugHandler` already exists: enable `DebugMode` and subscribe to `Step` events — opt-in, zero cost when disabled. If call-granular global tracing ever earns its place, it should be redesigned as a debugger-side observer (delegate cached in an engine field, ref-struct context over the existing call frame, no per-call allocation) — deliberately out of scope for the proxy work and best driven by concrete demand.

Thanks for the contribution and for surfacing the interception need — it directly shaped the requirements for #2678.

PR #1587, opened 2023-07-09
Readme Module section update

Zmarfan:
Changed from CreateModule to AddModule as it probably got renamed without changing readme. Also added a semi-colon which was missing.


lahma:
Thank you!

PR #1590, opened 2023-07-28
Bytes as ArrayBuffer

Scooletz:
This PR introduces a capability to convert raw byte chunks like `byte[]` as an `ArrayBuffer`. It does it by implementing an object converter directly in `Jint` core project. This is done to not make the `ArrayBufferInstance` public and keep the API change as minimal.

Open questions:

1. Why the test is failing? I tried to debug it and I fail to understand why an ArrayBuffer is not an instance of ArrayBuffer. I think it's not realm connected. I observed that Object.prototype.toString returns something different than expected as well. I'd appreciate some help.
2. Are we ok with introduction of a custom object converter in the core?
3. Are we ok to implement a similar approach for `Memory<byte>` which would require offset bookkeeping in the ArrayBuffer (I can implement it).


lahma:
> This PR introduces a capability to convert raw byte chunks like `byte[]` as an `ArrayBuffer`. It does it by implementing an object converter directly in `Jint` core project. This is done to not make the `ArrayBufferInstance` public and keep the API change as minimal.

Nice! Here's some feedback from the top of my head.

I think it would also be fine if we rename `ArrayBufferInstance` to more idiomatic `JsArrayBuffer` and expose it as public, generally things are internal if they haven't been API-checked for general consumption. We've already changed `DateInstance` to `JsDate` etc to get clean API. This would be for the case if you want to return from your c# code something like `engine.Realm.Intrinsics.ArrayBuffer.Construct(bytes);` (see below).

> 1. Why the test is failing? I tried to debug it and I fail to understand why an ArrayBuffer is not an instance of ArrayBuffer. I think it's not realm connected. I observed that Object.prototype.toString returns something different than expected as well. I'd appreciate some help.

The constructed object is not properly initialized, I'd suggest adding new `Construct` method to `ArrayBufferConstructor`:

```csharp
public ArrayBufferInstance Construct(byte[] data)
{
    var obj = OrdinaryCreateFromConstructor(
        this,
        static intrinsics => intrinsics.ArrayBuffer.PrototypeObject,
        static (engine, _, state) => new ArrayBufferInstance(engine, state!),
        data);

    return obj;
}
```

And changing your converter logic to:

```csharp
if (value is byte[] bytes)
{
    result = engine.Realm.Intrinsics.ArrayBuffer.Construct(bytes);
    return true;
}
```

> 2. Are we ok with introduction of a custom object converter in the core?

I think there might be small problem with ambiguity here. You could also want to have your `byte[]` to be converted to `Uint8Array` or `Uint8ClampedArray` - so automatic conversion during runtime might not be obvious, or one could want to have both possibilities (converting to both ArrayBuffer or Uint8Array).

> 3. Are we ok to implement a similar approach for `Memory<byte>` which would require offset bookkeeping in the ArrayBuffer (I can implement it).

Everything what makes Jint more usable in interop scenarios sounds good to me.

Scooletz:
Thank you for a prompt and in-depth reply @lahma . This is my first more serious work with Jint codebase so that I'm stepping into unfamiliar territory. Let me dive deeper with your comments in mind and get this PR altered over the weekend.

lahma:
With #1591 I did some preliminary work and did the rename and expose of type. If you find it useful for future, could also unseal it if there's some benefit with different backing store implementations.

sfmskywalker:
@lahma It looks like that if I can somehow instantiate the renamed `JsArrayBuffer` from my custom object converter, I would be good to go.

I'm also wondering if it would make sense to update `DefaultObjectConverter` with a type map between `byte[]` and `JsArrayBuffer`?

lahma:
I'll try to create a intermediate step PR to improve APIs to allow passing in existing items.

lahma:
@sfmskywalker please see #1890 and whether it would help you achieve your goals. The automatic `byte[]` to `ArrayBuffer` would be a breaking change as it would also mean all other number array types would/could get the same conversion.

sfmskywalker:
@lahma Your PR is spot-on to achieving my goals 🙌🏻  Thank you for the super fast turnaround!

Scooletz:
I don't have capacity to push it further. Should I close?

lahma:
I'll close this as I think the most important bits are now supported. Thanks for bringing this up and pushing the capabilities further.

PR #1596, opened 2023-07-29
Upgrade to Esprima 3.0.0-rc-03

lahma:


Jint/Native/RegExp/RegExpConstructor.cs
```diff
@@ -104,17 +104,14 @@ private ObjectInstance RegExpInitialize(JsRegExp r, JsValue pattern, JsValue fla
...
 
-                // seems valid
-                r.Value = scanner.ParseRegex(p, f, options.RegexTimeout);
-
-                var timeout = _engine.Options.Constraints.RegexTimeout;
-                if (timeout.Ticks > 0)
+                if (regExp is null)
                 {
-                    r.Value = new Regex(r.Value.ToString(), r.Value.Options, timeout);
+                    ExceptionHelper.ThrowSyntaxError(_realm, $"Could not parse regex '/{p}/{flags}'");
```
adams85, line 111:
```suggestion
                    ExceptionHelper.ThrowSyntaxError(_realm, $"Unsupported regular expression: '/{p}/{flags}'");
```

Jint/Runtime/Interpreter/Expressions/JintLiteralExpression.cs
```diff
@@ -76,7 +76,12 @@ private JsValue ResolveValue(EvaluationContext context)
             if (expression.TokenType == TokenType.RegularExpression)
             {
                 var regExpLiteral = (RegExpLiteral) _expression;
-                return context.Engine.Realm.Intrinsics.RegExp.Construct((System.Text.RegularExpressions.Regex) regExpLiteral.Value!, regExpLiteral.Regex.Pattern, regExpLiteral.Regex.Flags);
+                if (regExpLiteral.Value is not System.Text.RegularExpressions.Regex regex)
+                {
+                    ExceptionHelper.ThrowSyntaxError(context.Engine.Realm, $"Unsupported regular expression format '{regExpLiteral.Regex.Pattern}/{regExpLiteral.Regex.Flags}'");
```
adams85, line 81:
```suggestion
                    ExceptionHelper.ThrowSyntaxError(context.Engine.Realm, $"Unsupported regular expression: '{regExpLiteral.Regex.Pattern}/{regExpLiteral.Regex.Flags}'");
```

adams85 (commented):
I'm not sure whether we should also address this Unicode mode (flag u) problem:

> There could be false positive empty string matches in the middle of surrogate pairs. Patterns as simple as /a?/ have this issue when the input string contains astral Unicode chars. I didn't find a workaround for this but I'm pretty sure that, if it's exists, it's very complicated. This is a big problem but can be mitigated somewhat using a bit of "post-processing", that is, filtering out the false positive matches after evaluation like it's done [here](https://github.com/sebastienros/esprima-dotnet/blob/22eab845300564fcf756aab548ff635c58c4cd18/test/Esprima.Tests/RegExpTests.cs#L296). I'm afraid we can't do anything better than this until .NET adds the options to treat the input string as Unicode code points.

If the issue is not clear, just check the output of this expression: `[..."💩".matchAll(/i?/ug)]`. It should return only 2 matches (with index 0 and 2) but Jint will report 3 matches (an extra fake one with index 1) if we don't do something about it.

What do you think? Should we, as a best effort, remove such false positive matches during regex evaluation? I'm not familiar with the Jint internals but I suppose it could be done by adjusting `RegExpPrototype`.

lahma:
I think you are the best authority to assess this as you've gone so deep with regex parsing. Jint generally just tries its best with produced regexes. Many seem as quite advanced scenarios so not sure how often people will hit limitations.

Maybe after this has been merged you could check if you can work some magic on Jint side? No pressure as Jint probably isn't your main scenario to tweak Esprima 🙂

adams85:
Sure thing. I'll have a look soon if the situation can be improved.

PR #1603, opened 2023-07-30
Implement workarounds for regex parsing known issues

adams85:
Implements workarounds for [regex parsing known issues](https://github.com/sebastienros/esprima-dotnet/pull/364#issuecomment-1606045259).

Also, fixes [a bug in unicode mode regex handling](https://github.com/sebastienros/jint/pull/1603/files#diff-c07eb2081a7c5a8fa331c39b2e1d3dec9afa1c9807f1e131ec83629c115df5b2R945-R949) (code point -> code unit index translation is not needed as .NET regexes return code unit indices).

### DIFF
--- a/Jint/Native/RegExp/RegExpPrototype.cs
+++ b/Jint/Native/RegExp/RegExpPrototype.cs
@@ -942,11 +942,7 @@ namespace Jint.Native.RegExp
 
             var e = match.Index + match.Length;
 
-            // .NET regexes doesn't returns code point indices
-            //if (fullUnicode)
-            //{
-            //    e = GetStringIndex(s, e);
-            //}
+            // NOTE: Even in Unicode mode, we don't need to translate indices as .NET regexes always return code unit indices.
 
             if (global || sticky)
             {
@@ -955,35 +951,6 @@ namespace Jint.Native.RegExp
 
             return CreateReturnValueArray(R.Engine, matcher, match, s, fullUnicode, hasIndices);
         }
-
-        /// <summary>
-        /// https://tc39.es/ecma262/#sec-getstringindex
-        /// </summary>
-        //private static int GetStringIndex(string s, int codePointIndex)
-        //{
-        //    if (s.Length == 0)
-        //    {
-        //        return 0;
-        //    }
-
-        //    var len = s.Length;
-        //    var codeUnitCount = 0;
-        //    var codePointCount = 0;
-
-        //    while (codeUnitCount < len)
-        //    {
-        //        if (codePointCount == codePointIndex)
-        //        {
-        //            return codeUnitCount;
-        //        }
-
-        //        var isSurrogatePair = char.IsSurrogatePair(s, codeUnitCount);
-        //        codeUnitCount += isSurrogatePair ? 2 : 1;
-        //        codePointCount += 1;
-        //    }
-
-        //    return len;
-        //}
 
         private static JsArray CreateReturnValueArray(
             Engine engine,
@@ -1113,7 +1080,7 @@ namespace Jint.Native.RegExp
             {
                 try
                 {
-                    var bytes = Shims.BytesFromHexString(groupNameFromNumber.AsSpan(encodedGroupNamePrefix.Length));
+                    var bytes = groupNameFromNumber.AsSpan(encodedGroupNamePrefix.Length).BytesFromHexString();
                     groupNameFromNumber = Encoding.UTF8.GetString(bytes);
                 }
                 catch { /* intentional no-op */ }
--- a/Jint/Native/RegExp/RegExpPrototype.cs
+++ b/Jint/Native/RegExp/RegExpPrototype.cs
@@ -1,4 +1,5 @@
 ﻿using System.Diagnostics.CodeAnalysis;
+using System.Text;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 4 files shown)


Jint/Native/RegExp/RegExpPrototype.cs
```diff
@@ -902,31 +903,51 @@ private static JsValue RegExpBuiltinExec(JsRegExp R, string s)
...
+            if (!success)
             {
-                e = GetStringIndex(s, e);
+                R.Set(JsRegExp.PropertyLastIndex, JsNumber.PositiveZero, true);
+                return Null;
             }
 
+            var e = match.Index + match.Length;
+
+            // .NET regexes doesn't returns code point indices
```
lahma, line 945: maybe can remove? (though this was translated from spec I guess)

Jint/Native/RegExp/RegExpPrototype.cs
```diff
@@ -938,31 +959,31 @@ private static JsValue RegExpBuiltinExec(JsRegExp R, string s)
...
-                }
-
-                var isSurrogatePair = char.IsSurrogatePair(s, codeUnitCount);
-                codeUnitCount += isSurrogatePair ? 2 : 1;
-                codePointCount += 1;
-            }
-
-            return len;
-        }
+        //private static int GetStringIndex(string s, int codePointIndex)
```
lahma, line 962: maybe can remove this too?

lahma (commented):
Nice, thanks for helping out here! Seems that following test cases can also removed from `Test262Harness.settings.json` exclude now as they pass after the fixes:

```
"built-ins/RegExp/match-indices/indices-array-unicode-property-names.js",
"built-ins/RegExp/named-groups/non-unicode-property-names.js",
"built-ins/RegExp/named-groups/unicode-property-names.js",
"built-ins/RegExp/prototype/Symbol.replace/named-groups.js",
```

Jint/Native/RegExp/RegExpPrototype.cs
adams85: We can, I was just not sure whether to keep this piece of code around as it's an exact translation of the spec.

lahma:
> PR updated. There's a failing macos test but doesn't seem to have anything to do with my changes: 
> 
> ```
> [xUnit.net 00:00:01.75]     Jint.Tests.PublicInterface.PublicInterfaceTests.BindFunctionInstancesArePublic [FAIL]
> 
> Error: System.AggregateException : One or more errors occurred. (Unable to cast object of type 'Jint.Runtime.Environments.GlobalEnvironmentRecord' to type 'Jint.Runtime.Environments.FunctionEnvironmentRecord'.)
> ---- System.InvalidCastException : Unable to cast object of type 'Jint.Runtime.Environments.GlobalEnvironmentRecord' to type 'Jint.Runtime.Environments.FunctionEnvironmentRecord'.
> ```
> 
> Is this really something unrelated?

Unrelated. Engine is not thread safe and for some reason that test case is racy on macos.

lahma (approved):
Looks good. I guess we can merge?

adams85:
Yep, I think this one is ready.

However, I just discovered another issue:
* `"bab".replace(/.(?<a>a)(.)/, "$1$2")` results "ab", while
* `new Regex(".(?<a>a)(.)", RegexOptions.ECMAScript).Replace("bab", "$1$2")` results "ba".

It seems that capturing group numbering is messed up in .NET, at least it doesn't work in the same way as in JS (which assigns strictly increasing numbers to capturing groups, regardless of being a named group or not). This is a pretty sad situation and I'm not sure if we can do anything about it. I'll try to find a workaround but will do that in a separate PR (if it exists at all - but probably that will need to be fixed in Esprima...)

lahma:
It's almost like that `RegexOptions.ECMAScript` wasn't doing it's job 😉

lahma:
For some reason I cannot neither update or rebase this branch via GitHub UI, could you please do the rebase locally and force push?

lahma:
Thanks once again!

adams85:
Glad to help!

I did some investigation on capturing group numbering. According to MSDN, this "anomaly" is by design:

> [...] Captures that use parentheses are numbered automatically from left to right based on the order of the opening parentheses in the regular expression, starting from 1. However, [named capture groups](https://learn.microsoft.com/en-us/dotnet/standard/base-types/grouping-constructs-in-regular-expressions#named-matched-subexpressions) are always ordered last, after non-named capture groups. The capture that's numbered 0 is the text matched by the entire regular expression pattern.

Crazy design decision. I never thought I would say this, but in this regard JS is the sane one ... If only MS had got the ECMAScript compatibility mode right at least... But that ship has sailed for sure.

I can see only one viable workaround: in the regex parser we rewrite all named capturing groups to numbered ones (and, of course, named backrefs as well), like `/.(?<a>a)(.)\k<a>/` -> `@".(a)(.)\1".` Only problem is that Jint will (and other consumers may) need the group names, so we also need to return that information in some form (probably a group name -> group number lookup), besides the `Regex` instance. The regex parser gathers this information, so technically this workaround should be possible.

I think we can't get away without addressing this because this is a big issue. Without fixing this, we can't even call legacy regex handling okayish... So looks like we're in for another RC. I'll get back to you soon.

PR #1613, opened 2023-08-05
Support explicit interface and hidden member of super class

### LINKED ISSUE #1611: Support explicit interface
**Is your feature request related to a problem? Please describe.**
```c#
    public interface I1
    {
        string Value();
    } 

    public class CI1 : I1
    {
        public I1 I1 { get { return this; } }
        public CI1 Self { get { return this; } }

        string I1.Value()
        {
            return "CI1 as I1";
        }

        public string Value()
        {
            return "CI1";
        }

        public void Run()
        {
            System.Console.WriteLine(this.I1.Value());
            System.Console.WriteLine(this.Self.Value());
        }
    }
```

```js
var ci1 = new CI1();
ci1.I1.Value(); // CI1 in js, but CI1 as I1 in c#
ci1.Self.Value(); // CI1 in js
```

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

viruscamp:
Close #1611 Support explicit interface and hidden member of super class
Add field `_clrType` to `ObjectWrapper`
Add `unwrapClr` helper method

### DIFF
--- a/Jint.Tests/Runtime/InteropExplicitTypeTests.cs
+++ b/Jint.Tests/Runtime/InteropExplicitTypeTests.cs
@@ -1,0 +1,186 @@
+﻿using System.Reflection;
+
+namespace Jint.Tests.Runtime
+{
+    public class InteropExplicitTypeTests
+    {
+        public interface I1
+        {
+            string Name { get; }
+        }
+
+        public class Super
+        {
+            public string Name { get; } = "Super";
+        }
+
+        public class CI1 : Super, I1
+        {
+            public new string Name { get; } = "CI1";
+
+            string I1.Name { get; } = "CI1 as I1";
+        }
+
+        public class Indexer<T>
+        {
+            private readonly T t;
+            public Indexer(T t)
+            {
+                this.t = t;
+            }
+            public T this[int index]
+            {
+                get { return t; }
+            }
+        }
+
+        public class InterfaceHolder
+        {
+            public InterfaceHolder()
+            {
+                var ci1 = new CI1();
+                this.ci1 = ci1;
+                this.i1 = ci1;
+                this.super = ci1;
+
+                this.IndexerCI1 = new Indexer<CI1>(ci1);
+                this.IndexerI1 = new Indexer<I1>(ci1);
+                this.IndexerSuper = new Indexer<Super>(ci1);
+            }
+
+            public readonly CI1 ci1;
+            public readonly I1 i1;
+            public readonly Super super;
+
+            public CI1 CI1 { get => ci1; }
+            public I1 I1 { get => i1; }
+            public Super Super { get => super; }
+
+            public CI1 GetCI1() => ci1;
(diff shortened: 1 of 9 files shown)


lahma (approved):
Nice work and kudos for the great test coverage!

CreepGin:
I think this PR is making all `public extern` members null. In my case, it's making a lot of Unity APIs inaccessible now.

CreepGin:
The following Assertion will fail for this PR.

```cs
public class Container
{
    public Child Child => _child;

    Child _child = new Child();

    public BaseClass Get()
    {
        return _child;
    }
}

public class BaseClass
{
    public int index = 123;
}

public class Child : BaseClass { }

[Fact]
public void CustomTest()
{
    var engine = new Engine().SetValue("container", new Container());
    var res = engine.Evaluate(@"container.Child === container.Get()"); // These two should be the same object. But this PR makes `container.Get()` return a different object

    Assert.True(res.AsBoolean());
}
```

Also in the context of Unity, if Child and BaseClass contains `public extern` members, these will now be null if the object is retrieved via `container.Get()`.

viruscamp:
@CreepGin , I think this behavior is as design.  
Maybe we can improve it a little:
```c#
Assert.False(engine.Evaluate(@"container.Child === container.Get()")); // as design
Assert.True(engine.Evaluate(@"container.Child == container.Get()")); // TODO
```

BTW, it provides `unwrap` method:
```c#
Assert.True(engine.Evaluate(@"container.Child === clrHelper.unwrap(container.Get())"));
```

CreepGin:
So now the "wrapped" version of the object is making all the `public extern` members null. I'm assuming that's not by design right?

viruscamp:
@CreepGin 
```c#
public interface I1
{
  string MethodFromInterface();
}
public class C1 : I1
{
  public extern string MethodFromInterface();
  public extern string MethodInClass();

  public static I1 InterfaceInstance => new C1();
}
```
Prepare js tests: ```engine.SetValue("C1", TypeReference.CreateTypeReference(engine, typeof(C1)));```
```javascript
var i1 = C1.InterfaceInstance; // typed ObjectWrapper with type I1
i1.MethodFromInterface(); // should success, if you cannot call it, fire an issue
i1.MethodInClass(); // failed as design, you wanna this works?
clrHelper.unwrap(i1).MethodInClass(); // should success, if you cannot call it, fire an issue
clrHelper.wrap(i1, C1).MethodInClass(); // it's better than just unwrap, like explicit conversion in C# ((C1)i1).MethodInClass();
```

CreepGin:
```cs
i1.MethodInClass(); // failed as design, you wanna this works?
```

Yes. But maybe make explicit interface optional (via Jint.Options)? Or provide a way to disable it? (If this is already the case, please let me know; I must've missed it)

I do think the old behaviour should also be useful for a dynamic language like JS. (A good chunk of my codebase depends on it too)

PR #1614, opened 2023-08-05
Fix capturing group numbering bug and a few more regex-related issues

adams85:
Counterpart of https://github.com/sebastienros/esprima-dotnet/pull/392

Besides the issues mentioned in the linked PR, this one also fixes the fast path implementation of `RegExpPrototype.Split`, which is a bit broken.

Unfortunately, the remaining 2 test cases excluded [here](https://github.com/sebastienros/jint/compare/main...adams85:regexp-fixes?expand=1#diff-524efe3f200cee8c6ad344ec2f94e7f14b789909f63be3c1f6b6a9441ee7a5fe) cannot be fixed with our current regex handling approach due to the differences/limitations of the .NET regex engine.

Oh, and one more thing. I'm not sure about this: "https://github.com/adams85/jint/blob/95029c6cbbfc09e81749e39916841296a2e52f4d/Jint/Runtime/Interop/DefaultObjectConverter.cs#L31-L34" I'm afraid that passing a non-adapted, external `Regex` instance to Jint may lead to some pretty confusing behavior.

### DIFF
--- a/Jint.Tests.Test262/Test262Harness.settings.json
+++ b/Jint.Tests.Test262/Test262Harness.settings.json
@@ -44,14 +44,8 @@
     "language/literals/regexp/named-groups/forward-reference.js",
 
     // RegExp handling problems
-    "built-ins/RegExp/named-groups/non-unicode-match.js",
-    "built-ins/RegExp/named-groups/non-unicode-property-names-valid.js",
-    "built-ins/RegExp/named-groups/unicode-match.js",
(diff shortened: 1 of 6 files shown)


adams85:
Just tested with the newly released Esprima version, fixes work as expected. So, as for me, PR is ready for merge.

lahma:
Looking good, thanks for the help 👍🏻

PR #1616, opened 2023-08-08
More improvements for CLR Interop

viruscamp:
- Simplize nested CLR Type, can be got directly from parent type
- Add more CLR helper functions: clrUnwrap, clrWrap, clrToString, clrType, clrTypeToObject, clrObjectToType. The last 3 need Options.Interop.AllowGetType == true.
- Make CLR objects ToString meaningful: NamespaceReference, TypeReference, ClrFunctionInstance, MethodInfoFunctionInstance
- Make Jint.Repl call ToString when Serialize failed

### DIFF
--- a/Jint/Options.cs
+++ b/Jint/Options.cs
@@ -120,39 +120,9 @@ namespace Jint
                         (thisObj, arguments) =>
                             new NamespaceReference(engine, TypeConverter.ToString(arguments.At(0)))),
                     PropertyFlag.AllForbidden));
-                engine.Realm.GlobalObject.SetProperty("clrToString", new PropertyDescriptor(new ClrFunctionInstance(
-                        engine,
-                        "clrToString",
-                        ClrHelper.ClrToString),
+                engine.Realm.GlobalObject.SetProperty("clrHelper", new PropertyDescriptor(
+                    new ObjectWrapper(engine, new ClrHelper(Interop)),
                     PropertyFlag.AllForbidden));
-                engine.Realm.GlobalObject.SetProperty("clrUnwrap", new PropertyDescriptor(new ClrFunctionInstance(
-                        engine,
-                        "clrUnwrap",
-                        ClrHelper.ClrUnwrap),
-                    PropertyFlag.AllForbidden));
-                engine.Realm.GlobalObject.SetProperty("clrWrap", new PropertyDescriptor(new ClrFunctionInstance(
-                        engine,
-                        "clrWrap",
-                        ClrHelper.ClrWrap),
-                    PropertyFlag.AllForbidden));
-                if (Interop.AllowGetType)
-                {
-                    engine.Realm.GlobalObject.SetProperty("clrType", new PropertyDescriptor(new ClrFunctionInstance(
-                            engine,
-                            "clrType",
-                            ClrHelper.ClrType),
-                        PropertyFlag.AllForbidden));
-                    engine.Realm.GlobalObject.SetProperty("clrTypeToObject", new PropertyDescriptor(new ClrFunctionInstance(
-                            engine,
-                            "clrTypeToObject",
-                            ClrHelper.ClrTypeToObject),
-                        PropertyFlag.AllForbidden));
-                    engine.Realm.GlobalObject.SetProperty("clrObjectToType", new PropertyDescriptor(new ClrFunctionInstance(
-                            engine,
-                            "clrObjectToType",
-                            ClrHelper.ClrObjectToType),
-                        PropertyFlag.AllForbidden));
-                }
             }
 
             if (Interop.ExtensionMethodTypes.Count > 0)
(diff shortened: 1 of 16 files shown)


Jint/Options.cs
```diff
@@ -120,22 +120,39 @@ internal void Apply(Engine engine)
...
-                        if (arg is ObjectWrapper obj)
-                        {
-                            return new ObjectWrapper(engine, obj.Target);
-                        }
-                        else
-                        {
-                            return arg;
-                        }
-                    }),
+                engine.Realm.GlobalObject.SetProperty("clrToString", new PropertyDescriptor(new ClrFunctionInstance(
```
lahma, line 123: as the function name already changes, would it make sense to inject and object like `clrHelper` which would have these functions exposed? it could take `allowGetType` when constructed and throw error when `clrType` etc are being called. Usage would then be

`const unwrapped = clrHelper.unwrap(myObj);`

Jint/Runtime/Interop/ClrHelper.cs
```diff
@@ -0,0 +1,64 @@
+﻿using Jint.Native;
+
+namespace Jint.Runtime.Interop
```
lahma, line 3: for new files, can also use file-scoped namespaces, at some point when there's not so many pending PRs a mass change will be done towards file-scoped namespaces

Jint/Runtime/Interop/ClrHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    public static class ClrHelper
```
lahma, line 5: this can probably be internal? see also earlier comment about being instantiable and having `allowGetType` as constructor parameter

Jint/Runtime/Interop/MethodInfoFunctionInstance.cs
```diff
@@ -9,19 +9,18 @@ namespace Jint.Runtime.Interop
 {
     internal sealed class MethodInfoFunctionInstance : FunctionInstance
     {
-        private static readonly JsString _name = new JsString("Function");
         private readonly MethodDescriptor[] _methods;
         private readonly ClrFunctionInstance? _fallbackClrFunctionInstance;
 
-        public MethodInfoFunctionInstance(Engine engine, MethodDescriptor[] methods)
-            : base(engine, engine.Realm, _name)
+        public MethodInfoFunctionInstance(Engine engine, MethodDescriptor[] methods, string name)
```
lahma, line 15: nit: name feels more natural to come before methods

Jint/Runtime/Interop/ObjectWrapper.cs
```diff
@@ -35,6 +34,7 @@ public ObjectWrapper(Engine engine, object obj, Type? type = null)
         }
 
         public object Target { get; }
+        public Type ClrType { get; }
```
lahma, line 37: do we need this to be `public` or is `internal` enough?

Jint/Runtime/Interop/Reflection/NestedTypeAccessor.cs
```diff
@@ -0,0 +1,23 @@
+
+namespace Jint.Runtime.Interop.Reflection
+{
+    internal sealed class NestedTypeAccessor : ReflectionAccessor
+    {
+        private readonly TypeReference _typeReference;
+        public NestedTypeAccessor(TypeReference typeReference, string name) : base(typeof(Type), name)
```
lahma, line 7: nit: should have new line before constructor

Jint/Runtime/Interop/TypeReference.cs
```diff
@@ -346,5 +346,10 @@ private static JsValue HasInstance(JsValue thisObject, JsValue[] arguments)
 
             return derivedType != null && baseType != null && (derivedType == baseType || derivedType.IsSubclassOf(baseType));
         }
+
+        public override string ToString()
+        {
+            return "[clr type: " + ReferenceType + "]";
```
lahma, line 352: should it be upper-case `CLR`?

lahma:
Nice work and appreciate you improving the interop story! Added some review comments, please take a look. Branch also needs a rebase as it's a merge requirement that all PRs are against latest `main` before merge.

Jint/Runtime/Interop/MethodInfoFunctionInstance.cs
viruscamp: I will add a `Type targetType` to `MethodInfoFunctionInstance`.
Would you like the arguments sequence?
```c#
public void MakeProperty(
  Engine engine,
  Type targetType,
  object target,
  MemberInfo memberInfo,
  string memberName,
  MethodDescriptor[] methods,
  ClrFunctionInstance? fallbackClrFunctionInstance = null
) {}
```

lahma: It doesn't matter that much as types are internal, we can always shuffle order later on without breaking API

Jint/Runtime/Interop/ClrHelper.cs
viruscamp: I will make the class public but with an internal constructor.

Jint/Options.cs
viruscamp: Almost done except Tests.

Jint/Runtime/Interop/ClrHelper.cs
viruscamp: I have made `ClrHelper` having `InteropOptions` as constructor parameter, to make it possible to change `Engine.Options.Interop.AllowGetType` after Engine created.

lahma (approved):
Great work, thank you!

Issue #1637, opened 2023-09-19
error No public methods with the specified arguments were found


hyzx86:
**Version used**
 [3.0.0-beta-2052](https://www.nuget.org/packages/Jint/3.0.0-beta-2052)
**Describe the bug**

Hi ,

I used Jint and [FreeSql ](https://github.com/dotnetcore/FreeSql/) to encapsulate the changes of different databases and operate on the data. Recently, I encountered a problem.
Update from beta-2049 to beta-2052
An exception is thrown. `No public methods with the specified arguments were found. `

See the last line for an example

```c#
  var db = await new FreeSqlMethodsProvider().UseDbAsync(scope.ServiceProvider);
  var total = db.SelectTypedQuery("UserProfile").Where("1=1").Any();

  engine.SetValue("objSelect", objSelect);
  var result1 = engine.Evaluate("objSelect.where('1=1')");//pass
  engine.SetValue("result1", result1);
  var result2 = engine.Evaluate("result1.any()");//error No public methods with the specified arguments were found
```

ISelect<object> SelectTypedQuery(string)

 **Where Method:**

```C#
        /// <summary>
        /// 原生sql语法条件，Where("id = @id", new { id = 1 })<para></para>
        /// 提示：parms 参数还可以传 Dictionary&lt;string, object&gt;
        /// </summary>
        /// <param name="sql">sql语法条件</param>
        /// <param name="parms">参数</param>
        /// <returns></returns>
        TSelect Where(string sql, object parms = null);
```
https://github.com/dotnetcore/FreeSql/blob/be6f2fa9c2cc04ce5bf01fc38d2b01a493021088/FreeSql/Interface/Curd/ISelect/ISelect0.cs#L291-L298

**Any Method:**

```c#
        /// <summary>
        /// 执行SQL查询，是否有记录
        /// </summary>
        /// <returns></returns>
        bool Any();
```
https://github.com/dotnetcore/FreeSql/blob/be6f2fa9c2cc04ce5bf01fc38d2b01a493021088/FreeSql/Interface/Curd/ISelect/ISelect0.cs#L195-L200

When I debug in jint, I find that I want to call `bool Any(); `

But always call to`  bool Any(Expression<Func<T1, bool>> exp); `
And it's the only one of the methods

```c#
/// <summary>
        /// 执行SQL查询，是否有记录
        /// </summary>
        /// <param name="exp">lambda表达式</param>
        /// <returns></returns>
[...]

PR #1638, opened 2023-09-19
fix interface methods

hyzx86:
fix https://github.com/sebastienros/jint/issues/1637


Genteure:
I'm not familiar with this area so I can't comment on whether the fix itself is appropriate or not.

Please use simple loops instead of LINQ as LINQ use much more memory and CPU cycles compared to simple for loops.
Please consider adding tests to verify your changes and to prevent regression.

hyzx86:
@Genteure , That object is too complicated. So I don't know how to instantiate it....

```C#
using System.Linq.Expressions;

namespace Jint.Tests.Runtime.TestClasses
{
    public class InterfaceTestModel
    {
        public int Age { get; set; }
    }



    public class InterfaceTest<TSelect, T1> : ISelect0<TSelect, T1>
    {
        public bool Any()
        {
            return true;
        }
    }


    public interface ISelect0 { }
    public partial interface ISelect0<TSelect, T1> : ISelect0
    {
        bool Any();
    }

    public interface ISelect<T1> : ISelect0<ISelect<T1>, T1>
    {
        bool Any(Expression<Func<T1, bool>> exp);

    }
}

```

hyzx86:
@2881099 ，Can you help me see how this thing can be instantiated?

hyzx86:
Hi @lahma ,Could you please look at this PR? Thank you.

I don't know why PR Check tasks don't pass to windows tasks,

But on my computer, I passed most of the unit tests and failed only a few that involved language and culture.
image

viruscamp:
@hyzx86 , I provide you the tests. But there is still something unresolved, maybe you should mark `CallSubPropertySuperMethod_SubProperty` and `CallSubPropertySuperMethod_SuperMethod` as `[Fact(Skip = "TODO")]`.

hyzx86:
> `CallSubPropertySuperMethod_SubProperty`

Hi @viruscamp , Thanks for your support ,  I have done the `UnitTest` , 
But I added precompiled instructions at the beginning and end of the class file to solve the following problem.
"The default interface implementation is not supported by the target runtime."

viruscamp:
I wrote the test in Linux, so I wrote interface default methods which won't run in net462.

Just copy the 5 methods with body in `I0` and `I1` to class `C1`,  add `I1.` or `I0.` to methods name, and then remove the methods body `=>"..."` in interfaces.

hyzx86:
@viruscamp  I have created a new file as Jint.Tests/Runtime/InterfaceTests.cs

I ran the 4 tests you mentioned on my computer and they all passed ,Whether it's net462 or net6.0

image
image

Jint.Tests/Runtime/InterfaceTests.cs
viruscamp, line 1: Would you like to rename to a more clear name like `InteropInterfaceExtendTests`?

hyzx86: Done

Jint/Native/WeakMap/JsWeakMap.cs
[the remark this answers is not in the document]
hyzx86: Done

viruscamp (approved):
OK

hyzx86:
Hi @lahma ,Could you please review this PR? Thank you.

lahma:
Thank you.

PR #1642, opened 2023-10-01
Fixed rest/spread bug with empty object

CreepGin:
Fixed the NullReferenceException for the following code:
```js
function test({ ...props }) { }
test({})
```


lahma (approved):
Nice catch 👍🏻

PR #1645, opened 2023-10-12
Proxy on clr

### LINKED ISSUE #1643: JS Proxy on Clr Object
Hi,
I am using Jint 3.0.0-beta-2052. I am testing the Js Proxy where I get exceptions if I access a property that exists on a clr object that the proxy was created on. If the property does not exist on the clr object, everything works as expected.
Also everything works fine if the proxy is created on a class created in JS instead of a clr class.

Example

```C#
class TestClass
{
     public string Name => "My Name is Test";
}

```

```C#
new Engine()
    .SetValue("T", new TestClass())
    .Execute("""
        const handler = {
            get(target, property, receiver) {

                if (!target[property]) {
                    return (...args) => "Not available";
                }

                // return target[property];
                return Reflect.get(target, property, receiver);
            }
        };

        const p = new Proxy(T, handler);
        const t = p.GetX();
        // const name = p.Name; // When commenting this in, an exception is thrown
    """);
```

Am I doing something wrong here or did I understand something wrong? Or isn't it supported to use a Proxy on a clr object?

viruscamp:
fix #1643 

provide detailed message for exceptions in JsProxy
donot use ReferenceEquals on data property value comparing

### DIFF
--- a/Jint.Tests/Runtime/ProxyTests.cs
+++ b/Jint.Tests/Runtime/ProxyTests.cs
@@ -1,4 +1,7 @@
-﻿namespace Jint.Tests.Runtime;
+﻿using Jint.Native.Error;
+using Jint.Runtime;
+
+namespace Jint.Tests.Runtime;
 
 public class ProxyTests
 {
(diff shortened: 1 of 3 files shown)


viruscamp:
Other exceptions in JsProxy should be attached messages, and make tests. It's so boring, I won't do it recently.  

Next, I may rewrite `PropertyDescriptor.IsDataDescriptor`, as currently CLR field property method and etc are treat as data property, in JsProxy it triggers multiple property getter calls.

Jint/Native/Proxy/JsProxy.cs
```diff
@@ -134,25 +135,55 @@ public override JsValue Get(JsValue property, JsValue receiver)
...
+                    {
+                        ExceptionHelper.ThrowTypeError(_engine.Realm, $"'get' on proxy: property '{property}' is a non-configurable accessor property on the proxy target and does not have a getter function, but the trap did not return 'undefined' (got '{result}')");
+                    }
                 }
             }
 
             return result;
         }
 
+        private static bool IsSameValue(JsValue expected, JsValue actual)
```
lahma, line 159: there's already `JsValue.IsSameValue` method that probably should be used, it doesn't handle `ObjectWrapper` though

Jint/Native/Proxy/JsProxy.cs
```diff
@@ -134,25 +135,55 @@ public override JsValue Get(JsValue property, JsValue receiver)
...
+        {
+            if (ReferenceEquals(expected, actual))
+            {
+                return true;
+            }
+            if (expected.GetType() != actual.GetType())
```
lahma, line 165: this probably wouldn't work correctly for `ConcatenatedString` vs `JsString`, hence the `JsValue.IsSameValue`

lahma (changes requested):

Jint.Tests/Runtime/ProxyTests.cs
```diff
@@ -183,4 +186,218 @@ public void ConstructHandlerInvariant()
...
+            });
+            const handler = {
+              get(target, property, receiver) {
+                return 32;
+              }
+            };
+            let p = new Proxy(o, handler);
+        """);
+        var ex = Assert.Throws<JavaScriptException>(() => _engine.Evaluate("p.value"));
+        AssertJsTypeError(_engine, ex, "'get' on proxy: property 'value' is a read-only and non-configurable data property on the proxy target but the proxy did not return its actual value (expected '42' but got '32')");
```
viruscamp, line 294: Should we comment all `AssertJsTypeError`?  
As it depends internal implementations and unstable error messages.  
Just comment out these for develop usage.

lahma: I think you did great work replicating the V8 error messages and they can stay, we can always tweak later if comes a maintenance burden 👍🏻

lahma (approved):
Great work! 👍🏻

PR #1652, opened 2023-10-21
Remove exposed CLR type equality requirement from ObjectWrapper.Equals

lahma:
I can't wrap my head around why we would need to have `Equals` to check object wrapper's exposed type as current `ReferenceEquals` is already a strong guarantee. There might be cases to only expose member of exposed type but we shouldn't state that pointer to same object isn't the same object based on what was exposed.

This should return the behavior expected with @CreepGin 's Unity integration.

/cc @CreepGin @viruscamp

### DIFF
--- a/Jint.Tests/Runtime/InteropExplicitTypeTests.cs
+++ b/Jint.Tests/Runtime/InteropExplicitTypeTests.cs
@@ -87,10 +87,7 @@ public class InteropExplicitTypeTests
     public void EqualTest()
     {
         Assert.Equal(_engine.Evaluate("holder.I1"), _engine.Evaluate("holder.i1"));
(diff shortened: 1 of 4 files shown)


viruscamp:
Actually, It's a design choice.  
I selected type `Equals`, because even same CLR object with different type may have different behavior.  
Additional I think it should use `==` to check without type, and `===` to check with type.

lahma:
In .NET `ReferenceEquals` always wins and this is interop, in JavaScript there's no such comparison as the inheritance model is just so different. So I guess this change of behavior back to original is still a design choice too? For me it makes more sense.

lahma:
I'll merge this to create a new release, we can iterate based on more feedback if needed.

PR #1666, opened 2023-10-30
Fix parsing of empty date string

### LINKED ISSUE #1665: Fails to parse a date with blank value
**Version used**

3.0.0-beta-2053

**Describe the bug**

It fails to run `Date.parse('')`

image

**To Reproduce**

Evaluate the following:
```js
Date.parse('')
```

**Expected behavior**

It should return `NaN`

lahma:
fixes #1665


sebastienros:
Surprising there is no ecma test for this.

lahma:
I was quite surprised too, but it's usually the easy ones - you just need to handle negative string pointer to deleted prototype that's bound to another realm's error constructor inside late bound function.

PR #1685, opened 2023-11-18
Code changes to keep NET 8 analyzers happy

lahma:


Jint/Native/Global/GlobalObject.cs
```diff
@@ -605,7 +605,7 @@ public JsValue Escape(JsValue thisObject, JsValue[] arguments)
             for (var k = 0; k < strLen; k++)
             {
                 var c = uriString[k];
-                if (WhiteList.IndexOf(c) != -1)
+                if (AllowList.Contains(c))
```
sebastienros, line 608: This list can be ordered, so we should be able to use a binary search instead of Contains.

We also need to use SearchValues in some places, probably mostly in Esprima. But that means a net8.0 target.

lahma: Added naive shim for this https://github.com/lahma/jint/blob/6553ee63a87139a8da0e0fbda3f2bfa6a8a6e9ef/Jint/Extensions/SearchValues.cs so can use with both NET 8 and older.

PR #1691, opened 2023-11-24
Update benchmark results for NET 8

lahma:


SebastianStehle:
It got up to 50% faster just because of .NET 8?

sebastienros:
Changes + jump from .net6, didn't realize it was not on 7 before.

lahma:
There might be some usual fluctuation caused by *something* while running the tests (please try running the suite on your machine!), but Jint tries its best to use CLR smarts and should benefit like any other using plain old code. `SeachValues` shim introduced here shouldn't show up in benchmarks AFAIK.

SebastianStehle:
I know that you are doing a awesome job :)

lahma:
I'm happy to see Jint faring so well, but the results should show a balance between things one needs from the engine. Jint's strong suites are running simple scripts fast and being *conformant to the spec*. I'd argue others might have trouble fulfilling the spec requirements and at the same time there are few consumers that would actually need to worry about such weird edge cases  / conformance.

PR #1693, opened 2023-11-26
Adding cultural check to the "ToUpperCase" string prototype.

LuisMerinoP:
Adding some easy win test fixes. 
Before:
image
With fix:
image

Comments:

- Lithuanian case is quite tricky. Only fixed the tests. Maybe there is some more work to do to check larger texts are upper-case-ed correctly.
- Maybe some very specific functionality in internationalization (such as `LithuanianStringProcessor`) needs its place in the project, kind of some utils folder or similar. Please advice if that needs to be moved somewhere else. 
- I only wanted to fix "ToLocaleUpperCase" tests because it seemed an easy win. Apparently, more tests get fixed. Don't know why and I did not dig into that.

### DIFF
--- a/Jint/Native/String/StringInlHelper.cs
+++ b/Jint/Native/String/StringInlHelper.cs
@@ -1,0 +1,55 @@
+﻿using System.Text;
+
+namespace Jint.Native.String
+{
+    /// <summary>
+    /// Some internacionalization logic that is special or specific to determined culture.
+    /// </summary>
+    internal class StringInlHelper
+    {
+        private static List<int> GetLithuaninanReplaceableCharIdx(string input)
+        {
+            List<int> replaceableCharsIdx = new List<int>();
+            for (int i = 0; i < input.Length; i++)
+            {
+                if (input[i].Equals('\u0307'))
+                {
+                    replaceableCharsIdx.Add(i);
+                }
+            }
+
+            // For capital I and J we do not replace the dot above (\u3017).
+            replaceableCharsIdx
+                .RemoveAll(idx => (idx > 0) && input[idx - 1] == 'I' || input[idx - 1] == 'J');
+
+            return replaceableCharsIdx;
+        }
+
+        /// <summary>
+        /// Lithuanian case is a bit special. For more info see:
+        /// https://github.com/tc39/test262/blob/main/test/intl402/String/prototype/toLocaleUpperCase/special_casing_Lithuanian.js
+        /// </summary>
+        public static string LithuanianStringProcessor(string input)
+        {
+            var replaceableCharsIdx = GetLithuaninanReplaceableCharIdx(input);
+            if (replaceableCharsIdx.Count > 0)
+            {
+                StringBuilder stringBuilder = new StringBuilder(input);
+
+                // Remove characters in reverse order to avoid index shifting
+                for (int i = replaceableCharsIdx.Count - 1; i >= 0; i--)
+                {
+                    int index = replaceableCharsIdx[i];
+                    if (index >= 0 && index < stringBuilder.Length)
+                    {
+                        stringBuilder.Remove(index, 1);
+                    }
+                }
+
+                return stringBuilder.ToString();
+            }
+
+            return input;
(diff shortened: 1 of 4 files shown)


LuisMerinoP:
> But it seems there are other edge cases for tr, el and az

All but "el" cases are fixed on this PR, because there are tests for those: https://github.com/tc39/test262/tree/main/test/intl402/String/prototype/toLocaleUpperCase.

Greeks will need to wait :).

lahma:
What worries me here is that this seems to just target the test case. Like input `"Hello I\u0307".toLocaleUpperCase("lt")` won't behave the same way with Jint and NodeJS.

LuisMerinoP:
Yeah totally. Thanks for noticing. I focused on fixing the tests only that are character based and the Lithuanian case took all of my bandwidth, lol. That is what I meant with the 1st point of the PR description.
I will check the behaviour for larger strings and fix when I find the time :)

LuisMerinoP:
Arranged the character filtering for larger strings. I can add some tests in [StringTests](https://github.com/sebastienros/jint/blob/main/Jint.Tests/Runtime/StringTests.cs) to verify corner cases and quirky scenarios provide the same output as node if that is useful. Let me know if that sounds good.
On the other hand, macos check is failing, don't know why.
Cheers :).

LuisMerinoP:
> The more tests the better

Perfect, I'll add some.

LuisMerinoP:
Adding tests for the string processing for all the special character cases.

lahma:
So is this ready for merge or still some work pending?

LuisMerinoP:
According to me, its ready for review and eventual merge :)

Jint/Native/String/StringInlHelper.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System.Text;
+
+namespace Jint.Native.String
```
lahma, line 3: nit: for new files should use file-scoped namespaces

Jint/Native/String/StringInlHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+    /// <summary>
+    /// Some internacionalization logic that is special or specific to determined culture.
+    /// </summary>
+    internal class StringInlHelper
```
lahma, line 8: nit: `internal` types should be either sealed or static

Jint/Native/String/StringInlHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <summary>
+        /// Lithuanian case is a bit special. For more info see:
+        /// https://github.com/tc39/test262/blob/main/test/intl402/String/prototype/toLocaleUpperCase/special_casing_Lithuanian.js
+        /// </summary>
+        public static string LithuanianStringProcessor(string input)
+        {
+            var replaceableCharsIdx = GetLithuaninanReplaceableCharIdx(input);
+            if (replaceableCharsIdx.Count > 0)
+            {
+                StringBuilder stringBuilder = new StringBuilder(input);
```
lahma, line 37: we have ValueStringBuilder for building these

Jint/Native/String/StringPrototype.cs
```diff
@@ -231,7 +232,32 @@ private JsValue ToLocaleUpperCase(JsValue thisObject, JsValue[] arguments)
         {
             TypeConverter.CheckObjectCoercible(_engine, thisObject);
             var s = TypeConverter.ToString(thisObject);
-            return new JsString(s.ToUpper(CultureInfo.InvariantCulture));
+            var culture = CultureInfo.InvariantCulture;
+            if (arguments.Length > 0 && arguments[0].IsString())
```
lahma, line 236: migth be cleaner to use `arguments.At(0) is JsString s`

lahma (approved):
Added some remarks, but as not in hot path we don't need to worry about them that much.

lahma:
Thank you for your contribution! At some point probably would be a good idea to enable some tests against these features.

PR #1695, opened 2023-11-30
Return old recursion limit capabilities

lahma:
We can now actually do better than the old limit `960` which was before NET 8


lahma:
I think it's probably the cold start scenario in tests at least. Benchmarks showed that NET 8 made Jint faster after all. Need to use magic number as older targets don't know about this enum member.

PR #1701, opened 2023-12-11
Add GetUtcNow to ITimeSystem

### LINKED ISSUE #1700: Support .NET 8 TimeProvider for virtualized time
I want to be able to control the current date / time inside of scripts. There is a new `TimeProvider` in .NET 8 that I am making use of to control time and I want my scripts to use that as well.

https://learn.microsoft.com/en-us/dotnet/api/system.timeprovider?view=net-8.0

ejsmith:
Fixes #1700

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -10,6 +10,7 @@
     <PackageVersion Include="Jurassic" Version="3.2.7" />
     <PackageVersion Include="Meziantou.Analyzer" Version="2.0.116" />
     <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="7.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.TimeProvider.Testing" Version="8.0.0" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.8.0" />
     <PackageVersion Include="MongoDB.Bson.signed" Version="2.19.0" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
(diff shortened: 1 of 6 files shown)


lahma:
I think the current implementation should suffice quite nicely. `TimeProvider`  is a bit problematic as it's a NET 8 feature and would require separate NuGet dependency added to get the feature to older targets (https://www.nuget.org/packages/Microsoft.Bcl.TimeProvider). This is why I would implement the support on consumer code instead of the library (it's very limited set of features where Jint needs current time as you can see).

Maybe a test case would be a nice addition showcasing the new possibilities?

lahma:
The `TimeProvider` uses `DateTimeOffset` as return value though: https://learn.microsoft.com/en-us/dotnet/api/system.timeprovider.getutcnow?view=net-8.0#system-timeprovider-getutcnow

The real use cases for Jint are for the timestamp I guess https://learn.microsoft.com/en-us/dotnet/api/system.timeprovider.gettimestamp?view=net-8.0#system-timeprovider-gettimestamp as we only need to subtract to get the different in milliseconds.

sebastienros:
> TimeProvider is a bit problematic as it's a NET 8 feature

No, it's been backported down to full framework. Will investigate why the doc says .NET 8 only.

sebastienros:
Oh, different assemblies, yes, probably.

ejsmith:
Yeah, might be nice to have the full TimeProvider available and could eventually be used to virtual timers, timeouts and such as well as providing current time. I was just trying to keep this PR simple.

ejsmith:
Added a test showing how to integrate TimeProvider

lahma:
Jint is is using central package management so you need to add the new dependency to `Directory.Packages.props` in root folder and reference it without version number in csproj. Can you also change the return type in `ITmeProvider` to DateTimeOffset? I know it doesn't matter for UTC but if we would ever add the `GetLocalNow()` we would be interested in offset part to and the methods would have parity (also parity with `TimeProvider`)

ejsmith:
@lahma I'm confused by the build error saying it can't target .NET 8. I didn't change any framework version. This test project was already targeting `.net8`.

lahma:
It's OK,  macos runners are lagging behind with net8.0 support.

lahma (approved):
Looks good, thank you!

ejsmith:
Thank you!

PR #1715, opened 2024-01-03
Debug StepInto stops at beginning of arrow function

tomatosalat0:
When using `StepMode.Into`, the debugger callback `DebugHandler.Step` did not get fired before the execution of a single expression arrow function.

For example, having this statement
```js
[1, 2, 3].map(v => v * 2)
```
The debugger would not stop before the `v * 2` expression - only after.

In contrast, having this statement
```js
[1, 2, 3].map(function(v) { return v * 2; })
```
The debugger would stop before and after the `return v * 2` expression.

Two tests were added, one for the `function` type and one for the `arrow` type.


lahma (approved):
LGTM, thank you!

PR #1722, opened 2024-01-05
Remove Engine.Realm, expose Engine.Intrinsics and Engine.Global directly

lahma:


viceice:
This change is casuing trouble, i can no longer extend `Function` because the only accessable constructor needs a `Realm` 😕 

It should probably changed to remove the explicit realm and use it from engine implicit?

viceice:
ok, found a workaround by using `ClrFunction` directly 🤗

PR #1727, opened 2024-01-05
Fix dynamic function names

scgm0:
It should be "anonymous" instead of "f".


lahma:
Can you please also add a test case?

scgm0:
> Can you please also add a test case?

This needs to be merged first #1726

lahma:
Let's wait then.

lahma:
fixed by #2394

PR #1731, opened 2024-01-08
Refactoring for custom module loader support

tomatosalat0:
# Motivation

To support fully custom `IModuleLoader` implementations, some code from the abstract `ModuleLoader` class has been moved into the public static `ModuleFactory` class.

With these public methods, custom `IModuleLoader` implementations can be created and used without deriving from the abstract class `ModuleLoader`. This allows custom code to implement certain features like a cache for parsed `Esprima.Ast.Module` instances which can be shared across multiple engine instances. The additional methods allow to use a different source specifier for modules instead of the default `Uri.LocalPath` variant used by the default implementation.

The behavior of the existing `DefaultModuleLoader` has not been changed.

# Implementation

The static class allows to create `Jint.Runtime.Modules.Module` instances without publishing the actual implementations like `SourceTextModule` or `SyntheticModule`, so these implementations can still be changed without affecting the public API.

### DIFF
--- a/Jint.Tests.PublicInterface/ModuleLoaderTests.cs
+++ b/Jint.Tests.PublicInterface/ModuleLoaderTests.cs
@@ -1,0 +1,264 @@
+﻿using System.Collections.Concurrent;
+using Jint.Native;
+using Jint.Native.Json;
+using Jint.Runtime.Modules;
+
+#nullable enable
+
+namespace Jint.Tests.PublicInterface;
+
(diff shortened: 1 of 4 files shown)


lahma (approved):
Splendid work, thank you for improving the public API promise and tweaking the logic!

sebastienros:
Looks like the project got itself a new MVP. Thanks @tomatosalat0 for you recent PRs!

PR #1739, opened 2024-01-13
Updates for v3 RTM

lahma:
* Update README.md

As generators and tweaks seem to be a forever project, might be easier to finally release the v3. I've gone through API from consuming viewpoint and I think it's quite near of not exposing too much and allows wiggle room in development - and there's always v4...

A week or so to ensure latest release doesn't break too much, or a year.. maybe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,11 +8,7 @@
 
 Jint is a __Javascript interpreter__ for .NET which can run on __any modern .NET platform__ as it supports .NET Standard 2.0 and .NET 4.6.2 targets (and later).
 
-💡 You should prefer 3.x beta versions from NuGet over the 2.x. All new features and improvements are targeted against version 3.x.
-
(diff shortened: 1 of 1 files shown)


sebastienros (approved):

lahma:
As @sebastienros basically gave us the go-live license, why not 🚀

sebastienros:
Do you think your unfinished generators PR is not worth it in its current state. As in better not do anything than not support all quircks?

lahma:
I think it can wait. Basically it suffers from the problem that it's now impossible from resume from *expression* state. Statements are easier now when it's just part of a list, but expressions are inside of a complex AST tree. The public APIs won't be affected from this when/if implemented.

lahma:
This is what you get from Duke Nukem projects

image

PR #1740, opened 2024-01-13
Engine execute module resolving

tomatosalat0:
# Problem

Non-module scripts can still import modules using the `import()` function. However, `Engine.Execute` and `Engine.Evaluate` always passed `string.Empty` as the source of the current script even if one was specified. 

# Change
With the change, `ScriptRecord` will get the value of `Script.Location.Source` as the source. Tests for `Engine.Execute` and `Engine.Evalute` with a custom `IModuleLoader` checking for a filled `referencingModuleLocation` were added.

The behavior for `Engine.Execute` without passing a `source` has not been changed (source is set to be `<anonymous>`)

### DIFF
--- a/Jint.Tests/Runtime/ModuleTests.cs
+++ b/Jint.Tests/Runtime/ModuleTests.cs
@@ -1,5 +1,6 @@
 using Jint.Native;
 using Jint.Runtime;
+using Jint.Runtime.Modules;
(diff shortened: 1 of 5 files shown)


lahma (approved):
Looks good, thank you for fixing this!

PR #1744, opened 2024-01-17
Add ValueTask and ValueTask<T> to Promise conversion

wazzamatazz:
This PR extends the work done in #1567 to add support for converting `ValueTask` and `ValueTask<T>` to promises.

This support is only available for .NET Standard 2.1 and .NET (non-Framework) targets. Support for .NET Framework and .NET Standard 2.0 targets could be enabled by adding a reference to [System.Threading.Tasks.Extensions](https://www.nuget.org/packages/System.Threading.Tasks.Extensions).

### DIFF
--- a/Jint.Tests/Runtime/AwaitTests.cs
+++ b/Jint.Tests/Runtime/AwaitTests.cs
@@ -78,4 +78,74 @@ public class AsyncTests
         engine.Execute("async function hello() {await myAsyncMethod();myAsyncMethod2();} hello();");
         Assert.Equal("12", log);
     }
+
+#if NETFRAMEWORK == false
+    [Fact]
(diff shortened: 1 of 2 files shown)


lahma (approved):
Thank you for the clean and concise PR!

wazzamatazz:
Would you be adverse to adding the System.Threading.Tasks.Extensions reference for .NET Framework and .NET Standard targets? Otherwise I can see the potential for confusion if someone is targeting .NET Framework and using `ValueTask` but finds that it doesn't work with Jint.

lahma:
Not sure if `ValueTask` is commonly used under .NET Framework, it's more targeted to modern .NET runtime. Maybe we should wait and see if there's some feedback suggesting that it is needed.

PR #1746, opened 2024-01-18
The getDate() method returns 0 for December 31st before 1970.

PavelPisotskiy:


lahma (approved):
Thanks for fixing this with test coverage!

PR #1753, opened 2024-01-20
Fix module builder loader interaction

### LINKED ISSUE #1747: IModuleLoader bugs
## Part 1. IModuleLoader interface has no information about calling module.
I try to write module loader for zip archives and it is not possible to implement relative module loading, because IModuleLoader interface has no information about calling module and i can not resolve module to import using following codes:

ibraty/execute.js file in module.zip archive:
```javascript
import * from '../relative_module.js'; 
```
C# importing code, that not possible to implement:
```csharp
var engine = new Engine(o => {
  o.EnableModules(new ZipModuleLoader("module.zip");
});
engine.Modules.Import("module.zip/libraty/execute.js");
```
## Part 2. IModuleLoader internal use bug.
I also noticed in file Engine.Modiles.cs following lines:
```csharp
internal Module Load(string? referencingModuleLocation, ModuleRequest request)
{
  var specifier = request.Specifier;
  var moduleResolution = ModuleLoader.Resolve(referencingModuleLocation, request);

  if (_modules.TryGetValue(moduleResolution.Key, out var module))
  {
    return module;
  }

  if (_builders.TryGetValue(specifier, out var moduleBuilder)) // IS THIS BUG???
  {
    module = LoadFromBuilder(specifier, moduleBuilder, moduleResolution);
  }
...
```
Seems following line contains bug:
```csharp
if (_builders.TryGetValue(specifier, out var moduleBuilder))
```
because  when i write custom ModuleLoader it is not possible to refer to certait ModuleBuilder after IModuleLoader.Resolve invocation. I guess thac correct line is:
```csharp
if (_builders.TryGetValue(moduleResolution.Key , out var moduleBuilder))
```
It will allow to resolve ModuleBuilder modules by custom ModuleLoader in Resolve method.

tomatosalat0:
Fixes #1747

* The internal dictionary of all available `ModuleBuilder` was used with the unresolved resolution key
* Modules created with the ModuleBuilder did not have any location, which prevented relative module location resolving.
* Public interface test for CustomModuleLoader adjusted to show the expected API usage of the `ResolvedSpecifier.Key` property.

### DIFF
--- a/Jint.Tests.PublicInterface/ModuleLoaderTests.cs
+++ b/Jint.Tests.PublicInterface/ModuleLoaderTests.cs
@@ -104,7 +104,7 @@ public class ModuleLoaderTests
         public ResolvedSpecifier Resolve(string? referencingModuleLocation, ModuleRequest moduleRequest)
         {
             Uri uri = Resolve(referencingModuleLocation, moduleRequest.Specifier);
-            return new ResolvedSpecifier(moduleRequest, moduleRequest.Specifier, uri, SpecifierType.Bare);
+            return new ResolvedSpecifier(moduleRequest, uri.ToString(), uri, SpecifierType.Bare);
         }
 
         private Uri Resolve(string? referencingModuleLocation, string specifier)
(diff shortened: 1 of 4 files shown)


lahma (approved):
That was a quick turnaround, awesome work!

PR #1767, opened 2024-01-30
Move interop tests to the public interface test project

BhaaLseN:
Most of those interop tests illustrate how Jint can interact with objects from popular libraries such as `Newtonsoft.Json`, `System.Text.Json` and dynamic objects such as `ExpandoObject`.

Because this can be a valuable reference for Jint users, they should not depend on any internals (which would be off limits for callers.) See #1766 for my endeavor on (successfully) integrating STJ.

This pretty much only moves the files over to the other project and updates the one location where internal code was used. Other tests like Newtonsoft and Dynamic didn't really use any internals, so it was just the namespace update there.

Initially, I moved over all tests that began with `InteropTests` and quickly realized how most of them don't _really_ mean interop in the sense of someone external trying to interface with it. At least I got stopped by various `internal` methods on `ObjectWrapper`, `ObjectInstance` etc. that I simply cannot justify changing just to move the tests.

Let me know if theres more (or possibly also less) you want included with this PR.

### DIFF
--- a/Jint.Tests.PublicInterface/InteropTests.Dynamic.cs
+++ b/Jint.Tests.PublicInterface/InteropTests.Dynamic.cs
@@ -1,9 +1,8 @@
 using System.Dynamic;
 using Jint.Native;
 using Jint.Native.Symbol;
-using Jint.Tests.Runtime.Domain;
 
-namespace Jint.Tests.Runtime
+namespace Jint.Tests.PublicInterface
 {
     public partial class InteropTests
     {
@@ -142,6 +141,12 @@ namespace Jint.Tests.Runtime
             {
                 return _properties.ContainsKey(key);
             }
+        }
+
+        private class Person
+        {
+            public string Name { get; set; }
+            public int Age { get; set; }
         }
     }
 }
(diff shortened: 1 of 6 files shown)


lahma (approved):
A great improvement, thank you!

hyzx86:
Hello @BhaaLseN , thanks for the PR, but I find that false for Boolean values doesn't seem to handle as expected

``` c#
        const string Json = """
        {
            "booleanValue":false,
            "trueValue":true,
            "employees": {
                "type": "array",
                "value": [
                    {
                        "firstName": "John",
                        "lastName": "Doe"
                    },
                    {
                        "firstName": "Jane",
                        "lastName": "Doe"
                    }
                ]
            }
        }
        """;

// ...other code..... 
        Assert.True(engine.Evaluate("variables.booleanValue==false").AsBoolean());
```

BhaaLseN:
I just moved the existing tests, I didn't make any functional changes. You should probably check Discussions and/or open an Issue instead.

sebastienros:
Check this PR https://github.com/sebastienros/jint/pull/1802, it passes but might not be the best impl.

PR #1769, opened 2024-01-31
Change GH Actions for PR to run on macos-14 instead of macos-latest

lahma:
Trying out the the new M1 offering.


sebastienros:
I remember a time when it took like 15 minutes to run all tests. And it was half of what is run today.

PR #1770, opened 2024-02-02
Json \u escape sequence was not case insensitive

tomatosalat0:
When a JSON strings contains the escape sequence `\uXXX`, the JSON parser only converted it properly if the hexadecimal characters `A-F` where in lower case.

This means, `\u003C` would get converted to a different character compared to `\u003c`.

The method `IsHexDigit` allows both versions - upper and lower case characters. The `.IndexOf()` did return -1 for upper case characters resulting in an invalid conversion.

Tests where added for this case.


lahma (approved):
LGTM, thank you!

PR #1775, opened 2024-02-07
Add ClearScript V8 to engine comparison benchmarks, trim suite, add script-to-host interop suite

### LINKED ISSUE #2706: Benchmarks comparing to Clearscript v8
### Pre-flight checks

- [x] I have searched [open](https://github.com/sebastienros/jint/issues) and [closed](https://github.com/sebastienros/jint/issues?q=is%3Aissue+is%3Aclosed) issues and could not find an existing proposal.
- [x] I have confirmed this is not already supported by Jint.
- [ ] For ECMAScript proposals — I have linked the [TC39 proposal](https://github.com/tc39/proposals) and noted its stage. Jint generally implements Stage 4 features.

### What problem are you trying to solve?

It would be great to add benchmarks if possible to ClearFoundry ClearScript. I'm aware of at least one project in the past that switched off of this library to that one for performace reasons.

Keep up the great work. Thank you

### Proposed solution

```csharp
Add benchmarks for https://github.com/ClearFoundry/ClearScript
```

### Alternatives considered

_No response_

### Affected area

None

### Contribution

- [ ] I am willing to submit a pull request for this if the proposal is accepted.

### Additional context

_No response_

viceice:
Resurrects this PR on top of current `main` (original commit by @viceice kept, packages moved to ClearScript 7.5.1) and extends it into the full benchmark-field refresh discussed in #2706.

Closes #2706.

## What's in here

1. **ClearScript V8 joins the comparison** (`Microsoft.ClearScript.V8` 7.5.1, now maintained by ClearFoundry):
   - The `ClearScript` lane now **disposes the engine** — the original lane leaked a native V8 isolate per iteration, which skews every subsequent measurement — and runs strict-prefixed source like the Okojo lanes (V8 has no engine-level strict switch).
   - New `ClearScript_Compiled` lane mirroring `Jint_ParsedScript` / `Okojo_Prepared`: a `V8Script` compiled once by a shared `V8Runtime`, executed in a fresh script engine (fresh V8 context) per operation — ClearScript's recommended production path.
2. **Suite trimmed to one variant per workload** (21 → 12 scripts): where a classic ES5 script had an ES2015+ `-modern` rewrite, only the modern variant remains. The duplicated rows ranked the engines nearly identically while doubling full-run wall clock. Classic scripts stay in `Scripts/` for the Jint-only benchmarks.
3. **New script ↔ host interop suite** (`EngineComparisonInteropBenchmark`): four byte-identical scripts (host method calls, property access, strings crossing the boundary, host array traversal) against Jint, NiL.JS, YantraJS, ClearScript and ClearScript's 7.5 **FastProxy** fast path. Design notes: host members are lowercase because YantraJS camel-cases CLR member names while the other engines don't; every script validates its aggregate and throws on mismatch so silent mis-marshaling can't post a time; Okojo is absent because 0.1.2-preview.1 has no public way to enable CLR access.
4. **README rewritten** for the new field, with freshly measured ClearScript and interop numbers merged in (measured on the same Ryzen 5950X / .NET 10.0.9 box as the 4.13.0 table).

## Headline results

Comparing production paths (`Jint_ParsedScript` vs `ClearScript_Compiled`), **the 12-script table splits 6/6 between Jint and native V8** — no other managed engine wins a row:

- Jint wins everything start-up- and eval-shaped: `minimal` **1.0 µs vs 461 µs (~450×)**, `evaluation-modern` ~107×, `linq-js` ~10×, `dromaeo-core-eval-modern` 1.35×, plus `dromaeo-object-array-modern` (1.07×) and `array-stress` (statistically tied).
- V8 wins the long compute rows: 3d-cube 2.5×, json-parse 1.7×, stopwatch 5.6×, object-string 8.8×, base64 9.7× (regexp within that row's variance).

**The interop suite inverts the picture — the managed engines win every row.** Plain ClearScript pays **5.7–7.0×** vs Jint on method calls / property access / string passing; the FastProxy path narrows that to **2.4–3.6×** at the cost of hand-registering every member (its zero-allocation marshaling is real, though: 13 KB vs 12.8 MB managed allocations on the method-call row).

The suite also surfaced an honest Jint finding on day one: re-reading a host array property inside a loop re-wraps it every iteration (19 ms / 32.9 MB on the traversal row). Hoisting the collection into a local drops Jint to **~1.1 ms / 0.3 MB — faster than every other engine on that row**; caching repeated wraps of the same host object is a follow-up improvement opportunity in Jint itself.

## Fairness notes baked into the README

- `Allocated` counts managed memory only — ClearScript's working memory lives on V8's native heap, so its rows are bridge overhead, not comparable memory figures; memory claims compare managed engines only.
- V8 runs background JIT/GC threads: `Mean` is wall-clock of the executing thread, total CPU is higher than the single-threaded managed engines.
- The full Dry-job smoke of all 116 cases passes; the measured tables come from default-job runs.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -13,6 +13,9 @@
     <PackageVersion Include="GitHubActionsTestLogger" Version="3.0.4" />
     <PackageVersion Include="ICU4N" Version="60.1.0-alpha.440" />
     <PackageVersion Include="Meziantou.Analyzer" Version="3.0.123" />
+    <PackageVersion Include="Microsoft.ClearScript.V8" Version="7.5.1" />
+    <PackageVersion Include="Microsoft.ClearScript.V8.Native.win-x64" Version="7.5.1" />
+    <PackageVersion Include="Microsoft.ClearScript.V8.Native.linux-x64" Version="7.5.1" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="5.6.0" PrivateAssets="all" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="5.6.0" PrivateAssets="all" />
     <PackageVersion Include="Verify.SourceGenerators" Version="2.5.0" />
--- a/Jint.Benchmark/EngineComparisonBenchmark.cs
+++ b/Jint.Benchmark/EngineComparisonBenchmark.cs
@@ -1,7 +1,8 @@
-﻿using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Configs;
 using BenchmarkDotNet.Order;
 using BenchmarkDotNet.Reports;
+using Microsoft.ClearScript.V8;
 using Okojo.Parsing;
 using Okojo.Runtime;
 using OkojoCompiler = Okojo.Compiler.JsCompiler;
@@ -26,9 +27,10 @@ public class EngineComparisonBenchmark
 
     private static readonly Dictionary<string, Prepared<Script>> _parsedScripts = new();
 
-    // Okojo has no engine-level strict switch (unlike the other engines' flags); the whole program is
-    // made strict by a leading "use strict" directive, so it runs in global strict mode like the rest.
-    private static readonly Dictionary<string, string> _okojoFiles = new();
+    // Okojo has no engine-level strict switch and V8 has no strict flag either (unlike the other
+    // engines); their whole program is made strict by a leading "use strict" directive, so both
+    // run in global strict mode like the rest.
+    private static readonly Dictionary<string, string> _strictFiles = new();
 
     // Okojo's realm-independent prepared artifact is the parsed program (JsProgram). Bytecode
     // (JsScript) is compiled against a specific realm, so — mirroring the fresh-engine-per-iteration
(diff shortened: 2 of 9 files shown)


lahma:
This is a bit apples to oranges comparison as interpreters vs truly compiled and inter-process vs invokes...

sebastienros:
@lahma what does it "inter-process" mean here?

I think it's good to represent compiled engines too, but maybe it should be an opportunity to call out the differences and introduce other benchmarks that will show the pros and cons. For instance the _Allocated_ column is only for managed memory, and compilation-based engines use mostly native memory so this won't show up. The other thing could be to add a benchmark where a script is used once, or with a way that scripts are re-compiled or re-parsed on each call. Some application behave this way but running a script just once. The last thing to take into account is how scripts are disposed, for instance you keep compiling/parsing new scripts, would it lead to resource exhaustions? Or can we release the script resources automatically (in-memory assemblies, loaded code, ...)?

So maybe it's about splitting this whole table and explaining what the test is about and why different engines do better than others, with the extra ones. I'd also hope that would make the Jint/Jint_Parsed duplication not necessary.

viceice:
yes, the table also shows that some tests are bad on clearscript and I assume it's because of the context switching between native V8 and dotnet. this is our main use case.

so we've a lot esm scripts which are also calling into dotnet a lot. so I think it's worth trying to switch to jint, or at least try it.

lahma:
> @lahma what does it "inter-process" mean here?

Just meant no interop between NET runtime and native code. It's the interop bridge which costs.

For V8 it might make sense to create a separate benchmark altogether comparing just it and Jint and keep V8 out from "interpreter" comparison. 

Benchmark could include something like.

1. Small eval (from comparison benchmark)
2. Large eval (from comparison benchmark)
3. for/of loop against large native JsArray
4. for loop against injected value and setting values to each index which is should show values getting back to NET runtime's original array
5. tight loop of calling engine, like C# loop calling `function(x) { return x; }` and providing call value and reading result each time
6. JS calling back to c#, so test 5 backwards
7. maybe something more

lahma:
There's also https://github.com/Taritsyn/JavaScriptEngineSwitcher if one wants to write their business code against abstraction and have possibility to change underlying engine later one (or on use case basic).

viceice:
> > @lahma what does it "inter-process" mean here?
> 
> Just meant no interop between NET runtime and native code. It's the interop bridge which costs.
> 
> For V8 it might make sense to create a separate benchmark altogether comparing just it and Jint and keep V8 out from "interpreter" comparison.
> 
> Benchmark could include something like.
> 
> 1. Small eval (from comparison benchmark)
> 2. Large eval (from comparison benchmark)
> 3. for/of loop against large native JsArray
> 4. for loop against injected value and setting values to each index which is should show values getting back to NET runtime's original array
> 5. tight loop of calling engine, like C# loop calling `function(x) { return x; }` and providing call value and reading result each time
> 6. JS calling back to c#, so test 5 backwards
> 7. maybe something more

Sounds good, will look into it later



> There's also https://github.com/Taritsyn/JavaScriptEngineSwitcher if one wants to write their business code against abstraction and have possibility to change underlying engine later one (or on use case basic).

To limmited for our use cases 🙃

lahma:
Resurrected this on top of current main and extended it for #2706: ClearScript 7.5.1 with a properly disposed fresh-isolate lane plus a `ClearScript_Compiled` lane (shared `V8Runtime` + cached `V8Script`), the comparison suite trimmed to 12 modern-variant scripts, and a new script↔host interop suite with a FastProxy lane. README rewritten with freshly measured numbers.

Short version: the compute table splits 6/6 between Jint and V8, while the managed engines win every interop row — plain ClearScript is ~6–7× slower than Jint there, FastProxy ~2.4–3.6×.

sebastienros:
Great job

PR #1786, opened 2024-02-21
feat: implicit convert DataView and ArrayBuffer to byte array

viceice:
- resolves #1785
- implicit convert `DataView` and `ArrayBuffer` to `byte[]` when calling `ToObject` for interop
- add `JsValue` extensions to handle and convert those types when return from engine


lahma (approved):
Looks good to me, thank you!

PR #1787, opened 2024-02-21
Add tasks to defaultobjectconverter

tom-b-iodigital:
With this PR it's possible to call object methods that return tasks.

For example:

```c#
var engine = new Engine();
engine.SetValue("log", new Action<object>(Console.WriteLine))
    .SetValue("getTest", new Func<TestClass>(() => new()))
    .SetValue("__result", 0);

engine.Execute("""

(async () => {
    __result = "Before call";
    let testClass = getTest();

    __result = await testClass.GetDelayedStringAsync();
})();

""");

Console.WriteLine(engine.GetValue("__result"));


public class TestClass
{
    public async Task<string> GetDelayedStringAsync()
    {
        await Task.Delay(5000).ConfigureAwait(false);

        return "This is a delayed string";
    }
}
```

Before PR:

image

After PR:

image

### DIFF
--- a/Jint.Tests/Runtime/AsyncTests.cs
+++ b/Jint.Tests/Runtime/AsyncTests.cs
@@ -1,3 +1,5 @@
+using Jint.Tests.Runtime.TestClasses;
+
 namespace Jint.Tests.Runtime;
 
 public class AsyncTests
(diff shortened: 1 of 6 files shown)


lahma:
Thanks for the PR, we also need tests to cover this new functionality - it's easy to break things later if behavior hasn't been tested.

tom-b-iodigital:
I'll add some test cases

tom-b-iodigital:
test cases added, let me know if you anything else

lahma (approved):
LGTM, thank you for your contribution!

uleelx:
This feature breaks the code before. In version 3.0.0, we can get the value synchronously: `var result = sometask.result`, and it seems much faster than `Promise.then` in v3.0.1 @tom-b-iodigital @lahma

PR #1790, opened 2024-02-24
Issue 1768: make behaviour of SetValue consistent when using interfaces

tom-b-iodigital:
When passing an instance of a C# object using one of its interfaces into the engine using SetValue, the interpreter incorrectly allowed method calls to class methods not defined in the interface. This behavior violates the principle of least privilege and can lead to unintended side effects or security vulnerabilities, as users could execute methods that were meant to be inaccessible through the interface abstraction. This PR fixes that and ads a more descriptive exception message to avoid users considering it a bug.


tom-b-iodigital:
@lahma you closed the issue, but I think this is how SetValue should behave

lahma:
Closing due to inactivity.

PR #1791, opened 2024-02-24
Ensure modules have always source specified

### LINKED ISSUE #1755: IModuleLoader bug 2 (continuation)
# Returning to issue #1747.
Thanks for fast bug fix, it is solves basic problem, but i faced with next behaiviour:
In some cases, engine is not transmit the source modile reference, like shown in test below. And it is not possible to solve module.

```csharp
[TestFixture]
public class AbsolutePathResolveBug
{
    public class TestModuleLoader : ModuleLoader
    {
        public override ResolvedSpecifier Resolve(string? referencingModuleLocation, ModuleRequest moduleRequest)
        {
            var moduleSpec = moduleRequest.Specifier;

            // to resolve this statement requires information about source module
            if (moduleSpec == "../execute.js") {
                Assert.That(!string.IsNullOrEmpty(referencingModuleLocation)); // bug here?
                moduleSpec = $"code/execute.js";
            }

            return new ResolvedSpecifier(
                moduleRequest,
                moduleSpec,
                Uri: null,
                SpecifierType.RelativeOrAbsolute
            );
        }
        protected override string LoadModuleContents(Engine engine, ResolvedSpecifier resolved)
        {
            if (resolved.Key == $"code/lib/module.js")
                return $"import * as m from '../execute.js'; log('code/lib/module.js')";
            if (resolved.Key == $"code/execute.js") {
                return $"import * as m from 'archive/lib/module1.js'; log('code/execute.js')";
            }

            throw new NotImplementedException(); // no need in this test
        }
    }

    [Test]
    public void Test2()
    {
        var engine = new Engine(o => o.EnableModules(new TestModuleLoader()));

        var logs = new List<string>();
        engine.SetValue("log", logs.Add);

        engine.Modules.Import($"code/lib/module.js");

        CollectionAssert.AreEqual(new[] {
            $"code/execute.js",
            $"code/lib/module.js"
        }, logs);
    }
}
```

kkohno:
fixes #1755 
Sorry it took so long to fix - a lot of work.
I hope you accept the pull request.

### DIFF
--- a/Jint.Tests/Runtime/ModuleTests.cs
+++ b/Jint.Tests/Runtime/ModuleTests.cs
@@ -614,4 +614,51 @@ export const count = globals.counter;
         protected override string LoadModuleContents(Engine engine, ResolvedSpecifier resolved)
             => throw new InvalidOperationException();
     }
+
+    [Fact]
+    public void EngineShouldTransmitSourceModuleForModuleLoader()
+    {
+        var engine = new Engine(o => o.EnableModules(new ModuleLoaderForEngineShouldTransmitSourceModuleForModuleLoaderTest()));
+
+        var logs = new List<string>();
+        engine.SetValue("log", logs.Add);
+
+        engine.Modules.Import($"code/lib/module.js");
+
+        Assert.Collection(logs,
+            s => Assert.Equal("code/execute.js", s),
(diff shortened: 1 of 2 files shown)


lahma:
Thank you!

PR #1792, opened 2024-02-25
Improve interop performance against CLR list types

### LINKED ISSUE #1772: Unexpected memory overflow/leak
**Version used**

3.0
(the same concept worked with 3.0.0-beta-2049 without raising this issue)

**Describe the bug**

When setting an iterable variable in the engine, we changed the prototype of the said variable to the array prototype in order to access methods like `.map()`, `.filter()`, ... However, since the migration from `3.0.0-beta-2049` to `3.0`, the same methodology have a huge impact on the memory for unknown reasons (to us at least). 

As an alternative to illustrate a normal memory usage, we made a test case using `Array.from` instead.

**To Reproduce**
See attachment, the code was too big because of the sample data payload.
[array_bug_test.txt](https://github.com/sebastienros/jint/files/14166317/array_bug_test.txt)

**Expected behavior**

The expression to be evaluated without having to raise the memory limit.

**Additional context**

Related to https://github.com/sebastienros/jint/discussions/1733 and https://github.com/sebastienros/jint/issues/1030#issuecomment-1919747531

lahma:
fixes #1772 

## Jint.Benchmark.ListInteropBenchmark

| **Diff**|Method|Toolchain|Mean|Error|Allocated|
|------- |-------|-------|-------:|-------|-------:|
| Old |Filter|Default|967.84 μs|12.909 μs|1935.49 KB|
| **New** |	| **Default** | **704.80 μs (-27%)** | **13.149 μs** | **973936 B (-51%)** |
| Old |Indexing|Default|305.84 μs|4.030 μs|113.96 KB|
| **New** |	| **Default** | **263.17 μs (-14%)** | **3.922 μs** | **92667 B (-21%)** |
| Old |HasProperty|Default|18.40 μs|0.246 μs|30.97 KB|
| **New** |	| **Default** | **11.17 μs (-39%)** | **0.090 μs** | **32 B (-100%)** |

### DIFF
--- a/Jint.Benchmark/ListInteropBenchmark.cs
+++ b/Jint.Benchmark/ListInteropBenchmark.cs
@@ -1,0 +1,114 @@
+﻿using System.Collections;
+using BenchmarkDotNet.Attributes;
+using Jint.Native;
+using Jint.Runtime.Interop;
+
+namespace Jint.Benchmark;
+
+[MemoryDiagnoser]
+public class ListInteropBenchmark
+{
+    private static bool IsArrayLike(Type type)
+    {
+        if (typeof(IDictionary).IsAssignableFrom(type))
+        {
+            return false;
+        }
+
+        if (typeof(ICollection).IsAssignableFrom(type))
+        {
+            return true;
+        }
+
+        foreach (var typeInterface in type.GetInterfaces())
+        {
+            if (!typeInterface.IsGenericType)
+            {
+                continue;
+            }
+
+            if (typeInterface.GetGenericTypeDefinition() == typeof(IReadOnlyCollection<>)
+                || typeInterface.GetGenericTypeDefinition() == typeof(ICollection<>))
+            {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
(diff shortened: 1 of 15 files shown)


Jint.Benchmark/ListInteropBenchmark.cs
```diff
@@ -0,0 +1,114 @@
...
+                || typeInterface.GetGenericTypeDefinition() == typeof(ICollection<>))
+            {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    private const int Count = 10_00;
```
sebastienros, line 40: I used [sharplab ](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK7wCYgNQB8ACATAAwCwAUEQIwX7EAE+VALANwW1UCcAFFcQH1ixAJSsgA)to be sure what it was ;) Now I am even more confused by the location of the `_`

lahma: Hah, probably did some testing and forgot to remove the underscore which should be used as a thousand separator 😆

Issue #490, opened 2018-03-28
Source map support


Enerccio:
Is there a support for source maps in jint?

lahma:
What would this mean exactly? We'll close this if there's no concrete action plan.

lahma:
Closing after six months of silence.

thautwarm:
Source map maps the source code position from javascript to some "original" file.

This is usually useful when debugging javascript code generated from other languages(ReasonML, TypeScript, etc).

I'd like to request this feature, and if possible I'd like to help.

thautwarm:
This can be well leveraged: https://github.com/microsoft/sourcemap-toolkit
Just add an option to `ParserOption` and remap the source code position.

lofcz:
It would be nice if we could reopen this as even if this is done on Esprima's side, there is still integration TBD in Jint. Recently, there's been good progress in enabling high I/O scenarios (async <-> Task discussion, initial work on it by @lahma, perf).

The other thing such scenarios (think server back-end) need to be maintainable is typing, in the case of JS, there are multiple choices (TS prominently). These languages transpiling to JS generate a standardized source map as pointed out by @thautwarm

Implementing this would hence enable debugging of TS code - along with the proposed non-blocking async interop Jint could be effectively used in many new ways.

lahma:
We can reopen if someone wants to take the implementation task and offer PRs, otherwise it's quite moot to keep a forever issue open.

PR #1793, opened 2024-02-25
Add Options.InteropOptions.BuildCallStackHandler

scgm0:
A preliminary idea, this would allow one to modify the stack string generation, such as using a map of the source map.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -22,6 +22,7 @@
     <PackageVersion Include="NUnit" Version="4.2.2" />
     <PackageVersion Include="NUnit3TestAdapter" Version="4.6.0" />
     <PackageVersion Include="SharpZipLib" Version="1.4.2" />
+    <PackageVersion Include="SourceMaps" Version="0.3.0" />
     <PackageVersion Include="Spectre.Console.Cli" Version="0.45.0" />
     <PackageVersion Include="System.Text.Json" Version="8.0.5" />
     <PackageVersion Include="Test262Harness" Version="1.0.1" />
--- a/Jint.Tests.PublicInterface/CallStackTests.cs
+++ b/Jint.Tests.PublicInterface/CallStackTests.cs
@@ -1,3 +1,6 @@
+using Jint.Runtime;
+using SourceMaps;
+
 namespace Jint.Tests.PublicInterface;
 
 public class CallStackTests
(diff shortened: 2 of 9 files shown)


scgm0:
@lahma What do you think?

lahma:
Probably would need better documentation and tests, `ValueStringBuilder` shouldn't be made public. If there's a solid use case for such configuration, why not.

scgm0:
@lahma This is my first time writing a test, is it enough?

lofcz:
> If there's a solid use case for such configuration, why not.

We've discussed this over at Acornima https://github.com/adams85/acornima/issues/6
This was previously requested here as well: https://github.com/sebastienros/jint/issues/490

This should warrant a sufficient use case. As hinted in the included test the feature aims to enable running transpiled TypeScript (and possibly other JS dialects) through Jint.

lahma:
This is starting to look nice with the example use case described with tests! I'm traveling until Saturday but probably with minor tweaks we can merge.

lofcz:
@scgm0 the PR is missing XML doc for the newly added public members, if you want I can add that for you.

lahma:
One thing that came to mind is that would it be enough just to have callback `MapSourcePosition` which would by default just return same values back or be null and call stack builder would just call it and all string manipulation would remain in original call stack builder.

scgm0:
> One thing that came to mind is that would it be enough just to have callback `MapSourcePosition` which would by default just return same values back or be null and call stack builder would just call it and all string manipulation would remain in original call stack builder.

The source map allows mapping function names(Although the typescript compiler does not seem to support it), and there may be users who prefer to use other formats for stack strings. For example, I like to use 4 spaces before `at` instead of 3.

lahma:
The current format is node format I guess. We also need to consider if CallStackBuilder could be inherited and methods overridden, maybe configurable properties for indent. Need to tinker a bit

lofcz:
@scgm0 I've opened a PR against your branch with the docs. Getting win/fw462 to pass tests requires resigning the Tests assembly due to the added dependency on `SourceMaps` I believe, win/net8 builds fine.

scgm0:
> @scgm0 I've opened a PR against your branch with the docs. Getting win/fw462 to pass tests requires resigning the Tests assembly due to the added dependency on `SourceMaps` I believe, win/net8 is ok.

Thank you so much!

scgm0:
I'm sorry, I'm using linux and still don't know why the win test fails (

lahma (approved):
Sorry it took so long, I rebased this and made minor adjustments, thank you for the contribution!

PR #1799, opened 2024-03-07
Remove Newtonsoft package

MikeAlhayek:


lahma (approved):
This only changes NJS dependency to transitive, there's still tests using it, but as the test suite is happy we can merge this.

Thank you.

MikeAlhayek:
yeah. just one less dependency in your tests.

Issue #1801, opened 2024-03-12
Boolean value in JsonObject cannot be work as expected


hyzx86:
Reproduction code

``` c#
        const string Json = """
        {
            "booleanValue":false,
            "trueValue":true
        }
        """;
       var variables = JsonNode.Parse(Json);

       var engine = new Engine(options =>
       {
           // make JsonArray behave like JS array
           options.Interop.WrapObjectHandler = static (e, target, type) =>
           {
               var wrapped = new ObjectWrapper(e, target);
               if (target is JsonArray)
               {
                   wrapped.Prototype = e.Intrinsics.Array.PrototypeObject;
               }
               return wrapped;
           };

           // we cannot access this[string] with anything else than JsonObject, otherwise itw will throw
           options.Interop.TypeResolver = new TypeResolver
           {
               MemberFilter = static info =>
               {
                   if (info.ReflectedType != typeof(JsonObject) && info.Name == "Item" && info is PropertyInfo p)
                   {
                       var parameters = p.GetIndexParameters();
                       return parameters.Length != 1 || parameters[0].ParameterType != typeof(string);
                   }

                   return true;
               }
           };
       });

       engine.SetValue("variables", variables);

        Assert.False(engine.Evaluate("variables.trueValue==false").AsBoolean());
        var boolObject = engine.Evaluate("variables.trueValue");
        Assert.NotNull(boolObject);// not a boolean value
        Assert.True(engine.Evaluate("variables.trueValue").AsBoolean());
        Assert.True(engine.Evaluate("variables.trueValue==true").AsBoolean());
        Assert.False((bool) variables["booleanValue"]);
        Assert.True(engine.Evaluate("variables.booleanValue==false").AsBoolean());
```

_Originally posted by @hyzx86 in https://github.com/sebastienros/jint/issues/1767#issuecomment-1990342006_

lahma:
A pull request fixing the issue is most welcome.

hyzx86:
> A pull request fixing the issue is most welcome.

It's no problem at all, but I haven't found a solution to this problem yet. Can you give me a hint?
 
var boolObject = engine.Evaluate("variables.trueValue");
boolObject.GetType();

```

{Name = "ObjectWrapper" FullName = "Jint.Runtime.Interop.ObjectWrapper"}
    Assembly: {Jint, Version=3.0.1.0, Culture=neutral, PublicKeyToken=2e92ba9c8d81157f}
[...]

PR #1803, opened 2024-03-13
Add test code to validate boolean vlue in the if condition.

hyzx86:
relate #1801

### DIFF
--- a/Jint.Tests.PublicInterface/InteropTests.SystemTextJson.cs
+++ b/Jint.Tests.PublicInterface/InteropTests.SystemTextJson.cs
@@ -55,7 +55,6 @@ public partial class InteropTests
                     if (jsonValue.TryGetValue<bool>(out var boolValue))
                     {
                         return e.Construct("Boolean", boolValue ? JsBoolean.True : JsBoolean.False);
-                        //return e.Construct("Boolean", boolValue ? JsBoolean.True : JsBoolean.False);
                     }
 
                     if (jsonValue.TryGetValue<double>(out var doubleValue))
--- a/Jint.Tests.PublicInterface/InteropTests.SystemTextJson.cs
+++ b/Jint.Tests.PublicInterface/InteropTests.SystemTextJson.cs
@@ -2,9 +2,54 @@ using System.Reflection;
 using System.Text.Json.Nodes;
 using Jint.Native;
 using Jint.Runtime.Interop;
-
+using System.Text.Json;
 namespace Jint.Tests.PublicInterface;
 
+public class TestJsonValueConverter : IObjectConverter
+{
+    public bool TryConvert(Engine engine, object value, out JsValue result)
+    {
+        result = JsValue.Undefined;
+        if (value is JsonValue jsonValue)
+        {
+            var valueKind = jsonValue.GetValueKind();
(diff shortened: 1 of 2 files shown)


lahma:
Why don't you just investigate the problem instead making others do the leg work?

hyzx86:
> Why don't you just investigate the problem instead making others do the leg work?

Sorry, I've already made some attempts. But none of them works as expected and you are not very familiar with these configuration items

hyzx86:
Hi @lahma , I tried to use AddObjectConverter to solve this problem, so far the test passes, but I don't know if the method is correct.

Jint.Tests.PublicInterface/InteropTests.SystemTextJson.cs
```diff
@@ -2,19 +2,74 @@
 using System.Text.Json.Nodes;
 using Jint.Native;
 using Jint.Runtime.Interop;
-
+using System.Text.Json;
 namespace Jint.Tests.PublicInterface;
 
+#if NET8_0_OR_GREATER
```
BhaaLseN, line 8: I don't think this is supposed to be here, is it? At least I don't see any .NET 8 specific code at a glance?

hyzx86:
``` c#
            var valueKind = jsonValue.GetValueKind();
```
This line ,the `GetValueKind()` is warning not support net462

BhaaLseN: Yes, that method was introduced with .NET 8...but the available NuGet package that ships `System.Text.Json` still supports .NET Framework 4.6.2 upwards; so it _should_ work there as well. If this wasn't the case, the old code should've errored out already too.

Granted, I don't have any STJ interop scenarios on .NET Framework at the moment, but I still use this in a .NET 6 integration that is waiting to get upgraded to .NET 8. And as far as I can tell, things _should_ work there if you override the package version. I'd have to test this though, and that might be a while.

Maybe gate the converter behind this, and leave the old code as alternative for anyone that isn't on .NET 8 yet? Remember those tests serve as documentation, and having them work for _fewer_ people could be a problem (at least it would be in my head, not sure if others disagree.)

hyzx86: > Remember those tests serve as documentation, and having them work for fewer people could be a problem (at least it would be in my head, not sure if others disagree.)

Yes, I agree, I think it's enough to provide a concrete example that people can use to transform different complex objects, so we shouldn't worry too much about whether it works in all versions，people will always have a way to make it work😁

lahma:
Please note that this way does not lead to optimal performance as it's still reflection based - and it's been before too also with Newtonsoft.Json, but if you would want optimal solution performance-wise then a custom `ObjectWrapper` should be created which would use known types instead of reflection when accessing properties.

hyzx86:
Thanks for the suggestion, I just don't know why this expression doesn't work with `ObjectWrapper` , I found out through debugging that it does differ from `JsBoolean.False` 

 var result2 = engine.Evaluate("!variables.falseValue");

lahma (approved):

PR #1818, opened 2024-03-26
Readonly accessor.

SebastianStehle:
Introduces a new filter to mark objects and accessors as readonly. This is handy when you want to have pure evaluation functions.


lahma:
How does this differ from `options.AllowClrWrite(false)` / `options.Interop.AllowWrite = false`?

SebastianStehle:
There is probably not a difference, except the filtering. To be honest, I have just not seen the option. I was looking for something  and could not find it :D

PR #1822, opened 2024-03-30
Add support for wrapped lists to push and pop items

ejsmith:
Test that is showing that trying to push to an interop list does not work. Should this work? Is there some rule that I'm not aware of that is trying to keep interop to read only operations?

### DIFF
--- a/Jint.Tests/Runtime/InteropTests.cs
+++ b/Jint.Tests/Runtime/InteropTests.cs
@@ -2914,6 +2914,48 @@ namespace Jint.Tests.Runtime
         }
 
         [Fact]
+        public void ArrayPrototypeFindWithInteropList()
+        {
+            var engine = new Jint.Engine();
+            var list = new List<string> { "A", "B", "C" };
+ 
+            engine.SetValue("list", list);
+
+            Assert.Equal(1, engine.Evaluate("list.findIndex((x) => x === 'B')"));
+            Assert.Equal('B', engine.Evaluate("list.find((x) => x === 'B')"));
+        }
+
+        [Fact]
+        public void ArrayPrototypePushWithInteropList()
+        {
+            var engine = new Jint.Engine();
+
+            var list = new List<string> { "A", "B", "C" };
+
+            engine.SetValue("list", list);
+
+            engine.Evaluate("list.push('D')");
+            Assert.Equal(4, list.Count);
+            Assert.Equal("D", list[3]);
+            Assert.Equal(3, engine.Evaluate("list.lastIndexOf('D')"));
+        }
+
+        [Fact]
+        public void ArrayPrototypePopWithInteropList()
+        {
+            var engine = new Jint.Engine();
+
+            var list = new List<string> { "A", "B", "C" };
+            engine.SetValue("list", list);
+
+            Assert.Equal(2, engine.Evaluate("list.lastIndexOf('C')"));
+            Assert.Equal(3, list.Count);
+            Assert.Equal("C", engine.Evaluate("list.pop()"));
+            Assert.Equal(2, list.Count);
+            Assert.Equal(-1, engine.Evaluate("list.lastIndexOf('C')"));
+        }
+
+        [Fact]
         public void ShouldBeJavaScriptException()
         {
             var engine = new Engine(cfg => cfg.AllowClr().AllowOperatorOverloading().CatchClrExceptions());
--- a/Jint/Native/Array/ArrayOperations.cs
+++ b/Jint/Native/Array/ArrayOperations.cs
@@ -493,7 +493,25 @@ namespace Jint.Native.Array
 
             public override ulong GetLongLength() => GetLength();
 
-            public override void SetLength(ulong length) => throw new NotSupportedException();
+            public override void SetLength(ulong length)
+            {
+                if (_list == null)
+                {
+                    throw new NotSupportedException();
+                }
+                
+                while (_list.Count > (int) length)
+                {
+                    // shrink list to fit
+                    _list.RemoveAt(_list.Count - 1);
+                }
+                
+                while (_list.Count < (int) length)
+                {
+                    // expand list to fit
+                    _list.Add(null);
+                }
+            }
 
             public override void EnsureCapacity(ulong capacity)
             {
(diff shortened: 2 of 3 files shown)


lahma:
Jint doesn't automatically attach array prototype when wrapping objects ([same as with STJ](https://github.com/sebastienros/jint/blob/23a2a2587750a98cd55e4697e4da8270bcf460db/Jint.Tests.PublicInterface/InteropTests.SystemTextJson.cs)). Here's a good example which it would a bad decision by default. JS expects that you can index outside of collection size to create a new entry. Probably a custom `ObjectWrapper` implementation catching the outside-of-bounds indexer usage and calling `Add` instead for target might work.

ejsmith:
Hmm... I guess that is kind of confusing that it implements things like `indexOf` but not `push`. I will experiment with a `IObjectConverter` to see if I can make it support `push`.

lahma:
IndexOf is  a List method, it doesn't need JS array prototype

ejsmith:
Hmm... yeah, I guess a lot of the JS Array methods line up with the list methods. I tried what you said and it didn't work either.

```csharp
        [Fact]
        public void ArrayPrototypePushWithInteropList()
        {
            var engine = new Jint.Engine(options =>
            {
                // make List behave like JS array
                options.Interop.WrapObjectHandler = static (e, target, type) =>
                {
                    if (target is IList)
                    {
                        var wrapped = ObjectWrapper.Create(e, target);
                        wrapped.Prototype = e.Intrinsics.Array.PrototypeObject;
                        return wrapped;
                    }

                    return ObjectWrapper.Create(e, target);
                };
            });

            engine.SetValue("list", new List<string> { "A", "B", "C" });

            engine.Evaluate("list.Add('D')");
            Assert.Equal(3, engine.Evaluate("list.LastIndexOf('D')"));
        }
```

lahma:
You probably need to implement something custom deriving from ObjectWrapper which would intercept problematic calls.

ejsmith:
`ObjectWrapper` is sealed.

ejsmith:
This seems like something that should work out of the box. Do you agree with that or is against the direction you want to go with the project?

I am going to investigate more and try and figure out the correct spot to implement this. The code is quite complex so trying to figure out the correct way to implement it will probably take some time. I'd be willing to send some sponsor money your way to either give me a tour of the code and point me in the right direction or if you just want to implement yourself.

ejsmith:
Kept digging and found a way that I think might be acceptable to get `push` and `pop` on wrapped `IList` objects.

Jint.Tests/Runtime/InteropTests.cs
```diff
@@ -2913,6 +2913,76 @@ public void ArrayPrototypeIndexOfWithInteropList()
...
+            Assert.Equal('B', engine.Evaluate("list.find((x) => x === 'B')"));
+        }
+
+        [Fact]
+        public void ArrayPrototypePushWithInteropList()
+        {
+            var engine = new Jint.Engine(options =>
+            {
+                // make List behave like JS array
+                options.Interop.WrapObjectHandler = static (e, target, type) =>
```
lahma, line 2934: with latest main this is probably no longer needed

Jint.Tests/Runtime/InteropTests.cs
```diff
@@ -2913,6 +2913,76 @@ public void ArrayPrototypeIndexOfWithInteropList()
...
+            Assert.Equal(3, engine.Evaluate("list.lastIndexOf('D')"));
+        }
+
+        [Fact]
+        public void ArrayPrototypePopWithInteropList()
+        {
+            var engine = new Jint.Engine(options =>
+            {
+                // make List behave like JS array
+                options.Interop.WrapObjectHandler = static (e, target, type) =>
```
lahma, line 2963: probably not needed

Jint/Native/Array/ArrayOperations.cs
```diff
@@ -493,7 +493,31 @@ public IndexWrappedOperations(ObjectWrapper wrapper)
 
             public override ulong GetLongLength() => GetLength();
 
-            public override void SetLength(ulong length) => throw new NotSupportedException();
+            public override void SetLength(ulong length)
+            {
+                if (_list == null)
+                    throw new NotSupportedException();
```
lahma, line 499: we always use braces

Jint/Native/Array/ArrayOperations.cs
```diff
@@ -493,7 +493,31 @@ public IndexWrappedOperations(ObjectWrapper wrapper)
...
+                    // shrink list to fit
+                    for (var i = _list.Count - 1; i >= (int) length; i--)
+                    {
+                        _list.RemoveAt(i);
+                    }
+
+                    return;
+                }
+                
+                if (length >= (ulong) _list.Count)
```
lahma, line 512: for loop should cover this if already

Jint/Native/Array/ArrayOperations.cs
```diff
@@ -493,7 +493,31 @@ public IndexWrappedOperations(ObjectWrapper wrapper)
 
             public override ulong GetLongLength() => GetLength();
 
-            public override void SetLength(ulong length) => throw new NotSupportedException();
+            public override void SetLength(ulong length)
+            {
+                if (_list == null)
+                    throw new NotSupportedException();
+                
+                if (length < (ulong) _list.Count)
```
lahma, line 501: could probably be just while loop, next for loop for addition will be skipped

Jint/Native/Array/ArrayOperations.cs
```diff
@@ -493,7 +493,31 @@ public IndexWrappedOperations(ObjectWrapper wrapper)
...
+
+                    return;
+                }
+                
+                if (length >= (ulong) _list.Count)
+                {
+                    // expand list to fit
+                    for (var i = _list.Count; i < (int) length; i++)
+                    {
+                        _list.Add(JsValue.Undefined.ToObject());
```
lahma, line 517: would be clearer just to call with `null`

Jint/Native/Object/ObjectInstance.cs
```diff
@@ -702,6 +703,11 @@ internal bool DeletePropertyOrThrow(JsValue property)
         /// </summary>
         public virtual bool Delete(JsValue property)
         {
+            if (this is ObjectWrapper objectWrapper && objectWrapper.Target is IList && property.IsNumber())
```
lahma, line 706: should override in ObjectWrapper instead of breaking the whole idea of OOP...

ejsmith:
Updated. Thanks for the feedback!

ejsmith:
> Thank you!

No, thank you!

PR #1826, opened 2024-03-31
Support interop against System.Text.Json types on NET8+

lahma:
On NET 8+ we can support reading and accessing STJ types without complex and hard to figure out configuration.

### DIFF
--- a/Jint.Tests.PublicInterface/InteropTests.SystemTextJson.cs
+++ b/Jint.Tests.PublicInterface/InteropTests.SystemTextJson.cs
@@ -1,4 +1,3 @@
-using System.Reflection;
 using System.Text.Json.Nodes;
 using Jint.Native;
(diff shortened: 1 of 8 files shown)


lahma:
> Can we make sure `push`ing and `pop`ping work on STJ arrays?

Yes we can. Anyone needing such a feature is welcome to submit the implementation via a PR. Here in investigating minimal read model OOTB experience.

ejsmith:
Thank you. I have been looking into it a bit and trying to understand the code and figure out where the right place to do so would be.

ejsmith:
Not sure if you will be interested to take my other PR for adding `push` and `pop` support to interop arrays, but I pulled this branch down and tried to get it working on `JsonArray`. I did get it working, but it's pretty hacky. If this PR lands, then I will create a new PR on top to discuss.

PR #1827, opened 2024-04-01
Add support for wrapped generic lists to push and pop items

ejsmith:
#1822 added support for `push` and `pop` on interop objects that implement `IList`. `JsonArray` from `System.Text.Json` implements `IList<T>` and doesn't work.

My thoughts are that we implement a `GenericIndexWrappedOperations` that will have to use reflection to support adding and removing items.

I guess normal reflection in .NET has gotten fast enough where we don't need any special libraries, right?

### DIFF
--- a/Jint.Tests.PublicInterface/InteropTests.SystemTextJson.cs
+++ b/Jint.Tests.PublicInterface/InteropTests.SystemTextJson.cs
@@ -5,63 +5,50 @@ using System.Text.Json;
 
 namespace Jint.Tests.PublicInterface;
 
-public sealed class SystemTextJsonValueConverter : IObjectConverter
-{
-    public static readonly SystemTextJsonValueConverter Instance = new();
-
-    private SystemTextJsonValueConverter()
-    {
-    }
-
-    public bool TryConvert(Engine engine, object value, out JsValue result)
-    {
-        if (value is JsonValue jsonValue)
-        {
-            var valueKind = jsonValue.GetValueKind();
-            switch (valueKind)
-            {
-                case JsonValueKind.Object:
-                case JsonValueKind.Array:
-                    result = JsValue.FromObject(engine, jsonValue);
-                    break;
-                case JsonValueKind.String:
-                    result = jsonValue.ToString();
-                    break;
-                case JsonValueKind.Number:
-                    if (jsonValue.TryGetValue<double>(out var doubleValue))
-                    {
-                        result = JsNumber.Create(doubleValue);
-                    }
-                    else
-                    {
-                        result = JsValue.Undefined;
-                    }
-                    break;
-                case JsonValueKind.True:
-                    result = JsBoolean.True;
-                    break;
-                case JsonValueKind.False:
-                    result = JsBoolean.False;
-                    break;
-                case JsonValueKind.Undefined:
-                    result = JsValue.Undefined;
-                    break;
-                case JsonValueKind.Null:
-                    result = JsValue.Null;
-                    break;
-                default:
-                    result = JsValue.Undefined;
-                    break;
-            }
-            return true;
-        }
-        result = JsValue.Undefined;
-        return false;
-
-    }
-}
 public partial class InteropTests
 {
+
+    [Fact]
+    public void ArrayPrototypeFindWithInteropJsonArray()
+    {
+        var engine = GetEngine();
+        
+        var array = new JsonArray { "A", "B", "C" };
+        engine.SetValue("array", array);
+
+        Assert.Equal(1, engine.Evaluate("array.findIndex((x) => x === 'B')"));
+        Assert.Equal('B', engine.Evaluate("array.find((x) => x === 'B')"));
+    }
+
+    [Fact]
+    public void ArrayPrototypePushWithInteropJsonArray()
+    {
+        var engine = GetEngine();
+
+        var array = new JsonArray { "A", "B", "C" };
+        engine.SetValue("array", array);
+
+        engine.Evaluate("array.push('D')");
+        Assert.Equal(4, array.Count);
+        Assert.Equal("D", array[3]?.ToString());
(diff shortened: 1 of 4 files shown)


lahma:
I would probably strive towards something like:

* `ListWrapper extends ObjectInstance` or (ObjectWrapper) `implements IObjectWrapper`
  * has virtual (overridable) services like `Get(int index)` and `Set(int index, object value)`, `int Length { get; }`
* GenericListWrapper extends ListWrapper`, overridering methods and uses reflection to access generic instance

Current ArrayOperations specialization could just detect that wrapper is `ListWrapper` and delegate to it. We could detect that type is definitely not both dictionary and and a list and then allow creating specific wrapper. This is way I originally introduced `ObjectWrapper.Create` for such decisions. Need to keep in mind that target could implement both dictionary and list interfaces and it shouldn't allowed to be wrapped to specific list wrapper unless it's meant to be used in int-indexing context.

ejsmith:
I took at stab at an approach where I've added abilities to the `TypeDescriptor` and I'm using those from `IndexWrappedOperations`. I've got some reflection issues to resolve, but I think it can work. If you have any thoughts or you don't like this approach, please let me know.

ejsmith:
@lahma Would you be able to take a look at this code? All but 1 test are passing. The one that is failing is trying to do a `indexOf` check on `JsonArray` and I think there is some sort of type coercion missing somewhere that I'm not able to track down. I'm also just interested in general if this is going to be an acceptable approach.

ejsmith:
I ended up removing the `indexOf` test because I tried the same thing outside of Jint and it didn't work on `JsonArray` either. So that test was invalid.

lahma:
With this approach there's the problem that the custom usage is only available via common array operations (push, pop) but that's of course basically what PR's title says. Now TypeDescriptor is no longer only describing the type and its capabilities but also operating on the types. 

With the `ListWrapper` approach I suggested above a call to `person.Cars[0].color` could also use the faster custom object wrapper implementation for indexing without trying to figure out which one to use in which case - List<Car> should only allow indexing value with positive integer, otherwise call prototype (Array.prototype).

ejsmith:
I get what you are saying and I definitely think this can be done a lot cleaner. I am just not familiar enough with everything yet. When I looked at taking your approach, it seemed like it was going to be a lot of work and I'd have a lot of learning to do.

TypeDescriptor already had behavior in it with `TryGetValue` and `Remove` and `GetKeys`.

I am going to try and do some more digging and see what I can come up with. It looks like I would need to derive from `ObjectInstance` and `IObjectWrapper` and then copy a lot of code from `ObjectWrapper`, but specialize it for just `IList<T>`? I guess this would be `ListWrapper<T>`? And then who would construct this wrapper object? Would we use reflection to automatically wrap `IList<T>` implementors?

ejsmith:
Do you think something like this is on the right path:

```csharp
    public interface IArrayLikeWrapper : IObjectWrapper
    {
        int Length { get; }
        object? Get(int index);
        void Set(int index, object value);
        void Add(object value);
        void RemoveAt(int index);
        void EnsureCapacity(int capacity);
    }
    
    public class ArrayLikeWrapper : ObjectInstance, IArrayLikeWrapper
    {
        private readonly ICollection? _collection;
        private readonly IList? _list;
        
        internal ArrayLikeWrapper(Engine engine, object? target)
            : base(engine)
        {
            if (target == null)
            {
                ExceptionHelper.ThrowArgumentNullException(nameof(target));
            }
            
            Target = target;
            _collection = target as ICollection;
            _list = target as IList;
            Prototype = engine.Intrinsics.Array.PrototypeObject;
        }
        
        public static IArrayLikeWrapper Create(Engine engine, object? target)
        {
            if (target is IList<object> list)
                return new ArrayLikeWrapper<object>(engine, list);

            // TODO: Figure out how to check for IList<T> and ICollection<T> and construct wrapper
            
#if NET8_0_OR_GREATER
            if (target is JsonArray array)
                return new ArrayLikeWrapper<JsonNode?>(engine, array);
#endif
            
            return new ArrayLikeWrapper(engine, target);
        }
        
        public object Target { get; }
        public virtual int Length => _collection?.Count ?? 0;

        public virtual object? Get(int index)
        {
            if (_list is not null && index >= 0 && index < _list.Count)
            {
                return _list[index];
            }

            return null;
        }

        public virtual void Set(int index, object value)
        {
            EnsureCapacity(index);
            
            if (_list is not null)
            {
                _list[index] = value;
            }
        }

        public virtual void Add(object value)
        {
            _list?.Add(value);
        }

        public virtual void RemoveAt(int index)
        {
            _list?.RemoveAt(index);
        }

        public virtual void EnsureCapacity(int capacity)
        {
            if (_list is null)
            {
                return;
            }

            while (_list.Count < capacity)
            {
                _list.Add(null);
            }
        }

        public override object ToObject()
        {
            return Target;
        }
    }
    
    public class ArrayLikeWrapper<T> : ArrayLikeWrapper
    {
        private readonly ICollection<T> _collection;
        private readonly IList<T>? _list;
        
        public ArrayLikeWrapper(Engine engine, ICollection<T> target)
            : base(engine, target)
        {
            _collection = target;
            _list = target as IList<T>;
        }
        public override int Length => _collection?.Count ?? 0;

        public override object? Get(int index)
        {
            if (_list is not null && index >= 0 && index < _list.Count)
            {
                return _list[index];
            }

            return null;
        }

        public override void Set(int index, object value)
        {
            EnsureCapacity(index);
            
            if (_list is not null)
            {
                _list[index] = (T)value;
            }
        }

        public override void Add(object value)
        {
            _list?.Add((T)value);
        }

        public override void RemoveAt(int index)
        {
            _list?.RemoveAt(index);
        }

        public override void EnsureCapacity(int capacity)
        {
            if (_list is null)
            {
                return;
            }

            while (_list.Count < capacity)
            {
                _list.Add(default!);
            }
        }
    }

```

ejsmith:
Trying this approach here: #1828

PR #1828, opened 2024-04-04
Adding ArrayLikeWrapper to support more interop array operations

ejsmith:
There are failing tests, but this approach does have less reflection going on. What are your thoughts @lahma ?


lahma:
This is closer to what I had in mind. If you wish I can push commit(s) to this branch to tweak things.

ejsmith:
Yeah, absolutely! Go for it.

ejsmith:
I added you to the repository so you can push commits there if you like.

lahma:
Just an update, I made some progress with bit of a refining proposal, hopefully have some time coming days to finish.

lahma:
I've implemented some changes, all tests now green and also supports Newtonsoft array map operations that were failing earlier in OOTB experience because of problematic indexers. Custom configuration for non .NET 8 is now quite minimal.

ejsmith:
Looks great!

PR #1820, opened 2024-03-29
Acorn-based parsing

adams85:
It was unavoidable, I had to see how my new acorn-based JS parser works with Jint... 😄 

It seems that pretty well: it resolves most of the parser-related issues in test262 and brings a tiny perf. improvement in overall (see benchmark results below).

So, I'm curious to know what do you think about an acorn-based Jint? Of course, absolutely no pressure, plus this would mean some potentially annoying breaking changes, so if you'd rather stick with Esprima, I'm totally fine with that. (Even then this work won't have been done completely in vain as this is a great integration test for my parser, which uncovered some minor bugs and showed me where the AST has shortcomings currently.)

However, if you'd like to do the switch, that would make my very proud and gladly do the minor adjustments which are needed to finalize this PR!

PR #1831, opened 2024-04-09
Prevent users from passing arbitrary parser options to Jint

adams85:
As suggested [here](https://github.com/sebastienros/jint/pull/1820#issuecomment-2035194119), this PR extracts some changes from #1820 to make them happen earlier.

This means some breaking changes to the public API which prevent users from passing arbitrary parser options to Jint. Only a subset of the parser options should be adjustable by end users, while the rest should be reserved for Jint to achieve the desired parser behavior when it needs to parse code dynamically (`eval`, `Function` ctor, shadow realms, etc.)

The PR also...
* makes sure that the options used to parse the root code are preserved in `ExecutionContext` for dynamic code parsing - e.g. when the script is parsed in non-tolerant mode, dynamic code should be as well. (In the case of `eval` this doesn't work currently as Esprima doesn't provide the necessary settings, so tolerant mode must be enabled no matter what. This will be solved though once the migration to Acornima happens.)
* fixes a minor bug when eval is used in a shadow realm,
* adds an alias for Esprima.Ast.Module to reduce further the amount of changes related to Acornima migration.


Jint.Tests/Runtime/ParserOptionsPropagationTests.cs
```diff
@@ -0,0 +1,216 @@
+#nullable enable
+
+using Jint.Native;
+using Jint.Runtime;
+using Jint.Runtime.Modules;
+
+namespace Jint.Tests.Runtime
```
lahma, line 7: nit: for new files can use file-scoped namespaces

Jint/ParseOptions.cs
```diff
@@ -0,0 +1,116 @@
+﻿using System.Text.RegularExpressions;
+
+namespace Jint
+{
```
lahma, line 4: nit: could use file-scoped namespace

Jint/ParseOptions.cs
```diff
@@ -0,0 +1,116 @@
...
+    public interface IParseOptions
```
lahma, line 5: I'm not a native speaker, but would `IParsingOptions` be more natura, also `ScriptParsingOptions` - just a thought

Jint/ParseOptions.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// Gets or sets whether to parse the source code in tolerant mode.
+        /// Defaults to <see langword="true"/>.
+        /// </summary>
+        /// <remarks>
+        /// Please note that there is a plan to change the default to <see langword="false"/> in the next major version.
+        /// </remarks>
+        bool Tolerant { get; init; }
+    }
+
+    public record class ScriptParseOptions : IParseOptions
```
lahma, line 36: maybe sealed?

Jint/ParseOptions.cs
```diff
@@ -0,0 +1,116 @@
...
+        };
+
+        public static readonly ScriptParseOptions Default = new();
+
+        /// <summary>
+        /// Gets or sets whether to allow return statements at the top level.
+        /// Defaults to <see langword="true"/>.
+        /// </summary>
+        /// <remarks>
+        /// Please note that there is a plan to change the default to <see langword="false"/> in the next major version.
```
lahma, line 51: next major of Acornima? so should this just be `true`?

Jint/ParseOptions.cs
```diff
@@ -0,0 +1,116 @@
...
+        TimeSpan? RegexTimeout { get; init; }
+
+        /// <summary>
+        /// Gets or sets whether to parse the source code in tolerant mode.
+        /// Defaults to <see langword="true"/>.
+        /// </summary>
+        /// <remarks>
+        /// Please note that there is a plan to change the default to <see langword="false"/> in the next major version.
+        /// </remarks>
+        bool Tolerant { get; init; }
```
lahma, line 33: live below (came back to this), should we just initialize with `true`?

Jint/PrepareOptions.cs
```diff
@@ -0,0 +1,52 @@
+﻿namespace Jint
+{
+    public interface IPrepareOptions<out TParseOptions>
```
lahma, line 3: not my grammar, but `IPreparationOptions`?

lahma (commented):
Lookin good, thanks for doing this! I added some comments/thoughts. This could very well be next 3.1 release, I usually scan through compilation problems with the usual suspects utilizing Jint so I can check if further PRs needed to refine some things.

Jint/ParseOptions.cs
adams85: I meant the next major version of Jint.

Or we could just make this BC in v3.1 as the related APIs will be broken anyway. What do you think?

Jint/ParseOptions.cs
lahma: I'm fine with either way, I guess we need an Oxford scholar for this 😆

Jint/ParseOptions.cs
lahma: I'm not just quite sure if it's best approach on interpreter side to be strict-by-default. For parsing it makes sense, but for anyone just "having a good time with JS" do we need to restrict parsing and allowed constructs?  But again seems that NodeJS throws error on console from using return so maybe it makes sense.

Jint/ParseOptions.cs
adams85: Naming things and cache invalidation, the two hard problems in software engineering :D 

I ran a quick search on the ASP.NET Core repo. Although there are some counter-examples but it's mostly `[...]ingOptions` or `[...]tionOptions`, so I'm convinced. :D

Jint/ParseOptions.cs
adams85: I wrote about this in the other PR: "[...] I can only speak for myself, but I'd be surprised if I fed some script to Jint which is invalid per spec and get some results instead of SyntaxError. I mean I think relaxing the rules should be a deliberate choice, not the default. [...]"

> But again seems that NodeJS throws error on console from using return so maybe it makes sense.

Disabling tolerant mode doesn't affect top level return because that's controlled by another switch than tolerant mode. The option would be on for scripts by default: https://github.com/sebastienros/jint/pull/1831/files#diff-54aceb67398ce1901835b606789cff7f0ac2abb5136a75480469b5c732d19b11R40

adams85: Let's not break this behavior in a minor version, so I'd say let's leave it as is now and reconsider it when v4 is on the table.

Jint/ParseOptions.cs
adams85: I'm not sure I get this. This is an interface, we can't initialize this here, can't we? It's initialized here: https://github.com/sebastienros/jint/pull/1831/files#diff-5196d539da48e2bc4acd76456adfff1ca098ad422ce71e90849e98258c066633R62

lahma: Lazy eyes, sorry.

lahma (approved):

lahma:
Thank you once again!

adams85:
Sure thing!

PR #1837, opened 2024-04-18
Fix dynamic importing of JSON modules

adams85:
While doing the parser migration, I spotted a bug: dynamically importing JSON modules (e.g. `await import("package.json", { with: { type: "json" } })`) results in an error ("default is not defined").

This PR attempts to fix the issue.


lahma (approved):
Thanks for spotting and fixing!

PR #1868, opened 2024-05-22
Make LazyPropertyDescriptor generic and some prototypes more lazy

lahma:
Using `LazyPropertyDescriptor` for objects that have many properties that can be seldomly used. Generic interface makes usage simpler.


sebastienros:
Take this line

```
["copyWithin"] = new LazyPropertyDescriptor<ArrayPrototype>(this, static prototype => new ClrFunction(prototype._engine, "copyWithin", prototype.CopyWithin, 2, PropertyFlag.Configurable), PropertyFlags),
```

What state and what behavior in the actual implementation could vary across engines and scripts that it would prevent it from being a static `ClrFunction` instance?

Rephrased, should there be something else that an `Engine` instance (which a script can mutate), which could change how this method behaves? If not then there could be a way to have these more stable and shared across engine instances.

lahma:
The problematic piece is the actual `Engine` instance (and the active `Realm` instance). Engine configuration can alter some behaviors and as engine is not thread-safe it cannot shared between process-wide instances. Alternative would be to change every `Call` method etc to to take the active `engine` as parameter.

PR #1878, opened 2024-05-29
Enable grouped version updates

sebastienros:
Prevents several PRs to be created. This can be improved to group dependencies (test libs, dev libs ...) in case a lib might block others.


lahma (approved):
Sounds logical, thanks for tweaking!

Issue #1945, opened 2024-08-16
Inconsistent ObjectInstance.GetOwnProperties


lofcz:
**Version used**

4.0.1

**Describe the bug**

I'm working on a Blazor debugger for Jint. Recently I've encountered a weird inconsistency - if the thing is intentional, feel free to yell at me. `ObjectInstance.GetOwnProperties()` behaves (at least from my perspective) differently for CLR `ObjectWrapper` and native `Object` instances. 

For CLR objects, only properties and fields get reported:
```cs
// we take public properties and fields
foreach (var p in ClrType.GetProperties(BindingFlags.Static | BindingFlags.Instance | BindingFlags.Public))
{
    var indexParameters = p.GetIndexParameters();
    if (indexParameters.Length == 0)
    {
        var jsString = JsString.Create(p.Name);
        yield return jsString;
    }
}

foreach (var f in ClrType.GetFields(BindingFlags.Static | BindingFlags.Instance | BindingFlags.Public))
{
    var jsString = JsString.Create(f.Name);
    yield return jsString;
}
```

While native objects report properties, which are methods as well. For example:

```js
var myTest = {
  myFn: () => { 
    return 4
  }
}
```

reports `myFn` when calling `GetOwnProperties` on a corresponding `JsValue`, while only props and fields are reported for CLR objects as per the code above. 

While I understand that `myFn` is a property from the JS perspective and `public void MyFn()` is a method from CLR perspective, it still seems inconsistent, given we report fields from CLR objects already.
My understanding is that `GetOwnProperties` is the tool to check from the CLR side, how the JS interpreter sees the given object, hence my suggestion to include methods too.

**Expected behavior**

Methods of CLR objects are reported from `GetOwnProperties` (we include `GetMethods` result in the code above).

lahma:
Adding public methods sounds reasonable as they can be invoked. Would you like to create a PR with some test coverage?

lofcz:
Sure, I'll go ahead. Technically, I would consider this breaking however, there might be consumers relying on the fact that the method in question works the way it does now, so this might warrant a 4.1 release.

lahma:
One option is to add new enum flags property to interop options defining returned properties and omit methods for now.

lofcz:
solved in https://github.com/sebastienros/jint/pull/1947

PR #1946, opened 2024-08-16
Make ObjectWrapper ClrType public

lofcz:
In the scenario described in https://github.com/sebastienros/jint/issues/1945 I need to get the CLR type of `ObjectWrapper`. This PR makes the corresponding property public.


lahma:
So is this still needed if/when #1947 gets merged?

lofcz:
Yes, I still need to get the CLR type somehow; this seems the easiest way to do it.

PR #1947, opened 2024-08-16
Add ObjectWrapperReportedMemberTypes to Options

lofcz:
Implements the feature discussed in https://github.com/sebastienros/jint/issues/1945.

### DIFF
--- a/Jint.Tests/Runtime/InteropTests.cs
+++ b/Jint.Tests/Runtime/InteropTests.cs
@@ -3499,7 +3499,7 @@ try {
     {
         var engine = new Engine(opt =>
         {
-            
+            opt.Interop.ObjectWrapperReportedMemberTypes = MemberTypes.Field | MemberTypes.Property;
         });
         
         engine.SetValue("clrInstance", new ClrMembersVisibilityTestClass());
@@ -3510,25 +3510,43 @@ try {
          var props = obj.GetOwnProperties().ToList();
          
          Assert.Equal(props.Count, 1);
+         props.Should().ContainSingle(x => x.Key == "A");
     }
     
     [Fact]
     public void ShouldSeeClrMethods()
     {
-        var engine = new Engine(opt =>
-        {
-            opt.Interop.ObjectWrapperReportedMemberTypes = MemberTypes.Method | MemberTypes.Field | MemberTypes.Property;
-        });
+        var engine = new Engine();
         
         engine.SetValue("clrInstance", new ClrMembersVisibilityTestClass());
-
-        var getValue = engine.Evaluate("clrInstance.get_A()");
-        Assert.Equal(getValue, 10);
         
         var val = engine.GetValue("clrInstance");
         var obj = val.AsObject();
         var props = obj.GetOwnProperties().ToList();
         
-        Assert.Equal(props.Count, 8); // A get/set + F + GetType + ToString + Equals + GetHashCode
+        Assert.Equal(props.Count, 2);
+        props.Should().ContainSingle(x => x.Key.AsString() == "A");
+        props.Should().ContainSingle(x => x.Key.AsString() == "F");
+    }
+    
+    private class ClrMembersVisibilityTestClass2
+    {
+        public int Get_A { get; set; } = 5;
+    }
+    
+    [Fact]
+    public void ShouldSeeClrMethods2()
+    {
+        var engine = new Engine();
+        
+        engine.SetValue("clrInstance", new ClrMembersVisibilityTestClass2());
+        
+        var val = engine.GetValue("clrInstance");
+
+        var obj = val.AsObject();
+        var props = obj.GetOwnProperties().ToList();
+         
+        Assert.Equal(props.Count, 1);
+        props.Should().ContainSingle(x => x.Key == "Get_A");
     }
 }
--- a/Jint.Tests/Runtime/InteropTests.cs
+++ b/Jint.Tests/Runtime/InteropTests.cs
@@ -3507,10 +3507,9 @@ try {
          var val = engine.GetValue("clrInstance");
 
          var obj = val.AsObject();
-         var props = obj.GetOwnProperties().ToList();
+         var props = obj.GetOwnProperties().Select(x => x.Key.ToString()).ToList();
          
-         Assert.Equal(props.Count, 1);
-         props.Should().ContainSingle(x => x.Key == "A");
+         props.Should().BeEquivalentTo(["A"]);
     }
     
     [Fact]
@@ -3522,11 +3521,9 @@ try {
         
         var val = engine.GetValue("clrInstance");
         var obj = val.AsObject();
-        var props = obj.GetOwnProperties().ToList();
-        
-        Assert.Equal(props.Count, 2);
-        props.Should().ContainSingle(x => x.Key.AsString() == "A");
-        props.Should().ContainSingle(x => x.Key.AsString() == "F");
+        var props = obj.GetOwnProperties().Select(x => x.Key.ToString()).ToList();
+
+        props.Should().BeEquivalentTo(["A", "F"]);
     }
     
     private class ClrMembersVisibilityTestClass2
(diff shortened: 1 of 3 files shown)


Jint.Tests/Runtime/InteropTests.cs
```diff
@@ -3483,4 +3483,52 @@ void Debug(object o)
...
+        {
+            return 4;
+        }
+    }
+
+    [Fact]
+    public void ShouldNotSeeClrMethods()
+    {
+        var engine = new Engine(opt =>
+        {
```
lahma, line 3501: could omit the defaults and have just `new Engine();`

Jint.Tests/Runtime/InteropTests.cs
```diff
@@ -3483,4 +3483,52 @@ void Debug(object o)
...
+    public void ShouldSeeClrMethods()
+    {
+        var engine = new Engine(opt =>
+        {
+            opt.Interop.ObjectWrapperReportedMemberTypes = MemberTypes.Method | MemberTypes.Field | MemberTypes.Property;
+        });
+        
+        engine.SetValue("clrInstance", new ClrMembersVisibilityTestClass());
+
+        var getValue = engine.Evaluate("clrInstance.get_A()");
```
lahma, line 3525: Maybe just `engine.Evaluate("clrInstance.get_A()").Should().Be(10);`, but I'm goin to argument that the current behavior is wrong

Jint.Tests/Runtime/InteropTests.cs
```diff
@@ -3483,4 +3483,52 @@ void Debug(object o)
...
+        {
+            
+        });
+        
+        engine.SetValue("clrInstance", new ClrMembersVisibilityTestClass());
+        
+         var val = engine.GetValue("clrInstance");
+
+         var obj = val.AsObject();
+         var props = obj.GetOwnProperties().ToList();
```
lahma, line 3510: I would prefer asserting the actual keys: 

```c#
var props = obj.GetOwnProperties().Select(x => x.Key.ToString()).ToList();
props.Should().ContainSingle("A");
```

Jint.Tests/Runtime/InteropTests.cs
```diff
@@ -3483,4 +3483,52 @@ void Debug(object o)
...
+        });
+        
+        engine.SetValue("clrInstance", new ClrMembersVisibilityTestClass());
+
+        var getValue = engine.Evaluate("clrInstance.get_A()");
+        Assert.Equal(getValue, 10);
+        
+        var val = engine.GetValue("clrInstance");
+        var obj = val.AsObject();
+        var props = obj.GetOwnProperties().ToList();
```
lahma, line 3530: Rather:

```c#
var props = obj.GetOwnProperties().Select(x => x.Key.ToString()).ToList();
props.Should().BeEquivalentTo("A", "get_A", "set_A", "F", "GetType", "ToString", "Equals", "GetHashCode");

var jsProps = engine.Evaluate("Object.getOwnPropertyNames(clrInstance)").AsArray().Select(x => x.ToString()).ToList();
jsProps.Should().BeSameAs(["A", "get_A", "set_A", "F", "GetType", "ToString", "Equals", "GetHashCode"]);
```

But I'm going to argument that this behavior is currently wrong.

Jint/Options.cs
```diff
@@ -364,6 +364,13 @@ public class InteropOptions
         /// Whether the engine should throw an error when a member is not found on a CLR object. Defaults to false.
         /// </summary>
         public bool ThrowOnUnresolvedMember { get; set; }
+
+        /// <summary>
+        /// Types of CLR members reported by <see cref="ObjectWrapper"/> when enumerating properties/serializing <see cref="ObjectWrapper.ToObject"/>.
+        /// Supported values are: <see cref="MemberTypes.Field"/>, <see cref="MemberTypes.Property"/>, <see cref="MemberTypes.Method"/>.
+        /// All other values are ignored.
+        /// </summary>
+        public MemberTypes ObjectWrapperReportedMemberTypes { get; set; } = MemberTypes.Field | MemberTypes.Property;
```
lahma, line 373: As the current behavior is wrong, I'd say jus use all three types and I'll add release note info that this can be reverted back by modifying the configuration

lahma (changes requested):
I think this needs some discussion.

The JavaScript behavior for `get`/`set` etc:

```javascript
var base = {
  baseField: "foo",
  funcBase() {
  },
  get propBase() {
  },
  set propBase(val) {
  },
};

var a = {
  aField: 0,
  funcA() {
  },
  get propA() {
  },
  set propA(val) {
  },
}

Object.setPrototypeOf(a, base);

console.log("own: " + Object.getOwnPropertyNames(a));

console.log("enumerable properties");

for (var k in a) {
  console.log(k);
}
```
Output:

```
own: aField,funcA,propA
enumerable properties
aField
funcA
propA
baseField
funcBase
propBase
```

So the correct output for the "full" test case should be:

```
jsProps.Should().BeSameAs(["A", "F"]);
```

It should not contain inherited ones and property get/set and not listed as it's implied by the property. `GetType` is already special cased in configuration anyway not to be allowed as it opens reflection.

Jint.Tests/Runtime/InteropTests.cs
lahma: well not sure if the possibility to call is wrong, but the idiomatic way is to evaluate `clrInstance.A`

lofcz:
Thanks for the feedback, the requested changes are in a0a36c597bdc4045743aa7cae1ae0f3559ecb9cb, we can squash it after your review.

lofcz:
I have no idea why or how the Linux tests are failing:

> Passed!  - Failed:     0, Passed: 64848, Skipped:  7253, Total: 72101, Duration: 1 m 1 s - Jint.Tests.Test262.dll (net8.0)
> Error: Process completed with exit code 1.

lahma:
No worries, I'll look into this and from quick look seems good. Thanks for iterating and discussing 👍🏻

lahma (approved):
Looks good, I tweaked a bit for readability. Thanks for fixing this!

lofcz:
@lahma why are we taking `DeclaredOnly` fields and methods but not properties?

lahma:
> @lahma why are we taking `DeclaredOnly` fields and methods but not properties?

It would break existing test case, if you want to check the feasibility for the change, a PR and investigation is always welcome 🙂

lofcz:
Could you point me to the failing test? I'm using Windows box and tests were passing here without the `DeclaredOnly` constraint. I'd like to look into lifting it, as the current version won't report inherited fields/methods, which I need for the debugger scenario described in the linked issue.

lahma:
For me it's `EngineShouldStringifyAnJObjectListWithValuesCorrectly` and it's enough to have `foreach (var p in ClrType.GetProperties(BindingFlags | BindingFlags.DeclaredOnly))` when runnin on NET 8.

lahma:
So I guess it's a case of `dynamic` where nobody knows where the properties originate from. It's been a PITA for so long but a valid use case - probably `ExpandoObject` related or something.

lofcz:
I've run the two tests included in this PR and they fail for net462 which I've not noticed before:
image
This is without any changes. The `EngineShouldStringifyAnJObjectListWithValuesCorrectly` seems to be another thing.

lahma:
This is my result when running on a Windows box with the flag changed for properties against latest `main`

image

lofcz:
Those are ok here, both with and without the `DeclaredOnly` flag:
image
`dotnet --list-sdks` returns `8.0.302` for me.

image

lofcz:
This one is passing for me, both with and without the `DeclaredOnly` flag:
image

lahma:
Let build server be the judge: https://github.com/sebastienros/jint/pull/1948

Failed test summary: https://github.com/sebastienros/jint/actions/runs/10434541945?pr=1948

lofcz:
I'll try to get my environment to be consistent with the build server then and look into the failing test.

PR #1948, opened 2024-08-17
Make own properties for declared only method and properties configurable

lahma:


lahma:
As the earlier change was breaking change as users could rely on having also public fields being part of JSON serialization, we need to revert that and make this declared-only behavior configurable.

/cc @lofcz

lofcz:
Thank you!

PR #1952, opened 2024-08-21
Don't overwrite default export from `SourceTextModule`

### LINKED ISSUE #1951: Exporting a default function without name isn't possible
**Version used**

4.0.1

**Describe the bug**

Can't get the default export of a module if it's a function without a name.

**To Reproduce**

```csharp
var engine = new Engine();
engine.Modules.Add("a", "export default function () { return 1; }"
engine.Modules.Add("b", "export default function main () { return 1; }"
engine.Modules.Import("a").Get("default"); //-> undefined
engine.Modules.Import("b").Get("default")(); //-> 1

```

**Expected behavior**

Both test cases above should work.

**Additional context**

none

viceice:
Before this `JintExportDefaultDeclaration.ExecuteInternal` always overwrites the probably existing binding from `SourceTextModule.InitializeEnvironment`.

- Fixes #1951

### DIFF
--- a/Jint.Tests/Runtime/ModuleTests.cs
+++ b/Jint.Tests/Runtime/ModuleTests.cs
@@ -44,6 +44,22 @@ public class ModuleTests
     }
 
     [Fact]
+    public void ShouldExportDefaultFunctionWithoutName()
+    {
+        _engine.Modules.Add("module1", "export default function main() { return 1; }");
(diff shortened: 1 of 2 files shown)


viceice:
@lahma I don't think the failed test is related to the PR changes. 🤔

lahma (approved):
I don't see any harm in this change, thanks for investigating and fixing!

PR #1956, opened 2024-08-26
Tweak object wrapper reported members logic

lahma:
* to have parity, we shouldn't limit to declared members only as generic search will try to find from all levels
* obey member filter while reporting
* cleanup


lofcz:
thank you

Issue #1962, opened 2024-09-04
TypeResolver should resolve to non-hidden property over property it hides


mspertus:
**Version used**

4.0.1

**Describe the bug**

When `TypeResolver:TryFindMemberAccessor` resolves a property, it might find one that was hidden instead of the non-hidden one that would be selected by accessing the object's member of that name in C#.

**To Reproduce**
Create C# classes like the following. When Jint accesses the `A` member of a `Bar` object from JavaScript, you may get the hidden one (I say "may" because `Type:GetProperties` does not specify an order in a number of versions of .NET supported by jint).

```
class Foo
{
    public int A {get;set;}
    public string B {get;set;}
}

class Bar :  Foo
{
public new float A { get; set; }
}
```

([This dotnet fiddle](http://dotnetfiddle.net/NQRXIJ) shows that both `A` elements are iterated by `GetProperties` even though if `b` is a `Bar`, `b.A` unambiguously refers to `Bar:A`)

**Expected behavior**
You should get `Bar:A` instead of an indeterminate `A`

**Additional context**
I have created a simple pull request that fixes this

PR #1963, opened 2024-09-04
Don't find hidden property instead of the hiding property

mspertus:
resolves sebastienros/jint#1962

When `TypeResolver:TryFindMemberAccessor` resolves a property, it might find one that was hidden where the non-hidden one would normally be found by C++ member access.

For example, when you access the `A` member from JavaScript, you may get the hidden one (I say "may" because `Type:GetProperties` does not specify an order in a number of versions of .NET supported by jint).

```
class Foo
{
    public int A {get;set;}
    public string B {get;set;}
}

class Bar :  Foo
{
	public new float A { get; set; }
}
```

([This dotnet fiddle](http://dotnetfiddle.net/NQRXIJ) shows that both `A` elements are iterated by `GetProperties` even though if `b` is a `Bar`, `b.A` unambiguously refers to `Bar:A`)

This change ensures that in the case where a property in a derived class hides another, the derived one is resolved to.

### DIFF
--- a/Jint.Tests/Runtime/InteropTests.MemberAccess.cs
+++ b/Jint.Tests/Runtime/InteropTests.MemberAccess.cs
@@ -280,4 +280,25 @@ public partial class InteropTests
 
         engine.Invoking(e => e.Evaluate("obj.AgeMissing")).Should().Throw<MissingMemberException>();
     }
+
+    public class ClassWithPropertyToHide
+    {
+        public int x { get; set; } = 2;
+        public int y { get; set; } = 3;
+    }
+
+    public class ClassThatHidesProperty : ClassWithPropertyToHide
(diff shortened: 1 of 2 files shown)


lahma (changes requested):
Thank you for the PR, can you also add a test case ensuring this works as expected and won't break later on by accident.

mspertus:
> Thank you for the PR, can you also add a test case ensuring this works as expected and won't break later on by accident.

Thanks, Lahma. I've added a test and confirmed that it fails without the PR (dotnet462 windows) and  passes with it.

lahma (approved):
Looking good, thank you!

mspertus:
Thanks, @lahma!

PR #1976, opened 2024-10-04
Fix Array.prototype.toString() stackoverflow

### LINKED ISSUE #1975: Calling Array.prototype.toString() can result in a stack overflow exception if the array contains a circular reference.
**Version used**

4.0.3

**Describe the bug**

Calling  `Array.prototype.toString()` or `Array.prototype.join()` produces a Stackoverflow exception when it contains a circular reference.

**To Reproduce**

```cs
 var _engine = new Engine();
 var result = _engine.Evaluate("Array.prototype.toString.call((c = [1, 2, 3, 4], c[1] = c, c))").AsString();
```
**Expected behavior**

```cs
 var _engine = new Engine();
 var result = _engine.Evaluate("Array.prototype.toString.call((c = [1, 2, 3, 4], c[1] = c, c))").AsString();
 Assert.Equal("1,,3,4", result);
```

I expect to get an empty string where the circular reference is set.

**Additional context**
Node and Firefox produce `1,,3,4`

xBaank:
Closes #1975

### DIFF
--- a/Jint/Native/Array/ArrayPrototype.cs
+++ b/Jint/Native/Array/ArrayPrototype.cs
@@ -1273,14 +1273,14 @@ public sealed class ArrayPrototype : ArrayInstance
             return JsString.Empty;
         }
 
-        static string StringFromJsValue(JsValue value)
+        static string StringFromJsValue(JsValue value, JsValue thisObject)
         {
-            return value.IsNullOrUndefined()
+            return value.IsNullOrUndefined() || thisObject == value
                 ? ""
                 : TypeConverter.ToString(value);
         }
 
-        var s = StringFromJsValue(o.Get(0));
+        var s = StringFromJsValue(o.Get(0), thisObject);
         if (len == 1)
         {
             return s;
@@ -1294,7 +1294,7 @@ public sealed class ArrayPrototype : ArrayInstance
             {
                 sb.Append(sep);
             }
-            sb.Append(StringFromJsValue(o.Get(k)));
+            sb.Append(StringFromJsValue(o.Get(k), thisObject));
         }
 
         return sb.ToString();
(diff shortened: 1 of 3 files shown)


Jint/Native/Array/ArrayPrototype.cs
```diff
@@ -1273,14 +1273,14 @@ private JsValue Join(JsValue thisObject, JsValue[] arguments)
             return JsString.Empty;
         }
 
-        static string StringFromJsValue(JsValue value)
+        static string StringFromJsValue(JsValue value, JsValue thisObject)
         {
-            return value.IsNullOrUndefined()
+            return value.IsNullOrUndefined() || thisObject == value
```
lahma, line 1278: for clarity, I think this should be `ReferenceEquals(thisObject, value)`, the operator overloads can be a bit quirky

xBaank:
I modified the code because `Array.prototype.toString.call((c = [1, 2, 3, 4], b = [1, 2, 3, 4], b[1] = c, c[1] = b, c))` could still produce a StackOverflow. Now it checks if the array is already visited.

lahma:
This will make join calls very costly as now every time another array is being created, or did I misunderstand?

xBaank:
> This will make join calls very costly as now every time another array is being created, or did I misunderstand?

Yes, maybe a better aproach could be to pass down the `visited` hashset along with `thisObject` and `arguments` so no arrays are created?

lahma:
I haven't dug deep into this problem, but I would other types would also be problematic, not on concrete JsArrays if called with different thisObject. Like objects having indexes pointing you circular references, or interop array-like wrappers.

lahma:
By looking at V8 code, it seems that at least following are vulnerable to cyclic behavior (V8 has safeguards):

* `ArrayPrototype.join`
* `ArrayPrototype.toLocaleString`
* `TypedArrayPrototype.join`
* `TypedArrayPrototype.toLocaleString`

lahma:
There's `ObjectTraverseStack` that could probably be used for this, adding `TryEnter` which wouldn't throw like the `Enter` does (`Enter` would `TryEnter` and throw).

lahma (approved):
Looking good, thank you for taking the time iterating!

Issue #1971, opened 2024-09-24
caller is undefined,but it's not undefined in broswer.


sgf:
**Version used**

4.0.2

**Describe the bug**

caller is undefined,but it's not undefined in broswer.

**To Reproduce**

  ```html
  <script type="text/javascript">
    function Test(Lf_) {
        function h() {
            return getName();
        };
        return h();
        return 'Lf'
    };
    function getName() {
        debugger;
        var caller = getName.caller;
        if (caller.name) {
            return caller.name
        }
        var str = caller.toString().replace(/[\s]*/g, "");
        var name = str.match(/^function([^\(]+?)\(/);
        if (name && name[1]) {
            return name[1];
        } else {
            return '';
        }
    }
    Test();
</script>
```

**Expected behavior**

caller should be not undefined,in up code,its should be "h()";

**Additional context**

im using AngleSharp.Js. 
https://github.com/AngleSharp/AngleSharp.Js/issues/98

sgf:
https://github.com/sebastienros/jint/issues/819

Although this is not a standard definition feature. But I think Jint should be closer to the actual behavior of browsers while respecting the standard.
After all, in the end, we all aim for practicality.

lahma:
Would you like to propose a PR to implement this?

sgf:
> Would you like to propose a PR to implement this?

I will propose an RP later.

lahma:
Any updates?

sgf:
I only added one line of code, and the local compilation was fine.
I looked at the logs, there were a lot of logs, but the most important failure was not obvious, I didn't even know where the error was.
Forget it, I gave up on making a PR.
I think it's too much effort. This is not my forte.

Genteure:
The compilation is fine, yes. The problems are tests.

I know you said you give up, but the PR workflow failed because there are tests (https://github.com/tc39/test262) that check jint whether is actually spec compliant, meaning following the requirements of what is allowed and not allowed.

For example, according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/caller (note: MDN web docs is not the spec):

> In strict mode, accessing caller of a function throws an error — the API is removed with no replacement.

That means you at the minimum needs to check if the current context is in strict mode before returning any value, and there are tests in test262 that check this.

You can not "fix" a behavior mismatch by creating another 100+ specification violations, that would only make the behavior of jint even further from browsers.

Have a nice day.

PR #1980, opened 2024-10-18
fixed issues/1971 caller is undefined,but it's not undefined in broswer.

sgf:
https://github.com/sebastienros/jint/issues/1971


lahma:
Seems that a proper, a spec compliant solution, might require a bit more work I'm afraid.

sgf:
I only added one line of code, and the local compilation was fine.
I looked at the logs, there were a lot of logs, but the most important failure was not obvious, I didn't even know where the error was.
Forget it, I gave up on making a PR.
I think it's too much effort. This is not my forte.

PR #1984, opened 2024-10-23
Made test of cancellation of engine execution more robust

tomatosalat0:
Found a flaky test in #1983 . This PR makes this one test more robust (fixes a race condition). 

The test will now be green, no matter when the provided `ThreadPool.QueueUserWorkItem` will get executed. This has been simulated by adding `Thread.Sleep(10000)` everywhere within the `threadPoolAction` locally. Without the adjustments, the sleep additions would break the tests.


sebastienros (approved):

lahma (approved):
Thanks for taking the time to investigate, this test case has been a real PITA.

PR #1986, opened 2024-10-24
Optimize some character checks and ValueStringBuilder

lahma:


sebastienros:
I would have assumed SearchValues to be quicker in any cases. It can vectorize and will find the best comparison logic based on inputs.

lahma:
You probably are correct, I can see a slight regression, I'll maybe do another round..

PR #1987, opened 2024-10-24
Make JsSet public

kenlyon:
Allow use of the `Jint.Native.JsSet` class outside of expressions, allowing consumers to cast to this type and call its members.

For example, you might want to convert it to a `HashSet<T>` to use it elsewhere without being tied to using Jint types.

The `OrderedSet` class also had to be made `public` due to its use in the `JsSet` constructor.

### DIFF
--- a/Jint/Native/JsSet.cs
+++ b/Jint/Native/JsSet.cs
@@ -1,3 +1,4 @@
+using System.Collections;
 using System.Diagnostics.CodeAnalysis;
 using Jint.Native.Object;
 using Jint.Runtime;
@@ -5,11 +6,11 @@ using Jint.Runtime.Descriptors;
 
 namespace Jint.Native;
 
-public sealed class JsSet : ObjectInstance
+public sealed class JsSet : ObjectInstance, IEnumerable<JsValue>
 {
     internal readonly OrderedSet<JsValue> _set;
 
-    public JsSet(Engine engine) : this(engine, new OrderedSet<JsValue>(SameValueZeroComparer.Instance))
+    internal JsSet(Engine engine) : this(engine, new OrderedSet<JsValue>(SameValueZeroComparer.Instance))
     {
     }
 
@@ -53,7 +54,7 @@ public sealed class JsSet : ObjectInstance
 
     public bool Has(JsValue key) => _set.Contains(key);
 
-    public bool SetDelete(JsValue key) => _set.Remove(key);
+    public bool Remove(JsValue key) => _set.Remove(key);
 
     internal void ForEach(ICallable callable, JsValue thisArg)
     {
@@ -74,4 +75,8 @@ public sealed class JsSet : ObjectInstance
     internal ObjectInstance Entries() => _engine.Realm.Intrinsics.SetIteratorPrototype.ConstructEntryIterator(this);
 
     internal ObjectInstance Values() => _engine.Realm.Intrinsics.SetIteratorPrototype.ConstructValueIterator(this);
+
+    public IEnumerator<JsValue> GetEnumerator() => _set.GetEnumerator();
+
+    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
 }
--- a/Jint/Native/JsSet.cs
+++ b/Jint/Native/JsSet.cs
@@ -5,7 +5,7 @@ using Jint.Runtime.Descriptors;
 
 namespace Jint.Native;
 
-internal sealed class JsSet : ObjectInstance
+public sealed class JsSet : ObjectInstance
 {
     internal readonly OrderedSet<JsValue> _set;
 
@@ -47,17 +47,17 @@ internal sealed class JsSet : ObjectInstance
         return base.TryGetProperty(property, out descriptor);
     }
 
-    internal void Add(JsValue value) => _set.Add(value);
+    public void Add(JsValue value) => _set.Add(value);
 
-    internal void Remove(JsValue value) => _set.Remove(value);
+    public void Remove(JsValue value) => _set.Remove(value);
 
-    internal void Clear() => _set.Clear();
+    public void Clear() => _set.Clear();
 
-    internal bool Has(JsValue key) => _set.Contains(key);
+    public bool Has(JsValue key) => _set.Contains(key);
 
-    internal bool SetDelete(JsValue key) => _set.Remove(key);
+    public bool SetDelete(JsValue key) => _set.Remove(key);
 
-    internal void ForEach(ICallable callable, JsValue thisArg)
+    public void ForEach(ICallable callable, JsValue thisArg)
     {
         var args = _engine._jsValueArrayPool.RentArray(3);
         args[2] = this;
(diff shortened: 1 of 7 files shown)


Jint/Native/JsSet.cs
```diff
@@ -47,17 +47,17 @@ protected override bool TryGetProperty(JsValue property, [NotNullWhen(true)] out
...
+    public void Remove(JsValue value) => _set.Remove(value);
 
-    internal void Clear() => _set.Clear();
+    public void Clear() => _set.Clear();
 
-    internal bool Has(JsValue key) => _set.Contains(key);
+    public bool Has(JsValue key) => _set.Contains(key);
 
-    internal bool SetDelete(JsValue key) => _set.Remove(key);
+    public bool SetDelete(JsValue key) => _set.Remove(key);
```
lahma, line 58: `Remove` (which should be `Delete`) already handles this so can be removed altogether.

Jint/Native/JsSet.cs
```diff
@@ -73,7 +73,7 @@ internal void ForEach(ICallable callable, JsValue thisArg)
         _engine._jsValueArrayPool.ReturnArray(args);
     }
 
-    internal ObjectInstance Entries() => _engine.Realm.Intrinsics.SetIteratorPrototype.ConstructEntryIterator(this);
+    public ObjectInstance Entries() => _engine.Realm.Intrinsics.SetIteratorPrototype.ConstructEntryIterator(this);
```
lahma, line 76: Are `Entries` and `Values` really needed, they don't expose meaningful API outside of Jint's logic at the moment?

Jint/Runtime/OrderedSet.cs
```diff
@@ -1,6 +1,6 @@
 namespace Jint.Runtime;
 
-internal sealed class OrderedSet<T>
+public sealed class OrderedSet<T>
```
lahma, line 3: I think this can remain as `internal`, unfortunate need for JS usage where they come up with oddities

Jint/Native/JsSet.cs
```diff
@@ -47,17 +47,17 @@ protected override bool TryGetProperty(JsValue property, [NotNullWhen(true)] out
         return base.TryGetProperty(property, out descriptor);
     }
 
-    internal void Add(JsValue value) => _set.Add(value);
+    public void Add(JsValue value) => _set.Add(value);
 
-    internal void Remove(JsValue value) => _set.Remove(value);
+    public void Remove(JsValue value) => _set.Remove(value);
```
lahma, line 52: I think this should be just plain `Delete`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

Jint/Native/JsSet.cs
```diff
@@ -47,17 +47,17 @@ protected override bool TryGetProperty(JsValue property, [NotNullWhen(true)] out
...
+    public void Clear() => _set.Clear();
 
-    internal bool Has(JsValue key) => _set.Contains(key);
+    public bool Has(JsValue key) => _set.Contains(key);
 
-    internal bool SetDelete(JsValue key) => _set.Remove(key);
+    public bool SetDelete(JsValue key) => _set.Remove(key);
 
-    internal void ForEach(ICallable callable, JsValue thisArg)
+    public void ForEach(ICallable callable, JsValue thisArg)
```
lahma, line 60: Maybe better keep this internal, very JS spec specific and not a natural c# API

lahma (changes requested):

Jint/Runtime/OrderedSet.cs
lahma: So `SetConstructor` would handle all constructor requirements and `JsSet` constructor should be internal, like I indicated in earlier discussion, `SetConstructor.Construct()` would have the needed overloads

Jint/Native/JsSet.cs
kenlyon: No, right enough. They're not really needed elsewhere.

Jint/Runtime/OrderedSet.cs
kenlyon: Unfortunately, I wasn't able to compile without this change. I think the alternative would be to downgrade the `JsSet` constructor that uses it to also be `internal`. This could potentially be a breaking change as I don't know where all it is used.

Jint/Native/JsSet.cs
kenlyon: You mean rename the method itself to `Delete(JsValue value)`? Given that it was previously internal, am I safe to assume that all existing consumers that could break as a result are in the same repo? Hopefully refactoring it globally is adequate.

kenlyon: OK, so I see it's only used in SetPrototype, however I wonder if the name "SetDelete" was used to avoid conflicting with ObjectInstance.Delete() which would attempt to delete a property by that name from the object (presumably like javascript's `delete foo.bar`) whereas we're wanting `mySet.delete("abc")` to take a value out of the set.

Jint/Native/JsSet.cs
kenlyon: I kept this name for now given that "Delete" is inherited. Alternatively I could use `new` to hide the inherited version although I think that's generally discouraged.

Jint/Native/JsSet.cs
kenlyon: Changed back to internal.

Jint/Runtime/OrderedSet.cs
kenlyon: I was able to revert this by changing the affected JsSet constructor to internal. Hopefully that doesn't break something elsewhere. I only see it used by SetPrototype.

Jint/Native/JsSet.cs
kenlyon: Done. This also eliminates a warning I was getting because `ICallable` is `internal`.

lahma (approved):
I did small adjustments and a public API test to show usage (and ensure API will be kept), did a 180 and went with the `Remove` method name as `SetDelete` felt too clunky...

lahma:
Thanks for iterating a publishing the API!

lahma:
In #1988 I did another 180 and came to conclusion that in this case the `new` hiding actually makes sense as you are working with `Set` and `Map` semantics with direct type, you probably don't want to do "generic object operations".

Jint/Runtime/OrderedSet.cs
```diff
@@ -1,6 +1,8 @@
+using System.Collections;
+
 namespace Jint.Runtime;
 
-internal sealed class OrderedSet<T>
+internal sealed class OrderedSet<T> : IEnumerable<T>
 {
     internal List<T> _list;
```
sebastienros, line 7: OrderedDictionary<T, T> has been added recently, or is it soon? https://learn.microsoft.com/en-us/dotnet/api/system.collections.specialized.ordereddictionary?view=net-8.0

lahma: The generic version ( https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.ordereddictionary-2 ) has been added NET 9, maybe not yet worth to add another TFM just for that..

kenlyon:
Thanks for finishing this. I've upgraded to Jint 4.1.0 and these changes are really helpful.

Jint/Runtime/OrderedSet.cs
sebastienros: That's what was confusing me then. I knew the generic one was added to 9, but on the docs it said 2,3....9. I was pointing to the wrong one. It all makes sense and I agree not worth it if just for 9.

PR #1997, opened 2024-11-04
Use Array.Resize when growing ArrayInstance

lahma:


SebastianStehle:
I am just curious. Is it that much faster than just using a list?

SebastianStehle:
I am not questioning your decisions, you are doing a wonderful job here ;) ... I am pretty sure, there is a good reason to use an array.

lahma:
> I am not questioning your decisions, you are doing a wonderful job here ;) ... I am pretty sure, there is a good reason to use an array.

Sure, no worries. I have wondered this myself as managing all manually is a real pain. But the thing is that List<T> brings abstraction and all the checks it has inside of it during method calls, Jint already does a lot of checking and accounting for length so it would be just another (abstraction) layer slowing down. Probably some silly milliseconds, but it's there.

PR #2000, opened 2024-11-09
Update benchmark results

lahma:


lofcz:
Why are most of the benchmarks looking worse? 👀

lahma:
> Why are most of the benchmarks looking worse? 👀

Different CPU, the results are meant to compare the engines, don't mind the absolute values.

PR #2028, opened 2025-01-16
Pin FluentAssertions to version 7.0.0

lahma:


sebastienros:
What made you change your mind? Logs?

lahma:
More realistic NuGet usage statistics for them 😉 7.x should receive bug fixes etc so probably better baseline. Haven't looked into other options but I really don't like xUnit Assert syntax.

PR #2033, opened 2025-01-20
Ensure that relative imports from module import are module-relative

rmja:
Consider the Engine configuration with enabled modules:

```
options.EnableModules("/modules-directory");
```

where the contents of `/modules-directory` is:
```
/modules-directory/my-module/index.js
/modules-directory/my-module/helper-file.js
```
The content of the index file references the helper file is:
```
export * from "./helper-file.js";
```

Note that the path to `./some-file.js` is relative to the index file.

If we then do
```
engine.Modules.Add("my-module", File.ReadAllText("/modules-directory/my-module/index.js"));
```

Then when calling into `my-module`, the `some-file.js` file should be found from the `/modules-directory/my-module` directory, and not from `/modules-directory`.

### DIFF
--- a/Jint.Tests/Runtime/Modules/DefaultModuleLoaderTests.cs
+++ b/Jint.Tests/Runtime/Modules/DefaultModuleLoaderTests.cs
@@ -21,6 +21,20 @@ public class DefaultModuleLoaderTests
         Assert.Equal(SpecifierType.RelativeOrAbsolute, resolved.Type);
     }
 
+
+    [Fact]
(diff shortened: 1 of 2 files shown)


rmja:
Okay, I can see from the failing test that this is actually the intended behavior?

lahma:
Generally the test cases establish the current API contract so I guess this would be a breaking change, unless the old behavior can be considered incorrect. 

One option is to provide more extension points if needed so that you can achieve your goal.

PR #2035, opened 2025-01-21
Fix statement counting code in MaxStatementsConstraint

nicolay-f:
Current implementation of MaxStatementsConstraint allows one more statement to be executed. The increment must be in prefix form so value is updated before being used.


lahma (approved):
Thank you!

PR #2049, opened 2025-01-29
Engine InvokeAsync is implemented for task/valuetask structure

baytekink:
Engine InvokeAsync is implemented for task/valuetask structure

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -25,6 +25,7 @@
     <PackageVersion Include="SourceMaps" Version="0.3.0" />
     <PackageVersion Include="Spectre.Console.Cli" Version="0.45.0" />
     <PackageVersion Include="System.Text.Json" Version="9.0.1" />
+    <PackageVersion Include="System.ValueTuple" Version="4.5.0" />
     <PackageVersion Include="Test262Harness" Version="1.0.1" />
     <PackageVersion Include="xunit" Version="2.9.3" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="3.0.1" PrivateAssets="all" />
--- a/Jint.AotExample/Jint.AotExample.csproj
+++ b/Jint.AotExample/Jint.AotExample.csproj
@@ -6,6 +6,7 @@
     <Nullable>enable</Nullable>
     <IsPackable>false</IsPackable>
     <PublishAot>true</PublishAot>
+    <NoWarn>$(NoWarn);NU1507</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Jint.Repl/Jint.Repl.csproj
+++ b/Jint.Repl/Jint.Repl.csproj
@@ -4,6 +4,7 @@
     <OutputType>Exe</OutputType>
     <IsPackable>false</IsPackable>
     <PublishAot>true</PublishAot>
+    <NoWarn>$(NoWarn);NU1507</NoWarn>
   </PropertyGroup>
   <ItemGroup>
     <ProjectReference Include="..\Jint\Jint.csproj" />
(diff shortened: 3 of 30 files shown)


baytekink:
What did I do : Remove ManualResetEvent, instead TaskCompletionSource<bool> is added. So we give the thread back to thread pool. After that, I added new methods for InvokeAsync, I generally copied the sequantial code and convert it into Task structure. Most of the components/classes uses sequantial methods, but JintAwaitExpression or JintAwaitxxx classes override the async methods. I may call ...Async methods from the sequantial ones (using GetAwaiter().GetResult()), but I did not do that, so we can discuss if we call Async methods by GetAwaiter().GetResult() or do we copy the nearly same code with ...Async ending method.

SebastianStehle:
I will give you a bonus if this does work and would make async calls easier. But I am afraid it will kill performance. Have you added a benchmark and have you considered ValueTask?

baytekink:
Thanks Sebastian, if you will give a chance, I will work with ValueTask if possible for performance, and would improve the code. I did not add any benchmarks cause I did not work enough on this (I suspect if you reject my work actually, so I dont want to work much on this 🙂 ) Hence, if you are serious about adding native Async, I will reengineer again about ValueTask, benchmark etc.

________________________________
Gönderen: Sebastian Stehle ***@***.***>
Gönderildi: 2 Şubat 2025 Pazar 20:54
Kime: sebastienros/jint ***@***.***>
Bilgi: baytekink ***@***.***>; Author ***@***.***>
Konu: Re: [sebastienros/jint] Engine InvokeAsync is implemented for task/valuetask structure (PR #2049)


I will give you a bonus if this does work and would make async calls easier. But I am afraid it will kill performance. Have you added a benchmark and have you considered ValueTask?

lahma:
My main concern here is all the copy-paste, I'm not willing to maintain two copied versions.

lofcz:
A lot of the code currently copy pasted could be generated, I can look into that if the other issues are solved.

baytekink:
Ok, I try to write that i copied the code, but i can eliminate and generate one code, at the same time i will try to use Valuetask. After that i will write to you to comment. Thx for your comments.

baytekink:
> My main concern here is all the copy-paste, I'm not willing to maintain two copied versions. Also of course, performance should not degrade.

Returning Task is changed by ValueTask for performance. 

Duplicate methods reduced to single: While calling async code from sequantial code, Preserve() method is used for performance (This method is similar to [AsTask()], but it returns the same [ValueTask] instance when this [ValueTask] represents a successful synchronously completed operation ). For ex:

 public JsValue Invoke(JsValue value, object? thisObj, object?[] arguments)
 {
     return InvokeAsync(value, thisObj, arguments).Preserve().GetAwaiter().GetResult();
 }

If javascript code does not include any async/await code, this code will continue to work synchronously. The only difference it will create a ValueTask on stack. But it wont be a problem for GC since it was created on stack.

I have not found time to benchmark this code, I will try on this week, but while I try to benchmark, It would be nice if you could give comments on PR. Thx in advance

baytekink:
> I will give you a bonus if this does work and would make async calls easier. But I am afraid it will kill performance. Have you added a benchmark and have you considered ValueTask?

I now changed Task to ValueTask. Actually i do not think (I wrote think because teorically ValueTask wont be a problem since it is created on stack) it will decrease the performance, but we must see it by benchmarks of course. Thx in advance

lahma:
Async state machinery will add a lot of strain for performance. So I wouldn't keep my hopes up with this async API...

lahma:
We have a less invasive solution with #2274 now, hopefully it addresses the requirements without negative performance impact. Than you for the PR though, please propose improvements if still unaddressed.

PR #2053, opened 2025-02-11
Add support for object return values in Func callbacks

martinbu:
If a Func defines a none value-type return type that does not implement IConvertible a InvalidCastException was thrown. This update skips Convert.ChangeType if the returned value is not IConvertible and just casts the result.


lahma (approved):
Thank you!

Issue #2057, opened 2025-02-18
Incorrect calculation of `JSDate` flags for `DateTime.MaxValue`?


nil4:
**Version used**

4.2.0

**Describe the bug**

The `JsDate(Engine engine, DateTime value)` constructor sets flags for `DateTime.{Min,Max}Value`:

https://github.com/sebastienros/jint/blob/f012c8398fbfd0dddd377114169d4d67be210817/Jint/Native/JsDate.cs#L23-L30

On line 29,  flags are set to `DateFlags.DateTimeMaxValue` (note <mark>`Max`</mark>), while the test on line 27 above actually checks for `DateTime.MinValue` (note <mark>`Min`</mark>).

It doesn't seem possible to end up with `DateFlags.DateTimeMaxValue` flags set with the current logic. I'm not sure if this may lead to issues later on, but thought it's worth reporting in case it matters.

lahma:
Could you send a PR fixing the issue with test backing the fix to ensure no regressions later?

PR #2058, opened 2025-02-19
Remove apparently-redundant `DateFlags` calculation in `JsDate` constructor

nil4:
Added tests for `DateFlags.DateTime{Min,Max}Value` ref. https://github.com/sebastienros/jint/issues/2057#issuecomment-2666563547

Tests did not fail, indicating this code may be redundant.  

The `JsDate` constructor calls `engine.Realm.Intrinsics.Date.FromDateTime` i.e. this bit which already sets the flags correctly:

https://github.com/sebastienros/jint/blob/f012c8398fbfd0dddd377114169d4d67be210817/Jint/Native/Date/DateConstructor.cs#L197-L207

No other tests failed locally with this change. I hope I haven't overlooked anything else and this resolves https://github.com/sebastienros/jint/issues/2057

### DIFF
--- a/Jint.Tests/Runtime/DateTests.cs
+++ b/Jint.Tests/Runtime/DateTests.cs
@@ -1,3 +1,5 @@
+using Jint.Native;
+
 namespace Jint.Tests.Runtime;
 
 public class DateTests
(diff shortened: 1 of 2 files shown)


lahma (approved):
Thank you for your contribution!

PR #2060, opened 2025-02-21
Improve function arguments parameter expression detection

lahma:
Currently we don't detect expressions properly which causes extra environment to be created even though it isn't necessary. Requires more AST analysis.

### DIFF
--- a/Jint.Benchmark/FunctionBenchmark.cs
+++ b/Jint.Benchmark/FunctionBenchmark.cs
@@ -1,0 +1,27 @@
+using BenchmarkDotNet.Attributes;
+
+namespace Jint.Benchmark;
+
+[MemoryDiagnoser]
+public class FunctionBenchmark
+{
+    private readonly Engine _engine;
+
+    public FunctionBenchmark()
+    {
+        _engine = new Engine();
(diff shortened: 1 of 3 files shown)


lahma:
A modest ~4% improvement to to https://github.com/adams85/babeljintpoc/blob/master/3-benchmark/ParseJsBenchmark.cs 

### main

| Method        | Mean         | Error      | StdDev     | Gen0       | Gen1       | Gen2      | Allocated |
|-------------- |-------------:|-----------:|-----------:|-----------:|-----------:|----------:|----------:|
| BabelParse    | 3,699.159 ms | 19.5511 ms | 15.2642 ms | 92000.0000 | 21000.0000 | 3000.0000 | 1432.5 MB |

## this branch

| Method        | Mean         | Error      | StdDev    | Gen0       | Gen1       | Gen2      | Allocated  |
|-------------- |-------------:|-----------:|----------:|-----------:|-----------:|----------:|-----------:|
| BabelParse    | 3,561.074 ms | 10.9282 ms | 8.5320 ms | 92000.0000 | 21000.0000 | 3000.0000 | 1429.67 MB |

PR #2100, opened 2025-05-09
Fix ObjectAccessBenchmark test

mac-michael:
Removed excessive semicolon


sebastienros:
V8 complains too:

image

@adams85

lahma (approved):
This was actually broken as-is so Acornima doesn't parse this either (as expected). Thanks for the PR.

PR #2103, opened 2025-05-18
Add default 10 second timeout for promise unwrapping

lahma:
This protects from errors that could hang the engine.


qihangnet:
The timeout set on the engine is ineffective, and many of our tasks cannot run properly.
@lahma Can you fix it? 😂

Issue #2099, opened 2025-05-08
[Bug] Closure values are also cached with function instances and lead to wrong behavior


warappa:
**Version/platform/runtime used**

Jint: 4.2.2

**Describe the bug**
It seems the last performance optimization for dictionaries (#2087) may have introduced a hard to catch closure caching issue (or was an un-hider).

As seen in the reproduction below, when caching the C# representation of a Jint Javascript function, the first call with it with the initial parameters succeeds. But the second call with another search value (but same list to search) fails, although it should also succeed.

Logging shows, that the outer search value `kind` and the captured value of `kind` inside the function (or lambda) of the `find(...)` function.

**To Reproduce**
It was not easy to get a condensed reproduction. Somehow it was important to also pass the `values` list from C# down to Javascript and not only the search value. 🤷‍♂️

`outer kind` and `closure kind` should always show the **same values**.

```csharp
using Jint;
using Jint.Native;

var engine = new Engine();
engine.SetValue("log", (string message) => Console.WriteLine(message));
engine.Evaluate(
    """
    globalThis.findIt = function(array, kind) {           
        log("find start");
        log("outer kind:         " + kind);
        //return array.find(x => x.kind == kind);                    
        let found = array.find(function (x) {
            log("closure kind:       " + kind);
            log("x.kind:             " + x.kind);
            log("");
            return x.kind == kind;
        });
        return found;
    };
    """);

var func = (Func<JsValue, JsValue[], JsValue>)engine.Evaluate("""
    function functionToBeConsumedByCSharp(values, blub) {
    
    let found = globalThis.findIt(values, blub);
    if (!found) { 
        log("blub " + blub);
        throw new Error('value not found!');
    }
    
    return found;
}
functionToBeConsumedByCSharp;
""").ToObject()!;

var values = new List<object> { new { kind = 'a' }, new { kind = 'b' } };

var found = func(JsValue.Undefined, [JsValue.FromObject(engine, values), JsValue.FromObject(engine, "a")]).ToObject();
if (found != values[0])
{
    throw new InvalidOperationException("Should have found first element");
}

found = func(JsValue.Undefined, [JsValue.FromObject(engine, values), JsValue.FromObject(engine, "b")]).ToObject();
if (found != values[1])
{
    throw new InvalidOperationException("Should have found second element");
}
```

Output:
```log
find start
outer kind:         a
closure kind:       a
x.kind:             a

find start
outer kind:         b
closure kind:       a
x.kind:             a

[...]

PR #2105, opened 2025-05-21
#2099: Fix erroneous caching by providing each function instance with it own interop variant

warappa:
- Effectively revert delegate cache optimization
  - Each function instance _requires_ its own .NET representation to handle parameters correctly (-> function environment)
- Use a factory method to reduce `System.Linq.Expression` tree usage and therefore also the need for compiling them.
  - Compile an expression tree that allows to retarget a delegate to the actual Function instance (1 per AST, not per function instance).
  - Ensure returning the same delegate instance to allow for example unregistering handlers
- Add 2 tests
  - Ensure multiple invokations using .NET interop don't cache function environment
  - Ensure multiple invokations using JavaScript-only functions don't cache function environment


lahma (approved):
Awesome work, great job with test coverage!

PR #2123, opened 2025-06-22
Bugfix/2106 make eval syntax checks not descend into function declarations

### LINKED ISSUE #2106: eval descends into functions, resulting in errors about new.target or super use.
**Version/platform/runtime used**

Git tag v4.2.2

**Describe the bug**

When using eval to load a class that checks constructor use, I'm getting the error
`Error: new.target expression is not allowed here`
This is thrown inside PerformEval after checking the script not to contain things like (in this case) new.target
However, according to https://tc39.es/ecma262/#sec-static-semantics-contains function declarations should not be considered for contains.

**To Reproduce**

Run this script:
```Javascript
const code = `function MyClass() {
   if (!new.target) throw new Error("Use MyClass as constructor!");
   console.log("constructor ok");
}`

eval(code);
const x = new MyClass();
console.log("ok");
```

**Expected behavior**

Instead of the error, the two log messages should appear. This is e.g. what happens when running the script using node.

**Additional context**

None

rg1:
According to https://tc39.es/ecma262/#sec-static-semantics-contains, function declarations should not be considered for contains checks. 
I'm adding visitor methods in the eval check, that simply skip those and a test showing the now working eval.

fixes #2106

### DIFF
--- a/Jint.Tests/Runtime/EngineTests.cs
+++ b/Jint.Tests/Runtime/EngineTests.cs
@@ -598,6 +598,19 @@ public partial class EngineTests : IDisposable
     }
 
     [Fact]
+    public void EvalFunctionWithTargetNewParse()
+    {
+        RunTest(@"
+                const code = `function MyClass() {
+                   if (!new.target) throw new Error('Use MyClass as constructor!');
+                }`;
+
+                eval(code);
+                const x = new MyClass();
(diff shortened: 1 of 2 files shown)


lahma (approved):
LGTM, thank you for your contribution!

PR #2124, opened 2025-06-23
Ability to use Script and Module ASTs directly

ejsmith:
I would like to be able to have the ability to use other parsers or augment existing parsers and provide Jint with the AST directly. The main use case I'm trying to solve for is I would like to have a TypeScript parser that strips out any TypeScript code while preserving all source locations and pass that directly into Jint as a normal JavaScript AST. Thoughts?


SGStino:
A use case for this would be to inspect an AST, modify it, and execute that. Instead of going back to string code after the modification, and parse that again?

sebastienros:
Integrated against the current `main` branch in #3889 and merged as `733a65d429325b103435b85ab07cd334faafc0fa`. The replacement preserves Eric J. Smith's contribution with a `Co-authored-by` trailer. Thank you for the original implementation.

ejsmith:
For anyone interested, I created this project to support TypeScript in Jint now that Jint has opened their API to allow using ASTs parsed outside of their library. https://github.com/FoundatioFx/Jint.TypeScript

sebastienros:
Of course we know ;) Being the initial designer of the Orchard Workflows (which was the base of Elsa Workflows) I was interested. Your website also led me to create the Jint static documentation, I just asked an agent to copy the tech stack (it even copied "Foundatio" somwhere in the logo)

I hope that this PR getting merged will help you

ejsmith:
Thank you @sebastienros for opening up the API and making this possible! Assuming you are talking about the [Jint.Workflows](https://github.com/FoundatioFx/Jint.Workflows) project? I am a big fan of Elsa, Orchard, Jint and just about everything you've worked on. :)

PR #2151, opened 2025-07-23
Fix iterating a clr list proxy and length property

martinbu:
When iterating over a JsProxy of a clr list or access its length property, jint fails with `System.InvalidCastException : Unable to cast object of type 'Jint.Native.JsProxy' to type 'Jint.Runtime.Interop.ObjectWrapper'.`

This PR fixes the issue - but I am pretty sure, that there is a "better" more solid solution for it?


lahma (approved):
I think this a good solution for now, there are always corner cases for interop. Thank you for your contribution!

PR #2183, opened 2025-10-15
Fix Object.is when providing BigInt values

tomatosalat0:
`Object.is` uses `JsValue.SameValue()` to compare two values. However, `JsValue.SameValue()` did not compare `JsBigInt` values. 

The new test method `ObjectIsReturnsTrueForSameBigInts` fails without the change to `JsValue.SameValue()`;


tomatosalat0:
Looks like that fixed it - almost 😆. Looks like a flaky test now got in my way.

lahma (approved):
LGTM, thank you!

PR #2196, opened 2025-11-09
Allow garbage collector to collect unused memory

tomatosalat0:
When having a long living `Engine` object, a few internal of Jint caches can prevent .NET from garbage collecting values which are not used anymore. 

The added test fails without the changes.

## RefStack

This stack is a simple array internally, but removing items from it does not clear the value within the array itself, meaning that the value will not be garbage collected by .NET until the value within the array will be overwritten by a new one - which is not guaranteed to happen.

## JsArrayPool

When an array is cached within the JsArrayPool, its content was not cleaned, meaning that the values of the array would not be able to get garbage collected by the .NET runtime. 

I've benchmarked the use of `Array.Fill()` vs

```cs
array[0] = null!;
array[1] = null!;
array[2] = null!;
```

and `array[0] = null!; ...` won by a large margin. The same optimization is not worth it within `RefStack.Clear()` in my opinion because that one is only used in case an Exception happened during script execution, so I wouldn't call that a hot path.

## Test

I tried to make the test not flaky, fingers crossed that it works. The expected before&after behavior of the changes was verified by using dotMemory.

### DIFF
--- a/Jint.Tests/Runtime/AsyncTests.cs
+++ b/Jint.Tests/Runtime/AsyncTests.cs
@@ -1,5 +1,6 @@
 using System.Collections.Concurrent;
 using Jint.Native;
+using Jint.Runtime;
 using Jint.Tests.Runtime.TestClasses;
 
 namespace Jint.Tests.Runtime;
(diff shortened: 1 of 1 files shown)


tomatosalat0:
I am using the MyGet version anyway, so personally I don't _need_ these fixes from Nuget. At the moment, I don't have other open issues I can create PRs for, so from my side it is good to go.

lahma:
Jint 4.4.2 has been published with your recent improvements 🚀

PR #2208, opened 2025-12-12
Fix: Handle unicode escape sequences in regex literals when Acornima returns null Regex

scm1400:
## Summary
  - Add fallback handling for regex literals containing unicode escape sequences (e.g., `\u2028`, `\u2029`) that fail to compile via Acornima's `ParseResult.Regex`
  - This fixes a regression from Jint 3.x (Esprima) to Jint 4.x (Acornima) where certain regex patterns would throw `NullReferenceException`

  ## Problem
  When parsing regex literals like `/[\u2028\u2029]/g`, Acornima's `ParseResult.Regex` returns `null` even though `ParseResult.Success` is `true`. This causes a `NullReferenceException` when the regex is used in operations like `String.prototype.replace()`.

  **Error:**
  System.NullReferenceException: Object reference not set to an instance of an object.
     at Jint.Native.RegExp.RegExpPrototype.Replace(JsValue thisObject, JsValue[] arguments)

  ## Solution
  Added a fallback in `JintLiteralExpression.ResolveValue()` that:
  1. Detects when `ParseResult.Regex` is null
  2. Manually converts unicode escape sequences (`\uXXXX`) to actual characters
  3. Creates the .NET `Regex` object with appropriate options

  ## Changes
  - `Runtime/Interpreter/Expressions/JintLiteralExpression.cs`
    - Added `UnescapeUnicode()` helper method
    - Added fallback logic when `regExpParseResult.Regex` is null

  ## Test
  ```javascript
  // Previously failed, now works
  const result = "test\u2028value".replace(/[\u2028\u2029]/g, "");

### DIFF
--- a/Jint/Runtime/Interpreter/Expressions/JintLiteralExpression.cs
+++ b/Jint/Runtime/Interpreter/Expressions/JintLiteralExpression.cs
@@ -7,8 +7,24 @@ internal sealed class JintLiteralExpression : JintExpression
 {
     private static readonly object _nullMarker = new();
 
+    // Regex to match unicode escape sequences like \u2028, \u2029
+    private static readonly Regex UnicodeEscapePattern = new Regex(@"\\u([0-9A-Fa-f]{4})", RegexOptions.Compiled);
+
     private JintLiteralExpression(Literal expression) : base(expression)
     {
+    }
+
+    /// <summary>
+    /// Converts unicode escape sequences (e.g., \u2028) to actual unicode characters.
+    /// </summary>
+    private static string UnescapeUnicode(string pattern)
+    {
+        return UnicodeEscapePattern.Replace(pattern, match =>
+        {
+            var hex = match.Groups[1].Value;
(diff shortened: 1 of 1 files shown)


lahma:
Can you please also add a test case showing the problem?

adams85:
> Add fallback handling for regex literals containing unicode escape sequences (e.g., \u2028, \u2029) that fail to compile via Acornima's ParseResult.Regex

Can't reproduce the issue.

Acornima converts `/[\u2028\u2029]/` to an equivalent .NET `Regex` as expected.

Running `const result = "test\u2028value".replace(/[\u2028\u2029]/g, "");` in Jint.Repl returns the expected result. (Tested with Jint v4.4.2 + .NET 4.7.2/.NET 8/.NET 10 + without AOT / with AOT + Windows/Linux.)

> This fixes a regression from Jint 3.x (Esprima) to Jint 4.x (Acornima) where certain regex patterns would throw NullReferenceException

Please share a minimal reproducible example with us.

Also, describe your exact environment (OS, .NET version, Jint version, etc.)

lahma:
Closing due to inactivity, we can reopen if new information arrives.

PR #2251, opened 2026-01-17
Implement Internalization API

lahma:


lahma:
Iterating and testing approaches, unfortunately BCL doesn't seem to have all the needed CLDR data avaliable...

PR #2286, opened 2026-02-15
Add Copilot instructions

lahma:


sebastienros (commented):
Ideally there should be a section on his to run the benchmarks and add new ones, so it can also irritate iterate on these. Preferably with short jobs.

PR #2293, opened 2026-02-20
Refactor version extraction in release workflow

MikeAlhayek:


lahma:
Thanks for this, and sorry it sat open so long.

This has since been fixed as part of #2748, which reworked the release workflow for nuget.org trusted publishing and included the same `${GITHUB_REF_NAME#v}` change you proposed here — the old `${VERSION//v}` stripped *every* `v` in the tag rather than just the leading one, so a tag like `v5.0.0-preview.1` would have been mangled.

The version is now derived in a dedicated step and, beyond `Version`, also feeds `AssemblyVersion` and `FileVersion`, so the tag is the single source of truth for a release:

https://github.com/sebastienros/jint/blob/main/.github/workflows/release.yml

Closing as the change has landed — thanks for spotting it.

PR #2299, opened 2026-03-01
Fix accessing properties defined by a bound function

lkosson:
When an object property is defined as a getter, which in turn is a bound function, accessing the property results in `InvalidCastException` in `ObjectInstance.UnwrapFromGetter`, because `getter` variable is `Jint.Native.Function.BindFunction` instead of `Jint.Native.Function.Function`.

Explicit test for `BindFunction` is not ideal and perhaps `ICallable` should be used, but I'm not sure how to properly pass it to `Engine`.

Test code to reproduce the behavior:

```
Console.WriteLine(new Engine().Evaluate("""
	const holder = {
		x: 42,
		getter() { return this.x; }
	};

	const target = {};

	Object.defineProperty(target, "prop", { get: holder.getter.bind(holder) });

	target.prop;
"""));
```

Before fix: throws `InvalidCastException`
After fix: prints out `42`

### DIFF
--- a/Jint/Native/Object/ObjectInstance.cs
+++ b/Jint/Native/Object/ObjectInstance.cs
@@ -419,6 +419,8 @@ public partial class ObjectInstance : JsValue, IEquatable<ObjectInstance>
             return Undefined;
         }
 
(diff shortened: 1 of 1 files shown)


lahma:
Let's see copilot's review, also each bug fic should contain also a test case proving the fix.

PR #2393, opened 2026-04-12
Regexp parsing cleanup

adams85:
I was playing with Jint to test the brand new flag v mode and when I looked at the code, I noticed some minor problems.

This PR addresses my findings:
- suggests reverting to Acornima's JS to .NET regex conversion, which would make it possible to remove a big chunk of code,
- reworks NeedCustomEngine, which was pretty sloppy and inefficient,
- cleans up and improves `ParsingOptions` and `PreparationOptions`,
- improves performance by deferring regexp conversion until execution,
- updates readme (as Jint can now match emojis like the green water pistol or gender-neutral purple-skinned families :D)

(As a matter of fact, I also found some previously unknown bugs in the JS to .NET regex conversion logic. I put workarounds in place so tests pass, but I'm going to fix these on Acornima's side soon.)


lahma (approved):
Awesome, thank you!

PR #2398, opened 2026-04-13
Improve fallback to custom engine detection logic for regexps

adams85:
As it turned out, the JS to .NET regexp conversion had a few bugs so blatant that I'm considering backporting the fixes to Esprima... 😅 Anyway, the correction is released in [Acornima v1.5.0](https://github.com/adams85/acornima/releases/tag/v1.5.0).

This PR upgrades to this version and cleans up workarounds that were necessary because of these bugs.


lahma (approved):
Great, thank you!

PR #2404, opened 2026-04-19
Update README.md

adams85:
Updates and clears up an outdated part of the readme.


lahma:
Good catch, thank you!

PR #2410, opened 2026-04-25
Drop redundant out-var reset after int.TryParse failure

lahma:
## Summary

\`int.TryParse\` already guarantees the out variable is set to \`0\` on failure (.NET BCL contract since v1), so the explicit \`displayMonth = 0;\` reset inside the failure branch of \`NonIsoCalendars.TryValidateMonthCode\` is dead code. One-line cleanup caught in the #2409 review.

## Test plan
- [x] \`dotnet build --configuration Release Jint/Jint.csproj\` clean (4 target frameworks).
- [x] No behavior change.


lahma:
Subsumed by #2411 — the one-line cleanup is now part of that PR's commits.

PR #2415, opened 2026-04-25
Upgrade to Acornima v1.6.1

adams85:
In flag v regexp validation, I found a bug that incorrectly rejects quantified modifier groups (e.g., `/(?i:x)?/v`).

I released the [fix](https://github.com/adams85/acornima/commit/6ebd1d2792cb7460d7d80ed3c33b3338af0eb667) in [Acornima v1.6.1](https://github.com/adams85/acornima/releases/tag/v1.6.1).

This PR upgrades Acornima to the patched version.

PR #2418, opened 2026-04-26
Drop Interlocked.Exchange in single-threaded pool hot paths

lahma:
## Summary

The slot/env/block-env pools added in #2413 (`FunctionEnvironment` pool, `JintBlockStatement._cachedEnv`, `JintFunctionDefinition.State._cachedSlots/_cachedEnv`) all rent and release using `Interlocked.Exchange`. But `Engine` instances are documented as not thread-safe for concurrent script execution — those atomic ops are pure defensive overhead in the only supported usage.

This PR replaces the atomic with plain field access in the three pool sites:
- `JintEnvironment.NewFunctionEnvironment` — env + slot rent
- `ScriptFunction.Call` finally — env + slot release
- `JintBlockStatement.ExecuteBlock` — block-env rent + release

## Why it's safe

Worst-case race in a multi-thread misuse: a tiny extra allocation when two threads both rent and one allocates fresh; never corruption. The existing `ReferenceEquals(_engine, engine)` guard at rent sites already protects against cross-engine reuse if a pooled object happens to land in the wrong engine.

If thread-safety is ever needed, a single boolean on `Engine` plus a branch at the pool sites would be a cleaner reintroduction than the current always-on atomics.

## Performance

Per-op savings of ~2–3 atomic operations are nanoseconds — well below the noise floor of the `stopwatch.js` benchmark on this hardware (~1–2 ms StdDev within a run, ±5% run-to-run variance). The change is **free** at this level of measurement; no regression visible. Ship for code-quality / removing-misleading-thread-safety reasons rather than a chase-able number.

## Validation

- `Jint.Tests` (net10.0): 2861 passed, 0 failed.
- `Jint.Tests.Test262` (net10.0): 98567 passed, 506 skipped, 0 failed (flaky decodeURI timeouts under concurrent test load — known pre-existing pattern, irrelevant to this change).

Rebased on top of latest `main` (post-#2412 / #2413 / #2415 / #2416 / #2417).

## Stacked work

Fourth of six related performance changes. Remaining: `perf/skip-empty-callee-env`, `perf/block-env-reset` queued on `lahma/jint`.

### DIFF
--- a/Jint/Native/Function/ScriptFunction.cs
+++ b/Jint/Native/Function/ScriptFunction.cs
@@ -1,4 +1,3 @@
-using System.Threading;
 using Jint.Native.Object;
 using Jint.Runtime;
 using Jint.Runtime.Descriptors;
@@ -122,17 +121,16 @@ public sealed class ScriptFunction : Function, IConstructor
             {
                 if (funcEnv is not null)
                 {
-                    // Cache the slot array for reuse by next call to the same function (thread-safe)
+                    // Pool returns: plain writes are sufficient — the engine is single-threaded for a
+                    // given execution; a concurrent misuse would at worst lose a pooled object (one
+                    // extra allocation), never produce stale state.
                     if (funcEnv._slots is { } slots)
                     {
                         System.Array.Clear(slots, 0, slots.Length);
-                        Interlocked.Exchange(ref state!._cachedSlots, slots);
+                        state!._cachedSlots = slots;
                         funcEnv._slots = null;
                     }
-
-                    // Return the env itself to the per-State pool so the next call to a function
-                    // sharing this State (typically the same Function instance) avoids the allocation.
-                    Interlocked.Exchange(ref state!._cachedEnv, funcEnv);
+                    state!._cachedEnv = funcEnv;
                 }
                 _engine.LeaveExecutionContext();
             }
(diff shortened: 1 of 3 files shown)


lahma:
Closing — CI revealed a real correctness regression on the Linux ARM and macOS runners. Root cause:

`Jint.Tests.Test262/TestHarness.cs` prepares harness scripts (assert.js, sta.js, etc.) **once** into a static dictionary and reuses them across all parallel-running test fixtures. AST `UserData` holds `JintFunctionDefinition.State`, so the per-State pool fields (`_cachedSlots`, `_cachedEnv`, and the equivalents on `JintBlockStatement.BlockState`) end up genuinely shared between parallel engines.

Without `Interlocked.Exchange` two parallel threads can read the same cached env reference and reset/use it concurrently — `LexicalEnvironment._dictionary.Clear()` racing with another thread's identifier read produced the NREs in `JintIdentifierExpression.GetValue` and `ScriptFunction.Call` traces.

The "engine is single-threaded" doc applies to a single `Engine` instance. It does **not** apply to the State objects, which can be shared via prepared-script reuse — exactly the supported pattern (`Engine.PrepareScript` once, multiple engines).

The Interlocked.Exchange is correct as it stands. Per-call savings of ~3 atomic ops are nanoseconds, not worth introducing this kind of subtle hazard. Abandoning.

PR #2424, opened 2026-04-26
Add Engine.ResetState() for lightweight engine reuse

desty2k:
## Summary

- Add `Engine.ResetState()` that resets an engine to a clean execution state while preserving Intrinsics, Options, Parser, and CLR metadata caches
- Capture execution generation at promise registration time so late-arriving `Task.ContinueWith` callbacks from a prior cycle are silently dropped
- Track outstanding async waiters (`_pendingAsyncOperations`) — `ResetState` throws `InvalidOperationException` if called while an `EvaluateAsync` Task is pending
- Add `Host.RebuildGlobals()` internal helper that re-invokes `PostInitialize()` on reset, keeping the `Host` extension point working
- Prevent extension method wrapper chain buildup on repeated resets
- Move `ModuleOperations` initialization before configuration callbacks in `Options.Apply` so `options.Configure(e => e.Modules.Add(...))` works
- Extract `ClearObjectWrapperCache()` private helper (used by both `Dispose()` and `ResetState()`)
- Update README with usage guidance in Performance and Thread-safety sections

## Motivation

When embedding Jint in scenarios that require many short-lived engine instances (e.g., executing user scripts in a server loop, tool dispatch), the cost of constructing a new `Engine` per invocation adds up — especially the Intrinsics (Object/Function constructors + prototypes), Parser, ExtensionMethodCache, and option-derived fields. None of these change between invocations.

`ResetState()` allows callers to reuse an engine instance from a pool by clearing all execution state and rebuilding only the GlobalObject and GlobalEnvironment.

## Benchmark — three workload profiles

Light scripts (tool dispatch: parse args, format output), heavy scripts (5K-record processing), and I/O-bound scripts (async CLR calls — the most realistic server-side pattern):

```
| Method               | Mean          | Gen0       | Gen1      | Allocated   | Alloc Ratio |
|--------------------- |--------------:|-----------:|----------:|------------:|------------:|
| Light_NewEngine_10x  |     293.14 us |    86.9141 |    5.3711 |   713.38 KB |        1.00 |
| Light_ResetState_10x |     242.18 us |    58.5938 |   19.5313 |   479.42 KB |        0.67 |
| Heavy_NewEngine_10x  | 154,994.82 us | 13000.0000 | 5250.0000 | 94338.21 KB |      132.24 |
| Heavy_ResetState_10x | 153,620.74 us | 13250.0000 | 5750.0000 | 93971.49 KB |      131.73 |
| IO_NewEngine_10x     |     141.11 us |    73.2422 |    2.9297 |   605.42 KB |        0.85 |
| IO_ResetState_10x    |      92.73 us |    44.9219 |    4.8828 |   369.95 KB |        0.52 |
```

- **Light scripts**: 17% faster, 33% fewer allocations — setup is a meaningful fraction
- **Heavy scripts**: ~0% difference — script execution completely dominates, setup is noise
- **I/O-bound (async CLR calls)**: 34% faster, 39% fewer allocations — JS CPU work is minimal, engine setup is the main cost

The feature is most valuable for I/O-bound and light workloads where engine construction is a significant fraction of CPU time per invocation.

## What ResetState clears

- Execution state: `_completionValue`, `_activeEvaluationContext`, `_error`, `_lastSyntaxElement`
- Execution context stack and call stack
- Event loop (drains pending events)
- GlobalObject and GlobalEnvironment (rebuilt fresh via `Host.RebuildGlobals`)
- `Host.PostInitialize()` re-invoked
- Realm template map
- Global symbol registry (user-created `Symbol.for()` entries)
- Module cache (via `Options.Apply` which creates fresh `ModuleOperations`)
- Object wrapper identity cache

## What ResetState preserves

- Intrinsics (ObjectConstructor, FunctionConstructor, all lazy-initialized built-in constructors and prototypes)
- Options, Parser, ParserOptions
- ExtensionMethodCache
- TypeCache and reflection accessor cache (deterministic CLR metadata)
- Constraint definitions
- Debugger state (breakpoints, step mode)

## Stale async callback protection

When `ResetState()` is called after an async execution completes, the previous execution's `Task.ContinueWith` callbacks may still fire from background threads. These closures capture the execution generation at promise registration time. If the generation has been bumped by `ResetState`, the settle closure returns early without enqueueing stale work onto the event loop.

## Safety guards

- Throws `InvalidOperationException` if called during synchronous script execution (`_activeEvaluationContext` check)
- Throws `InvalidOperationException` if called while an `EvaluateAsync`/`ExecuteAsync`/`InvokeAsync` Task is still pending (`_pendingAsyncOperations` check)

## Caveats

- Built-in prototype mutations from prior scripts persist (e.g., `Array.prototype.custom = ...`). For untrusted scripts, create a new engine instead.
- Programmatic module registrations (`engine.Modules.Add(...)`) are cleared on reset. Modules registered via `options.Configure(e => e.Modules.Add(...))` are automatically re-applied.
- Custom `Host` subclasses should place per-execution setup in `PostInitialize` (re-invoked on reset), not `InitializeHostDefinedRealm` (construction-only).

## Test plan

- [x] 52 dedicated tests covering sync and async paths
- [x] Global var/let/const/function clearing
- [x] Built-in object and prototype preservation
- [x] Prepared scripts work after reset
- [x] CLR delegates, objects, and type references after reset
- [x] Multiple reset cycles (10 iterations)
- [x] Mixed sync/async cycles
- [x] EvaluateAsync, InvokeAsync, Promise chains, Promise.all
- [x] Task interop (ExperimentalFeature.TaskInterop) after reset
- [x] Stale async callback dropping — observable flag assertion
- [x] Error recovery (throw → reset → new script)
- [x] Recursion overflow recovery
- [x] eval(), closures, generators, async generators, Proxy/Reflect
- [x] Strict mode preservation
- [x] Guard: throws InvalidOperationException during sync execution
- [x] Guard: throws InvalidOperationException during pending async evaluation
- [x] Custom Host with PostInitialize survives reset
- [x] Module registrations via Options.Configure survive reset
- [x] Module registrations via engine.Modules.Add cleared on reset
- [x] Full test suite regression check (2914 passed, 0 failures)

### DIFF
--- a/Jint.Benchmark/EngineResetBenchmark.cs
+++ b/Jint.Benchmark/EngineResetBenchmark.cs
@@ -1,0 +1,75 @@
+using BenchmarkDotNet.Attributes;
+using Jint.Native;
+
+namespace Jint.Benchmark;
+
+[MemoryDiagnoser]
+public class EngineResetBenchmark
+{
+    private Engine _reusableEngine = null!;
+    private Engine _reusableEngineWithOptions = null!;
+    private Prepared<Script> _setupScript;
+    private Prepared<Script> _workScript;
+
+    [GlobalSetup]
+    public void GlobalSetup()
+    {
+        _setupScript = Engine.PrepareScript(@"
+            function processInput(input) {
+                return JSON.parse(input);
+            }
+            function formatOutput(obj) {
+                return JSON.stringify(obj, null, 2);
+            }
+            var CONFIG = { maxRetries: 3, timeout: 5000 };
+        ");
+
+        _workScript = Engine.PrepareScript(@"
+            var result = processInput('{""key"":1}');
+            formatOutput(result);
+        ");
+
+        _reusableEngine = new Engine();
+        _reusableEngine.Execute(_setupScript);
+        _reusableEngine.Evaluate(_workScript);
+
+        _reusableEngineWithOptions = new Engine(options =>
+        {
+            options.Strict = true;
+        });
+        _reusableEngineWithOptions.Execute(_setupScript);
+        _reusableEngineWithOptions.Evaluate(_workScript);
+    }
+
+    [Benchmark(Baseline = true)]
+    public JsValue NewEngine_WithSetup()
+    {
+        var engine = new Engine();
+        engine.Execute(_setupScript);
+        return engine.Evaluate(_workScript);
+    }
+
+    [Benchmark]
+    public JsValue ResetState_WithSetup()
+    {
+        _reusableEngine.ResetState();
+        _reusableEngine.Execute(_setupScript);
+        return _reusableEngine.Evaluate(_workScript);
+    }
+
+    [Benchmark]
+    public JsValue NewEngine_WithOptions_WithSetup()
+    {
+        var engine = new Engine(options => options.Strict = true);
+        engine.Execute(_setupScript);
+        return engine.Evaluate(_workScript);
+    }
+
+    [Benchmark]
+    public JsValue ResetState_WithOptions_WithSetup()
+    {
+        _reusableEngineWithOptions.ResetState();
+        _reusableEngineWithOptions.Execute(_setupScript);
+        return _reusableEngineWithOptions.Evaluate(_workScript);
+    }
+}
--- a/Jint.Benchmark/EngineResetBenchmark.cs
+++ b/Jint.Benchmark/EngineResetBenchmark.cs
@@ -3,73 +3,302 @@ using Jint.Native;
 
 namespace Jint.Benchmark;
 
+/// <summary>
+/// Simulates a server-side script execution loop where each iteration creates/reuses
+/// an engine, loads setup scripts (utility functions), binds CLR objects,
+/// then executes user-provided code.
+///
+/// Three workload profiles:
+/// - Light: short scripts typical of tool dispatch (parse args, format output) — setup cost dominates
+/// - Heavy: CPU-intensive scripts (large data processing) — script execution dominates
+/// - IO: scripts that call async CLR methods (simulated) — allocations matter, wall-clock is I/O-bound
+/// </summary>
 [MemoryDiagnoser]
 public class EngineResetBenchmark
 {
-    private Engine _reusableEngine = null!;
-    private Engine _reusableEngineWithOptions = null!;
-    private Prepared<Script> _setupScript;
-    private Prepared<Script> _workScript;
+    private Engine _reusableLight = null!;
+    private Engine _reusableHeavy = null!;
+    private Engine _reusableIO = null!;
+    private Prepared<Script> _consoleShim;
+    private Prepared<Script> _utilityLibrary;
+    private Prepared<Script>[] _lightScripts = null!;
+    private Prepared<Script> _heavyScript;
+    private Prepared<Script> _ioScript;
+
+    private static void ConfigureEngine(Options options)
+    {
+        options.LimitMemory(32_000_000);
+        options.TimeoutInterval(TimeSpan.FromSeconds(10));
+        options.Constraints.PromiseTimeout = TimeSpan.FromSeconds(10);
+        options.LimitRecursion(64);
+        options.Strict();
+        options.Interop.AllowGetType = false;
+        options.Interop.AllowSystemReflection = false;
+        options.Interop.AllowWrite = false;
+        options.ExperimentalFeatures = ExperimentalFeature.TaskInterop;
+    }
+
+    private static void BindClrObjects(Engine engine)
+    {
+        engine.SetValue("__log", new Action<string>(_ => { }));
+        engine.SetValue("config", new { MaxRetries = 3, Timeout = 5000, Region = "us-east-1" });
+    }
+
+    private static void BindIOClrObjects(Engine engine)
+    {
+        BindClrObjects(engine);
+        engine.SetValue("fetchData", new Func<string, Task<string>>(key =>
+            Task.FromResult("{\"id\":1,\"name\":\"test\",\"values\":[1,2,3,4,5]}")));
+        engine.SetValue("sendResult", new Func<string, Task<bool>>(data =>
+            Task.FromResult(true)));
+    }
 
     [GlobalSetup]
     public void GlobalSetup()
     {
-        _setupScript = Engine.PrepareScript(@"
-            function processInput(input) {
-                return JSON.parse(input);
+        _consoleShim = Engine.PrepareScript("""
+            var console = {
+                log(...a) { __log(a.map(String).join(' ')); },
+                error(...a) { __log('ERROR: ' + a.map(String).join(' ')); },
+                warn(...a) { __log('WARN: ' + a.map(String).join(' ')); },
+                info(...a) { __log('INFO: ' + a.map(String).join(' ')); }
+            };
+            """, "console-shim.js");
+
+        _utilityLibrary = Engine.PrepareScript("""
+            function parseCSV(text) {
+                var lines = text.split('\n');
+                var headers = lines[0].split(',');
+                var result = [];
+                for (var i = 1; i < lines.length; i++) {
+                    if (!lines[i].trim()) continue;
+                    var values = lines[i].split(',');
+                    var obj = {};
+                    for (var j = 0; j < headers.length; j++) {
+                        obj[headers[j].trim()] = values[j] ? values[j].trim() : '';
+                    }
+                    result.push(obj);
+                }
+                return result;
             }
-            function formatOutput(obj) {
-                return JSON.stringify(obj, null, 2);
+            function transformRecords(records, filterFn, mapFn) {
+                return records.filter(filterFn).map(mapFn);
             }
-            var CONFIG = { maxRetries: 3, timeout: 5000 };
-        ");
+            function summarize(data) {
+                var total = data.reduce(function(s, r) { return s + (r.value || 0); }, 0);
+                return JSON.stringify({ count: data.length, total: Math.round(total * 100) / 100, items: data });
+            }
+            """, "utility-library.js");
 
-        _workScript = Engine.PrepareScript(@"
-            var result = processInput('{""key"":1}');
-            formatOutput(result);
-        ");
-
-        _reusableEngine = new Engine();
-        _reusableEngine.Execute(_setupScript);
-        _reusableEngine.Evaluate(_workScript);
-
-        _reusableEngineWithOptions = new Engine(options =>
+        _lightScripts = new[]
         {
-            options.Strict = true;
-        });
-        _reusableEngineWithOptions.Execute(_setupScript);
-        _reusableEngineWithOptions.Evaluate(_workScript);
+            Engine.PrepareScript("""
+                var csv = 'id,name,score\n1,Alice,95\n2,Bob,82\n3,Charlie,91';
+                var data = parseCSV(csv);
+                var result = transformRecords(data,
+                    function(r) { return parseInt(r.score) > 85; },
+                    function(r) { return { name: r.name, value: parseInt(r.score) }; });
+                summarize(result);
+                """),
+            Engine.PrepareScript("""
+                var items = [];
+                for (var i = 0; i < 100; i++) {
+                    items.push({ id: i, value: Math.sin(i) * 100, label: 'item_' + i });
+                }
+                var filtered = items.filter(function(x) { return x.value > 0; });
+                JSON.stringify({ count: filtered.length, avg: filtered.reduce(function(s,x) { return s + x.value; }, 0) / filtered.length });
+                """),
+            Engine.PrepareScript("""
+                var text = 'The quick brown fox jumps over the lazy dog';
+                var words = text.split(' ');
+                var freq = {};
+                words.forEach(function(w) { freq[w] = (freq[w] || 0) + 1; });
+                var sorted = Object.keys(freq).sort();
+                JSON.stringify({ wordCount: words.length, uniqueWords: sorted.length, frequencies: freq });
+                """),
+            Engine.PrepareScript("""
+                var matrix = [];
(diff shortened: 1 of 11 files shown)


lahma:
I would think that all other things during script evaluation would be the cost of running Jint. What if you include real world script scenario and see how much the reset helps then?

lahma (commented):
## Overview

Adds a public `Engine.ResetState()` that preserves Intrinsics / Options / Parser / ExtensionMethodCache / TypeCache / reflection accessors and rebuilds `Realm.GlobalObject` + `Realm.GlobalEnv`, then re-applies `Options.Apply`. The architectural idea is sound, the perf wins are plausible, and the sync test coverage is good. A few correctness issues around the async story and a couple of extensibility edges look worth addressing before merge.

---

## Critical

### 1. The generation counter does not actually drop late `Task.ContinueWith` callbacks

`EventLoop.Enqueue` reads `_generation` **at enqueue time**:

```csharp
public void Enqueue(Action continuation)
{
    _events.Enqueue(new TaggedEvent(continuation, _generation));
    ...
}
```

When a stale `ContinueWith` fires *after* `Reset()` has already incremented `_generation` to N+1, the resulting `TaggedEvent` is tagged with N+1 — the **new** generation. `RunAvailableContinuations` sees `taggedEvent.Generation == currentGeneration` and runs it.

The generation tag only protects events already in the queue at reset time, and even those are explicitly drained by the `while (_events.TryDequeue(out _)) { }` loop, making the tag mostly redundant for that case.

`ResetState_StaleAsyncCallbacksDroppedAfterReset` passes only because the stale `resolve(999)` operates on an orphaned `JsPromise` that nobody observes — the test never verifies the callback was dropped. To detect the bug, the stale callback would need to do something observable on the engine (set a global, throw, etc.).

A real fix would capture the generation at *registration* time (in `RegisterPromiseWithClrValue`'s closure) and skip the enqueue if `_generation` has changed.

### 2. The "during execution" guard is too narrow for async paths

```csharp
if (_activeEvaluationContext is not null)
{
    Throw.InvalidOperationException("ResetState cannot be called while script is executing");
}
```

`_activeEvaluationContext` is set/cleared in the **synchronous** `ExecuteWithConstraints` scope. `EvaluateAsync` looks like:

```csharp
public Task<JsValue> EvaluateAsync(string code, ...)
{
    var result = Evaluate(code, source);                  // sync — clears _activeEvaluationContext on exit
    return UnwrapResultAsync(result, cancellationToken);  // returns a Task
}
```

By the time the user's next line of C# (`engine.ResetState()`) runs, `_activeEvaluationContext` is already null even though `AwaitPromiseSettlementAsync` is suspended on `eventLoop.WaitForEventAsync`. The guard does not fire, and `ResetState`:

- Calls `Interlocked.Exchange(ref _eventAvailable, null)`, orphaning the TCS that the previous `AwaitPromiseSettlementAsync` is awaiting. That `await` never wakes up; the previous `EvaluateAsync` Task hangs forever.
- Allows the stale callbacks described in (1) to bleed into the next run.

This is exactly the scenario `ResetState_StaleAsyncCallbacksDroppedAfterReset` constructs — it silently leaks the orphan Task. The XML doc says "between Execute / EvaluateAsync calls" but the guard cannot enforce this. Either track outstanding async waiters (a counter incremented before `await` in `AwaitPromiseSettlementAsync` and decremented in `finally`), or be much more explicit in the docs that calling `ResetState` while a returned `Task` is still pending is undefined behavior.

### 3. Custom `Host` post-init logic is silently dropped

`ResetState` rebuilds `GlobalObject`/`GlobalEnv` directly, not via `_host.Initialize(this)`. This skips:

- `Host.InitializeHostDefinedRealm` (overridable)
- `Host.PostInitialize` (an explicit extension point used to register host-specific globals)

After reset, anything set up in `PostInitialize` is gone, with no signal to the user. `Options._configurations` *is* re-applied via `Options.Apply`, so `Configure(engine => engine.SetValue(...))` survives — but the parallel `Host` extension point does not. Either re-run `_host.PostInitialize()` on reset (and document that `InitializeHostDefinedRealm` overrides are honored only at construction), or split `Host.Initialize` into a "create realm" phase and a "reset-safe global setup" phase that `ResetState` can re-invoke.

---

## Moderate

### 4. Extension-method wrapper chain grows on every reset

`Options.Apply` calls `AttachExtensionMethodsToPrototypes`, which does:

```csharp
if (prototype.HasOwnProperty(key) && prototype.GetOwnProperty(key).Value is Function functionInstance)
{
    descriptorWithFallback = CreateMethodInstancePropertyDescriptor(functionInstance);
    prototype.SetOwnProperty(key, descriptorWithFallback);
}
```

First apply: `functionInstance` is the built-in (e.g. `Array.prototype.find`), wrapped by a `MethodInfoFunction`. Because `ResetState` preserves prototypes, the second apply finds the previous `MethodInfoFunction` as `functionInstance` and wraps that in another `MethodInfoFunction`. After N resets you have an N-deep wrapper chain on every extended method — slow leak plus per-call overhead. Worth a tag-and-skip on the descriptor or short-circuiting `AttachExtensionMethodsToPrototypes` on the rebuild path.

### 5. `Modules.Clear()` is dead code

`Options.Apply(this)` ends with `engine.Modules = new Engine.ModuleOperations(engine, Modules.ModuleLoader);` — it always replaces `Modules` entirely. The `Modules.Clear()` call right before `Options.Apply(this)` (and the new `ModuleOperations.Clear()` helper) is unreachable in this path. Either drop the helper + call, or stop replacing the instance in `Options.Apply` and rely on `Clear()` instead.

### 6. User-registered `ModuleBuilder`s are silently dropped

Same root cause as (5): `engine.Modules.Add("lib", b => b.ExportType<X>())` registers into the old `ModuleOperations._builders`, which is discarded by the replacement. The XML doc lists "Module cache" as cleared but doesn't call out that *programmatic module registrations* are also gone — a user trying the pooling pattern from the README will be surprised. Either preserve `_builders` across reset or document this explicitly.

---

## Minor

### 7. `Host.CreateGlobalObject` widened from `protected` to `protected internal`

`Host` is public, so this is a public-API surface change made only because `Engine.ResetState()` wants to call it from the `Jint` namespace. Cleaner: add an internal-only `Host.RebuildGlobals(realm)` helper (which would also help (3)) and keep `CreateGlobalObject` `protected virtual`. Once shipped, `protected internal` is harder to walk back than `protected`.

### 8. Test names overstate what is verified

`ResetState_StaleAsyncCallbacksDroppedAfterReset` and `ResetState_MultipleStaleCallbacksAllDropped` only assert the *new* evaluation succeeds. To actually test "dropped", make the stale callback's body something observable (e.g. set a global it shouldn't see, or flip a flag and assert it stays unflipped).

### 9. `_objectWrapperCache` clear path differs from `Dispose()`

```csharp
#if SUPPORTS_WEAK_TABLE_CLEAR
    _objectWrapperCache.Clear();
#else
    _objectWrapperCache = null;
#endif
```

`Dispose()` uses reflection to invoke an internal `Clear` on legacy frameworks; this path nulls the field. Functionally OK, slightly inconsistent — a shared helper would keep them in sync.

### 10. `_typeReferences` / `_reflectionAccessors` / `TypeCache` / `_debugger` are correctly preserved

These are configuration/CLR metadata. One small caveat: `_debugger` is stateful — if a user has been driving a debugging session, internal state from that session survives reset. Worth a sentence in the XML doc.

---

## Style / conventions

- Matches existing style; `TaggedEvent` as a `readonly record struct` follows CLAUDE.md guidance.
- Visibility is appropriately `internal` for the new helpers (`EventLoop.Reset`, `ExecutionContextStack.Clear`, `ModuleOperations.Clear`, `GlobalSymbolRegistry.Reset`).
- The XML doc acknowledges the prototype-mutation caveat clearly.

## Performance

- Numbers are believable. Tagging each event with `(Action, int)` adds a few bytes per queue item plus the struct write — minor overhead, but the correctness it claims to buy isn't actually delivered (issue 1). If the design is fixed properly, the cost is justified.

## Test coverage gaps worth filling

- A test where a `ContinueWith` body does observable engine work after `ResetState`, asserting it does not run.
- A test with a `Host` subclass that registers a global in `PostInitialize`, asserting whether it survives reset (and decide what the contract is).
- A test that calls `ResetState` while an `EvaluateAsync` Task is still pending and asserts the previous Task is canceled/faulted rather than hanging.

---

## Recommendation

The synchronous-reset story is implemented well and the benchmarks justify the addition, but the async-safety claim is not delivered by the current design. Before merge:

1. Fix or remove the generation-counter "stale callback drop" claim — either capture the generation at registration time, or stop advertising the protection and document async-pending-task usage as undefined behavior.
2. Extend the guard to detect outstanding async waiters, or cancel them gracefully in `Reset`.
3. Decide what to do about `Host.PostInitialize` on reset and document the contract.
4. Address the extension-method wrapper buildup before this gets used in long-lived pools.
5. Drop the redundant `Modules.Clear()` + helper, or invert the relationship with `Options.Apply`.
6. Tighten the test names that overstate verified invariants.

desty2k:
Good question — I've added three workload profiles to the benchmark to show the full picture:

```
| Method               | Mean          | Gen0       | Gen1      | Allocated   | Alloc Ratio |
|--------------------- |--------------:|-----------:|----------:|------------:|------------:|
| Light_NewEngine_10x  |     293.14 us |    86.9141 |    5.3711 |   713.38 KB |        1.00 |
| Light_ResetState_10x |     242.18 us |    58.5938 |   19.5313 |   479.42 KB |        0.67 |
| Heavy_NewEngine_10x  | 154,994.82 us | 13000.0000 | 5250.0000 | 94338.21 KB |      132.24 |
| Heavy_ResetState_10x | 153,620.74 us | 13250.0000 | 5750.0000 | 93971.49 KB |      131.73 |
| IO_NewEngine_10x     |     141.11 us |    73.2422 |    2.9297 |   605.42 KB |        0.85 |
| IO_ResetState_10x    |      92.73 us |    44.9219 |    4.8828 |   369.95 KB |        0.52 |
```

**Heavy scripts** (5000 records, grouping, sorting, aggregation — 15ms of JS work per iteration): ResetState makes no meaningful difference to time. Script execution completely dominates the ~6μs of setup savings. This confirms your intuition — for CPU-bound scripts the time improvement vanishes.

**Light scripts** (tool dispatch: parse a few rows, format JSON — ~5μs of JS work): 17% faster, 33% less allocation. Setup is a meaningful fraction of total work.

**I/O-bound scripts** (async CLR calls via TaskInterop — scripts that fetch data from external services, process the response, and send results back): **34% faster, 39% less allocation**. The JS CPU work is light (parse a response, transform, format output), but the engine setup cost is paid on every invocation. In high-concurrency server scenarios where many engines are created per second, the allocation savings (234 KB over 10 iterations) reduce GC pressure.

The time improvement depends on the ratio of JS CPU work to engine setup cost. For scripts that mostly coordinate I/O (call gRPC/HTTP, process the response, call another service), engine setup is a significant fraction of CPU time per request. For scripts that do heavy computation in JS itself, it's noise. The allocation savings are constant regardless — ~23 KB per engine reuse cycle.

lahma:
First — thank you for the persistence on this one, and apologies that it sat as long as it did. You took a demanding review and addressed every point of it properly: the generation counter is now captured at registration time where it actually protects something, the `_pendingAsyncOperations` guard closes the suspended-`EvaluateAsync` hole cleanly, `Host.RebuildGlobals` + re-running `PostInitialize` is the right resolution of the Host-contract question, and the observable-flag tests verify what their names claim. That is real machinery and real diligence, and it directly shaped what happens next.

That said, I'm going to close this in favor of #2834 (`CaptureGlobalSnapshot`/`RestoreGlobalSnapshot`, shipped in 4.15.0), and I want to give you the substantive reasons rather than a "superseded" label:

**1. What pooled hosts measurably need is warm restore, not rebuild.** We audited the large fresh-engine-per-evaluation embedders (Elsa, OrchardCore, docfx). For a short script, engine construction is ~2–5 % of the evaluation; **re-establishing the host surface — dozens of `SetValue`/`TypeReference` registrations and setup scripts — is 85–95 %**. `ResetState` rebuilds the global object, so a pooled host re-pays that dominant cost every cycle (your own benchmark re-runs `BindClrObjects` + the two setup scripts per iteration — correctly, because it must). The snapshot design inverts it: configure once, capture, restore per cycle; the captured surface — including functions declared by setup scripts and still-lazy globals — comes back without any registration code running.

**2. Replacing the global object's identity fights the engine's cache invariants.** Since April, main has landed the builtin-shape global, the global-binding identifier caches, and version-witnessed prototype-method caches. All of them validate against version counters under an assumption of stable identity. A rebuild path makes "the realm's global was swapped mid-life" a supported operation that every current and future cache lane must be audited against; the snapshot design keeps identity stable and only ever *bumps* counters, which can only cause a safe miss. (Relatedly: because intrinsics are preserved, `ResetState` as-is leaks the Annex B `RegExp.$1`/`RegExp.input` statics across resets — one more piece of state the restore path clears explicitly.)

**3. Neither design is an isolation boundary** — prototype mutations survive both, as your docs honestly say — so the rebuild's extra hardness (Symbol.for, module registry) buys hygiene, not security, and none of the audited embedders needs it. If you have a concrete scenario that does, I'd genuinely like to hear it — both are expressible as opt-ins on the snapshot surface, and I'd take that follow-up.

**What shipped from this PR, with credit:**
- The **async machinery went into #2834 essentially as you designed it** — the pending-async-operations guard and registration-time generation stamping are exactly what the restore path was missing, and that commit carries `Co-authored-by`. Your two observable-flag tests came along in adapted form.
- The **`Options.Apply` ordering fix** (`ModuleOperations` created before configuration callbacks) was an independent bug fix and went in as its own credited PR (#2837).
- Several of your doc caveats (programmatic module registrations, debugger state) are prior art for #2834's docs.

I'd value your eyes on the shipped surface — you know this problem space as well as anyone now. Thanks again for the contribution and for sticking with it since April; the reuse story Jint ships in 4.15.0 is better because this PR existed.

PR #2427, opened 2026-04-27
Fix EventLoop async API: signal every concurrent WaitForEventAsync wa…

Ellested:
…iter

Repro: any caller pattern that issues two concurrent WaitForEventAsync calls against a single EventLoop drops the wake signal to the second caller.

Root cause: the loop tracked the wake TCS in a single volatile field (_eventAvailable). When a second caller arrived while the first was still pending, the new TCS was rejected by CompareExchange and the caller fell into the "Active waiter exists — events may already be pending" branch, returning Task.CompletedTask without registering any wake signal. The caller's outer loop would then re-enter WaitForEventAsync on the next IsEmpty=true check and immediately get Task.CompletedTask again — a CPU-burning spin until the queue happened to drain by some other path.

Fix: hold a list of TCS waiters under a Lock, append per call, and broadcast TrySetResult to every entry on Enqueue. The list is cleared on each broadcast so subsequent Enqueues don't re-signal stale entries; spurious wakes (e.g. self-signaling on the post-registration double-check) are harmless because the caller's outer loop re-checks its own state.

Added Jint.Tests/Runtime/AsyncTests.cs::EventLoopShouldSignalAllConcurrentWaiters which fails on the previous code (waiter1 is signaled, waiter2 stays pending after Enqueue) and passes with the fix.

The single-waiter cancellation path (CancellationTokenRegistration disposal via tcs.Task.ContinueWith) is preserved unchanged for the common case where only one waiter is registered.

### DIFF
--- a/Jint.Tests/Runtime/AsyncTests.cs
+++ b/Jint.Tests/Runtime/AsyncTests.cs
@@ -2209,6 +2209,33 @@ public class AsyncTests
         Assert.Equal("7", result.AsString());
     }
 
+#if !NETFRAMEWORK
+    [Fact]
+    public async Task EventLoopShouldSignalAllConcurrentWaiters()
+    {
+        // Regression: a second concurrent caller of WaitForEventAsync used to receive
+        // Task.CompletedTask immediately because the loop only tracked a single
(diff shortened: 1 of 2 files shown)


lahma (approved):
LGTM, thank you!

PR #2448, opened 2026-05-02
[JsAccessible] source generator MVP: typed CLR interop accessors

lahma:
## Summary

Adds a second source-generator pipeline (alongside the existing `[JsObject]` one) for CLR types annotated with the new public `[JsAccessible]` attribute. The generator emits per-member typed accessor classes that bypass `PropertyInfo.GetValue` / `MethodInfo.Invoke` + boxing on the hot path, registered into a static `GeneratedAccessorRegistry` via `[ModuleInitializer]`. `TypeResolver` consults the registry before falling back to reflection; un-attributed types are completely unaffected.

```csharp
[JsAccessible]
public sealed class Player
{
    public int Score { get; set; }
    public string Name { get; set; } = string.Empty;
    public bool Alive { get; set; }
    public int AddPoints(int delta) { Score += delta; return Score; }
    public string Describe() => $""{Name}:{Score}"";
}
```

The generator emits a `Player.JsAccessible.g.cs` partial with `Player_Score_Accessor`, `Player_AddPoints_Accessor`, etc., plus a `[ModuleInitializer]` that wires them into the registry at assembly load.

**MVP scope**: instance properties (int/double/bool/string/JsValue) and 0/1-argument instance methods. Anything outside scope (overloaded methods, indexers, generic methods, `ref`/`out`, multi-arg methods, complex parameter types) silently falls back to the existing reflection path with zero behavior change.

## Performance

`InteropAccessibleBenchmarks` (warm, prepared scripts via `Engine.PrepareScript`, AMD Ryzen 9 5950X / .NET 10):

| Method                | Reflection | Generated  | Speedup |
|---------------------- |-----------:|-----------:|--------:|
| PropertyGet_Int       |  179.7 ns  |  140.6 ns  |  1.28× |
| PropertyGet_String    |  175.7 ns  |  139.5 ns  |  1.26× |
| PropertyGet_Bool      |  233.8 ns  |  142.1 ns  |  1.65× |
| PropertySet_Int       |  194.2 ns  |  171.4 ns  |  1.13× |
| MethodInvoke_OneArg   |  426.2 ns  |  200.3 ns  |  2.13× |
| MethodInvoke_NoArgs   |  401.1 ns  |  221.9 ns  |  1.81× |
| Mixed_GetMethodInvoke |  393.5 ns  |  329.0 ns  |  1.20× |

Wins are smaller than the planning doc's 5–10× estimate because the wrapper's descriptor cache already amortizes most reflection cost on warm reads — the per-call `PropertyInfo.GetValue` overhead is ~50 ns warm, not the 2–5 µs cold cost. The genuine win is method invocation, where `MethodInfo.Invoke` + arg/return boxing are fully eliminated. Property gets benefit from skipping the box-then-unbox round-trip in `JsValue.FromObjectWithType`.

The bench was refactored from `Engine.Execute(string)` (parses each iteration, parse cost dominated total) to prepared scripts so the per-iteration measurement isolates the accessor hot path.

## Implementation notes

**Public surface added**:
- `Jint.JsAccessibleAttribute` — class-level marker.
- `Jint.Runtime.Interop.GeneratedAccessorRegistry.Register(...)` — called from generator's `[ModuleInitializer]`.
- `Jint.Runtime.Interop.Reflection.GeneratedReflectionAccessor` / `GeneratedMethodAccessor` — abstract base classes that the generator-emitted subclasses inherit from. Required to be public so external assemblies can host generator output.
- `Jint.Runtime.Interop.Reflection.ReflectionAccessor` — class promoted from `internal` to `public`; `GetValue`/`SetValue` made `virtual`. Compatible change — existing callers unchanged.

**Subtleties**:
- `ObjectWrapper.Set` bypasses the descriptor and calls `accessor.SetValue(...)` directly, so the override on `GeneratedReflectionAccessor.SetValue` is required (in addition to overriding the descriptor's `DoSet`).
- `JsCallArguments` is a global-using alias for `JsValue[]` defined in `Jint/GlobalUsings.cs`. External assemblies don't have this alias, so the generator emits `JsValue[]` literally for `Function.Call` overrides.
- Adding the source-gen analyzer reference to `Jint.Benchmark` / `Jint.Tests` triggers CS0436 because the post-init source re-emits Jint's internal attributes that are also visible via same-SNK `InternalsVisibleTo`. Suppressed via `<NoWarn>CS0436</NoWarn>` — the duplicates are functionally harmless (`[Conditional(""JINT_SOURCE_GENERATORS"")]` strips them from release IL).

## Test plan

- [x] `dotnet test` Jint.Tests: **2893 / 0 / 5 skipped** (added 2 new `JsAccessibleSmokeTests` for registry population + read/write/invoke validation)
- [x] `dotnet test` Jint.Tests.Test262: **98,959 / 0 / 131 skipped** (matches main baseline, no regressions)
- [x] `dotnet test` Jint.Tests.PublicInterface: **79 / 0 / 0**
- [x] `dotnet test` Jint.Tests.CommonScripts: **28 / 0 / 0**
- [x] InteropAccessibleBenchmarks: before/after measured with default BDN config (numbers above)


lahma:
Superseded — `[JsAccessible]` shipped in v5, in four pull requests, and this branch's approach could not have got there.

**What this MVP got right, and what carried over:** the architecture. A registry consulted in `TypeResolver`, a typed accessor subclassing the existing accessor abstraction, un-annotated types completely untouched. That shape is what shipped.

**Why it was rewritten rather than merged.** An audit of this branch found the emitted code used three internal members — `Function.Call`'s `protected internal`, `ObjectInstance._prototype`, and `Engine.Realm` — so it had only ever been compiled by assemblies on the `InternalsVisibleTo` grant list. It could not compile in a consumer's project, and the generator was not in the NuGet package either, so no consumer could have obtained it. `[ModuleInitializer]` also does not exist downlevel without a polyfill a consumer has no reason to carry.

**What shipped instead:**

- **#3333** — the emitter targets the public surface: a static `Invoke(JsValue, JsValue[])` handed to the public `ClrFunction`, with an explicit generated `RegisterAll()` in place of `[ModuleInitializer]`. It also emits the same four delegates the engine builds at runtime with expression trees, so a generated accessor *is* `CompiledMemberAccessor` resolved at compile time — which means it helps on `net472`, `netstandard2.0` and AOT, where the compiled lanes decline outright.
- **#3352** — generated members are filtered and named exactly as reflected ones are, so `Options.Interop.MemberFilter` and the hardened profile apply to both lanes.
- **#3373 / #3376** — seven `JINT03x` diagnostics naming every member the generator could not take, promotable per-id through `.editorconfig`.
- **#3375** — the generator ships in the package, verified by a CI job that packs, restores into a throwaway consumer on **net472 and net10.0**, runs it, and requires a severity promotion to fail the build.

**On the claim:** this PR's title says "typed CLR interop accessors" and its body claimed a 1.13–2.13× interop speedup measured under BenchmarkDotNet's default job. `Jint.Benchmark/AGENTS.md` requires `JINT_BENCH_MODE=gate` for any number quoted in a PR, and four of those seven rows sat inside the regime a single launch cannot resolve. v5 therefore ships the feature as **"opt-in, trimming-safe, no reflection on the hot path"** rather than as an AOT or speed claim — the supported member set reaches none of the value-type generic-instantiation sites that are Jint's actual AOT gaps.

Two defects this branch's differential testing would not have caught, found while rewriting: `p.Name = null` stored the string `"null"` where reflection stores `null`, and `null` into a value-type member would have thrown `NullReferenceException` where the runtime binder writes `default(T)`.

Thanks — the idea and the shape were right, and both are in v5.

PR #2466, opened 2026-05-11
Allow hosts to hide selected global built-ins

nfcampos:
## Summary

Allow hosts to hide selected built-ins from globalThis.

## Details

- Add Options.Builtins.DisabledGlobalBuiltIns and a fluent DisableGlobalBuiltIns helper.
- Add GlobalBuiltIn flags for ArrayBuffer, SharedArrayBuffer, DataView, TypedArray constructors, Atomics, and ShadowRealm.
- Remove disabled generated global properties after globalThis source-generated initialization.
- Keep each flag independent so disabling ArrayBuffer does not implicitly hide DataView, typed arrays, or SharedArrayBuffer.

## Linked issue

Refs #

## Test plan

- [ ] Ran dotnet test --configuration Release locally

Additional local validation:

- dotnet test Jint.Tests/Jint.Tests.csproj --filter "FullyQualifiedName~GlobalTests"

## Breaking change?

No.

### DIFF
--- a/Jint.Tests/Runtime/GlobalTests.cs
+++ b/Jint.Tests/Runtime/GlobalTests.cs
@@ -3,6 +3,64 @@ namespace Jint.Tests.Runtime;
 public class GlobalTests
 {
     [Fact]
+    public void CanDisableSelectedGlobalBuiltIns()
+    {
+        var engine = new Engine(options => options.DisableGlobalBuiltIns(
+            GlobalBuiltIn.ArrayBuffer
+            | GlobalBuiltIn.SharedArrayBuffer
+            | GlobalBuiltIn.DataView
+            | GlobalBuiltIn.TypedArray
+            | GlobalBuiltIn.Atomics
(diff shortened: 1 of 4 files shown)


nfcampos:
windows ci fail appears be a flaky timing-related test assertion, but let me know if caused by my changes

lahma:
Why would you want to disable/hide these? I think it would be more natural to have *enabled* built-ins which defaults to `All` and you can surgically create a list of items/categories when you actually want to narrow it down.

nfcampos:
Because these allow allocating arbitrary amounts of memory and have no valid use in my environment, but I agree this list should eventually cover all, whether its opt-in starting from no globals or opt-out starting from all globals doesn't make much of a difference to me

lahma:
All globals are lazy, if they are not being used they should only cost you the lazy descriptor and dictionary slot, how much performance benefit are you seeing here?

PR #2467, opened 2026-05-11
Return pooled temporaries on abrupt completion

nfcampos:
## Summary

Return pooled temporary values when evaluation exits abruptly.

## Details

- Wrap rented argument arrays and temporary element buffers in finally blocks so callbacks, constructors, and iterator operations return pooled storage even when they throw.
- Cover array, map, set, typed array, object property, call, new, tagged-template, function environment, and expression-cache paths.
- Keep this independent of native recursion hardening by using the existing call APIs on main.
- This is a resource-lifecycle change; intended JavaScript behavior is unchanged except for avoiding leaked pooled temporaries after abrupt completion.

## Linked issue

Refs #

## Test plan

- [x] Ran dotnet test --configuration Release locally

Additional local validation:

- dotnet test Jint.Tests/Jint.Tests.csproj --filter "FullyQualifiedName!~EngineTests.ShouldSetYearBefore1970"

Note: full dotnet test Jint.Tests/Jint.Tests.csproj also ran; only EngineTests.ShouldSetYearBefore1970 failed locally due the existing timezone-sensitive main-branch test covered separately by nc/date-test-utc.

## Breaking change?

No.

### DIFF
--- a/Jint/Engine.cs
+++ b/Jint/Engine.cs
@@ -858,33 +858,39 @@ public sealed partial class Engine : IDisposable
                 items[i] = JsValue.FromObject(this, arguments[i]);
             }
 
-            // ensure logic is in sync between Call, Construct, engine.Invoke and JintCallExpression!
-            JsValue result;
             var thisObject = JsValue.FromObject(this, thisObj);
-            if (callable is Function functionInstance)
+            try
             {
-                var callStack = CallStack;
-                callStack.Push(functionInstance, expression: null, ExecutionContext);
-                try
+                // ensure logic is in sync between Call, Construct, engine.Invoke and JintCallExpression!
+                JsValue result;
+                if (callable is Function functionInstance)
                 {
-                    result = functionInstance.Call(thisObject, items);
-                }
-                finally
-                {
-                    // if call stack was reset due to recursive call to engine or similar, we might not have it anymore
-                    if (callStack.Count > 0)
+                    var callStack = CallStack;
+                    callStack.Push(functionInstance, expression: null, ExecutionContext);
+                    try
                     {
-                        callStack.Pop();
+                        result = functionInstance.Call(thisObject, items);
+                    }
+                    finally
+                    {
+                        // if call stack was reset due to recursive call to engine or similar, we might not have it anymore
+                        if (callStack.Count > 0)
+                        {
+                            callStack.Pop();
+                        }
                     }
                 }
-            }
-            else
-            {
-                result = callable.Call(thisObject, items);
-            }
+                else
+                {
+                    result = callable.Call(thisObject, items);
+                }
 
-            _jsValueArrayPool.ReturnArray(items);
-            return result;
+                return result;
+            }
+            finally
+            {
+                _jsValueArrayPool.ReturnArray(items);
+            }
         }
 
         return ExecuteWithConstraints(Options.Strict, DoInvoke);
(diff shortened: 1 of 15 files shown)


nfcampos:
I have another stacked change ready to open a PR once this one is merged in https://github.com/nfcampos/jint/commit/6ef3cc44eab0a5eb5d202ccef33540777f2d6d25

lahma:
Thanks for the PR — pushing back on the premise though: I don't think these guardrails are needed, and in the hot paths they're likely net-negative.

**The pool is opportunistic by design**, not a managed resource. From `Jint/Pooling/ObjectPool.cs` (the pool's own XML doc, lines 30–40):

> 1. it is not the goal to keep all returned objects. Pool is not meant for storage. If there is no space in the pool, extra returned objects will be dropped.
>
> Not returning objects to the pool is not detrimental to the pool's work, but is a bad practice.

Concretely: `JsValueArrayPool` is fixed-size (15 slots per arity). When full, `FreeSlow` silently bails. So drops-on-exception are indistinguishable from drops-because-the-pool-was-already-full — there is no leak (the array is just GC'd like any normal allocation), no correctness implication, and the next caller allocates a fresh `new JsValue[N]`. The phrase "leaked pooled temporaries" in the description is a bit misleading for that reason: a missed return is a missed reuse, not a leak.

**Concerns with the change as written:**

- **JIT overhead in the call dispatch hot path.** Wrapping the body of `JintCallExpression.EvaluateInternal` in `try/finally` blocks RyuJIT's promotion/inlining/register-allocation across the protected region. That's exactly where you don't want EH scaffolding for an opportunistic optimization. No benchmarks accompany the PR, but I'd expect a small but real regression in `FunctionCallBenchmark`-style scenarios. We have a per-change benchmark convention for performance-sensitive code; this should have before/after numbers before landing.
- **`ExpressionCache.cs` uses `try/catch + rethrow`** rather than `try/finally`. That's strictly worse — `catch/throw` re-touches the exception runtime context. If kept, it should be a `finally` with the `rented` flag, not catch-and-rethrow.
- **Existing convention is the opposite.** Of the ~16 files calling `ReturnArray`, only `IteratorProtocol.cs` wraps in `try/finally`, and only because the iterator itself must be closed on throw per spec — the array return just rides along in the same `finally`. Everywhere else the pattern is "rent, use, return on success." The PR breaks that consistency without a corresponding correctness gain.
- **PR template says tests were added/updated, but the diff has no test changes** — probably a missing commit or a checkbox slip; worth correcting either way.

**Salvageable bits.** A few of the changes do fix real (tiny) missed returns on *non-throw* paths — `JintCallExpression.HandleEval` and `SuperCall` early returns, and the `IsConstructor == false` path in `JintNewExpression`. Those are one-line fixes worth keeping as a small targeted PR. The bulk refactor I'd decline.

Happy to be wrong if benchmarks show otherwise — but the burden of proof for adding EH to hot dispatch paths should be a measurement, not a defensiveness argument, given the pool's documented contract.

nfcampos:
Sorry not sure I understand how the pool is designed such that the intended usage contract is "try to return leased objects, but dont try too hard" either return them in all code paths or in none, returning in some (and especially returning in the happy path where everything went well, but not when an exception is thrown or some other unexpected thing happens).

Note that in all places where the change was made the objects were already being returned to the pool in some cases. I understand there is sometimes overhead to try-catch or try-finally, is that really more relevant than knowing that code behaves the same in the unhappy paths, that are very frequently in this codebase not even covered by the unit tests?

PR #2469, opened 2026-05-11
Fix async resume through control-flow statements

nfcampos:
## Summary

Fix async resume through control-flow statements.

## Details

- Resume async execution inside catch/finally without rerunning the try block or leaking catch bindings.
- Resume selected if/while/do-while/for/switch paths without rerunning already-completed guards, bodies, discriminants, or case tests.
- Preserve accumulated loop/switch completion values and binary-expression left operands across await suspension.
- Add focused AsyncTests coverage for catch/finally, branch, loop, switch, and binary-expression resume behavior.

## Linked issue

Refs #

## Test plan

- [x] Ran dotnet test --configuration Release locally

Additional local validation:

- dotnet test Jint.Tests/Jint.Tests.csproj --filter "FullyQualifiedName~AsyncTests"

## Breaking change?

No.

### DIFF
--- a/Jint.Tests/Runtime/AsyncTests.cs
+++ b/Jint.Tests/Runtime/AsyncTests.cs
@@ -358,6 +358,60 @@ public class AsyncTests
     }
 
     [Fact]
+    public void ShouldNotReevaluateCompoundAssignmentLhsAfterAwait()
+    {
+        var result = EvaluateAsyncJson("""
+            const obj = { 0: 0 };
+            let i = -1;
+            obj[++i] += await Promise.resolve(5);
+            return { obj, i };
+            """);
+
+        Assert.Equal("""{"obj":{"0":5},"i":0}""", result);
+    }
+
+    [Fact]
+    public void ShouldNotReevaluatePropertyAssignmentLhsAfterAwait()
+    {
+        var result = EvaluateAsyncJson("""
+            const obj = { x: 10 };
+            let touches = 0;
+            const accessor = () => (touches++, obj);
+            accessor().x -= await Promise.resolve(3);
+            return { obj, touches };
+            """);
+
+        Assert.Equal("""{"obj":{"x":7},"touches":1}""", result);
+    }
+
+    [Fact]
+    public void ShouldPreserveCompoundAssignmentLhsAcrossNullishCoalescing()
+    {
+        var result = EvaluateAsyncJson("""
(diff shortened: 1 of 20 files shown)


lahma:
Rebased onto current `main` (post #2474 / AwesomeAssertions migration) and pushed 5 follow-up commits on top of the original fix, each with test coverage. The original commit is unchanged.

Stack on top of `eb677356c`:

| | SHA | |
|---|---|---|
| 1 | `c15749dc8` | Fix async resume through control-flow statements *(original)* |
| 2 | `894b43081` | Async resume helpers: tighten visibility to internal |
| 3 | `a2be00fef` | Async resume: preserve left operand of logical and conditional expressions |
| 4 | `ba86a7fdd` | Async resume: preserve compound assignment LHS across await |
| 5 | `13dd98334` | Async resume: preserve already-evaluated call/new arguments across await |
| 6 | `455e5e099` | Async resume: parity tests for generators and async generators |

Why these specifically: the same "left half re-evaluated on resume" bug the original fix repairs for binary expressions also exists in `&&` / `||` / `?:` (commit 3), in compound-assignment LHS (`obj[++i] += await y` re-running `++i` — commit 4), and in call/new argument lists (`foo(++i, ++i, await x)` re-running the earlier `++i`s — commit 5). All four use the same `ISuspendable.Data` infrastructure introduced by the original commit; commit 3 generalizes `BinaryExpressionSuspendData` → `LeftOperandSuspendData` and the others add focused suspend-data types (`AssignmentSuspendData`, `CallArgumentsSuspendData`).

Commit 6 adds 11 new tests in `GeneratorTests.cs` that mirror the AsyncTests with `yield`, since sync/async generators share the same resume machinery — these are regression guards against drift.

Verification on the rebased branch:
- `dotnet build --configuration Release` clean (warnings-as-errors)
- `dotnet test Jint.Tests/Jint.Tests.csproj`: 2973 passed / 0 failed / 5 skipped (net10.0); 2925 / 0 / 5 (net472) — 21 of those are new tests added in commits 3-6
- `dotnet test Jint.Tests.Test262/Jint.Tests.Test262.csproj`: 98957 / 0 / 133 — no test262 regressions

One follow-up explicitly left out of this PR: `ExpressionCache.BuildArgumentsWithSpreads` has the same suspend-and-replay bug, but iterator state on resume is more involved. Worth a separate PR.

PR #2480, opened 2026-05-17
Fix MemoryLimitConstraint skipping check when async resumes on a different thread

brunodeuner:
GC.GetAllocatedBytesForCurrentThread() is per-thread. When an async continuation resumes on a different thread pool thread, Reset() and Check() read counters from different threads, which can silently bypass the limit or produce false positives.

Fix: store the ManagedThreadId in Reset() and skip the check in Check() when the current thread differs. Safe degradation: the constraint is not enforced across thread hops, but no wrong results are produced.


lahma (approved):
Thank you!

PR #2482, opened 2026-05-18
Correct escaping of regexp patterns created at run-time

adams85:
## Summary
Corrects escaping of regexp patterns created at run-time via `new RegExp("\n")`, `RegExp("\n")`, `/x/.compile("\n")`, `"x".match("\n")`, etc.

## Details

The `/` character is not valid without escaping in regexp literals outside class sets (however, it's valid inside them). New line characters are not valid  in regexp literals either. There are cases which aren't escaped correctly (e.g. `new RegExp("[/]")`, `new RegExp("\r[\n]")`, etc.) This PR aims to fix this.

Also, .NET Regex objects can be exposed to the engine. However, it's incorrect to return the .NET pattern as if it were a JS regexp pattern. This could be misleading. The PR suggests a placeholder (which is an invalid pattern BTW) instead: `?[native regex]`. This is a breaking change though. I don't insist on it if you don't think it's a good idea.

Finally, makes a minor improvement to the Repl app: in interactive mode, it prints the regexp literal (like e.g. Chrome or Firefox dev tools do) instead of the meaningless `{}` text.

## Linked issue
n/a

## Test plan

- [ ] For ECMAScript spec changes: ran `Jint.Tests.Test262` and confirmed no regressions

## Breaking change?
Yes. Minor behavioral breaking change to `JsRegExp.Source` and `ToString` for native .NET Regex objects exposed to the engine, as explained above.

### DIFF
--- a/Jint.Repl/Program.cs
+++ b/Jint.Repl/Program.cs
@@ -221,10 +221,17 @@ while (true)
         JsValue str = result;
         if (!result.IsPrimitive() && result is not IJsPrimitive)
         {
-            str = serializer.Serialize(result, JsValue.Undefined, "  ");
-            if (str == JsValue.Undefined)
+            if (result is JsRegExp jsRegExp)
             {
-                str = result;
+                str = result.ToString();
+            }
+            else
+            {
+                str = serializer.Serialize(result, JsValue.Undefined, "  ");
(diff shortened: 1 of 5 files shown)


lahma (approved):
Thank you!

PR #2489, opened 2026-05-31
Update test262 suite and implement newly-surfaced features

lahma:
Bumps the Test262 suite to `4249661388e5d3f92a85186213da140a6481490f` and implements the spec gaps the new tests surface. The full conformance suite goes from **133 failing cases to 0** (no regressions; `Jint.Tests` and `Jint.Tests.PublicInterface` remain green, no public API changes).

### Features / fixes

- **`Error.prototype.stack` accessor** (error-stack-accessor proposal). `stack` is now a `get`/`set` accessor on `%Error.prototype%` (`{ enumerable: false, configurable: true }`) rather than an own data property on each instance. The construction-time trace is stored on the `JsError` and returned by the getter; the setter uses `SetterThatIgnoresPrototypeProperties`. The accessor lives only on the root `%Error.prototype%` — the NativeError prototypes inherit it.
- **Immutable `ArrayBuffer` write rejection.** `ValidateTypedArray` gains a write access mode that throws a `TypeError` *before* observable side effects when the backing buffer is immutable. Wired into `%TypedArray%.prototype` `copyWithin`/`fill`/`reverse`/`set`/`sort`, the `map`/`filter`/`slice` species-create destination, `Atomics` `add`/`and`/`compareExchange`/`exchange`/`or`/`store`/`sub`/`xor`, and `Uint8Array` `setFromBase64`/`setFromHex` (checked before reading arguments).
- **Intl legacy-constructor mode** (normative-optional) for `NumberFormat` and `DateTimeFormat`: a per-realm `%Intl%.[[FallbackSymbol]]` (`"IntlLegacyConstructedSymbol"`), `Chain*` on `[[Call]]`, and `Unwrap*` in the prototype methods. `Collator` is intentionally left without it (its `this-value-ignored` test still requires a fresh instance and it has no legacy-constructed-symbol test).
- **`Intl.PluralRules` `compactDisplay` option** — read after `notation`, surfaced in `resolvedOptions` only when `notation` is `"compact"`.
- **`Temporal.ZonedDateTime`** — reject not-yet-adopted calendar annotations (e.g. `bangla`) in the string path, and correctly round to a start-of-day when midnight occurs twice across a fall-back transition (tc39/proposal-temporal#3312).
- **Test harness** — `$262.createRealm()` now exposes an `evalScript` that evaluates in the new realm, enabling cross-realm tests.

### Excluded tests

- `language/module-code/.../reexport-source-binding-*` and `ambiguous-export-bindings/namespace-unambiguous-if-import-source-and-export.js` — re-exporting a source-phase import requires a WebAssembly `<module source>`, which Jint does not support.
- `built-ins/TypedArray/prototype/slice/speciesctor-return-same-buffer-with-offset.js` — broadened to run the immutable-buffer arg factory while still expecting the element copy to succeed, which a spec-conformant `slice` (rejecting an immutable destination) cannot satisfy; the test does not declare the `immutable-arraybuffer` feature.

### DIFF
--- a/Jint.Tests.Test262/Test262Harness.settings.json
+++ b/Jint.Tests.Test262/Test262Harness.settings.json
@@ -1,5 +1,5 @@
 {
-  "SuiteGitSha": "673e9bacbe28590f501e2dcd817aadcc31899191",
+  "SuiteGitSha": "4249661388e5d3f92a85186213da140a6481490f",
   //"SuiteDirectory": "//mnt/c/work/test262",
   "TargetPath": "./Generated",
   "Namespace": "Jint.Tests.Test262",
@@ -12,6 +12,15 @@
   "ExcludedDirectories": [
   ],
   "ExcludedFiles": [
+
+    // %TypedArray%.prototype.slice creates its destination via TypedArraySpeciesCreate(..., write), which
+    // (correctly) rejects an immutable backing buffer — see the passing slice/speciesctor-destination-
+    // backed-by-immutable-buffer.js. This test was broadened to run against every buffer-arg factory
+    // (including makeImmutableArrayBuffer) yet still expects the element copy to succeed, so when the
+    // source — and therefore the same-buffer species destination — is immutable, a spec-conformant slice
+    // throws instead of producing the expected array. The test does not declare the immutable-arraybuffer
+    // feature; it is incompatible with an engine that implements it.
+    "built-ins/TypedArray/prototype/slice/speciesctor-return-same-buffer-with-offset.js",
 
     // === INTL402 EXCLUSIONS ===
     // The following tests require full CLDR locale data support which is not available in the default provider.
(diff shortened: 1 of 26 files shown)


nZeus:
⚠️ The first one (`Error.prototype.stack accessor`) is a breaking change. Our JavaScriptException handler read `stack` via `ObjectInstance.TryGetValue` (own-property lookup). After this commit that lookup misses

Could be helpful to metion it in the release notes 🙏

lahma:
> ⚠️ The first one (`Error.prototype.stack accessor`) is a breaking change. Our JavaScriptException handler read `stack` via `ObjectInstance.TryGetValue` (own-property lookup). After this commit that lookup misses
> 
> Could be helpful to metion it in the release notes 🙏

Release notes updated.

PR #2501, opened 2026-06-07
Parenthesize expressions to correctly pre-allocate `List<T>`

jnyrup:
## Summary

Parenthesize expressions to correctly pre-allocate `List<T>`.

## Details

As `+` binds stronger than `??` this PR parenthesizes the coalescing parts.

See this [SharpLab Example](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwAEAmARgFgAoIgBgAIiyUBua6pgTgAoBZWgIbcA2gF009MgEopbGmR79gI8ZJlyAlngz1+QgDIaAzhgA8WjAD56AiRfrAp9ALzWBAfgB0AYQj4d7u70DADUDpraurTKhibm2m52kY4u1twePn6RgcFOYcByHArcsuzyimRCYhLCZKLqRRXK1fS19aVU9ryV3LFmFon0ffFWDk6uNl6+/vQ5oQ5TWQFBtBE63THG/Qk2EsMDY6n06Yszc3nHwKfZK7JAA) for how missing parentheses affect the result.

## Test plan

- [N/A] Added or updated unit tests in `Jint.Tests`
- [N/A] For ECMAScript spec changes: ran `Jint.Tests.Test262` and confirmed no regressions
- [N/A] For interop changes: covered in `Jint.Tests/Runtime/Interop`
  - Can you advice which benchmark to run?

## Breaking change?

No

### DIFF
--- a/Jint/Native/Array/ArrayInstance.cs
+++ b/Jint/Native/Array/ArrayInstance.cs
@@ -315,7 +315,7 @@ public class ArrayInstance : ObjectInstance, IEnumerable<JsValue>
         }
 
         var temp = _dense;
-        var properties = new List<JsValue>((temp?.Length ?? 0) + 1);
+        var properties = new List<JsValue>((temp?.Length ?? _sparse!.Count) + 1);
         if (temp != null)
         {
             var length = System.Math.Min(temp.Length, GetLength());
--- a/Jint/Native/Array/ArrayInstance.cs
+++ b/Jint/Native/Array/ArrayInstance.cs
@@ -315,7 +315,7 @@ public class ArrayInstance : ObjectInstance, IEnumerable<JsValue>
         }
 
         var temp = _dense;
-        var properties = new List<JsValue>(temp?.Length ?? 0 + 1);
+        var properties = new List<JsValue>((temp?.Length ?? 0) + 1);
         if (temp != null)
         {
             var length = System.Math.Min(temp.Length, GetLength());
--- a/Jint/Native/Object/ObjectInstance.cs
+++ b/Jint/Native/Object/ObjectInstance.cs
@@ -305,7 +305,7 @@ public partial class ObjectInstance : JsValue, IEquatable<ObjectInstance>
 
     private List<JsValue> GetOwnPropertyKeysSorted(List<JsValue> initialOwnPropertyKeys, bool returningStringKeys, bool returningSymbols)
     {
-        var keys = new List<JsValue>(_properties?.Count ?? 0 + _symbols?.Count ?? 0 + initialOwnPropertyKeys.Count);
+        var keys = new List<JsValue>((_properties?.Count ?? 0) + (_symbols?.Count ?? 0) + initialOwnPropertyKeys.Count);
         if (returningStringKeys && _properties != null)
         {
             foreach (var pair in _properties)
--- a/Jint/Runtime/Environments/GlobalEnvironment.cs
+++ b/Jint/Runtime/Environments/GlobalEnvironment.cs
@@ -380,7 +380,7 @@ internal sealed class GlobalEnvironment : Environment
         // JT: Rather than introduce a new method for the debugger, I'm reusing this one,
         // which - in spite of the very general name - is actually only used by the debugger
         // at this point.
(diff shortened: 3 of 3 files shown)


lahma:
Can you describe the net effect?

jnyrup:
> Can you describe the net effect?

Potentially fewer resizes of `List<T>`.

E.g. for the first change, `temp?.Length ?? 0 + 1` is currently evaluated as `temp?.Length ?? (0 + 1)` so we either specify a capacity of `temp.Length` or `1`.
When adding `temp.Length + 1` elements to `properties`, it might cause `properties` having to resize.

With the proposed change to `(temp?.Length ?? 0) + 1` , we specify a capacity of `temp.Length + 1` or `1`

lahma:
I pushed a follow-up commit (375cb8f) on top of this. While benchmarking I noticed the `ArrayInstance.GetOwnPropertyKeys` capacity has the *same* class of bug in its other branch: in sparse mode `_dense` (`temp`) is `null`, so even after your parenthesization the capacity collapses to `(null ?? 0) + 1 == 1` and the list regrows while adding every `_sparse` entry. The commit sizes that branch from `_sparse!.Count`, matching the existing `_dense?.Length ?? _sparse!.Count` idiom used elsewhere in the file.

### Which benchmark

There isn't an existing one that isolates these paths, so I wrote a focused `[MemoryDiagnoser]` benchmark hitting the three fixed sites. Since `List<T>` capacity is purely a sizing hint, the effect is **allocation / GC pressure**, not wall-clock — so the **Allocated** column is the signal, and it's deterministic. Wall-clock moved down slightly on all three but stays within the error bars.

`before` = base `main`, `after` = this PR head incl. the follow-up. `AMD Ryzen 9 5950X`, `.NET 10.0.8`, `ShortRunJob`.

| Method (per call) | Allocated before | Allocated after | Δ | Gen0/1k before → after |
|---|---:|---:|---:|---|
| `Object.keys` — dense array (256) | 8.95 KB | 4.93 KB | **−45 %** | 0.542 → 0.298 |
| `Object.keys` — sparse array (500) | 13.03 KB | 8.74 KB | **−33 %** | 0.794 → 0.534 |
| `Reflect.ownKeys` — numeric + symbol object | 4.24 KB | 2.78 KB | **−34 %** | 0.244 → 0.153 |

Mapping to the changes:
- **dense** — your `(temp?.Length ?? 0) + 1` fix: the backing store length equals the element count for an array literal, so adding the non-enumerable `length` key tipped the list one past capacity and forced a 256 → 512 realloc. Now pre-sized.
- **sparse** — the follow-up commit (`_sparse!.Count`): removes the regrow chain from capacity 1.
- **numeric + symbol** — your `ObjectInstance` fix: the sorted path's old capacity dropped the symbol and initial-key counts, so adding the symbols forced a regrow. Now the sum is exact.

<details><summary>benchmark source (ad-hoc, not committed)</summary>

```csharp
[MemoryDiagnoser]
[ShortRunJob]
public class GetOwnPropertyKeysBenchmark
{
    private Engine _engine = null!;
    private Prepared<Script> _denseKeys, _sparseKeys, _numericOwnKeys;

    [GlobalSetup]
    public void Setup()
    {
        _engine = new Engine();
        _engine.Evaluate("var denseArr = [" + string.Join(",", Enumerable.Range(0, 256)) + "];");
        _engine.Evaluate("var sparseArr = []; for (var i = 0; i < 500; i++) Object.defineProperty(sparseArr, i, { value: i, enumerable: true, configurable: true, writable: true });");
        _engine.Evaluate("var numObj = {}; for (var i = 0; i < 100; i++) numObj[i] = i; for (var j = 0; j < 16; j++) numObj[Symbol('s' + j)] = j;");
        _denseKeys = Engine.PrepareScript("Object.keys(denseArr);");
        _sparseKeys = Engine.PrepareScript("Object.keys(sparseArr);");
        _numericOwnKeys = Engine.PrepareScript("Reflect.ownKeys(numObj);");
    }

    [Benchmark] public JsValue DenseArrayKeys() => _engine.Evaluate(_denseKeys);
    [Benchmark] public JsValue SparseArrayKeys() => _engine.Evaluate(_sparseKeys);
    [Benchmark] public JsValue NumericObjectOwnKeys() => _engine.Evaluate(_numericOwnKeys);
}
```
</details>

PR #2531, opened 2026-06-13
Remove unused StringBuilder FNV-hash overload; simplify RefStack.Clear

lahma:
## Cleanup (no behaviour change)

Two small hygiene changes in the internal collection helpers:

- **Remove the unused `Hash.GetFNVHashCode(System.Text.StringBuilder)` overload.** It has no callers anywhere in the codebase — only the `string` overload is used (via `Key`). Dead code.
- **Simplify `RefStack<T>.Clear()`** to `Array.Clear(_array, 0, _size)`. It previously used a manual zeroing loop on net462/netstandard2.0 and `Array.Fill(_array, default, …)` elsewhere; `Array.Clear` is a JIT intrinsic available on every target framework and lets the `#if` go away.

This is a cleanup PR with **no performance claim** (so no benchmark is attached). It came out of a broader data-structure performance pass; the substantive, benchmarked changes will land as separate PRs.

## Verification

Clean build across all five TFMs (net462, netstandard2.0, netstandard2.1, net8.0, net10.0) — 0 warnings (warnings-as-errors is on). `Jint.Tests` green on net10.0 and net472.

### DIFF
--- a/Jint/Collections/RefStack.cs
+++ b/Jint/Collections/RefStack.cs
@@ -146,14 +146,9 @@ internal sealed class RefStack<T> : IEnumerable<T> where T : struct
     {
         if (_size > 0)
         {
-#if NETFRAMEWORK || NETSTANDARD2_0
-            for (var i = 0; i < _size; i++)
-            {
-                _array[i] = default;
-            }
-#else
-            Array.Fill(_array, default, 0, _size);
-#endif
+            // Array.Clear zeroes the elements (== default(T) for the struct T) and is a JIT intrinsic on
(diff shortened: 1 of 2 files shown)


lahma:
Refocused this PR to a pure cleanup: while gathering before/after numbers I found the `StringBuilder` `GetFNVHashCode` overload has no callers, so I removed it rather than micro-optimising dead code. The `RefStack.Clear` simplification stays. The benchmarked data-structure changes (array-backed small property bag, etc.) will follow as separate PRs with before/after tables.

PR #2532, opened 2026-06-13
Back small property bags with a flat array instead of a linked list

lahma:
## What

`HybridDictionary<T>` is the property bag behind every JS object and every function scope. Below its cutover to `StringDictionarySlim` (≤ 8 entries — the overwhelmingly common object shape) it stored entries in a singly-linked `ListDictionary`, allocating **one heap node per key** and chasing pointers on every lookup.

This backs the small regime with `ArrayDictionary<T>` — a single contiguous `Entry[]`:
- **one allocation instead of N nodes**, with no per-node object-header overhead;
- a **cache-friendly linear scan** instead of pointer chasing;
- the constructor now **honors the size hint** in the small regime too (it previously only pre-sized the hash backing), so object literals and scopes — which know their member count up front — allocate the array **exactly once, with zero growth churn**.

Insertion order, the ref-returning add/get API (`GetValueRefOrAddDefault` / `GetValueRefOrNullRef`), `checkExistingKeys` semantics, and `ClearPreservingCapacity` (pooled function environments) are all preserved. `ListDictionary` is removed — `HybridDictionary` was its only consumer.

## Benchmark (before / after)

`SmallObjectAllocBenchmark` creates 50,000 objects via object literals in a loop (the hot `BuildObjectFast` path). BenchmarkDotNet **default job**, `[MemoryDiagnoser]`; the baseline (linked list) and the change (array) were each built from their **own git worktree** to keep the A/B honest:

| Properties | Allocated: linked → array | Δ mem | Mean: linked → array | Δ time |
|--:|--|--:|--|--:|
| 1 | 14.63 MB → 14.63 MB | 0.0% | 8.90 → 9.29 ms | +4% |
| 2 | 19.66 MB → 18.52 MB | **−5.8%** | 10.26 → 10.38 ms | ~0 |
| 3 | 24.69 MB → 22.40 MB | **−9.3%** | 12.63 → 12.11 ms | −4% |
| 5 | 34.75 MB → 30.17 MB | **−13.2%** | 15.78 → 15.35 ms | −3% |
| 8 | 49.83 MB → 41.82 MB | **−16.1%** | 20.43 → 19.40 ms | −5% |

Allocation savings grow with property count; time is neutral-to-better. The 1-property case is allocation-neutral (no regression). A direct micro-probe of a sized 3-entry dictionary confirms the structure-level drop independently: **216 B → 168 B (−22%)**.

## Verification

- **test262: 0 failures** (99,260 passed) on net10.0.
- **Jint.Tests** green on net10.0 and net472.
- Clean build across all five TFMs (net462, netstandard2.0, netstandard2.1, net8.0, net10.0), 0 warnings.


lahma:
Closing — benchmark-falsified on real workloads.

The array backing wins on object **literals** (sized path → exact allocation, no churn: the −6…−16% from SmallObjectAllocBenchmark). But real code builds objects mostly via the `this.x = …` **constructor pattern**, which takes the unsized/lazy path where the array's geometric growth churns (Entry[1]→[2]→[4]). A Dromaeo + SunSpider A/B (each side built from its own worktree, allocation = deterministic signal) showed the change is **net-flat on Dromaeo's object benchmarks** and **regresses SunSpider access-binary-trees by +8.8% allocation** (44.78 → 48.73 MB) — that benchmark's TreeNode is the canonical `this.left=…;this.right=…;this.item=…` constructor.

The linked list's churn-free per-node allocation is hard to beat for unsized dynamic construction, so this is a no-go rather than shipping a regression on a very common pattern.

PR #2394, opened 2026-04-12
Improve Function.prototype.toString() to return source text for non-native functions

adams85:
I took a stab at implementing `Function.prototype.toString()` according to [the spec](https://tc39.es/ecma262/#sec-function.prototype.tostring). So this is another attempt at what #1726 tried to achieve, it's just takes the correct path instead of trying to convert the AST back to source code.

However, this requires the full source text to be kept around as it would be a performance killer to materialize the source text for functions eagerly. So this is deferred until execution. Once the source text is created, the reference to the full code is released. This way GC can at least collect that once source text for every function is materialized.

It passes the relevant test262 tests (it's failing here because it requires a minor non-breaking change on the parser side, which hasn't been released to Nuget yet). According to my manual tests, it also produces correct results for crazy dynamic cases like

```js
eval('function fn() {}');
fn.toString()
```

or

```
(async () => {
  async function f() {}
  var AsyncFunction = f.constructor;
  var p1 = "a", p2 = "/*a */ b, c";
  var g = /* before */AsyncFunction(p1, p2, "return Promise.resolve(a+b+c) /* d */ //")/* after */; 
  console.log(g.toString());
  console.log(await g(1, 2, 3));
})()
```

If the PR is welcome, I'll finalize it soon.

PR #2562, opened 2026-07-02
Make Function.prototype.toString() source text retention opt-in (#2560)

### LINKED ISSUE #2560: Increased memory use compared to v4.4.0 when using many large prepared scripts
### Pre-flight checks

### Describe the bug

We have a product that uses Jint with a shared cache of prepared scripts that can then be used by multiple threads (up to 50) that each create their own engine. Some of our clients have hundreds of scripts, some of which are quite large at ~1MB. This can result in an in-memory cache (of AST objects) of, say, 400-500MB.

In v4.4.0 this was working fine, however after upgrading to v4.9.1 one of our clients reported OutOfMemoryExceptions were occurring (for non-Jint reasons, we are restricted to 32 bit).

From analyzing a .dmp file from the client, `Jint.Runtime.Interpreter.JintFunctionDefinition` and its `+State` were the main candidates being flagged by Visual Studio's hot path analysis. Initially I wondered if the changes in #2413 (v4.9.0) could have had an impact. The .dmp showed 57 occurrences of `Jint.Engine`, but I know from other information that the client only had up to 24 threads. The hot path did indeed show instances of the engine rooted back to the prepared script cache so I did wonder if disposal of the engine had an issue with references leftover within the cache causing a memory leak - though I'd caveat this by noting that the .dmp was taken after the OutOfMemoryExceptions.

However the .dmp also showed ~250MB of strings, and I think #2394 (also v4.9.0) is the reason for this because of `JintFunctionDefinition.SourceText`, set via the changes to `Engine.Ast`'s `NodeVisitor` when called by `PrepareScript`. This appears to be per-engine too, so I'm seeing 57 instances of the same ~500kb string. Can storing this string be made optional or configurable (we don't use `Function.prototype.toString()`)?

I'm attempting to create a repro, but hopefully the above is enough information to begin with.

### Minimal JavaScript repro

```js
TBC
```

### C# host code

```csharp
TBC
```

### Expected behavior

Similar memory use to v4.4.0 with many, large prepared scripts

```text
Much larger memory use, in particular with strings containing full JS contents
```

### Jint version

4.9.1

### .NET runtime

net10.0

### Operating system

Windows

### AOT / trimming

None

### Engine configuration

- [ ] Modules enabled (`EnableModules`)
- [ ] Strict mode

### Area

None

### Regression?

4.4.0

### Additional context

_No response_

lahma:
Fixes #2560.

## Problem

Since v4.9.0 (#2394), `Function.prototype.toString()` returns the real source text of user functions. To support this, **every parsed function node pins the entire source string** (`Node.UserData = ctx.Input`), released only lazily on the first `toString()` call. Applications that cache many large **prepared** scripts and never call `toString()` therefore retain the full source for the lifetime of the cache — the reporter observed **~250 MB of duplicated source strings** (the same ~500 KB string retained once per engine).

## Fix

Add an opt-in `RetainFunctionSourceText` option, **default `false`**:

- `IParsingOptions.RetainFunctionSourceText` on `ScriptParsingOptions` / `ModuleParsingOptions` — controls per-parse retention (prepared and non-prepared).
- `Options.RetainFunctionSourceText` (+ fluent `options.RetainFunctionSourceText()`) — engine-level toggle for the default parser, so bare `engine.Execute(string)` can opt in without passing parsing options.

When retention is disabled (the default), source text is not kept and `toString()` returns the pre-4.9 `function name() { [native code] }` placeholder. When enabled, real source is returned exactly as before.

```csharp
// lean by default now — no source retained:
new Engine().Evaluate("function f(){} f.toString()"); // "function f() { [native code] }"

// opt back in:
new Engine(o => o.RetainFunctionSourceText()).Evaluate("function f(){} f.toString()"); // real source
Engine.PrepareScript(code, options: new ScriptPreparationOptions {
    ParsingOptions = new ScriptParsingOptions { RetainFunctionSourceText = true } });
```

> **Behavior change:** this reverts the v4.9 default (real-source `toString()`) to opt-in, trading it for lower memory by default. The Test262 harness and the existing `toString` source tests opt in so conformance is unchanged.

## Verification

- **Jint.Tests:** 3147 passed / 0 failed (excluding the pre-existing, environment-sensitive `EngineLimitTests.ShouldAllowReasonableCallStackDepth`, which crashes on unmodified `main` here too).
- **Test262:** 99260 passed / 0 failed — the `built-ins/Function/prototype/toString/*` conformance tests stay green via the harness opt-in.
- **Retained memory** (`GarbageCollectionTests.PreparedScriptsDoNotRetainSourceTextByDefault`): holding 25 distinct ~800 KB cached prepared scripts retains **20.1 MB with retention on vs 1.8 MB off — ~18.3 MB / 91% freed**.

The secondary env-pooling suspect (#2413) is left for a separate, benchmarked PR; it is a bounded, self-healing-under-load retention, unlike the quantified source-string regression fixed here.

https://claude.ai/code/session_01Bf1LESD4X8dMNPwLKJCAh9

### DIFF
--- a/Jint.Tests.Test262/Test262Test.cs
+++ b/Jint.Tests.Test262/Test262Test.cs
@@ -138,14 +138,19 @@ public abstract partial class Test262Test
             if (file.Type == ProgramType.Module)
             {
                 var specifier = "./" + Path.GetFileName(file.FileName);
-                engine.Modules.Add(specifier, builder => builder.AddSource(file.Program));
+                engine.Modules.Add(specifier, builder => builder
+                    .WithOptions(static options => options with { RetainFunctionSourceText = true })
+                    .AddSource(file.Program));
                 engine.Modules.Import(specifier);
             }
             else
             {
                 var script = Engine.PrepareScript(file.Program, source: file.FileName, options: new ScriptPreparationOptions
                 {
-                    ParsingOptions = ScriptParsingOptions.Default with { Tolerant = false, AllowReturnOutsideFunction = false },
+                    // Retain source text so Function.prototype.toString() conformance tests
+                    // (built-ins/Function/prototype/toString/*) validate the real-source output, which is
+                    // opt-in since the engine no longer retains source by default (issue #2560).
+                    ParsingOptions = ScriptParsingOptions.Default with { Tolerant = false, AllowReturnOutsideFunction = false, RetainFunctionSourceText = true },
                 });
 
                 engine.Execute(script);
--- a/Jint.Tests/Runtime/FunctionTests.cs
+++ b/Jint.Tests/Runtime/FunctionTests.cs
@@ -772,12 +772,12 @@ assertEqual(booleanCount, 1);
         (
         )=>0
         """)]
-    public void ToStringShouldProduceStandardsCompliantResultByDefault(string code, string expectedResult)
+    public void ToStringShouldProduceStandardsCompliantResultWhenSourceTextRetained(string code, string expectedResult)
     {
         code = Regex.Replace(code, @"\r\n?", "\n", RegexOptions.CultureInvariant); // normalize line endings to '\n'
         expectedResult = Regex.Replace(expectedResult, @"\r\n?", "\n", RegexOptions.CultureInvariant); // normalize line endings to '\n'
 
-        var returnValue = new Engine().Evaluate(code);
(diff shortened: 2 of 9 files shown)


scgm0:
Can't we just keep the source text once, have the node store its own slice position, and only create the string slice from the source text based on the position when calling `toString()`?

lahma:
You can always propose a PR, but sounds that we are still pinning the source string.

PR #2565, opened 2026-07-03
Slot-backed strict-eval environments with per-source pooling

lahma:
Strict eval creates a fresh environment per call, and it was always dictionary-backed: every identifier access in the eval body paid a dictionary probe, and the per-AST-node slot caches (identifier reads, unboxed `i++` updates) could never engage — while the *identical* body under `new Function()` runs on fixed slots. `dromaeo-core-eval` spends about half of each op inside one such body.

### What this does

* The eval compile-cache entry now precomputes the environment's **fixed-slot layout** (deduplicated `var` + function-declaration names initialized to undefined, `let`/`const` as TDZ templates) plus the hoisting scope, so `PerformEval` instantiates the strict-eval environment directly from templates.
* `EvalDeclarationInstantiation` gains the precomputed `HoistingScope` parameter (kills a per-call AST walk) and a `bindingsPreInitialized` flag that skips the var/lexical passes; its function bookkeeping (`LinkedList`/`HashSet`) is now allocated lazily — a win for all function-free evals, sloppy included.
* Entries whose body cannot capture the environment (no closures, nested eval or `with` — the existing `EnvironmentEscapeAstVisitor.MayEscape` gate) **pool the environment across calls**, keeping the shared statement tree's per-node slot caches valid on reused engines.
* The layout is built only when it can amortize: a **loop in the body** (many accesses in one call) or a **promoted source** (repeats, pooled env carries caches across calls). One-shot loopless evals pay nothing.
* ShadowRealm's eval path reuses `CachedHoistingScope` for prepared scripts instead of re-walking the AST per call.

Non-strict direct eval (vars land in the caller's environment) is untouched.

### Measurements (same-window stash A/B, BenchmarkDotNet default job)

| Benchmark | main | PR | Δ |
|---|---:|---:|---|
| EvalHotReusedEngine | 1,313.2 µs / 79.2 KB | 1,069.6 µs / 77.0 KB | **−18.6%** |
| EvalFreshEngine | 1,279.7 µs / 174.6 KB | 1,056.3 µs / 174.7 KB | **−17.5%** |
| EvalSameSource | 264.9 µs / 80.7 KB | 199.4 µs / 42.0 KB | **−24.7% time, −47.9% alloc** |
| EvalDistinctSources (miss-path guard) | 101.4 µs / 201.2 KB | 100.9 µs / 201.2 KB | flat |
| NewFunctionHot / NewFunctionFresh / PlainFunctionLoopReference / ParseOnlyTmp | | | within noise |
| ShadowRealm | 2,900 ns | 2,824 ns | −2.6% |
| ShadowRealm_ParsedScript | 1,607 ns | 1,460 ns | **−9.2%** |
| dromaeo-core-eval (CpuProfileDriver ×400, source) | 4.021 ms/iter | 3.679 ms/iter | **−8.5%** |
| dromaeo-core-eval (CpuProfileDriver ×400, prepared) | 4.154 ms/iter | 3.917 ms/iter | **−5.7%** |

`EvalExecutionBenchmarks` (new in this PR) isolates the dromaeo-core-eval execution shape; `--profile-time` (new) splits dromaeo scripts per `test()` with Stopwatch, complementing `--profile-memory`.

Together with the follow-up compound-assignment PR, the EngineComparison `dromaeo-core-eval` row moves 2.507 → 1.941 ms (−22.6%) / prepared 2.480 → 1.909 ms (−23.0%), allocation flat (same-runtime worktree A/B).

### Gates

* `Jint.Tests` 3159/3097 (net10/net472), `Jint.Tests.PublicInterface` 79/79, `Jint.Tests.CommonScripts` 28/28 — all green
* Test262: 99,260 passed, 0 failed (133 skipped)
* All-TFM `Jint.csproj` Release build
* New `EvalTests` cover: pooled-env binding freshness, closures escaping eval (pooling gate), re-entrant same-source eval, function-declaration hoisting in slots, const TypeError, let TDZ, >cap fallback, sloppy leak semantics, indirect strict eval

### DIFF
--- a/Jint.Benchmark/EvalExecutionBenchmarks.cs
+++ b/Jint.Benchmark/EvalExecutionBenchmarks.cs
@@ -1,0 +1,95 @@
+using BenchmarkDotNet.Attributes;
+using Jint.Native;
+
+namespace Jint.Benchmark;
+
+/// <summary>
+/// Isolates the *execution* side of the dromaeo-core-eval shape (16 copies of a 1000-iteration
+/// string-append loop, ~1 KB source), complementing <see cref="EvalCacheBenchmarks"/> which
+/// isolates the compile pipeline. Hot rows reuse one engine with the eval compile cache warmed
+/// (pure body execution); Fresh rows create an engine per op like EngineComparisonBenchmark.
+/// PlainFunctionLoopReference runs the identical body as a regular function — the structural
+/// floor the eval rows should converge to. ParseOnlyTmp bounds the per-op parse share.
+/// </summary>
+[MemoryDiagnoser]
+[HideColumns("Error", "Gen0", "Gen1", "Gen2")]
+public class EvalExecutionBenchmarks
+{
+    private const string Cmd = """var str="";for(var i=0;i<1000;i++){str += "a";}ret = str;""";
+
+    private string _tmp16 = null!;
+
+    private Engine _evalEngine = null!;
+    private Engine _newFunctionEngine = null!;
+    private Engine _plainFunctionEngine = null!;
+
+    private Prepared<Script> _evalCall;
+    private Prepared<Script> _newFunctionCall;
+    private Prepared<Script> _plainFunctionCall;
+    private Prepared<Script> _freshEvalScript;
+    private Prepared<Script> _freshNewFunctionScript;
+
+    [GlobalSetup]
+    public void Setup()
+    {
+        _tmp16 = string.Concat(Enumerable.Repeat(Cmd, 16));
+
+        // Fresh-engine variants mirror the dromaeo script: build tmp by doubling, then run once.
+        const string BuildTmp = $"var ret; var cmd = '{Cmd}'; var tmp = cmd; for (var n = 0; n < 4; n++) tmp += tmp;";
+        _freshEvalScript = Engine.PrepareScript(BuildTmp + " eval(tmp);", strict: true);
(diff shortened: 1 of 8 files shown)


lahma:
Adversarial multi-agent review of the stack surfaced issues; fixed in fbd90a550:

* **Shadow-aware read walk**: the identifier slot-cache hit walk hopped intermediate environments by reference only, so a cached-eval body executing under a chain that shadows the name (a `let`/`const` block at a different call site, a nested eval of the same source, or sloppy eval var-injection) read a stale outer binding. The walk now probes each hopped declarative environment with `HasBinding` (pure; hop-0 hits — all the hot loops — pay nothing). Part of this was reproducible on main before this PR; slot-backed eval environments widen its reach, so the fix ships here.
* **Spec fix**: eval var names were validated with `CanDeclareGlobalFunction` instead of `CanDeclareGlobalVar` (sec-evaldeclarationinstantiation 8.a.ii.1) — `eval("var undefined;")` threw a spurious TypeError. Pre-existing; the lines live in this diff.
* **Pool hygiene**: pooled eval environments parked with the completed call's binding values still in the slots (engine-lifetime retention of arbitrary object graphs); slots now reset to templates at park time.
* Loops inside arrow bodies no longer mark the eval body loop-bearing for the slot-layout gate.

Five regression tests cover the repro shapes. Full suites + Test262 99,260/0 re-verified on the stack.

PR #2568, opened 2026-07-03
Per-engine environment caches: prepared scripts no longer pin engines; computed-key and escape-analysis fixes

lahma:
Follow-up to #2562, addressing the secondary cause of #2560 — plus adjacent correctness bugs surfaced by three adversarial review passes over the branch.

## Problem

Since #2413, a call's `FunctionEnvironment` is returned to a reuse slot on the **shared** `JintFunctionDefinition.State`, and block environments are similarly cached on `BlockState` (shared via AST `UserData`). An environment roots its creating engine (`Environment._engine` plus its outer chain to the global), and a prepared script's shared state is reused by every engine that runs it — so these caches pin the **last-caller engine** (realm, intrinsics, globals) until the same code is next executed by any engine. For idle functions/blocks in long-lived cached prepared scripts, the engine is pinned indefinitely.

### Environment caches become per-engine
- **Function envs** — cached on the `ScriptFunction` instance (`_envReuse`, one `object` field): the `FunctionEnvironment` directly for ordinary functions, or a bounded `RecursiveEnvPool` (LIFO, 16) for direct-recursive ones, dispatched by the existing `State.IsDirectRecursive`. Function instances are per engine, so the cache dies with the engine. The env-side `Interlocked` handshake and cross-engine identity checks are gone; renting happens only for plain calls (the construct path never returns its env, so renting there just drained the cache).
- **Block envs** — cached on the `JintBlockStatement` handler instance (built per statement list ⇒ per engine) instead of the shared `BlockState`. Parked envs are reset at park time (slots + outer chain detached) so they don't root the completed call's scope graph. `JintForStatement._cachedLoopEnv` needed no change — statement handlers are never cached on AST `UserData`, so it was already per-engine (verified by test).
- Cleared fixed-slot `Binding[]` hold no engine references and stay shared on `State._cachedSlots` (unchanged `Interlocked`), so freshly created instances of the same function still reuse the slot array across engines.

### Escape-analysis fixes (pre-existing, the caches rely on this gate)
- **Parameter defaults**: `MayEscape` scanned only the function body, so a closure in a parameter default expression escaped undetected and its environment was cached + rebound (`function f(a, get = function() { return a; }) {...}` — `g1()` returned the *next* call's `a`).
- **`this`/`new.target`/`super`**: the slot-aware analysis matched only identifiers, so an escaping arrow depending solely on `this` was classified non-capturing and its env rebound (`h1().name` returned `'o2'`). Any such node inside a closure now counts as an environment reference.

### Computed property keys evaluate correctly (pre-existing)
- `AstExtensions.TryGetComputedPropertyKey` had a node-type allowlist with a silent `undefined` fallback: keys like `{ [(sideEffect(), "k")]: v }` (SequenceExpression) or `{ [new Key()]: v }` bound the property `"undefined"` and never ran the key expression. The allowlist is gone — every expression type is evaluated.
- Decorator bookkeeping re-called `GetKey` after the class element was defined, re-running computed key side effects and applying decorators to the wrong name when the key changed between evaluations. The defined key is now threaded through and computed keys evaluate exactly once.
- Object literals check for generator suspension after evaluating a computed *method* key (mirrors the existing non-method check).

### Test infrastructure
- `EngineLimitTests.ShouldAllowReasonableCallStackDepth` crashed the whole test process intermittently on unmodified `main` (990 nested calls, no stack guard, unpredictable runner-thread stack). It now runs on a dedicated 16 MB-stack thread — deterministic and platform-independent (macOS skip removed).

## Verification

- **Engine retention** (`SharedPreparedScriptDoesNotRetainEngines`): one shared prepared script containing an ordinary function, a direct-recursive one, a `let/const` block and a `for-let` loop, run by 20 throwaway engines tracked via `WeakReference` — all must be collectable. On `main`, the function-env cache pins 1/20; after the function-env fix alone, the `let/const` block still pinned 1/20; now 0/20.
- **Behavior regression tests**: param-default closures, `this`/`new.target` capture, computed keys across expression types (params/literals/destructuring), decorated computed keys evaluated exactly once.
- **Suites**: Jint.Tests 3195 / 0 failed; PublicInterface clean; Test262 **99260 / 0** (multiple runs; the only failures ever observed were 4 AnnexB BMP eval-loop tests timing out under full parallel load on a busy host — reproduced identically on unmodified `main`, pass in seconds in isolation).
- **Benchmarks** (default BDN job, same-runtime worktree A/B vs `main`; measured before the block-env/computed-key commits, which are allocation-neutral — verified via deterministic alloc counters):

| Benchmark | main | PR | Δ time | Δ alloc |
|---|---|---|---|---|
| Recursion/Fib | 564.1 ms / 1 065 KB | 510.6 ms / 1 065 KB | −9.5% | = |
| Recursion/Tak | 15.19 ms | 14.96 ms | −1.5% | = |
| Recursion/DeepSum | 242.0 ms | 232.5 ms | −3.9% | = |
| EmptyClosureCall | 40.62 ms / 2.74 MB | 35.38 ms / 2.74 MB | −12.9% | = |
| CapturedVarReadWrite | 52.99 ms | 50.29 ms | −5.1% | = |
| ParamLocalCall | 106.1 ms / 11.74 MB | 104.9 ms / 11.74 MB | −1.1% | = |
| ManyLocalCall | 148.5 ms | 148.3 ms | −0.1% | = |
| DeclareMany | 73.84 µs / 166.7 KB | 73.05 µs / 170.6 KB | −1.1% | +3.9 KB |
| ConstructEach | 217.6 µs / 417.6 KB | 221.0 µs / 421.5 KB | +1.5% | +3.9 KB |
| Warm_Call | 590 ns / 1.64 KB | 582 ns / 1.76 KB | −1.4% | +0.12 KB |
| Warm_Apply | 699 ns / 1.91 KB | 692 ns / 2.02 KB | −1.1% | +0.11 KB |
| Warm_Bind | 488 ns / 1.73 KB | 476 ns / 1.74 KB | −2.5% | = |
| Warm_BindThenCall | 980 ns / 2.63 KB | 963 ns / 2.74 KB | −1.7% | +0.11 KB |

The recursion/closure/loop workloads are allocation-identical and consistently faster (no env-side `Interlocked`; LIFO recursive pool). The remaining alloc deltas are the direct cost of the leak fix: 8 bytes per `ScriptFunction` (the new field; `DeclareMany`/`ConstructEach` create 500 instances) and one env allocation when a *re-created* function instance is called — environments must not outlive their engine, which is the point of the fix.

## Notes for reviewers
- Retention bounds are now per live instance (16 recursive envs per live direct-recursive function instance vs 16 per State before); environments die with their instance/engine.
- Public `AstExtensions.GetKey(..., resolveComputed: true)` on host-built exotic nodes (e.g. `PreserveParens` ASTs, which Jint's own pipeline never produces) now surfaces `JintExpression.Build`'s "unsupported expression type" exception instead of silently returning an `"undefined"` key.

https://claude.ai/code/session_01Bf1LESD4X8dMNPwLKJCAh9


lahma:
Rebased onto latest `main` (059f1742) and ran a second adversarial review pass; two more pre-existing correctness issues in the escape-analysis family were found and fixed here, since this PR's caching relies on that gate:

- **`this`/`new.target`/`super` capture**: the slot-aware escape analysis only matched identifiers against slot names, so an escaping arrow whose only dependency is `this` was classified non-capturing and its environment was cached + rebound by the next call (`h1().name` returned `'o2'`; reproducible on `main` with the old `State._cachedEnv` too). Any such node inside a closure now counts as an environment reference. Regression tests added for both the `this` and `new.target` variants.
- **Construct rent gate**: the construct path rented from the reuse cache but never returns its environment (upstream parity), so `new f()` drained the cache and starved the next call's reuse. Renting now happens only for plain calls.
- `EngineLimitTests` uses a generous 16 MB dedicated stack after all: `StackOverflowException` is uncatchable and a tight budget could kill the whole test process on platforms with larger native frames.

Gates after the fixes: Jint.Tests 3163/0, Test262 99260/0, benchmarks vs `main`: Fib −10.3%, ParamLocalCall −12.7%, closures −6..7%, `Warm_*` −3..6%, allocations unchanged.

The review also confirmed a pre-existing, unrelated bug (computed destructuring keys with `SequenceExpression`/`NewExpression`/etc. silently bind `undefined` via `AstExtensions.TryGetComputedPropertyKey`'s allowlist) — will file separately.

lahma:
Post-merge observation, bisected: repeated `new Function()` on one engine regressed with this change.

`EvalExecutionBenchmarks` on adjacent worktrees (same window, only #2573 between the two commits):

| Row | f7935bbbe (pre) | eceb7ba5e (this PR) | Δ |
|---|---:|---:|---|
| NewFunctionHotReusedEngine | 1.052 ms | 1.390 ms | **+32%** |
| NewFunctionFreshEngine | 1.124 ms | 1.025 ms | −9% |

Likely mechanism: every `new Function(...)` call creates a fresh `ScriptFunction` (the compiled `JintFunctionDefinition` is shared via the dynamic-function cache, but the instance is new), so per-instance `_envReuse` never warms for this shape — each call pays the environment setup the previous per-State pool amortized. Fresh-engine and eval rows are unaffected/improved.

This also shows in the `dromaeo-core-eval` EngineComparison row (its `new Function` subtest is ~half the op). A possible direction: let dynamic functions (or single-invocation instances generally) fall back to definition-level reuse. Happy to pick this up as a follow-up PR if useful.

PR #2570, opened 2026-07-03
Fix Map keys()/values() throwing InvalidOperationException on mutation

svenrog:
Iterating a Map's keys() or values() while mutating the map, which is spec-legal, throws a CLR exception that escapes to the host. Because it's a raw InvalidOperationException (not a JavaScriptException), script-level try/catch can't catch it and CatchClrExceptions() doesn't wrap it.

An example of a test that would fail 

```
const m = new Map([['a', 1], ['b', 2], ['c', 3]]);
for (const k of m.keys()) m.delete(k);
// System.InvalidOperationException: Collection was modified;
// enumeration operation may not execute.
```

## Summary

Adjusted the `MapIteratorPrototype` to reuse the already working `MapIterator` for the `keys()` and `values()`-iterations.

## Details

`MapIteratorPrototype` was updated with:
- An enum, `MapIteratorKind` specifying the three possible types of iteration.
- An additional private field on `MapIterator` (passed via additional constructor parameter).
- Method `TryIteratorStep` was modified so that the `MapIteratorKind` can be used to determine the kind of `IteratorResult` to return.

## Test plan

## Breaking change?

No.

### DIFF
--- a/Jint.Tests/Runtime/MapTests.cs
+++ b/Jint.Tests/Runtime/MapTests.cs
@@ -27,6 +27,64 @@ public class MapTests
     }
 
     [Fact]
+    public void KeysIteratorToleratesDeletionDuringIteration()
+    {
+        const string Script = @"
+            var map = new Map([['a', 1], ['b', 2], ['c', 3]]);
+            var count = 0;
+            for (var key of map.keys()) { map.delete(key); count++; }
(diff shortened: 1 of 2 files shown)


lahma (approved):
Thank you!

PR #2591, opened 2026-07-06
Fix raw SetProperty on builtin-shaped hosts landing in an invisible dictionary

lahma:
\`SetProperty(Key, ...)\` deopted hidden-class \ShapeMode\ objects before storing a raw descriptor, but not \BuiltinShapeMode\ hosts: the write created a side \_properties\ dictionary that the shape-mode read paths never consult, so the property silently did not exist. Reachable through the public \FastSetProperty\ API on any shaped built-in (\Math\, \JSON\, prototypes, ...) since the shapes migration.

Regression test included (raw store on shaped \Math\: value visible, enumerable via \getOwnPropertyNames\, existing shape members still functional after the deopt).

## Verification

- \Jint.Tests\ green (net10.0 + net472), \Jint.Tests.PublicInterface\ green
- Full test262: 99,260 passed, 0 failed

### DIFF
--- a/Jint.Tests/Runtime/PropertyDescriptorTests.cs
+++ b/Jint.Tests/Runtime/PropertyDescriptorTests.cs
@@ -99,6 +99,21 @@ public class PropertyDescriptorTests
     }
 
     [Fact]
+    public void FastSetPropertyIsVisibleOnBuiltinShapedHost()
+    {
+        // Math uses builtin-shape storage; a raw property store must deopt it to dictionary
+        // mode — without that, the write lands in a side dictionary the shape-mode read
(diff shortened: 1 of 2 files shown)


lahma:
Note for reviewers: while writing the regression test I found a related but distinct pre-existing edge — a raw \SetProperty\/\FastSetProperty\ issued **before** a generated host's lazy \Initialize\ has run gets wiped when initialization later replaces the property bag (\SetProperties\), and that affects dictionary-path hosts too, not just shaped ones. This PR intentionally fixes only the post-initialization shaped-host case (the write landing in a dictionary the shape reads never consult); the pre-init store is left as a separate issue since it predates the shapes work.

PR #2607, opened 2026-07-09
Fix unlabeled break escaping a labeled switch statement

svenrog:
## Summary

Absorb an unlabeled `break` inside a labeled `switch` statement.

## Details

An unlabeled `break` inside a **labeled** `switch` was not absorbed by the switch:

```js
function f(e) {
    var o = 0;
    outer: switch (e) {
        case 1: o = 8; break;
        default: o = 29;
    }
    return o;
}
f(1); // returned undefined, should be 8
```

`JintSwitchStatement.HandleCompletion` only converted a `Break` completion to `Normal` when the break's target equalled the switch's own label:

```csharp
if (r.Type == CompletionType.Break
    && string.Equals(context.Target, _statement.LabelSet?.Name, StringComparison.Ordinal))
```

For an unlabeled `break`, `context.Target` is `null`. When the switch is unlabeled, `LabelSet?.Name` is also `null`, so `null == null` absorbed it. But once the switch carries a label, `LabelSet.Name` is non-null, so the unlabeled break (target `null`) failed the equality check and **propagated out** as a `Break` completion — skipping statements after the switch and, at function scope, making the body complete abnormally so the function returned `undefined`.

All loop statements (`for`, `for-in/of`, `while`, `do-while`) already guard this with `context.Target == null || string.Equals(...)`; the switch was the only breakable missing the null-target arm. This adds it, so a switch always absorbs an unlabeled break regardless of its own label — matching the loops and the spec (a `SwitchStatement` absorbs break completions with an empty target).

Found running minified React 19 inside a pure-JS DOM: `createFiberFromTypeAndProps` compiles to `e: switch (e) { case T: o = 8; l |= 24; break; ... }` and returned `undefined` for `StrictMode` elements, which then blew up downstream as `Cannot convert undefined or null to object`.

## Linked issue

No tracking issue — discovered while running minified React 19 on Jint. Refs #

## Test plan

Added three regression tests to `Jint.Tests/Runtime/SwitchTests.cs` (unlabeled break absorbed by a labeled switch; labeled break still escapes correctly; unlabeled break in a nested switch stays innermost). 3168 unit tests pass, and 737 Test262 `switch`/`labeled`/`break`/`continue` conformance tests pass, 0 failures.

## Breaking change?

No.


lahma (approved):
Thank you!

PR #2709, opened 2026-07-18
Fix spurious TDZ when a for-header reads a name the loop body shadows

### LINKED ISSUE #2708: Issue with lexical scoping in for-loop per-iteration environments
### Pre-flight checks

### Describe the bug

A `for` loop whose header (init, test, or update) references a variable that the loop body re-declares with `let`/`const` throws a spurious `ReferenceError` (`"<name> has not been initialized"`). Per spec the header reference resolves to the outer binding. The body's block-scoped declaration is not in scope for the header, so it should read the outer value, as V8/SpiderMonkey/JSC do.

Introduced by #2610 ("Flatten eligible loop-body lexicals into the pooled loop environment"): the flattening folds the body block's `let`/`const` slots into the loop environment, where the header then resolves the name to the still-uninitialized (TDZ) body slot. The eligibility gate only checks body names against the header's declared names, so it misses the case where a body name shadows an outer name the header references.`

### Minimal JavaScript repro

```js
(function () {
  let r = 5, o = 8, seen = [];
  for (let l = r; l < o; l++) {   // `r` here is the OUTER r (=5); body's `let r` is not in scope for the header
    let r = l * 10;
    seen.push(r);
  }
  return seen.join(',');
})()
```

### C# host code

```csharp
// Default engine, no options — reproduces out of the box.
var result = new Engine().Evaluate(
    "(function(){let r=5,o=8,s=[];for(let l=r;l<o;l++){let r=l*10;s.push(r);}return s.join(',');})()");
// throws Jint.Runtime.JavaScriptException: r has not been initialized
```

### Expected behavior

Returns `"50,60,70"`. The header's `r` reads the outer binding; the body's `let r` shadows only inside the body. Matches V8 (Node/Chrome), SpiderMonkey, and JavaScriptCore. Spec: ForBodyEvaluation (https://tc39.es/ecma262/#sec-forbodyevaluation) / CreatePerIterationEnvironment (https://tc39.es/ecma262/#sec-createperiterationenvironment) — the per-iteration lexical environment binds only the header's `BoundNames`; the body block is a separate, nested lexical scope.

```text
Jint.Runtime.JavaScriptException: r has not been initialized
The header reference resolves to the body's flattened slot while it is still uninitialized (TDZ). It also fires from the test (for (let l = 0; l < r; l++) { let r = ... }) and update (for (...; ...; step += ...) { let step = ... }) positions, not just init.
```

### Jint version

4.12.0

### .NET runtime

net10.0

### Operating system

Windows

### AOT / trimming

No — JIT, no trimming

### Engine configuration

- [ ] Modules enabled (`EnableModules`)
- [ ] Strict mode

### Area

ECMAScript language / parser

### Regression?

Yes, worked in 4.11.0, broke in 4.12.0 (#2610).

### Additional context

Real-world impact: the shape is common in minified bundler runtimes. Turbopack's module-factory loop (`for (let l = r; l < o; l++) { let r = e[l], o = t.get(r); … }`) hits it, which aborts execution of Turbopack/Next.js client bundles under Jint. A fix (guarding the flattening eligibility so it declines when a body slot name is referenced in the header) with unit tests is ready and I'll open a PR referencing this issue — full `Jint.Tests` and Test262 pass with no regressions.

svenrog:
## Summary
Fix spurious TDZ when a for-header reads a name the loop body shadows

## Details
Loop-body lexical flattening (#2610) folds an eligible body block's `let`/`const`
bindings into the pooled loop environment. The eligibility gate (`NamesOverlap`)
only checked body names against the header's *declared* names — it missed the case
where a body name shadows an *outer* binding that a header expression references:

```js
for (let l = r; l < o; l++) { let r = e[l], o = t.get(r); ... }
```

Here `r`/`o` in the header resolve to outer bindings, but flattening folded the
body's `r`/`o` slots into the loop env, so the header read them while still
uninitialized → `ReferenceError: r has not been initialized`. V8/SpiderMonkey/JSC
run it fine. This is the exact shape in Turbopack's module-factory runtime, so it
aborted rendering of Turbopack/Next.js bundles. The bug fires from the test and
update positions too, not just init.

Fix: add `HeaderReferencesAnyBodyName`, declining to flatten when any body slot
name appears as an identifier in the init/test/update. It over-approximates in the
safe direction (a matching property name or nested-scope declaration only forgoes
the optimization), mirroring `EnvironmentEscapeAstVisitor.ClosureReferencesAny`.
The scan runs once at statement construction, so there is no per-iteration cost,
and flattening still applies to loops whose headers reference no shadowed name.

Regression introduced in 4.12.0 (#2610); worked in 4.11.0.

## Linked issue

Fixes #2708

## Test plan

- [x] Added or updated unit tests in `Jint.Tests` (`LoopBodyFlatteningTests`: init/test/update header refs + a reduced Turbopack module-factory shape)
- [x] Ran `dotnet test --configuration Release` locally (3629 net10.0 / 3566 net472, 0 failures)
- [x] For ECMAScript spec changes: ran `Jint.Tests.Test262` and confirmed no regressions (99431 pass, 0 fail, 135 skip)
- [ ] For interop changes: covered in `Jint.Tests/Runtime/Interop` — n/a
- [ ] For perf changes: included before/after numbers from `Jint.Benchmark` — n/a (narrows an existing optimization; construction-time guard only)

## Breaking change?

No.

### DIFF
--- a/Jint.Tests/Runtime/LoopBodyFlatteningTests.cs
+++ b/Jint.Tests/Runtime/LoopBodyFlatteningTests.cs
@@ -143,6 +143,89 @@ public class LoopBodyFlatteningTests
     }
 
     [Fact]
+    public void HeaderInitReferencingOuterNameShadowedByBodyReadsOuter()
+    {
+        var engine = new Engine();
+        // the header init reads `r`; the body block declares its own `r`. The names do not overlap
+        // the header's declared name (`l`), so the old NamesOverlap gate let flattening fold the
+        // body's `r` slot into the loop env — where the init's `r` then resolved to the
+        // still-uninitialized slot and threw a spurious TDZ. The init's `r` is an OUTER reference.
+        var result = engine.Evaluate("""
+            (function () {
+                let r = 5, o = 8, seen = [];
+                for (let l = r; l < o; l++) {
+                    let r = l * 10;
+                    seen.push(r);
+                }
+                return seen.join(',');
+            })()
+            """).AsString();
+
+        Assert.Equal("50,60,70", result);
+    }
+
+    [Fact]
+    public void HeaderTestReferencingOuterNameShadowedByBodyReadsOuter()
+    {
+        var engine = new Engine();
+        // same shape, but the outer name is read from the loop test rather than the init
+        var result = engine.Evaluate("""
+            (function () {
+                let r = 3, seen = [];
+                for (let l = 0; l < r; l++) {
+                    let r = l * 10;
+                    seen.push(r);
+                }
+                return seen.join(',');
+            })()
+            """).AsString();
+
+        Assert.Equal("0,10,20", result);
+    }
+
+    [Fact]
+    public void HeaderUpdateReferencingOuterNameShadowedByBodyReadsOuter()
+    {
+        var engine = new Engine();
+        // the update reads outer `step`; the body shadows `step`
+        var result = engine.Evaluate("""
+            (function () {
+                let step = 2, seen = [];
+                for (let l = 0; l < 6; l += step) {
+                    let step = l * 10;
+                    seen.push(step);
+                }
+                return seen.join(',');
+            })()
+            """).AsString();
+
+        Assert.Equal("0,20,40", result);
+    }
+
+    [Fact]
+    public void TurbopackChunkShapeResolvesModuleFactory()
+    {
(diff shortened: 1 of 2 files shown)


lahma (approved):
Thank you!

PR #2750, opened 2026-07-23
Consult reference resolver for a call to an unresolvable identifier

### LINKED ISSUE #2749: Custom IReferenceResolver not consulted for a call to an unresolvable identifier since 4.6 (InvalidCastException, regresses #1041)
### Pre-flight
- [x] Searched open/closed issues (this is a regression of #1041, not a duplicate)
- [x] Reproduces against the latest release (verified on 4.13.0; `main`'s `JintCallExpression` still has the same guard)
- [x] Confirmed it's Jint, not Acornima (runtime, not parser)

### Describe the bug
A custom `IReferenceResolver` that returns a value for unresolvable references (the null-propagation helper deviation added for #1041) is no longer consulted when an unresolvable identifier is used in call position. The call throws a raw `InvalidCastException` instead. Reading the same identifier still goes through the resolver and returns the resolved value; only calls break.

### Minimal JavaScript repro
```js
foo()
```
This only reproduces with the custom resolver below. On a default engine it is a normal `ReferenceError`.

### C# host code
```csharp
var engine = new Engine(o => o.ReferenceResolver = new NullPropagationResolver());
engine.Evaluate("var x = foo;");  // fine, resolver returns undefined
engine.Evaluate("foo();");        // throws InvalidCastException

sealed class NullPropagationResolver : IReferenceResolver
{
    public bool TryUnresolvableReference(Engine e, Reference r, out JsValue v) { v = JsValue.Undefined; return true; }
    public bool TryPropertyReference(Engine e, Reference r, ref JsValue v) => false;
    public bool TryGetCallable(Engine e, object c, out JsValue v) { v = new ClrFunction(e, "noop", (_, _) => JsValue.Undefined); return true; }
    public bool CheckCoercible(JsValue v) => true;
}
```

### Expected behavior
Same as before #2266, and same as the read path today: the resolver is consulted, so the call resolves via `TryGetCallable` (here a no-op returning `undefined`). This is the behavior established by the #1041 fix (commit ff6765b, the "deviation from the spec to support null-propagation helper").

```text
System.InvalidCastException: Unable to cast object of type 'Jint.Native.JsString' to type 'Jint.Runtime.Environments.Environment'.
   at Jint.Runtime.Interpreter.Expressions.JintCallExpression.EvaluateInternal(EvaluationContext context)
   at Jint.Runtime.Interpreter.Expressions.JintExpression.GetValue(EvaluationContext context)
   at Jint.Engine.Evaluate(String code, String source)
```

### Jint version
4.13.0

### .NET runtime
net10.0

### Operating system
Windows

### Regression?
Worked through 4.5.x; broke in 4.6.0 (#2266). #2266 changed the base of an unresolvable reference to the `Reference.Unresolvable` sentinel, but `JintCallExpression` only consults the resolver when `baseValue.IsNullOrUndefined()`, so the sentinel skips the deviation and is cast to `Environment`.

### Additional context
One-line fix that restores the #1041 deviation under the new model:
```diff
- if (baseValue.IsNullOrUndefined()
+ if ((baseValue.IsNullOrUndefined() || referenceRecord.IsUnresolvableReference)
      && engine._referenceResolver.TryUnresolvableReference(engine, referenceRecord, out var value))
```
Happy to open a PR with this plus a `Jint.Tests` case.

poissoncorp:
## Summary

Consult the reference resolver for a call to an unresolvable identifier.

## Details

Since #2266 the base of an unresolvable reference is the `Reference.Unresolvable` sentinel rather
than `undefined`. `JintCallExpression` only consulted a custom `IReferenceResolver` when
`baseValue.IsNullOrUndefined()`, so a call to an undefined name (base = sentinel) skipped the
resolver and hit `(Environment) baseValue`, throwing `InvalidCastException`. A read of the same name
still went through the resolver, so only calls were affected.

This extends the existing "null-propagation helper" deviation to also fire for unresolvable
references (`referenceRecord.IsUnresolvableReference`), restoring the behavior the #1041 fix
established. A call to an undefined name is now routed through the resolver (e.g. resolved via
`TryGetCallable`) instead of throwing a CLR cast exception.

## Linked issue

Fixes #2749

## Test plan

- [x] Added a unit test in `Jint.Tests` (`Runtime/NullPropagation.CanCallUnresolvableReference`)
- [x] Ran `dotnet test --configuration Release` locally. `NullPropagation` passes on net10.0 and net472; the new test fails without the fix with the exact `InvalidCastException`
- [ ] ECMAScript spec change. N/A
- [ ] Interop change. N/A
- [ ] Perf change. N/A

## Breaking change?

No. It restores the pre-#2266 behavior for custom `IReferenceResolver` implementations and does not change the public API.

### DIFF
--- a/Jint.Tests/Runtime/NullPropagation.cs
+++ b/Jint.Tests/Runtime/NullPropagation.cs
@@ -120,6 +120,20 @@ function test2(arg) {
     }
 
     [Fact]
+    public void CanCallUnresolvableReference()
+    {
+        // A call to an unresolvable identifier must be routed through the reference resolver,
+        // like a read is, rather than throwing. Regression: the unresolvable reference base is a
+        // sentinel (not undefined), so the call path stopped consulting the resolver and cast the
+        // sentinel to an Environment, throwing InvalidCastException.
+        var engine = new Engine(cfg => cfg.SetReferencesResolver(new NullPropagationReferenceResolver()));
+
+        var act = () => engine.Execute("var read = missing; var called = missing();");
+
+        act.Should().NotThrow();
+    }
+
+    [Fact]
     public void NullPropagationShouldNotAffectOperators()
     {
         var engine = new Engine(cfg => cfg.SetReferencesResolver(new NullPropagationReferenceResolver()));
--- a/Jint/Runtime/Interpreter/Expressions/JintCallExpression.cs
+++ b/Jint/Runtime/Interpreter/Expressions/JintCallExpression.cs
@@ -120,8 +120,11 @@ internal sealed class JintCallExpression : JintExpression
                 {
                     var baseValue = referenceRecord.Base;
 
-                    // deviation from the spec to support null-propagation helper
-                    if (baseValue.IsNullOrUndefined()
+                    // deviation from the spec to support null-propagation helper;
+                    // since the unresolvable reference base is a sentinel (not undefined), also
+                    // consult the resolver for unresolvable references so a call to an undefined
+                    // name is routed through it instead of casting the sentinel to an Environment
+                    if ((baseValue.IsNullOrUndefined() || referenceRecord.IsUnresolvableReference)
                         && engine._referenceResolver.TryUnresolvableReference(engine, referenceRecord, out var value))
                     {
                         thisObject = value;


lahma (approved):
Thank you!

PR #2798, opened 2026-07-26
Share the resolved CLR member accessors through the type resolver

lahma:
Resolved CLR member accessors live in `Engine._reflectionAccessors`, so every engine
resolves every member it touches from scratch — and, since resolution is what builds
the `MethodDescriptor` set and the compiled read/write delegates hang off the
resolved accessor, re-does that work too. For the common
fresh-engine-per-operation embedding pattern that is the whole cost, paid on every
operation, for a result that is a pure function of the reflected type and the interop
configuration.

`Options.Interop.TypeResolver` already documents itself as cache-holding state that
"should be shared between engines". Move the accessor cache there, so engines sharing
a resolver share the resolution. `TypeResolver.Default` is the process-wide default,
which is what engines constructed without an explicit resolver already share.

### The cache key merges two independently-motivated designs

This change and #2786 both concluded the old `(Type, PropertyName)` key was too
narrow, for different reasons, and **both components are necessary** — neither
subsumes the other:

- #2786 added `MemberResolutionRequirement`, because resolution is *filtered* by
  whether the caller needs a readable or a writable member. Without it a
  write-filtered resolution answered a later read, so a failed write to a getter-only
  property made every subsequent read of that member return `undefined`. That is a
  correctness fix and is preserved verbatim.
- Sharing across engines needs the interop configuration that steers resolution but
  does *not* live on the resolver, because it lives on `Options.Interop`:
  `AllowGetType`, the three `ObjectWrapperReported*BindingFlags` sets, the extension
  method cache, and whether the type converter is the stock one. Two engines
  differing in any of those would resolve the same member differently.

The reconciled key is `(Type, PropertyName, MemberResolutionRequirement, InteropResolutionProfile)`.

`InteropResolutionProfile` captures the second group. Nothing in it is affine to an
engine or a realm — the flags are values, the extension method lookup holds only
`MethodInfo`s and is itself meant to be shared (captured by reference identity), and
a host-installed `ITypeConverter` is reduced to "is it the stock one" rather than
captured, because those are routinely constructed per engine. It is hand-written
rather than a positional record struct so the hash is computed once, when an engine
captures its profile, instead of on every cache probe.

The engine captures its profile *after* `Options.Apply` runs the configuration
callbacks, since that is where a host-installed `ITypeConverter` is put in place.
Resolution while those callbacks are still running bypasses the cache entirely rather
than risk mislabeling an entry.

### Nothing engine-affine enters the shared cache

`IsShareable` keeps two accessor kinds out:

- `NestedTypeAccessor` holds a `TypeReference`, a `JsValue` owned by the engine that
  created it. Sharing it would hand one engine's object to another and pin that
  engine for the resolver's lifetime. This is the same hazard #2784 fixed for
  `TypeReference`'s own static-member cache; that guard covers `TypeReference`'s
  path, and this one covers `TypeResolver.GetAccessor`, which reaches
  `TryFindMemberAccessor`'s nested-type branch independently.
- `IndexerAccessor` bakes in the index key the engine's `ITypeConverter` produced
  from the member name. The stock converter only ever yields a plain CLR value, but a
  host-installed one may return anything, including something bound to its engine —
  so this one is only excluded when the converter is not the stock one.

Everything else a `ReflectionAccessor` holds is derived from the reflected type
alone, with the engine-affine parts built per call in
`ReflectionAccessor.CreatePropertyDescriptor` as before.

The cache keeps the reflected `Type`s alive for as long as the resolver lives, and
`Default` lives for the process — the same trade-off the other process-wide
reflection caches in the assembly already make. That is now documented on
`TypeResolver`, together with the advice to give a resolver of its own to any engine
that must not outlive the types it touches.

`InteropAccessorCacheSharingTests` proves the sharing with a counting `MemberFilter`
as the engagement probe — it is only consulted while a member is being resolved, so a
zero count on the second engine proves reuse rather than merely the same answer — and
guards the other half of the bargain: each configuration difference above partitions
the cache, and a nested type resolves to its own engine's `TypeReference`.

### Verification

Re-measured after rebasing onto current `main`. Release build clean (only the
pre-existing `MSB3277` in `Jint.Tests.CommonScripts` net472). `Jint.Tests` 4193
net10.0 / 4125 net472 (4183 / 4115 on `main` plus the 10 added here),
`Jint.Tests.PublicInterface` 256 — unchanged from `main`.

### DIFF
--- a/Jint.Tests/Runtime/InteropAccessorCacheSharingTests.cs
+++ b/Jint.Tests/Runtime/InteropAccessorCacheSharingTests.cs
@@ -1,0 +1,277 @@
+#nullable enable
+using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
+using Jint.Runtime.Interop;
+
+namespace Jint.Tests.Runtime;
+
+/// <summary>
+/// The resolved CLR member accessors live on the <see cref="TypeResolver"/>, so every engine configured with
+/// the same resolver resolves each member — and compiles the delegates that read it — exactly once.
+/// <para>
+/// The counting <see cref="TypeResolver.MemberFilter"/> below is the engagement probe: it is only consulted
+/// while a member is being resolved, so a zero count proves the second engine reused the first engine's work
+/// rather than merely arriving at the same answer. The rest of the file guards the other half of the
+/// bargain — that an entry is never served to an engine whose configuration would have resolved it
+/// differently, and that nothing engine-affine gets shared.
+/// </para>
+/// </summary>
+public class InteropAccessorCacheSharingTests
+{
+    #region hosts
+
+    public sealed class Host
+    {
+        public Host(int value) => Value = value;
+
+        public int Value { get; }
+
+        public string Name { get; set; } = "name";
+    }
+
+    public sealed class PrivateMemberHost
+    {
+        public int Visible => 1;
+
+        private int Hidden => 42;
+    }
+
+    public class Outer
+    {
+        public sealed class Inner
+        {
+            public int Value => 5;
+        }
+    }
+
+    public sealed class StringIndexed
+    {
+        public string this[string key] => key + "!";
+    }
+
+    private sealed class CountingResolver
+    {
+        private int _memberFilterCalls;
+
+        public TypeResolver Resolver { get; }
+
+        public CountingResolver()
+        {
+            Resolver = new TypeResolver
+            {
+                MemberFilter = _ =>
+                {
+                    _memberFilterCalls++;
+                    return true;
+                },
+            };
+        }
+
+        public int Reset()
+        {
+            var calls = _memberFilterCalls;
+            _memberFilterCalls = 0;
+            return calls;
+        }
+    }
+
+    private static Engine CreateEngine(TypeResolver resolver, object host, Action<Options>? configure = null)
+    {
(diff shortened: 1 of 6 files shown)


lahma:
## Measured

Idle machine, default BenchmarkDotNet job, process table verified empty immediately before each run. `InteropColdTypeBenchmark` — fresh engine per operation, single property read, which is the lane whose doc calls out "the per-engine reflection-accessor cache is what gets rebuilt here".

| | Mean | Allocated |
|---|---|---|
| base `285cc7fd1` | 4.947 µs ± 0.030 | 15.46 KB |
| this PR | **3.466 µs** ± 0.025 | **12.77 KB** |
| | **−29.9%** | **−17.4%** |

Well outside both the error bars and the ~4% cross-process floor I established from an untouched control row in a sibling run.

Worth noting this understates the real case: the lane touches one type with three members, whereas the change's benefit scales with how much reflection resolution a host repeats across engines. A host wrapping a large object model gets more. It was also the one PR in this batch with a plausible way to *lose* — it trades a per-engine `Dictionary` for a process-wide `ConcurrentDictionary` with a wider key, so a long-lived-engine host pays slightly more lookup for a sharing benefit it never collects. The cold-start win is large enough that the trade clearly pays.

PR #2804, opened 2026-07-26
Let host objects declare their property-access semantics

lahma:
An `ObjectInstance` subclass defined outside Jint can only reach the `protected ObjectInstance(Engine)` constructor, so it carries none of the internal type flags the read path keys off — neither `PlainObject` nor `ExoticGet`. Two consequences follow, one of them a live correctness hole.

### The correctness hole

`JintMemberExpression.ReadAfterOwnMissUncached` treats "lacks `InternalTypes.ExoticGet`" as "has ordinary `[[Get]]`". A host whose `Get` computes a value for names it owns no descriptor for is therefore silently bypassed for **every name that happens to resolve on its prototype**: the read is answered from the prototype's descriptor and the host's `Get` is never called. `Object.prototype` alone is enough to trigger it, and the host has no way to opt out, because the flag that would say so is internal.

```js
Object.prototype.onPrototype = 42;
host.onPrototype;   // 42 — the host's Get never runs
```

`HostObjectSemanticsTests.AHostThatOverridesGetIsHonouredWithoutDeclaringAnything` pins the fix: after this PR the host's computed value wins, and it wins without the host declaring anything, because overriding `Get` is itself the declaration.

### The cost

A read that hits an own property probes `GetOwnProperty` twice — once to prove the own property is not shadowed, discarding the descriptor, then again inside `Get`, whose fast path needs `PlainObject`. For a host that materializes descriptors lazily that is two virtual calls and two allocations per read, forever.

### The change

`PropertyAccessSemantics`, which the engine **derives per type** and never asks for:

* a subclass that **overrides `Get`** is `Exotic` — every read is routed through `Get`, which closes the correctness hole above;
* a subclass that **does not** is `Ordinary` — the read path resolves an own hit from a **single** probe.

Both directions are safe by construction. Overriding `Get` is the only way to have non-ordinary `Get` semantics, so a type that never overrides it cannot deviate; and a type that does override it is the only one that can. Overriding `GetOwnProperty` alone still lands on `Ordinary`, which is right — the single probe the interpreter then does is precisely the one `base.Get` would have done.

`protected ObjectInstance.SetPropertyAccessSemantics` survives for the two shapes the rule cannot see, and both have tests:

* a type that overrides `Get` and is nevertheless ordinary (tracing, metrics, a special case that still agrees with `GetOwnProperty`) declares `Ordinary` and gets the short lane back;
* a type that does **not** override `Get` yet still is not ordinary (a `GetOwnProperty` that is not stable across calls for the same name) declares `Exotic`.

`PropertyAccessSemantics.Unspecified` is gone; the enum is two-valued.

`OrdinaryGet` is deliberately not `PlainObject`. `PlainObject` is a *storage* claim — the engine reads `_properties` / the shape directly, bypassing virtual `GetOwnProperty` — which an object projecting lazily out of native state cannot honour. `OrdinaryGet` is only a *semantics* claim: `Get(p, receiver)` returns exactly `UnwrapJsValue(GetOwnProperty(p), receiver)` for an existing own property and otherwise walks the prototype chain.

**The lane order inside `ReadFromNonPlainReceiver` is load-bearing.** It consults the prototype-method inline cache *first* and only then probes the receiver. Putting the own probe first is a 10× regression on warm prototype reads: ten iterations of `host.protoMethod()` through one member node go from one probe to ten, because every warm hit re-probes a host it does not need to. `AWarmPrototypeReadDoesNotReprobeTheHost` is the guard.

Jint cannot verify the `Ordinary` claim cheaply at runtime, so a `[Conditional("DEBUG")]` check recomputes each such read through `Get` and compares. It is free in Release, and it turns a run of a host's own integration suite against a Debug build of Jint into a checker. It matters more under the derived model than it did under the opt-in one, since types now take the ordinary lane without a human having declared anything. The recomputation is skipped whenever it could be observable — an accessor, a custom-valued descriptor, or an exotic holder anywhere on the chain — so enabling it never changes what a script sees.

### Why not simply flip the default to `Ordinary`

Because it fails silently, and the failure is data loss rather than an exception.

There are shipping third-party hosts whose `Get` override returns a **lazily materialised** value for certain names: the backing property is not stored until first read, so `GetOwnProperty(p)` answers `Undefined` while `Get(p)` answers the parsed value. Under an `Ordinary` default the interpreter would resolve those reads from the descriptor and start returning `undefined` — no compile error, no exception. Their tests pass on the old version, they bump, and values quietly disappear. A release note does not help anyone in that position.

Deriving avoids it entirely, because overriding `Get` is exactly the signal that semantics may deviate.

This is still a behaviour change for `Get`-overriding hosts, but a **corrective** one: they move from today's state, where the prototype branch bypasses their `Get` for prototype-resident names, to `Exotic`, where it no longer does. That belongs in the release notes as a fix rather than as a break.

### Cost, and AOT

The probe is one `Type.GetMethod` per **type**, cached in a static `ConcurrentDictionary<Type, InternalTypes>` and resolved in the constructor, so nothing is added to the read path. `PropertyAccessSemanticsDerivationTests` pins it: 50 instances of a type cause one probe, and a subclass is resolved on its own rather than inheriting its base's entry. Jint's own hot types that reach the public constructor — `ObjectWrapper` (per interop wrap) and `IteratorResult` (per iterator step) — state their flags through the internal constructor instead, so neither pays even a cached lookup per instance. Their stated value is the one the derivation would have reached anyway.

Anything the probe cannot answer resolves to `Exotic`. That is what makes the trim/AOT story sound: routing every read through `Get` is always observably correct, so an environment that has stripped reflection metadata loses the optimization and never the semantics. The probe carries an `[UnconditionalSuppressMessage]` for `IL2070` so the warning does not surface in a downstream `PublishTrimmed` / `PublishAot` — Jint's own csproj already blanket-`NoWarn`s the IL2xxx family, so the Roslyn analyzer would not have flagged it either way. Under default trimming and NativeAOT the answer is correct: the trimmer must keep an override's body because the engine calls `Get` virtually, and metadata for kept members is preserved unless reflection-free mode is explicitly opted into. A host in that configuration that wants certainty declares `Exotic` explicitly.

### Probe-count guard, and the benchmark lanes

#2800 (now on `main`) added `HostObjectProbeCountTests` recording the two-probes-per-read cost with a note that the numbers should drop once this landed. They have, so this PR updates them rather than leaving the two inconsistent: exact counts per derived outcome, and the class comment records what changed. The assertions stay exact counts — a range would defeat the point of the test.

One count deliberately did **not** move, and the test records it so the difference is not mistaken for an oversight: an own-property *miss* still costs the ordinary host two probes, because the probe that establishes the miss cannot also produce a value, so the read still ends in a `Get` that re-probes the receiver. The base of a member call also still costs one, because it resolves straight through `ObjectInstance.Get` rather than through the member-read lane — which is precisely what showed the extra probe on the value lane was removable rather than intrinsic.

`HostObjectAccessBenchmark`'s receiver rows follow the model. `LazyHost` is now a host that overrides only `GetOwnProperty` and declares nothing, and `LazyHostExotic` is the identical projection behind a pass-through `Get` override — so the delta between them is the cost of being derived exotic with every other variable held fixed. Both sit alongside the `FixedLayout` row that has since landed on `main` rather than replacing it: they answer different questions, and every receiver kind now runs. No benchmark was run for this change; it does not touch the per-read mechanism the existing measurements cover, and each parameter combination was executed once through a temporary reflection driver (deleted, not part of this diff), in Release and under a Debug build where the verifier ran over the benchmark's own hosts and accepted them.

### Verification

Re-measured on current `main` (`a21cc6812`). Release build clean — exactly one warning, the pre-existing `MSB3277` in `Jint.Tests.CommonScripts` net472, confirmed on clean `main` first. `Jint.Tests` 4271 net10.0 / 4203 net472 (4268 / 4200 on `main` plus 3 added here), `Jint.Tests.PublicInterface` 312 (291 on `main` plus 21 added here), `Jint.Tests.Test262` 99441 passed / 0 failed / 157 skipped — identical to the baseline measured on clean `main` at the same commit. This PR touches the interpreter's member-read lanes, so Test262 is the run that matters.

### DIFF
--- a/Jint.Benchmark/HostObjectAccessBenchmark.cs
+++ b/Jint.Benchmark/HostObjectAccessBenchmark.cs
@@ -29,19 +29,19 @@ namespace Jint.Benchmark;
 /// <list type="bullet">
 /// <item><description>
 /// <see cref="ReceiverKind"/> — <see cref="HostReceiverKind.PlainObject"/> is the script-object
-/// floor. <see cref="HostReceiverKind.LazyHost"/> is the cost an embedder pays when it declares
-/// nothing: every own-property read goes through the virtual
-/// <see cref="ObjectInstance.GetOwnProperty"/> (the fast shape/dictionary lanes in
-/// <see cref="ObjectInstance.Get(JsValue, JsValue)"/> are not reachable for a
-/// non-<c>PlainObject</c> receiver) <b>twice</b>, allocating a fresh
-/// <see cref="PropertyDescriptor"/> each time and discarding one of them. Expect the LazyHost row
-/// to be several times the PlainObject row in both time and <c>Allocated</c>, with the allocation
-/// delta ≈ (descriptor size × own-property reads) — descriptor churn is half the story, so always
-/// read the <c>Allocated</c> column next to <c>Mean</c> here.
-/// <see cref="HostReceiverKind.LazyHostOrdinary"/> is the same host declaring
-/// <see cref="PropertyAccessSemantics.Ordinary"/>, which halves the probes on an own-property hit;
-/// it still allocates the descriptor the surviving probe returns, so the gap it closes against
-/// PlainObject should be visible on <c>Mean</c> and only partial on <c>Allocated</c>.
+/// floor. <see cref="HostReceiverKind.LazyHost"/> is what an embedder pays for a custom subclass
+/// that overrides <see cref="ObjectInstance.GetOwnProperty"/> and nothing else: the engine derives
+/// <see cref="PropertyAccessSemantics.Ordinary"/> for it, so an own-property read costs one virtual
+/// probe (the fast shape/dictionary lanes in <see cref="ObjectInstance.Get(JsValue, JsValue)"/> are
+/// not reachable for a non-<c>PlainObject</c> receiver) and the fresh
+/// <see cref="PropertyDescriptor"/> that probe returns. Descriptor churn is half the story, so
+/// always read the <c>Allocated</c> column next to <c>Mean</c> here — the allocation delta against
+/// PlainObject is ≈ (descriptor size × own-property reads).
+/// <see cref="HostReceiverKind.LazyHostExotic"/> is the identical projection behind an override of
+/// <c>Get</c> that only delegates to <c>base.Get</c>. Nothing about the values changes; the engine
+/// derives <see cref="PropertyAccessSemantics.Exotic"/> from the override alone and routes every
+/// read through the full <c>Get</c> pipeline. The LazyHost → LazyHostExotic delta is therefore the
+/// price of being exotic, paid by a host that genuinely needs to observe every read.
 /// <see cref="HostReceiverKind.FixedLayout"/> is the same host records built through
 /// <see cref="JsObject.Create(Engine, JsObjectLayout, ReadOnlySpan{JsValue})"/>: a declared layout
 /// resolves to one interned hidden class, so every item in the batch shares it and reads take the
@@ -115,7 +115,7 @@ public class HostObjectAccessBenchmark
     [Params(
         HostReceiverKind.PlainObject,
         HostReceiverKind.LazyHost,
-        HostReceiverKind.LazyHostOrdinary,
+        HostReceiverKind.LazyHostExotic,
         HostReceiverKind.FixedLayout)]
     public HostReceiverKind ReceiverKind { get; set; }
 
@@ -132,7 +132,7 @@ public class HostObjectAccessBenchmark
         {
             case HostReceiverKind.PlainObject:
             case HostReceiverKind.LazyHost:
-            case HostReceiverKind.LazyHostOrdinary:
+            case HostReceiverKind.LazyHostExotic:
             case HostReceiverKind.FixedLayout:
                 break;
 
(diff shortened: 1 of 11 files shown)


lahma:
## Measured

Idle machine, default BenchmarkDotNet job, process table verified empty immediately before each run. `HostObjectAccessBenchmark`, `ResolverKind=None`, `StatementLimit=False`. Baseline is `285cc7fd1`; the branch measured also contains #2808, so all three receiver kinds come from one binary.

| Receiver | Mean | Allocated |
|---|---|---|
| `PlainObject` (reference) | 53.16 µs | 26.93 KB |
| `LazyHost` on base | 71.59 µs | 67.55 KB |
| `LazyHost` on branch — **not opted in** | 70.24 µs | **67.55 KB** |
| `LazyHostOrdinary` — this PR | 69.27 µs | **48.80 KB** |

**Two things this settles.**

*No regression for hosts that don't opt in.* The `Ordinary` lane adds a flag test to `ReadFromNonPlainReceiver`, which every `ObjectWrapper` read reaches. `LazyHost` allocation is byte-identical before and after and the time delta is negative, so that cost is not observable.

*The win here is allocation, not latency.* −28% allocation is solid and matches the predicted probe drop exactly. But the −3.2% time delta is **at the measurement floor** — the untouched `PlainObject` control row swung ±5% across processes in the same comparison, so time deltas below ~4% aren't resolvable in this setup. I would not claim a speedup from this PR on its own.

The allocation ladder confirms the mechanism the description gives, including a nice tell: on base, `LazyHost` allocates *more* with no resolver (67.55 KB) than with one (48.80 KB), because the resolver-disabled route takes the double-probe path while the slow route probes once. The bug is visible in the allocation column.

Stacked with #2808 the combined effect is −21% time and −60% allocation against base `LazyHost` — see that PR.

PR #2808, opened 2026-07-26
Let a host answer an own read without materializing a descriptor

lahma:
Rebased onto current `main`, which now contains #2823 and #2824. #2823 was a **semantic** conflict rather than a textual one, and resolving it changed the design — the section after next explains what and why. Single commit.

Even resolving an own read from a single probe, a host that projects values out of native storage still allocates one `PropertyDescriptor` per read purely for `UnwrapJsValue` to throw away — including one per element of an indexed read, where the loop body never touches the interpreter's member lane at all.

`protected internal virtual bool ObjectInstance.TryGetOwnPropertyValue(JsValue property, JsValue receiver, out JsValue value)` lets such a host answer the own-property question and hand the value over in one call, with no descriptor in between.

### The contract, and why it is the stronger one

`false` means **there is no own property of that name**. It is authoritative: the engine trusts it, does not re-probe, and continues the read up the prototype chain. `true` means the value is exactly what `UnwrapJsValue(GetOwnProperty(property), receiver)` would have produced. In both directions the obligation is the one `ProbeOwnProperty` already carries — agree with `GetOwnProperty` for the same key at the same instant — so this is not a new kind of promise for a host in this area, it is the same one.

The weaker alternative was on the table and is what the pre-rebase version of this PR shipped: `false` meaning "I cannot answer, ask `GetOwnProperty`". That keeps the hook a pure optimization with no way to get it wrong, and it is the reason the choice has to be argued rather than assumed. It was rejected because of what #2823 landed in the meantime.

`ReadFromNonPlainReceiver`'s ordinary lane is exactly the code #2823 fixed. A host's own-property set lives outside the engine, so nothing moves `_propertiesVersion` when it changes and the version cannot stand in for "this receiver still has no own property of this name". #2823's fix re-establishes the own miss with a `GetOwnProperty` probe placed **before** the prototype-method cache is consulted, on every single read. That ordering is load-bearing — it is the entire reason `CanCacheAgainstVersions` may exempt a host *receiver* at all.

A hook whose `false` meant "ask `GetOwnProperty`" cannot discharge that. The probe would still have to run on every prototype-resolving read, so such a host would pay a hook call **plus** a probe where #2823 alone pays one probe — a regression on the warm-prototype path, bought with a win on own hits. An authoritative `false` re-establishes the *same* miss on its own, so #2823's ordering is preserved with no probe at all and every outcome improves instead.

The failure mode the stronger contract introduces is real: a host returning `false` because a value was awkward to produce would make the read resolve on the prototype, or evaluate to `undefined`, for a property that exists. Three things answer it:

* **There is no reason to write that `false`.** The base implementation *is* the descriptor-driven answer (`GetOwnProperty` plus the unwrap), so an override that cannot serve a key calls `base.TryGetOwnPropertyValue(...)` and is still exactly right, at exactly the cost of not having overridden anything. The XML doc states this in the imperative, names the failure mode, and points at `base` as the remedy; `HostObjectSemanticsTests` has a host that does it.
* **A Debug build catches it.** The `[Conditional("DEBUG")]` verifier checks *both* directions against `GetOwnProperty` on every read, and its `false`-direction message says what the return value means and what to call instead.
* **`property` is the same key `GetOwnProperty` receives**, unmodified and not through `ToPropertyKey`. So the contract has no clause about key forms at all: whatever key you would have had to handle in `GetOwnProperty`, you answer for here, and `base` handles the rest.

### The gate is derived, not declared

The lane is keyed on `InternalTypes.OwnValueHook`, derived from the runtime type by the same cached reflection that already derives `Ordinary`/`Exotic` in #2804, and orthogonal to it. A type that does not override the hook is therefore never asked: it pays one test on an already-loaded `_type`, not a virtual call that would only turn around and probe `GetOwnProperty`.

That is not a micro-optimization, it is what makes wiring `ObjectInstance.Get` acceptable. `Get`'s slow path is taken by every non-plain object in the engine — arrays, wrappers, functions, dates, every built-in — and an unconditional hook call there would have taxed all of them for a member none of them override. `Get` is wired because it is what covers the reads the interpreter's member lane never sees: computed keys (`list[i]`), `Reflect.get`, `super`, built-ins, and the base of a member call. Wiring only the member lane would have left the most allocation-heavy shape, the indexed read, untouched.

It also means the base implementation is never called by the engine at all — only by an override that defers to it — so its cost to everything else is one enum bit.

### Probe counts

`GetOwnProperty` calls per read, **Release (Debug)**. Both configurations are pinned, since the suite runs both.

| | ordinary | ordinary + value hook | exotic |
| --- | --- | --- | --- |
| own hit | 1 (3) | **0** (3) | 1 (1) |
| prototype hit, warm or cold | 1 (3) | **0** (3) | 1 (1) |
| absent everywhere | 2 (4) | **0** (4) | 1 (1) |
| member-call base (`host.prop.toUpperCase()`) | 1 (1) | **0** (1) | 1 (1) |

The ordinary and exotic columns are unchanged from #2823 — this PR moves nothing for a host that does not override the hook, which `AHostThatDoesNotOverrideTheValueHookIsNeverAsked` pins directly. The Debug numbers are the verifiers' own probes and are not costs; the ordinary and hook columns agree there because a Debug read recomputes through `Get`, which itself takes whichever lane the host is on.

The member-call base is the interesting one: it resolves through `ObjectInstance.Get` rather than the member-read lane, so it was the count #2804 could *not* move, and it moving here is what shows the hook is not member-read-only. `HostObjectProbeCountTests` gains the third column and a prototype-hit row it did not have.

### Regression coverage for the ordering

`HostObjectPropertySetChangeTests` gains the receiver-side cases for the hook lane — a host whose projection gains (or loses) a name that shadows a prototype member, read repeatedly through **one** AST node, asserting the shadow takes effect on the very next read. These are a different branch from the ones #2823 added, not a copy: there the own miss is re-established by a probe, here by the hook's `false`.

They were checked to bite. Reinstating the pre-#2823 ordering inside the hook branch (consult the prototype cache first) fails both "appearing" cases with `from-prototype,from-prototype,from-prototype`, on both target frameworks; the change was reverted afterwards and is not in this diff.

### Documentation

`AGENTS.md`'s "further levers" list under the subclassing cliff gains the hook, with the contract, the `base` remedy and the failure mode. The probe-count paragraph above it — the one #2824 has just corrected for #2823 — is **untouched**: the ordinary host's numbers do not move here, so there was nothing to correct, and the new bullet refers to that paragraph rather than restating it. `CLAUDE.md` is not touched; it is one import line.

### Benchmark

`HostObjectAccessBenchmark` gains the `LazyHostValueHook` receiver lane its `TODO` was waiting for, as one more subclass adding one more override to the identical projection code — so the `LazyHost` → `LazyHostValueHook` delta is the descriptor and nothing else, the way `LazyHost` → `LazyHostExotic` is the exoticness and nothing else. Every receiver kind now runs and no `TODO` is left in that switch.

**No benchmark was run.** The file is resolved to compile and to model the merged lanes; measuring it is not part of this change.

One note for anyone copying that host: it is spelled `protected internal override` there because `Jint.Benchmark` is a friend assembly. Outside the assembly the member is seen as `protected`, so an integrator writes `protected override` — which is what the hosts in `Jint.Tests.PublicInterface` do. Same member either way; the keyword is the only difference.

### Verification

Baselines measured on clean `main` (`5a925bfb1`) in a separate worktree first, then the same set on this branch.

| | clean `main` | this branch |
| --- | --- | --- |
| Release build | 0 errors, 1 warning | 0 errors, 1 warning |
| `Jint.Tests` Release net10.0 | 4290 passed / 4 skipped | 4290 passed / 4 skipped |
| `Jint.Tests` Release net472 | 4222 passed / 4 skipped | 4222 passed / 4 skipped |
| `Jint.Tests.PublicInterface` Release net10.0 / net472 | 390 / 390 | 405 / 405 |
| `Jint.Tests.PublicInterface` **Debug** net10.0 / net472 | 390 / 390 | 405 / 405 |
| `Jint.Tests.Test262` Release | 99441 passed / 0 failed / 157 skipped | 99441 passed / 0 failed / 157 skipped |

The single warning is the pre-existing `MSB3277` in `Jint.Tests.CommonScripts` net472, confirmed on clean `main` before dismissing it. The Debug row matters more than usual here: the probe-count verifier and the new both-directions hook verifier only exist in Debug, so that run is what proves the merged lane's contract holds rather than merely that its numbers were written down. The `PublicInterface` delta is exactly the tests added here — +7 theory rows in `HostObjectProbeCountTests` (third column plus the new prototype-hit row), +5 facts in `HostObjectSemanticsTests`, +3 in `HostObjectPropertySetChangeTests`.


lahma:
## Measured

Idle machine, default BenchmarkDotNet job, process table verified empty immediately before each run. `HostObjectAccessBenchmark`, `ResolverKind=None`, `StatementLimit=False`. Baseline `285cc7fd1`; this branch contains #2804, so the ladder comes from one binary.

| Receiver | Mean | Allocated | descriptors/read |
|---|---|---|---|
| `PlainObject` (reference) | 53.16 µs | 26.93 KB | — |
| `LazyHost` on base | 71.59 µs | 67.55 KB | 2 |
| `LazyHostOrdinary` (#2804) | 69.27 µs | 48.80 KB | 1 |
| `LazyHostValueHook` — this PR | **56.35 µs** | **26.93 KB** | 0 |

**−21% time and −60% allocation** against base `LazyHost`.

Two things worth calling out. The allocation column lands on exactly the predicted 2 → 1 → 0 descriptors per read, so the mechanism is visible rather than inferred. And `LazyHostValueHook` at 56.35 µs is within noise of `PlainObject` at 53.16 — a host object that projects its values lazily now costs about what a plain JavaScript object costs, which is the point of the extension point.

Measurement caveat for honesty: the untouched `PlainObject` control row swung ±5% across processes in this comparison, so I treat ~4% as the floor for time deltas here. −21% is well clear of it, and the allocation figures are deterministic counts rather than statistics.

PR #2810, opened 2026-07-26
Let host code build objects directly in the hidden-class representation

lahma:
Host code that projects data into JavaScript today reaches for `FastSetDataProperty` — the API whose *name* says "fast". It routes to `SetProperty(Key, PropertyDescriptor)`, which **deopts the object to dictionary mode** and forfeits the shape inline cache: a `PropertyDescriptor` allocation per property, a property dictionary per object, and a polymorphic `o.name` call site for the script reading the batch. Object literals and `JSON.parse` already build straight into the hidden-class representation; nothing let a host do the same.

This is not hypothetical: of the embedding integrations I looked at while investigating, four build objects exactly that way — one object per item in a result set, so the dictionary tax is paid per row and the reading script's call sites never stabilise.

This adds two entry points that build directly into the shaped representation. They are one PR because they share all of the machinery below (the per-prototype layout memo, the transition budget, the key classifier).

### `JsObject.Create(engine, layout, values)` — fixed, known layout

```csharp
private static readonly JsObjectLayout PointLayout = new("x", "y", "label");

JsObject ToJs(Engine engine, Point p) => JsObject.Create(
    engine, PointLayout,
    [JsNumber.Create(p.X), JsNumber.Create(p.Y), new JsString(p.Label)]);
```

`JsObjectLayout` is immutable, engine-agnostic and safe to share across engines and threads (typically a `static readonly` field). It validates **eagerly, once**, so creation itself has nothing left to check: names non-null/non-empty, distinct, not integer-index-like, and at most `Shape.MaxShapeProperties` (64) of them. The result is a completely ordinary object — `Object.prototype` prototype, configurable/enumerable/writable properties, own-key order identical to the equivalent literal.

### `JsObject.CreateFromEntries(engine, entries)` — runtime key set

For host data whose keys are only known at runtime. Entries go through the incremental hidden-class path, so repeated calls presenting the same key sequence share one interned hidden class. Later entries with a name already present overwrite in place, keeping the first occurrence's position — the rule object literals and `Object.fromEntries` follow.

**Both a `ReadOnlySpan<KeyValuePair<string, JsValue>>` and an `IEnumerable<…>` overload, deliberately.** A `Dictionary<string, object>` — the shape host data usually arrives in — has no span view, and materializing a temp array just to call the span overload would cost precisely the allocation the shaped representation is there to save.

### Three findings worth calling out

**1. The layout memo nests under the prototype, and lifetime then falls out for free.**
`Engine._emptyShapes` (`CWT<ObjectInstance, Shape>`) becomes `_shapeRoots` (`CWT<ObjectInstance, ShapeRoot>`), where `ShapeRoot` holds the root shape *plus* a `CWT<JsObjectLayout, Shape>` memo. Because a `Shape` references only its parent chain and its key strings — never a prototype, an engine or a layout — the ownership is unambiguous: **prototype dropped ⇒ its transition tree and its memo go; layout dropped ⇒ its entry goes.** No new lifetime rule to reason about, and no strong reference from a process-lifetime layout back into an engine.

**2. `Shape.MaxFanout` is enforced in `JsObject.TryShapeAdd`, not in `Shape.Add`** — so the wholesale layout path, which walks `Shape.Add` directly the way object literals do, bypasses it entirely. That matters because a transition tree is pinned by its prototype, and for `Object.prototype` that means the engine's lifetime; a host feeding wildly varying key sets (document fields, user-defined columns) could otherwise grow it without bound.

A per-engine `HostShapeTransitionBudget` bounds it: 16384 **newly interned** transitions. Same mechanism `JsonParser.ShapeTransitionBudget` already uses against an adversarial cold parse, at a different scope — the parser's 1024 is per `Parse` call, this one is over the engine's lifetime, because a host can call the factory as many times as it likes. Reused transitions — the entire point of the API — cost nothing, so the steady state never touches the budget. Once exhausted, host construction keeps working and simply produces ordinary dictionary-mode objects. This is currently the only bound on the layout path, and the comment on the constant says so.

**3. `setPrototypeOf` does not deopt, and that is correct.**
A `Shape` encodes name→slot and never the prototype, and the shape inline cache only serves *own-slot* reads. So a prototype change leaves the object shaped and correct; it costs nothing but a missed sharing opportunity with objects built under the new prototype. The tests assert **correctness** after `setPrototypeOf` — inherited reads, shadowing, `Object.keys`, `for...in` order — not a deopt.

### Degradation is automatic and total

Adding a property, deleting one, defining an accessor or a non-default attribute, freezing, sealing — all convert to the ordinary dictionary representation exactly as they do for an object literal. `CreateFromEntries` additionally falls back **mid-build, preserving everything added so far**, on an integer-index-like key, on exceeding the hidden-class property limit, or on the fan-out guard refusing the add. Every path degrades the representation, never the behaviour.

### Also in this PR

#2802 has already documented what the `Fast*` setters skip, including the dictionary-mode deopt, so this only **adds the forward pointer** those docs were missing: the class summary and `FastSetDataProperty` now name `JsObject.Create` / `JsObject.CreateFromEntries` as what to reach for when building a *new* object out of host data, so the next reader does not pick `FastSetDataProperty` for its name. Nothing #2802 wrote is rephrased or removed.

### The test that matters most

`HostObjectFactoryShapeTests.LayoutShapeIsTheSameShapeAnEquivalentObjectLiteralBuilds`: a `Create` result, a `CreateFromEntries` result, an object literal and a spread all reference **one and the same `Shape` instance**. That is the proof that the new entry points join the *existing* per-prototype transition tree rather than growing a parallel one — and it is what keeps a script's `o.name` call site monomorphic across objects from mixed sources.

Interning is not observable through the public API, so those assertions live in `Jint.Tests` (which sees internals); everything publicly observable is in `Jint.Tests.PublicInterface/FixedLayoutObjectTests.cs`.

### DIFF
--- a/Jint.Tests.PublicInterface/FixedLayoutObjectTests.cs
+++ b/Jint.Tests.PublicInterface/FixedLayoutObjectTests.cs
@@ -1,0 +1,573 @@
+using System.Collections.Generic;
+using Jint.Native;
+
+namespace Jint.Tests.PublicInterface;
+
+/// <summary>
+/// The host-facing object factories: <see cref="JsObjectLayout"/> + <see cref="JsObject.Create"/> for a
+/// fixed, known layout, and <see cref="JsObject.CreateFromEntries(Engine, System.ReadOnlySpan{KeyValuePair{string, JsValue}})"/>
+/// for a runtime key set. Everything asserted here is observable through the public API only; the
+/// hidden-class interning these build on is pinned separately (Jint.Tests, which can see internals).
+/// </summary>
+public class FixedLayoutObjectTests
+{
+    private static readonly JsObjectLayout Layout = new("id", "name", "active");
+
+    private static JsObject CreateSample(Engine engine) => JsObject.Create(
+        engine,
+        Layout,
+        [JsNumber.Create(1), new JsString("jint"), JsBoolean.True]);
+
+    private static Engine EngineWithSample(out JsObject obj)
+    {
+        var engine = new Engine();
+        obj = CreateSample(engine);
+        engine.SetValue("o", obj);
+        return engine;
+    }
+
+    // ---- layout validation ----
+
+    [Fact]
+    public void LayoutRejectsNullNameCollection()
+    {
+        Invoking(() => new JsObjectLayout((string[]) null)).Should().Throw<ArgumentNullException>();
+        Invoking(() => new JsObjectLayout((IReadOnlyList<string>) null)).Should().Throw<ArgumentNullException>();
+    }
+
+    [Fact]
+    public void LayoutRejectsDuplicateNames()
+    {
+        Invoking(() => new JsObjectLayout("a", "b", "a"))
+            .Should().Throw<ArgumentException>()
+            .WithMessage("*Duplicate property name 'a' at index 2*");
+    }
+
+    [Fact]
+    public void LayoutRejectsNullOrEmptyNames()
+    {
+        Invoking(() => new JsObjectLayout("a", null)).Should().Throw<ArgumentException>().WithMessage("*index 1*");
+        Invoking(() => new JsObjectLayout("a", "")).Should().Throw<ArgumentException>().WithMessage("*index 1*");
+    }
+
+    [Theory]
+    [InlineData("0")]
+    [InlineData("42")]
+    [InlineData("1x")]
+    [InlineData("0abc")]
+    public void LayoutRejectsIntegerIndexLikeNames(string name)
+    {
+        // Such a key must enumerate in ascending numeric order ahead of the string keys, which a fixed
+        // insertion-ordered layout cannot express.
+        Invoking(() => new JsObjectLayout("a", name))
+            .Should().Throw<ArgumentException>()
+            .WithMessage("*starts with a digit*");
+    }
+
+    [Fact]
+    public void LayoutAcceptsUpToSixtyFourNamesAndRejectsMore()
+    {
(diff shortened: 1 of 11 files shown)


lahma:
## Measured

Idle machine, default BenchmarkDotNet job, process table verified empty immediately before each run. `HostObjectAccessBenchmark`, `ResolverKind=None`, `StatementLimit=False`, measured on #2814's branch (which carries this PR's commit plus the benchmark row that exercises it).

| Receiver | Mean | Allocated |
|---|---|---|
| `PlainObject` (reference) | 53.15 µs | 26.94 KB |
| `LazyHost` | 73.62 µs | 67.56 KB |
| `FixedLayout` — this PR | **53.44 µs** | **26.94 KB** |

**−27% time and −60% allocation** against `LazyHost`, landing exactly on plain-object cost for both.

**The shaped lane provably engaged.** `JsObject.Create` falls back to dictionary mode when it cannot shape, and that fallback is invisible from script — so a green benchmark row proves nothing on its own. Allocation of 26.94 KB, identical to `PlainObject` and less than half of `LazyHost`'s 67.56 KB, is only reachable through the slot representation; a dictionary-mode fallback would have landed near 67 KB. Combined with the interning assertions already in the tests, the row is measuring what it claims to.

Measurement caveat: the untouched `PlainObject` control row swung ±5% across processes elsewhere in this batch, so ~4% is the floor for time deltas. −27% is well clear, and allocation figures are deterministic counts rather than statistics.

PR #2853, opened 2026-07-28
Fix calling and instanceof on bound functions whose target is itself bound

HermanusMuellerEU:
## Summary

Calling (or `instanceof` against) a bound function whose target is itself
bound threw `TypeError: Bind must be called on a function`.

## Details

`Function.prototype.bind` accepts any callable as its target, so binding an
already-bound function succeeds — but calling the result throws:

​```js
function f(a) { return this.x + ':' + a; }
const twice = f.bind({ x: 1 }).bind(null, 'arg');
twice(); // TypeError: Bind must be called on a function (Node: "1:arg")
​```

`x instanceof f.bind().bind()` fails the same way. Reproduces on every release
tested (4.9.3 → 4.14.0) and current `main`. We hit it in a real workload via a
common dispatch-map
pattern: `this.method = handlers.method.bind(this)` stored once, then
`this.method.bind(this)` registered into a command map — every dispatch threw.

### Root cause

`BindFunction` is an `ObjectInstance` (not a `Function` subclass). Its
`ICallable.Call` casts `BoundTargetFunction as Function` — for a bound-of-bound
function the target is another `BindFunction`, the cast yields null, and the
guard throws. Per
https://tc39.es/ecma262/#sec-bound-function-exotic-objects-call-thisargument-argumentslist
the target only needs to be callable. `OrdinaryHasInstance` had the same
Function-only cast; per InstanceofOperator it must follow the bound-target
chain. (`[[Construct]]` already cast to `IConstructor`, which `BindFunction`
implements, so bound-of-bound construction worked.)

### The fix

- `ICallable.Call`: cast the target to `ICallable` instead of `Function`.
- `OrdinaryHasInstance`: walk nested `BindFunction` targets to the underlying
  function before delegating.

## Linked issue

## Test plan

- [x] Added or updated unit tests in `Jint.Tests` — five tests in
  `Runtime/FunctionTests.cs`: call through a rebound function (first bind wins
  the this-binding), argument accumulation order across three binds, the
  dispatch-map double-bind pattern, construction through a rebound constructor
  (pins the already-working path), and `instanceof` against a rebound
  constructor. Four fail before the fix; all five pass after.
- [ ] For perf changes: n/a — same cast shape (`as ICallable` for `as
  Function`) on the same path.

## Breaking change?

No.

### DIFF
--- a/Jint.Tests/Runtime/FunctionTests.cs
+++ b/Jint.Tests/Runtime/FunctionTests.cs
@@ -86,6 +86,77 @@ public class FunctionTests
     }
 
     [Fact]
+    public void CanCallBoundFunctionWhoseTargetIsItselfBound()
+    {
+        // Per the spec a bound function's [[BoundTargetFunction]] only needs to be callable —
+        // binding an already-bound function must produce a function that still calls through
+        // (the first bind wins the this-binding).
+        var result = _engine.Evaluate("""
+            (function () {
+                function f(a) { return this.x + ':' + a; }
+                const once = f.bind({ x: 1 });
+                const twice = once.bind({ x: 2 }, 'arg');
+                return twice();
+            })()
+            """);
+        result.AsString().Should().Be("1:arg");
+    }
+
+    [Fact]
+    public void RepeatedBindAccumulatesArgumentsInOrder()
+    {
+        var result = _engine.Evaluate("""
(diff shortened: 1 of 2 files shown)


lahma (approved):
Thank you!

PR #2854, opened 2026-07-28
Pass the original receiver to inherited accessors in ObjectInstance.TryGetValue

HermanusMuellerEU:
## Summary

An inherited accessor's getter ran with the prototype as `this` instead of the
original receiver, breaking array destructuring of class instances.

## Details

OrdinaryGet(O, P, Receiver) calls an inherited accessor's [[Get]] with
Receiver — the object the lookup started on
(https://tc39.es/ecma262/#sec-ordinaryget). `ObjectInstance.TryGetValue`
instead recursed as `Prototype?.TryGetValue(property, out value)`, rebinding
`this` to the prototype at every hop, and then invoked the getter with that
prototype as its this value.

Any class whose prototype accessor reads instance state therefore saw an
"empty" object:

​```js
class List {
    constructor(items) { this._items = items; }
    get length() { return this._items.length; }
    [Symbol.iterator]() { return this._items[Symbol.iterator](); }
}
const [a, b] = new List(['a', 'b']);
// TypeError: Cannot read properties of undefined (reading 'length')
// (Node destructures fine)
​```

Array destructuring reaches this through `IsArrayLike`, which probes `length`
before choosing between the array-like fast path and the iterator protocol —
so the getter threw where the probe should simply have answered, and the
destructuring never fell through to the perfectly good `Symbol.iterator`.
`Array.prototype.keys`/`values` gate on the same probe. Reproduces on 4.12.0
and current `main`; we hit it in a real workload the moment any
domain-collection instance (`get length() { return this._arr.length; }`) was
destructured.

### The fix

Thread the receiver through the prototype walk via a private
`TryGetValue(property, receiver, out value)` overload; the public signature is
unchanged, own accessors still receive the object itself, and data-property
lookups are untouched.

Reviewers may want to double-check: `TryGetValue`'s other callers (e.g.
`IsArrayLike`, iteration helpers) all start the walk on the receiver they care
about, so the new overload only changes which `this` an inherited getter sees —
exactly the spec's Receiver.

## Linked issue

## Test plan

- [x] Added or updated unit tests in `Jint.Tests` —
  `Runtime/PrototypeAccessorReceiverTests.cs` (new, 5 tests): destructuring
  with an inherited `length` getter, a grandparent-inherited getter, the
  `Array.prototype.keys` route, an `Object.create` prototype chain, and an
  own-accessor control that must keep passing. Four fail before the fix; all
  five pass after.
- [ ] For perf changes: n/a — one extra argument on an already-recursive
  private path; the own-property fast path is unchanged.

## Breaking change?

No — `TryGetValue`'s public signature is unchanged. Behavior changes only for
inherited accessors, which now see the spec-mandated `this`.

### DIFF
--- a/Jint.Tests/Runtime/PrototypeAccessorReceiverTests.cs
+++ b/Jint.Tests/Runtime/PrototypeAccessorReceiverTests.cs
@@ -1,0 +1,99 @@
+namespace Jint.Tests.Runtime;
+
+/// <summary>
+/// https://tc39.es/ecma262/#sec-ordinaryget - an inherited accessor's getter runs with the
+/// ORIGINAL receiver as its <c>this</c>, not with the prototype object that happens to hold it.
+/// </summary>
+public class PrototypeAccessorReceiverTests
+{
+    private readonly Engine _engine = new();
+
+    [Fact]
+    public void ArrayDestructuringPassesTheReceiverToAnInheritedLengthGetter()
+    {
+        // Array destructuring probes 'length' via IsArrayLike before choosing between the
+        // array-like fast path and the iterator protocol. That probe walks the prototype
+        // chain; running the class's `get length()` with the prototype as `this` made it
+        // read an undefined backing field and throw instead of falling through to the
+        // iterator.
+        var result = _engine.Evaluate("""
+            class List {
+                constructor(items) { this._items = items; }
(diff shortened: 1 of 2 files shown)


lahma (approved):
Thank you!

PR #2856, opened 2026-07-28
Unwrap a JintStatement suspension node so for-await resumes into the right branch

HermanusMuellerEU:
## Summary

`GetSuspensionNode` never unwrapped a `JintStatement`, so it returned null for
every `for await...of` suspension — and an enclosing `if`/`else` re-evaluated its
test on resume, silently taking the other branch.

## Details

`for await...of` records its suspension against the loop STATEMENT: the implicit
await lives in `JintForInForOfStatement` itself, not in a `JintExpression`.
`JintStatement.GetSuspensionNode` unwrapped only `Node` and `JintExpression`, so
both casts failed and it returned null.

Resume-aware statements use that node to work out which branch they were in when
the frame suspended. With it null, `JintIfStatement` falls through to
re-evaluating its test — and when that test reads state the suspended code has
already mutated, it takes the opposite branch. Nothing throws; the caller just
gets a half-built result.

```js
async function* gen() {
    for await (const x of [1, 2, 3]) { yield x; }
}

const state = { built: false, items: [] };

async function build() {
    if (state.built) {
        return state.items;       // must NOT be reached on resume
    } else {
        state.built = true;       // the test above reads this
        for (const _ of ['a']) {
            for await (const v of gen()) { state.items.push(v); }
        }
        return state.items;
    }
}

(await build()).length; // 0 under Jint (Node: 3)
```

We hit this in a real workload (a large async-heavy codebase running under
Jint). An accounting config cache had exactly this guard-then-build shape: it
suspended inside the build loop, resumed into its own "already cached" test —
which its own partial execution had just made true — and returned an empty
environment. Every later config lookup then missed, so the run silently did less
work while still reporting success. That silence is the dangerous part: no
error, no rejection, just wrong results.

Reproduces on 4.12.0, 4.13.0, 4.14.0 and current `main`, so it is long-standing
rather than a regression.

### The fix

Add the missing `JintStatement` case to the unwrap chain, mirroring the existing
`JintExpression` one:

```csharp
return suspendable.LastSuspensionNode as Node
    ?? (suspendable.LastSuspensionNode as JintExpression)?._expression as Node
    ?? (suspendable.LastSuspensionNode as JintStatement)?._statement;
```

`JintStatement._statement` is a `Statement`, which derives from `Node`, so no
cast is needed. Behavior is unchanged wherever the node was already non-null —
this only replaces a null return with the correct node.

## Linked issue

## Test plan

- [x] Added a unit test in `Jint.Tests` —
  `Runtime/AsyncTests.cs::ShouldResumeForAwaitInsideElseWithoutReevaluatingIfTest`,
  the reduced form of the workload failure above. It fails on pristine
  `origin/main` (0 items) and passes with the one-line change.
- [x] Ran the full `Jint.Tests` suite locally — 4545 passed, no new failures
  (identical failure set to unpatched `main`: 5 timezone-dependent
  `DateTests.CanParseDate` cases from this machine's tzdata).
  regressions (identical results to unpatched `main`: 99,439 passed; the 2
  failures are pre-existing `intl402` timezone-canonicalization cases that fail
  identically on unpatched `main` on this machine).
- [ ] For perf changes: n/a — one extra type test, and only on a path that
  already established the frame is resuming.

## Breaking change?

No.

### DIFF
--- a/Jint.Tests/Runtime/AsyncTests.cs
+++ b/Jint.Tests/Runtime/AsyncTests.cs
@@ -47,6 +47,48 @@ public class AsyncTests
     }
 
     [Fact]
+    public void ShouldResumeForAwaitInsideElseWithoutReevaluatingIfTest()
+    {
+        // A `for await...of` records its suspension against the LOOP STATEMENT - its implicit
+        // await lives in JintForInForOfStatement itself, not in a JintExpression.
+        // GetSuspensionNode only unwrapped Node and JintExpression, so it returned null for every
+        // for-await suspension and JintIfStatement forgot which branch it was in, re-running its
+        // test on resume. Here the else-branch sets `built` before suspending, so a re-evaluated
+        // test takes the `if` branch and returns the half-filled array (0 items).
+        var engine = new Engine();
+
+        var result = engine.Evaluate("""
+            (async () => {
+                async function* gen() {
+                    for await (const x of [1, 2, 3]) {
+                        yield x;
+                    }
+                }
+
+                const state = { built: false, items: [] };
+
+                async function build() {
+                    if (state.built) {
+                        return state.items;       // must NOT be reached on resume
+                    } else {
+                        state.built = true;       // the test above reads this
+                        for (const _ of ['a']) {
+                            for await (const v of gen()) {
+                                state.items.push(v);
+                            }
+                        }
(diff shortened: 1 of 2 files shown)


lahma (approved):
Thank you!

PR #2869, opened 2026-07-30
Only bind a module's own lexical declarations at module level

gkurt:
`InitializeEnvironment` walks the `LexicallyScopedDeclarations` of the module body, which for `Module : ModuleItemList` are its top-level declarations only — a `let`, `const` or `class` inside a Block or a `for` head belongs to that statement's own scope. Binding one at module level as well creates a second binding that nothing ever initializes, so it stays in its temporal dead zone for the whole evaluation:

```js
{ let a = 1; }
var a = 2;
a;                       // ReferenceError, must be 2

for (let key in obj) {}
typeof key;              // ReferenceError, must be "undefined"

{ class C {} }
typeof C;                // ReferenceError, must be "undefined"
```

The first is the damaging one. `InitializeEnvironment` binds *and initializes* the module's var names before it creates the lexical bindings, and the lexical pass calls `CreateMutableBinding` unconditionally — so the extra binding replaces an initialized `var` of the same name with an uninitialized one, and every read of that name throws from then on.

The name need not appear twice in the same source file. A bundler concatenating one file's top-level `for (let key in ...)` with another's `for (var key in ...)` produces exactly this, which is where it turned up: running an ESM dev bundle, `css-vendor`'s `for (var key in jsCssMap)` threw because `immer` contributed a top-level `for (let key in objectTraps)` to the same module.

## Cause

`HoistingScope` collected module-level lexical declarations with `parent is null or AstModule`. `parent is AstModule` holds for any direct child of the module, which is what lets an exported declaration through — it sits one level down, inside the export declaration. But it equally matches any other top-level statement that owns a declaration one level down, and a Block and a `for` head both do.

Declarations in an `if` branch, a `switch` case or a function body escaped only because theirs sit two levels down, where the parent is the `IfStatement`, `SwitchCase` or function — which is why the failure looked so selective:

| At module top level | before | after | spec |
|---|---|---|---|
| `{ let a = 1; }` | ReferenceError | `undefined` | `undefined` |
| `{ class C {} }` | ReferenceError | `undefined` | `undefined` |
| `for (let a in x) {}` | ReferenceError | `undefined` | `undefined` |
| `if (1) { let a = 1; }` | `undefined` | `undefined` | `undefined` |
| `switch (1) { case 1: let a = 1; }` | `undefined` | `undefined` | `undefined` |
| `function f() { let a = 1; }` | `undefined` | `undefined` | `undefined` |

The check now also requires that intermediate node to be an export declaration. `ExportAllDeclaration` is deliberately left out: `export * from` binds no name locally, so it has no declaration to collect.

Scripts were never affected — there the parent of a top-level statement is a `Program`, so `parent is null` alone already selected the right set. This only makes modules agree with them.

## Tests

Five in `ModuleTests`. The three `ShouldNotBind…` ones fail without the change; `ShouldBindExportedLexicalDeclarationsAtModuleLevel` (covering `export const` / `export let` / `export class` / `export default class`, the case `parent is AstModule` exists to serve) and `ShouldStillApplyTheTemporalDeadZoneToModuleLevelDeclarations` pass either way and are there as guards.

test262 is green either way — 99484 passed, 0 failed, unchanged before and after — so nothing in it covers a lexical declaration inside a module's top-level block.

`Jint.Tests`: 4663 passed. The two failures on my machine are `ToStringUsesDaylightNameWhenInDst` and `DateToStringMethodsShouldUseCurrentTimeZoneAndCulture`, which assert OS-localised time zone display names and fail here regardless of this change.

### DIFF
--- a/Jint.Tests/Runtime/ModuleTests.cs
+++ b/Jint.Tests/Runtime/ModuleTests.cs
@@ -1193,4 +1193,56 @@ export const count = globals.counter;
 
         log.Should().Equal("mid");
     }
+
+    [Fact]
+    public void ShouldNotBindALexicalDeclarationFromANestedBlockAtModuleLevel()
+    {
+        // A block's `let` is not a module item, so binding it at module level would leave the `var` of the same name uninitialized.
+        _engine.Modules.Add("my-module", "{ let a = 1; } var a = 2; export const value = a;");
+        var ns = _engine.Modules.Import("my-module");
+
+        ns.Get("value").AsInteger().Should().Be(2);
+    }
+
+    [Fact]
+    public void ShouldNotBindAForHeadDeclarationAtModuleLevel()
+    {
+        // The `let` of a for head is scoped to the loop, so nothing named `key` reaches module scope.
+        _engine.Modules.Add("my-module", "for (let key in { x: 1 }) {} export const value = typeof key;");
+        var ns = _engine.Modules.Import("my-module");
+
+        ns.Get("value").AsString().Should().Be("undefined");
+    }
+
+    [Fact]
+    public void ShouldNotBindAClassDeclaredInABlockAtModuleLevel()
+    {
+        _engine.Modules.Add("my-module", "{ class C {} } export const value = typeof C;");
+        var ns = _engine.Modules.Import("my-module");
+
+        ns.Get("value").AsString().Should().Be("undefined");
+    }
+
+    [Fact]
+    public void ShouldBindExportedLexicalDeclarationsAtModuleLevel()
+    {
+        // The counterpart: an exported declaration sits inside the export declaration and is a module item.
+        _engine.Modules.Add("my-module", "export const c = 1; export let l = 2; export class K {} export default class D {}");
+        var ns = _engine.Modules.Import("my-module");
+
+        ns.Get("c").AsInteger().Should().Be(1);
+        ns.Get("l").AsInteger().Should().Be(2);
+        ns.Get("K").IsCallable.Should().BeTrue();
+        ns.Get("default").IsCallable.Should().BeTrue();
+    }
+
+    [Fact]
+    public void ShouldStillApplyTheTemporalDeadZoneToModuleLevelDeclarations()
+    {
+        // A module-level `const` is bound uninitialized until its declaration runs, so reading it early throws even through typeof.
+        _engine.Modules.Add("my-module", "globalThis.probe = typeof x; const x = 1;");
+
+        Invoking(() => _engine.Modules.Import("my-module")).Should().ThrowExactly<JavaScriptException>()
+            .Which.Message.Should().Be("Cannot access 'x' before initialization");
+    }
 }
--- a/Jint.Tests/Runtime/ModuleTests.cs
+++ b/Jint.Tests/Runtime/ModuleTests.cs
@@ -1198,29 +1198,40 @@ export const count = globals.counter;
     public void ShouldNotBindALexicalDeclarationFromANestedBlockAtModuleLevel()
     {
         // A block's `let` is not a module item, so binding it at module level would leave the `var` of the same name uninitialized.
-        _engine.Modules.Add("my-module", "{ let a = 1; } var a = 2; export const value = a;");
+        // The value is read from inside the block as well, because the post-block value alone cannot tell the fixed behaviour
+        // apart from the equally wrong one where the block's initializer lands on the module-level `var`.
+        _engine.Modules.Add("my-module", "let inner; { let a = 1; inner = a; } var a = 2; export const value = a; export { inner };");
         var ns = _engine.Modules.Import("my-module");
 
+        ns.Get("inner").AsInteger().Should().Be(1);
         ns.Get("value").AsInteger().Should().Be(2);
     }
 
     [Fact]
     public void ShouldNotBindAForHeadDeclarationAtModuleLevel()
     {
-        // The `let` of a for head is scoped to the loop, so nothing named `key` reaches module scope.
-        _engine.Modules.Add("my-module", "for (let key in { x: 1 }) {} export const value = typeof key;");
-        var ns = _engine.Modules.Import("my-module");
+        // The `let` of a for head is scoped to the loop, so nothing it binds reaches module scope, and a module-level
+        // `var` of the same name stays writable. Each head shape is a separate branch of the walk, and the two a
+        // bundler actually emits are `for` and `for-of`.
+        _engine.Modules.Add("for-in", "for (let key in { x: 1 }) {} export const value = typeof key;");
+        _engine.Modules.Add("for", "for (let i = 0; i < 1; i++) {} var i = 5; export const value = i;");
+        _engine.Modules.Add("for-of", "for (const o of [1]) {} var o = 6; export const value = o;");
 
-        ns.Get("value").AsString().Should().Be("undefined");
+        _engine.Modules.Import("for-in").Get("value").AsString().Should().Be("undefined");
+        _engine.Modules.Import("for").Get("value").AsInteger().Should().Be(5);
+        _engine.Modules.Import("for-of").Get("value").AsInteger().Should().Be(6);
     }
 
     [Fact]
     public void ShouldNotBindAClassDeclaredInABlockAtModuleLevel()
     {
-        _engine.Modules.Add("my-module", "{ class C {} } export const value = typeof C;");
-        var ns = _engine.Modules.Import("my-module");
+        // A class declaration is a separate branch of the walk from a variable declaration, so it needs both facts: the
+        // name is invisible outside its block, and it does not leave a module-level `var` of the same name uninitialized.
+        _engine.Modules.Add("invisible", "{ class C {} } export const value = typeof C;");
+        _engine.Modules.Add("clobbered", "{ class C {} } var C = 1; export const value = C;");
 
-        ns.Get("value").AsString().Should().Be("undefined");
+        _engine.Modules.Import("invisible").Get("value").AsString().Should().Be("undefined");
+        _engine.Modules.Import("clobbered").Get("value").AsInteger().Should().Be(1);
     }
 
     [Fact]
@@ -1237,10 +1248,22 @@ export const count = globals.counter;
     }
 
     [Fact]
+    public void ShouldBindADefaultExportedClassAtModuleLevel()
+    {
+        // `export default class D {}` binds `D` at module level like any other module item, so `D` is in scope - and in
+        // its temporal dead zone - for the module items above it. The namespace's `default` is supplied by the export
+        // machinery either way, so only an early read of the name itself pins this arm of the whitelist.
+        _engine.Modules.Add("my-module", "export const value = typeof D; export default class D {}");
+
+        Invoking(() => _engine.Modules.Import("my-module")).Should().ThrowExactly<JavaScriptException>()
+            .Which.Message.Should().Be("Cannot access 'D' before initialization");
+    }
+
+    [Fact]
     public void ShouldStillApplyTheTemporalDeadZoneToModuleLevelDeclarations()
     {
         // A module-level `const` is bound uninitialized until its declaration runs, so reading it early throws even through typeof.
-        _engine.Modules.Add("my-module", "globalThis.probe = typeof x; const x = 1;");
+        _engine.Modules.Add("my-module", "typeof x; const x = 1;");
 
         Invoking(() => _engine.Modules.Import("my-module")).Should().ThrowExactly<JavaScriptException>()
             .Which.Message.Should().Be("Cannot access 'x' before initialization");
--- a/Jint/HoistingScope.cs
+++ b/Jint/HoistingScope.cs
@@ -267,6 +267,10 @@ internal sealed class HoistingScope
 
             var effectiveLexicalNames = currentScopeLexicalNames ?? enclosingLexicalNames;
 
+            // Only a direct child of the module, or an export wrapping one, is a module item; a top-level block or `for` head is not.
+            var atTopLevel = parent is null
+                || (parent is AstModule && node.Type is NodeType.ExportNamedDeclaration or NodeType.ExportDefaultDeclaration);
+
             foreach (var childNode in node.ChildNodes)
             {
                 var childType = childNode.Type;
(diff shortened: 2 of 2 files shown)


lahma:
I ran a deep review of this against `main`, building both revisions and executing the cases rather than reading only. **The fix is correct** — I confirmed all three broken `for`-head shapes and the nested-block cases genuinely fail on `main` and pass here:

| case | `main` | this PR |
| --- | --- | --- |
| `for (let key in {x:1}) {} var key = 1;` | throws `Cannot access 'key' before initialization` | ok |
| `for (let i = 0; i < 1; i++) {} var i = 5;` | throws | ok, `5` |
| `for (const o of [1]) {} var o = 6;` | throws | ok, `6` |
| `{ let a = 1; } var a = 2;` | throws | ok, `2` |
| `{ class C {} } typeof C` | `"function"` | `"undefined"` |

Nothing below disputes the shipped behaviour. Two test-coverage points I'd fix before merge, then some adjacent observations.

## Worth addressing in this PR

**1. The `ExportDefaultDeclaration` arm of the whitelist is untested.** I deleted `or NodeType.ExportDefaultDeclaration` from the predicate and `dotnet test -c Release -f net10.0 --filter ~ModuleTests` still reported `Failed: 0, Passed: 67` — `ShouldBindExportedLexicalDeclarationsAtModuleLevel` passes either way, because its `ns.Get("default").IsCallable` assertion is satisfied by the export machinery rather than by lexical hoisting. The arm *is* load-bearing though: `export const value = typeof D; export default class D {}` throws `ReferenceError: Cannot access 'D' before initialization` with it and silently yields `"undefined"` without it. The mirror of the named-export test would pin it:

```csharp
_engine.Modules.Add("my-module", "export const value = typeof D; export default class D {}");
Assert.Throws<JavaScriptException>(() => _engine.Modules.Import("my-module"));
```

For contrast, removing *both* arms fails 35 of the 67 tests — the named-export arm is well covered, only the default one is bare.

**2. Only the `for-in` head shape is pinned.** `ShouldNotBindAForHeadDeclarationAtModuleLevel` covers `for (let key in …)`, which is the *rarest* of the three shapes the fix repairs. Over test262's module-parseable language tests, directly under a Module: `ForOfStatement -> VariableDeclaration` 169 occurrences, `ForStatement -> VariableDeclaration` 163, `ForInStatement` 23. test262 is green both before and after this change, so these Jint.Tests facts are the only regression net the fix has — and the two shapes an actual bundler emits are the uncovered ones. Two extra `Modules.Add` lines in the same test would do it.

**3. `ShouldNotBindALexicalDeclarationFromANestedBlockAtModuleLevel` reads only the post-block value,** so it cannot distinguish the fixed behaviour from the equally-wrong outcome where the block's `let a = 1` initializes the module-level `var a` — both leave `value === 2`. One observation from *inside* the block (e.g. `{ let a = 1; globalThis.inner = a; } var a = 2;` asserted as `1`) pins the property the test is named for.

## Smaller notes

- `globalThis.probe = typeof x;` in `ShouldStillApplyTheTemporalDeadZoneToModuleLevelDeclarations` never executes — `typeof x` throws first — and `probe` is never read. `"typeof x; const x = 1;"` proves the same thing without implying a captured value.
- The new comment paraphrases a specific production; a link to [`Module : ModuleItemList` LexicallyScopedDeclarations](https://tc39.es/ecma262/#sec-module-semantics-static-semantics-lexicallyscopeddeclarations) would match the convention in `AGENTS.md` and in `SourceTextModule.InitializeEnvironment`, and would help the next person auditing whether the two-node whitelist is complete.
- `ShouldNotBindAClassDeclaredInABlockAtModuleLevel` asserts only invisibility, where its `let` sibling asserts the `var`-clobbering symptom. `{ class C {} } var C = 1; export const value = C;` is what the bundle actually hits for a class, and the class path is a separate branch from the variable path.
- Minor: `atTopLevel` is now computed for every visited node, where the old predicate only ran once a lexical declaration had already matched. For scripts and function bodies the `parent is AstModule` test can never be anything but the `parent is null` root case. A `private readonly bool _module` on `ScriptWalker`, set from `GetModuleLevelDeclarations`, would collapse those walks back to a null check. Not a blocker, just noting it since `VisitCore` is on the prepare path.

## Pre-existing, adjacent — not this PR's doing

Flagging these because they are the same bug class one branch away, so it's worth an explicit decision on fold-in vs. follow-up:

**The hoisting walk still descends into a class `static {}` block.** `StaticBlock` is absent from both the block-position list and the recursion stop-list, so a static block's `function` and `var` declarations land in the enclosing scope. Identical on `main` and on this branch:

```js
class C { static { function f() {} } }
typeof f            // Jint: "function"   V8/spec: "undefined"

class C { static { var sv = 1; } }
Object.getOwnPropertyNames(globalThis).indexOf('sv') >= 0   // Jint: true
```

The `var` half is the quieter one: the leaked global stays `undefined` because the static block assigns into its own scope, so a downstream `if (sv)` feature-detect takes the false branch instead of throwing.

**Class names never reach `_lexicalNames`.** The `ClassDeclaration` branch adds the node to `_lexicalDeclarations` but, unlike the sibling `VariableDeclaration` branch, never adds the bound name under `_collectLexicalNames`. `JintFunctionDefinition.BuildState` uses exactly that list as the Annex B.3.3.1 conflict filter, so a function-body class fails to block Annex B hoisting of a same-named block function:

```js
(function(){ class f { static tag = 'class'; } { function f() {} } return String(f.tag); })()
// -> "undefined"  (the block function won; spec B.3.3.1 forbids the hoist)
```

Two controls confirm the mechanism: the `let` analogue correctly returns `lex`, and the same case at global scope correctly returns `class` (the script path consults `GlobalEnvironment.HasLexicalDeclaration`, built from `_lexicalDeclarations`, which does see classes).

Both would fit naturally as follow-ups; the static-block one in particular is squarely the theme of this PR.

gkurt:
Thanks — especially for building both revisions and running the cases rather than reading. Everything you asked for is in, pushed as an amend.

**1. The `ExportDefaultDeclaration` arm.** You were right that it was bare. New `ShouldBindADefaultExportedClassAtModuleLevel` reads the name early instead of reading the namespace:

```csharp
_engine.Modules.Add("my-module", "export const value = typeof D; export default class D {}");
Invoking(() => _engine.Modules.Import("my-module")).Should().ThrowExactly<JavaScriptException>()
    .Which.Message.Should().Be("Cannot access 'D' before initialization");
```

I verified it the way you verified the gap: with `or NodeType.ExportDefaultDeclaration` deleted, that test is the only one of the 68 in `ModuleTests` that fails.

**2. The two head shapes a bundler actually emits.** `ShouldNotBindAForHeadDeclarationAtModuleLevel` now covers all three, and the `for`/`for-of` cases assert the `var`-clobbering symptom rather than `typeof`, since that is the damaging one:

```csharp
_engine.Modules.Add("for-in", "for (let key in { x: 1 }) {} export const value = typeof key;");
_engine.Modules.Add("for", "for (let i = 0; i < 1; i++) {} var i = 5; export const value = i;");
_engine.Modules.Add("for-of", "for (const o of [1]) {} var o = 6; export const value = o;");
```

**3. The nested-block test now observes from inside the block**, so the outcome you describe — the block's `let a = 1` initializing the module-level `var a` — is excluded rather than indistinguishable:

```csharp
_engine.Modules.Add("my-module", "let inner; { let a = 1; inner = a; } var a = 2; export const value = a; export { inner };");
```

I used an exported `let` instead of `globalThis` so the fact stays inside the module.

**4. Class-in-a-block asserts both symptoms.** Added the `{ class C {} } var C = 1;` case next to the invisibility one, since the class path is its own branch.

**5. The dead `globalThis.probe` line is gone** — the TDZ test is now just `"typeof x; const x = 1;"`.

**6. Spec link added**, to `Module : ModuleItemList` LexicallyScopedDeclarations, per the convention.

**7. `atTopLevel` no longer costs anything off the module path.** `ScriptWalker` carries a `_module` flag set from `GetModuleLevelDeclarations`, so a script or function-body walk short-circuits to the root test:

```csharp
var atTopLevel = parent is null
    || (_module && node.Type is NodeType.ExportNamedDeclaration or NodeType.ExportDefaultDeclaration && parent is AstModule);
```

Taken together the regression net is real now rather than nominal: reverting the predicate to `parent is null or AstModule` fails three of the tests, where before it failed neither of the two shapes that turned this up.

**On the two pre-existing bugs** — thanks for flagging them, and for asking for an explicit decision rather than leaving it. I reproduced both exactly as you described (static blocks leak `function f` *and* `var sv`; the function-body `class f` loses to the block function while the `let` control correctly doesn't) and took the **follow-up** route, because both affect scripts and function bodies rather than modules and would have widened this PR's blast radius past what it is about.

They are fixed together in #2914. Two details from doing it that are worth having here: test262 misses the static-block `var` leak *because* of the symptom you identified — `static-init-scope-var-close.js` reads `undefined` either way, since the leaked binding is created outside and assigned inside — and its annexB `function-code` suite has no class-conflict variant at all, every `skip-early-err` case uses `let`. So neither had any coverage anywhere.

lahma:
Re-reviewed at `ed7423267`, building and executing again rather than reading the diff. Full `Jint.Tests` is green (Release, net10.0 — 4666 passed, 4 skipped), `ModuleTests` 67 → 68.

I checked your claims rather than taking them: reverting the predicate to `parent is null or AstModule` does fail exactly three tests (`ShouldNotBindAForHeadDeclarationAtModuleLevel`, `ShouldNotBindALexicalDeclarationFromANestedBlockAtModuleLevel`, `ShouldNotBindAClassDeclaredInABlockAtModuleLevel`), so the regression net is real now. #2914's diff does fix both pre-existing bugs — `NodeType.StaticBlock` in the recursion stop-list is sufficient on its own, since once the walk refuses to descend, the block-position list at line 312 can never see one and the static block's own walk roots on a `FunctionBody` with `parent == null`. Your test262 observations hold as well: all three `static-init-scope-var-*` files are blind to the leak, not just `-close`, and across all 280 annexB `skip-early-err*` files not one uses a class as the conflicting binding (six of the seven cases use `let`, the seventh a destructuring catch parameter).

Two things left.

## The `_module` flag has a second consumer

`GetProgramLevelDeclarations` still constructs its walker with the `module: false` default — and a module root does reach it. `Acornima.Ast.Module` reports `NodeType.Program`, so `AstAnalyzer` builds a `CachedHoistingScope` over it at prepare time. Probing `Engine.PrepareModule("export const a = 1; const b = 2;")` with the same internals probe on both revisions:

| walk | `92436389b` | `ed7423267` |
| --- | --- | --- |
| `GetModuleLevelDeclarations` | 2 | 2 |
| `GetProgramLevelDeclarations` | 2 | **1** |

The exported `const a` is gone from the cached scope, because `atTopLevel` short-circuits on `_module` before it can reach `parent is AstModule`. The probe also confirms the root really is `Acornima.Ast.Module`, `Type` is `Program`, and `UserData` is a `CachedHoistingScope`, so this list is genuinely built and stored for every prepared module.

It is not live today. `SourceTextModule.InitializeEnvironment` builds its own cache from `GetModuleLevelDeclarations` (`SourceTextModule.cs:295`/`321`), and every reader of the cached scope — `GetHoistingScope`, `GetVarNames`, `GetLexNames` at `Engine.cs:1717`/`1752`/`1774`/`2156` and `ShadowRealm.cs:173` — is `Script`-typed, which a `Module` cannot satisfy. But the walk's answer is now a function of which factory was called rather than of the AST, and the trap sits exactly where the obvious next cleanup lands: having modules consume the prepare-time scope instead of re-walking on every `InitializeEnvironment` would silently drop every `export const` / `export let` / `export class` binding, with the suite still green.

One expression restores the invariant while keeping the win you were after — the short-circuit stays, it just becomes one `isinst` per walk instead of per node:

```csharp
var treeWalker = new ScriptWalker(collectVarNames, collectLexicalNames, module: script is AstModule);
```

## The nested-block test still does not exclude the collapsed outcome

This one is my fault as much as anything — the phrasing I suggested (`{ let a = 1; globalThis.inner = a; } var a = 2;`) has the same hole, so asking for "an observation from inside the block" was the wrong ask.

Modelling both worlds in-language — dropping the `let` makes the block's initializer land on the module-level binding, which is precisely the collapsed semantics — the two are still indistinguishable in the shape the test uses:

```
--- `{ let a = 1; } var a = 2;`  (the shape the test uses) ---
own-binding : inner=1 value=2
collapsed   : inner=1 value=2      <- identical

--- `var a = 2; { let a = 1; }`  (var first) ---
own-binding : inner=1 value=2
collapsed   : inner=1 value=1      <- discriminates
```

In the collapsed world the block writes 1 into the module-level binding, `inner` reads 1, and `var a = 2` then overwrites it back to 2 — so both assertions are satisfied by an engine whose block never got its own binding. What discriminates is ordering the `var` *before* the block, so the outer binding already holds a different value while the block runs:

```csharp
_engine.Modules.Add("my-module", "var a = 2; { let a = 1; } export const value = a;");
// 2 when the block has its own binding, 1 when it does not
```

That is a second `Modules.Add` rather than a replacement — the existing script is still the one that pins the original TDZ symptom, which the reordered one does not reach.

Neither of these blocks the fix itself, which I'm satisfied is correct.

lahma:
Merging — thank you for this. Three rounds, and every one of them came back with the thing actually verified rather than argued: you mutation-checked your own new test the way I checked the gap, you took the two pre-existing bugs as #2914 instead of declaring them out of scope, and the note about `static-init-scope-var-close.js` being blind *for the reason the symptom predicts* is the kind of detail that only turns up when someone really goes and looks. The regression net is genuinely load-bearing now: reverting the predicate to `parent is null or AstModule` fails three tests, where before it failed neither of the two shapes that turned this up.

The Windows CI failure is `AsyncTests.ShouldPromiseBeResolved2` timing out, a known flake unrelated to this change — net472 passed the same test in the same run, and it passes locally on your head.

I'll take the two leftovers from my last comment myself rather than ask for a fourth round, since one of them is a consequence of a change I asked you to make:

- `GetProgramLevelDeclarations` deriving `module` from the node instead of defaulting it, so the walk's answer depends on the AST rather than on which factory was called.
- The `var a = 2; { let a = 1; }` ordering as an additional case, since the existing script cannot separate the two worlds on its own.

Will link the follow-up here.

lahma:
Follow-up is #2923 — the derived module flag plus the block test, both mutation-checked. Thanks again.

PR #2870, opened 2026-07-30
Find a registered module under the name it was registered with

gkurt:
`Modules.Add` files a builder under the specifier the host passed it, and `Load` looked one up by the key the module loader resolved that specifier to:

```csharp
_builders.Add(specifier, moduleBuilder);                                  // Engine.Modules.cs
if (_builders.TryGetValue(moduleResolution.Key, out var moduleBuilder))   // Engine.Modules.cs
```

Those are the same string often enough for the mismatch to go unnoticed. `DefaultModuleLoader` returns a bare specifier unchanged, and the existing coverage — `ModuleBuilderWithCustomModuleLoaderLoadsModulesProperly` — has its resolve handler return exactly the name it registered, so it satisfies the contract by construction.

A loader that canonicalizes cannot. One resolving urls builds a `Uri`, and `Uri` supplies the path a path-less url lacks:

```csharp
engine.Modules.Add("http://localhost", "export const value = 1;");
engine.Modules.Import("http://localhost");
// resolved key is "http://localhost/" — no builder under it, so the loader is asked
// to fetch a module whose source it was already handed
```

The host gets no error, which is what makes this worth fixing rather than documenting: the source it supplied is silently replaced by whatever the loader returns for that url. For a dev server, requesting its root answers with an HTML entry document, so it eventually surfaces as a syntax error in a module the host never wrote.

`_builders.Remove` was already being passed the resolved key rather than the registration name, so on a mismatch the entry was neither found nor removed.

## The fix

`TryGetBuilder` tries the request's own specifier when the resolved key finds nothing, and `LoadFromBuilder` is handed whichever name matched so the removal targets the right entry.

Canonicalizing inside `Add` instead would give a module one identity throughout, which is tidier. I did not, because `Add` is reachable without `EnableModules` — where `FailFastModuleLoader.Resolve` throws — so it would start throwing for hosts that register modules and never configure a loader. Happy to switch if you would rather have the stricter shape.

## Tests

One test, reusing the existing `LocationResolveOnlyModuleLoader`, whose `LoadModuleContents` throws — so bypassing the builder fails hard rather than quietly returning the wrong module. It fails on `main` and passes here.

`Jint.Tests` 4659 passed; test262 99484 passed, 0 failed, unchanged from `main`. The only failures on my machine are `ToStringUsesDaylightNameWhenInDst` and `DateToStringMethodsShouldUseCurrentTimeZoneAndCulture`, which assert OS-localised time zone display names and fail here regardless of this change.

### DIFF
--- a/Jint.Tests.PublicInterface/HostModuleBuilderIdentityTests.cs
+++ b/Jint.Tests.PublicInterface/HostModuleBuilderIdentityTests.cs
@@ -1,11 +1,6 @@
 #nullable enable
 
-using System.Collections.Generic;
-using System.IO;
-using Jint.Runtime;
 using Jint.Runtime.Modules;
-
-using Module = Jint.Runtime.Modules.Module;
 
 namespace Jint.Tests.PublicInterface;
 
@@ -126,8 +121,8 @@ public class HostModuleBuilderIdentityTests
         Invoking(() => engine.Modules.Add("http://localhost", "export const value = 'second';"))
             .Should().NotThrow("the registration was consumed by the load, so the name is free again");
 
-        // The re-registration is reachable through the index too, and the already-cached module still wins for
-        // the key it was cached under - a load consumes the builder, it does not evict the module.
+        // The already-cached module keeps answering imports of the name - a load consumes the builder, it
+        // does not evict the module - so the re-registration is never consulted for this key.
         engine.Modules.Import("http://localhost").Get("value").AsString().Should().Be("first");
     }
 
@@ -236,5 +231,124 @@ public class HostModuleBuilderIdentityTests
 
         engine.Modules.Add("http://elsewhere", "export const value = 'second';");
         engine.Modules.Import("http://elsewhere").Get("value").AsString().Should().Be("second");
+    }
+
+    [Fact]
+    public void TwoRegistrationsResolvingToTheSameKeyFailLoudly()
+    {
+        // Both spellings canonicalize to http://localhost/, so whichever registration lost would be silently
+        // unreachable under every name - its own name resolves to the shared key too, which now belongs to the
+        // winner. That is host-supplied source being discarded, the failure this identity scheme exists to
+        // eliminate, so the import that discovers the pair throws instead of picking a winner.
+        var engine = new Engine(options => options.EnableModules(new CanonicalizingModuleLoader()));
+
+        engine.Modules.Add("http://localhost", "export const value = 'first';");
+        engine.Modules.Add("HTTP://localhost", "export const value = 'second';");
+
+        Invoking(() => engine.Modules.Import("http://localhost"))
+            .Should().Throw<InvalidOperationException>().WithMessage("*resolve to the same specifier*");
+    }
+
+    [Fact]
+    public void ARegistrationCollidingWithOneFiledUnderTheCanonicalNameFailsLoudlyToo()
+    {
+        // The other collision shape: the earlier registration already carries the canonical spelling, so it is
+        // matched without ever being indexed, and only the later registration's resolution reveals the clash.
+        var engine = new Engine(options => options.EnableModules(new CanonicalizingModuleLoader()));
+
+        engine.Modules.Add("http://localhost/", "export const value = 'canonical';");
+        engine.Modules.Add("http://localhost", "export const value = 'short';");
+
+        Invoking(() => engine.Modules.Import("http://elsewhere"))
+            .Should().Throw<InvalidOperationException>().WithMessage("*resolve to the same specifier*");
+    }
+
+    [Fact]
+    public void CancellationInsideTheIndexingPassIsNotSwallowed()
+    {
+        // A refusal is the loader rejecting one name; cancellation is the host calling off the whole
+        // operation, and it must reach whoever cancelled rather than be treated as "leave that registration
+        // unindexed" while the triggering import carries on loading and evaluating.
+        var engine = new Engine(options => options.EnableModules(new CancellingResolveLoader()));
+
+        engine.Modules.Add("cancelled", "export const value = 1;");
+        engine.Modules.Add("http://localhost", "export const value = 2;");
+
+        Invoking(() => engine.Modules.Import("http://localhost"))
+            .Should().Throw<OperationCanceledException>();
+    }
+
+    /// <summary>
+    /// Canonicalizes like <see cref="CanonicalizingModuleLoader"/>, except that one specifier makes it throw
+    /// <see cref="OperationCanceledException"/> — the shape of a loader honoring a <c>CancellationToken</c>
+    /// mid-resolve.
+    /// </summary>
+    private sealed class CancellingResolveLoader : ModuleLoader
+    {
+        public override ResolvedSpecifier Resolve(string? referencingModuleLocation, ModuleRequest moduleRequest)
+        {
+            if (string.Equals(moduleRequest.Specifier, "cancelled", StringComparison.Ordinal))
+            {
+                throw new OperationCanceledException();
+            }
+
+            var uri = new Uri(moduleRequest.Specifier, UriKind.Absolute);
+            return new ResolvedSpecifier(moduleRequest, uri.AbsoluteUri, uri, SpecifierType.RelativeOrAbsolute);
+        }
+
+        protected override string LoadModuleContents(Engine engine, ResolvedSpecifier resolved)
+            => throw new InvalidOperationException("the loader was asked to load " + resolved.Key);
+    }
+
+    [Fact]
+    public void ALoaderHandingBackANullKeyDoesNotBreakAnUnrelatedImport()
+    {
+        // A loader compiled without nullable annotations can hand back a null key instead of throwing. That is
+        // a refusal in a different costume, and like the thrown kind it must not surface on the import that
+        // merely triggered the indexing pass.
+        var engine = new Engine(options => options.EnableModules(new NullKeyResolveLoader()));
+
+        engine.Modules.Add("broken", "export const value = 'unreachable';");
+        engine.Modules.Add("http://localhost", "export const value = 'fine';");
+
+        engine.Modules.Import("http://localhost").Get("value").AsString().Should().Be("fine");
+    }
+
+    /// <summary>
+    /// Canonicalizes like <see cref="CanonicalizingModuleLoader"/>, except that one specifier resolves to a
+    /// <c>null</c> key — which the annotations forbid, but a loader compiled without them can produce.
+    /// </summary>
+    private sealed class NullKeyResolveLoader : ModuleLoader
+    {
+        public override ResolvedSpecifier Resolve(string? referencingModuleLocation, ModuleRequest moduleRequest)
+        {
+            if (string.Equals(moduleRequest.Specifier, "broken", StringComparison.Ordinal))
+            {
+                return new ResolvedSpecifier(moduleRequest, Key: null!, Uri: null, SpecifierType.Bare);
+            }
+
+            var uri = new Uri(moduleRequest.Specifier, UriKind.Absolute);
+            return new ResolvedSpecifier(moduleRequest, uri.AbsoluteUri, uri, SpecifierType.RelativeOrAbsolute);
+        }
+
+        protected override string LoadModuleContents(Engine engine, ResolvedSpecifier resolved)
+            => throw new InvalidOperationException("the loader was asked to load " + resolved.Key);
+    }
+
+    [Fact]
+    public void APreCompiledModuleResolvesItsRelativeImportsAgainstItsPrepareTimeName()
+    {
+        // A pre-compiled module cannot be renamed by the registration - the AST is the host's - so its
+        // relative imports resolve against the name it was prepared with. Preparing it under the key the
+        // loader resolves the registration to, as the AddModule doc says, is what makes the two agree.
+        var engine = new Engine(options => options.EnableModules(new VirtualTreeLoader(new Dictionary<string, string>
+        {
+            ["/vfs/util.js"] = "export const value = 'from the virtual sibling';",
+        })));
+
+        var prepared = Engine.PrepareModule("export { value } from './util.js';", "/vfs/lib.js");
+        engine.Modules.Add("lib", builder => builder.AddModule(prepared));
+
+        engine.Modules.Import("lib").Get("value").AsString().Should().Be("from the virtual sibling");
     }
 }
--- a/Jint.Tests.PublicInterface/HostModuleBuilderIdentityTests.cs
+++ b/Jint.Tests.PublicInterface/HostModuleBuilderIdentityTests.cs
@@ -1,0 +1,240 @@
+#nullable enable
+
+using System.Collections.Generic;
+using System.IO;
+using Jint.Runtime;
+using Jint.Runtime.Modules;
+
+using Module = Jint.Runtime.Modules.Module;
+
+namespace Jint.Tests.PublicInterface;
+
+/// <summary>
+/// The name a programmatically registered module is found under — <c>Engine.Modules.Add</c> versus the key the
+/// <see cref="ModuleLoader"/> resolves an import to. A registration has to have exactly one identity, because a
+/// second one either hides it (the loader canonicalized and the engine looks under the canonical name only) or
+/// over-exposes it (the raw import text matches, which for a relative specifier names a different file in every
+/// importing directory).
+/// </summary>
+public class HostModuleBuilderIdentityTests
+{
+    /// <summary>
+    /// Resolves every specifier through <see cref="Uri"/>, the way a loader serving urls has to, and refuses to
+    /// load anything at all — so a test that reaches the loader fails loudly instead of silently getting the
+    /// wrong source.
+    /// </summary>
+    private sealed class CanonicalizingModuleLoader : ModuleLoader
+    {
+        public override ResolvedSpecifier Resolve(string? referencingModuleLocation, ModuleRequest moduleRequest)
+        {
+            var uri = Uri.TryCreate(referencingModuleLocation, UriKind.Absolute, out var referrer)
+                ? new Uri(referrer, moduleRequest.Specifier)
+                : new Uri(moduleRequest.Specifier, UriKind.Absolute);
+
+            return new ResolvedSpecifier(moduleRequest, uri.AbsoluteUri, uri, SpecifierType.RelativeOrAbsolute);
+        }
+
+        protected override string LoadModuleContents(Engine engine, ResolvedSpecifier resolved)
+            => throw new InvalidOperationException("the loader was asked to load " + resolved.Key);
+    }
+
+    [Fact]
+    public void ARegisteredModuleIsFoundWhenTheLoaderCanonicalizesItsName()
+    {
+        // `new Uri("http://localhost").AbsoluteUri` is "http://localhost/", so the registration name and the
+        // resolved key differ by the path Uri supplies. The loader throws if it is reached, which is what the
+        // mismatch used to cause.
+        var engine = new Engine(options => options.EnableModules(new CanonicalizingModuleLoader()));
+
+        engine.Modules.Add("http://localhost", "export const value = 'from the builder';");
+
+        engine.Modules.Import("http://localhost").Get("value").AsString().Should().Be("from the builder");
+    }
+
+    [Fact]
+    public void ARegisteredModuleIsFoundUnderTheCanonicalSpellingToo()
+    {
+        // The direction a raw-specifier fallback cannot reach: the import is written the way an import map or a
+        // generated specifier writes it, so it already equals the resolved key and never needs the fallback -
+        // but the registration is still filed under the shorter name.
+        var engine = new Engine(options => options.EnableModules(new CanonicalizingModuleLoader()));
+
+        engine.Modules.Add("http://localhost", "export const value = 'from the builder';");
+
+        engine.Modules.Import("http://localhost/").Get("value").AsString().Should().Be("from the builder");
+    }
+
+    [Fact]
+    public void EitherSpellingReachesTheSameModuleInstance()
+    {
+        // Both spellings resolve to one key, so they are one module - not two, and not one plus a failure once
+        // the registration has been consumed.
+        var engine = new Engine(options => options.EnableModules(new CanonicalizingModuleLoader()));
+
+        engine.Modules.Add("http://localhost", "export const value = {};");
+
+        var first = engine.Modules.Import("http://localhost");
+        var second = engine.Modules.Import("http://localhost/");
+
+        second.Should().BeSameAs(first);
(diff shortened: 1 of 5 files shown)


lahma (commented):
Thanks for digging into this — the bug you're fixing is real and worth fixing: a host that calls `Modules.Add("http://localhost", …)` against a canonicalizing loader currently has its registration silently bypassed in favour of a network/disk load. I went through the change carefully (including building the branch and running the scenarios below), and I think the fix picks the wrong key to fall back to. Details, most important first.

## The raw specifier is not a module identity

`TryGetBuilder`'s fallback matches on `moduleResolution.ModuleRequest.Specifier` — the raw, unresolved import text. For a relative import that text is relative to the **referencing** module, so it does not identify a module at all.

```csharp
var engine = new Engine(o => o.EnableModules(@"C:\Scripts"));
engine.Modules.Add("./config.js", "export const env = 'prod';");
// C:\Scripts\sub\thing.js contains:  import './config.js'   → means C:\Scripts\sub\config.js
```

`DefaultModuleLoader` resolves that import to `file:///C:/Scripts/sub/config.js`, which misses `_builders`; the new fallback then looks up the raw text `./config.js` and hits the host's registration. `sub/thing.js` silently gets the host module instead of its own sibling — wrong exports, wrong side effects, no error.

It gets worse on the second import, because `LoadFromBuilder` **consumes** the registration:

```csharp
engine.Modules.Add("./shared.js", src);
// a/main.js:  import { v } from './shared.js'   → resolves to file:///base/a/shared.js
// b/main.js:  import { v } from './shared.js'   → resolves to file:///base/b/shared.js
```

The first one to be traversed matches via the fallback, caches under *its* resolved key, and runs `_builders.Remove("./shared.js")`. The second misses `_modules` (different key), misses `_builders` (now removed), and falls through to the loader — `Module Not Found`. The identical import statement succeeds in one module and throws in another, decided purely by graph traversal order. The host's own later `Modules.Import("./shared.js")` is dead too.

## Three more consequences of the same widened lookup

**Nested imports get the wrong base.** `ModuleBuilder.Parse()` parses with `source: _specifier`, so a `BuilderModule`'s `Location` is the *registration* name while the module is cached under the resolved key. `Host.GetImportedModule` passes `referrer.Location` as `referencingModuleLocation`. With a loader mapping `"lib"` → `"/vfs/lib.js"`, `Modules.Add("lib", "import './util.js'; …")` now resolves via the fallback, and its nested `./util.js` is resolved against `"lib"` rather than `/vfs/lib.js`. Before this PR that module came from the loader with `Location == Key` and the nested import resolved correctly.

**`ArgumentNullException` on a shape that used to work.** This is the first place `Load` dereferences `ModuleRequest.Specifier`. `ModuleRequest` is a `readonly record struct`, so a host loader returning `new ResolvedSpecifier(default, key, null, SpecifierType.Bare)` yields `null` there and `_builders.TryGetValue(null, …)` throws before any module loads. Note the adjacent `ModuleCacheKey.From` is explicitly written to survive a partially-populated request (`resolved.ModuleRequest.Attributes ?? []`), so this is a real shape, not a hypothetical.

**Import attributes are dropped.** `_modules`/`ModuleCacheKey` are attribute-aware; `_builders` is not. Widening which requests can reach a builder therefore widens the set of requests whose `with { type: … }` is silently ignored. Reproduced on this branch with a canonicalizing loader:

```csharp
engine.Modules.Add("data", """{ "message": "hello" }""");
// import json from "data" with { type: "json" }
// → JavaScriptException: error in module 'data': Unexpected token ':' (data:1:12)
```

Before the change that request missed the builder map and went to `LoadFromModuleLoader`, which honours `IsJsonModule()` / `IsTextModule()`.

## The fix is also incomplete in two directions

**Canonical spelling still misses.** The guard is `!string.Equals(specifier, moduleResolution.Key)`, so when the import is written the canonical way — `import 'http://localhost/'` with the trailing slash, which is exactly what an import map or a generated specifier produces — `specifier == Key`, the fallback short-circuits, and the builder is never consulted. The same module resolves to a different source depending on spelling. Ordering compounds it: once the loader-loaded module is cached under `http://localhost/`, a later `import 'http://localhost'` hits `_modules` first and the registration becomes permanently unreachable.

**Loaders that canonicalize by *rejecting* are untouched.** The fallback runs after `ModuleLoader.Resolve`, so it only helps loaders that canonicalize by rewriting. The shipped `DefaultModuleLoader` frequently canonicalizes by throwing:

```csharp
var engine = new Engine(o => o.EnableModules(basePath));
engine.Modules.Add("./config", "export const x = 1;");
engine.Modules.Import("./config");   // Unsupported Directory Import — _builders never consulted
```

Same for `Unauthorized Module Path`, `Invalid Module Specifier`, and the `#`-import `NotSupportedException`. The stated goal ("a registered module is found under the name it was registered with") is met only for the subset of loaders whose canonicalization succeeds.

## The reason given for rejecting the deeper fix doesn't hold

The PR description rules out canonicalizing inside `Add` because `Add` is reachable without `EnableModules`, where `FailFastModuleLoader.Resolve` throws. It doesn't throw — `Jint/Runtime/Modules/FailFastModuleLoader.cs:18-21`:

```csharp
public ResolvedSpecifier Resolve(string? referencingModuleLocation, ModuleRequest moduleRequest)
{
    return new ResolvedSpecifier(moduleRequest, moduleRequest.Specifier, Uri: null, SpecifierType.Bare);
}
```

Only `LoadModule` and `BasePath` throw. So `Add` can call `ModuleLoader.Resolve(null, new ModuleRequest(specifier, []))` and file the builder under `Key` — or `Load` can lazily re-key `_builders` once on first use if you'd rather not resolve at registration time. Either way modules get **one** identity instead of two, and that single change fixes the canonical-spelling direction, the relative-specifier cross-match, the `Location` mismatch and the attribute blindness together, instead of leaving `_builders` and `_modules` permanently disagreeing about what identifies a module.

The one case that shape still doesn't cover is the reject-at-`Resolve` loader above; if you want that too, matching builders before/independently of `Resolve` is the only thing that reaches it — which is worth deciding deliberately, since it changes whether a registration shadows an on-disk module.

## Smaller points

- **Public documentation.** `Modules.Add` now has two effective key namespaces, and the only place that says so is the `<summary>` of a private method. The public `Add` overloads have no XML doc and the README module section is untouched — it still tells hosts that registration is keyed by one name. The module APIs are third-party integration surface, where observable behaviour changes are breaking even with an untouched signature, so the rule wants stating where an embedder will see it.
- **`_builders.Remove` isn't pinned.** The other half of the fix — passing the matched specifier rather than `moduleResolution.Key` into `LoadFromBuilder` — has no test. `LoadFromBuilder` caches into `_modules` before removing, so reverting that argument leaves the whole suite green while leaking a builder entry for the engine's lifetime. Importing twice, or asserting a re-`Add` of the same specifier succeeds rather than throwing from `_builders.Add`, would pin it.
- **No negative test.** Nothing asserts that a builder registered under a relative or bare name is *not* handed to an unrelated import spelling the same text. That ambiguity is the riskiest part of the change and has zero coverage — whichever way it's resolved, a test should record the intent.
- **`ModuleBuilder` already holds the specifier** (`_specifier`, used as the parse source name), so the threaded out-parameter duplicates existing state — `LoadFromBuilder` could call `_builders.Remove(moduleBuilder.Specifier)` and `TryGetBuilder` could drop to one out-param. Worth checking the `Add(string, ModuleBuilder)` overload first, since it can file a builder under a second, different name.
- **Test placement.** Per AGENTS.md, integrator-facing contract tests belong in `Jint.Tests.PublicInterface` (the only project without `InternalsVisibleTo`). Weak point though — the two directly analogous tests already live in `ModuleTests.cs` and share the `LocationResolveOnlyModuleLoader` helper, so co-locating is defensible.

gkurt:
Thanks — this one changed the design rather than the tests. The fallback is replaced, not extended, in a follow-up commit on top of the reviewed one (the branch also carries a third commit that removes a README paragraph the second one added; that was my call to revert, not a review point).

## The raw specifier is gone

You were right that it is not a module identity, and the consuming-registration ordering hazard is the part that convinced me: the same import statement succeeding in one module and throwing in another, decided by graph traversal order, is not a behaviour worth shipping to fix a spelling mismatch.

So a builder now has exactly **one** identity — the key its registration resolves to. `TryGetBuilder` looks under `moduleResolution.Key`, and `IndexBuilderKeys` puts each registration through `ModuleLoader.Resolve` once and indexes it under the resulting key. A registration is resolved against no referrer, the way `Modules.Import` resolves one, so `Add("./dep.js")` names `<base>/dep.js` and an import of `'./dep.js'` inside `sub/entry.js` resolves elsewhere and does not match.

That also fixes the direction the fallback structurally could not reach — the `!string.Equals(specifier, Key)` short-circuit you found, where an import written canonically never consulted the builder at all.

**On attribute blindness:** the fix is that the lookup no longer widens. `_builders` is still not attribute-aware where `ModuleCacheKey` is, so a request for a registered name carrying `with { type: 'json' }` still reaches the builder and still drops the attribute when the two names already agree — that is pre-existing on `main` and I have not touched it. What is gone is the *widened* set of requests that acquired the problem.

## Lazy, not at registration time

I checked `FailFastModuleLoader` and you are right — `Resolve` returns a `ResolvedSpecifier`, only `BasePath` and `LoadModule` throw. So the PR description's stated reason for rejecting the deeper fix was simply wrong, and I have said so in the follow-up rather than leaving it standing.

I still took the lazy half of your suggestion over resolving inside `Add`, because it keeps `Add` independent of the loader: registration neither requires `EnableModules` nor gains a failure mode for a specifier the loader would reject.

One consequence worth flagging, since it is the only place the implementation is deliberately permissive: the indexing pass runs inside some *other* module's load, so a `Resolve` that throws is swallowed and that registration is left unindexed. Letting it out would fail an unrelated import that merely happened to trigger the pass. Such a registration stays reachable only under its own name — exactly as reachable as before the index existed. It is attempted once, not on every load. `ARegistrationTheLoaderRefusesToResolveDoesNotBreakAnUnrelatedImport` pins it.

## The nested-import base

Fixed, and it needed a source change rather than just a lookup change: `ModuleBuilder.Parse` now takes the location, and `LoadFromBuilder` passes the resolved key. Your `"lib"` → `"/vfs/lib.js"` example is exactly right — without this, every registration the new index makes reachable would have had broken relative imports, since `Location` is the referrer its own imports resolve against. The parse-error message still names the module by its registration name, on the grounds that the diagnostic is for the host rather than for the resolver.

## Deliberately out of scope

The reject-at-`Resolve` loader. The index runs after `Resolve`, so it only helps loaders that canonicalize by rewriting, and `DefaultModuleLoader` throws instead for `Add("./config")` — plus `Unauthorized Module Path`, `Invalid Module Specifier`, and the `#`-import `NotSupportedException`. Reaching those means matching builders before or independently of `Resolve`, which makes a registration unconditionally shadow an on-disk module of the same name. That is a semantics change that deserves deciding on its own rather than arriving as a side effect, so the stated goal now holds for the subset of loaders whose canonicalization succeeds, and the gap is written down instead of implied.

## Tests

Moved to `Jint.Tests.PublicInterface`, including the original one. I went the other way from your "defensible to co-locate" on this, mainly for consistency with the sibling module PR; `LocationResolveOnlyModuleLoader` stays in `ModuleTests.cs` for the tests that already use it.

Each decision is pinned against the mutation that reverses it, and I made each mutation to check rather than assuming:

- Restoring the raw-specifier fallback fails `ARegisteredRelativeNameIsNotHandedToAnUnrelatedImportSpellingTheSameText` and `ARegisteredModuleIsFoundUnderTheCanonicalSpellingToo` — the correctness bug and the incompleteness, and nothing else.
- Passing `moduleResolution.Key` to `_builders.Remove` fails `ARegistrationIsConsumedUnderTheNameItWasFiledUnder`. You were right that this had no coverage at all, and right about why it is invisible: the module is cached before the removal, so the leak never surfaces as a failed import. Re-registering the same specifier is what surfaces it.
- Parsing with the registration name fails `ARegisteredModuleResolvesItsOwnRelativeImportAgainstItsResolvedKey`.

That last one is worth admitting: my first version of it **did not** pin the change. It registered under an already-canonical url, where the registration name and the resolved key coincide, so the mutation survived it and passed. I only caught that by running the mutation, and rewrote the test around your virtual-tree example. It is the same failure mode you identified in the original tests — coverage that cannot distinguish the policy it is supposed to be about.

Plus the negative and edge facts: `Add` still works with modules disabled, a registration added after an earlier import already triggered the index is still picked up, and both spellings of a canonicalized url reach the same module *instance* rather than two.

## Documentation

The two-key-namespaces problem is moot now there is one namespace, but your point that the rule was documented only on a private method stands. The three public `Add` overloads had no XML doc at all and now carry it: the resolved-key rule, both consequences, that resolution is lazy and why, and that a registration is consumed by its first load. I initially added a README paragraph too and then removed it again — the XML docs are where an embedder reading the API meets this, and a second copy is a second thing to keep correct.

## CI

The red windows job is not from this change: it is three `intl402/DateTimeFormat/prototype/format/temporal-objects-*` test262 tests, and all 500 `DateTimeFormat` tests pass locally on the branch. Runner ICU data. Locally green otherwise — test262, `Jint.Tests.CommonScripts`, both host suites, and the host-contract verification leg — apart from two `DateTests`/`EngineTests` timezone failures that reproduce on unmodified `main` on this machine.

lahma (commented):
This is the right design, and the response is a model of how to take a review — especially running each mutation to check the pin actually pins, and admitting the first version of the virtual-tree test didn't. I re-verified the round-one findings against the new head and they are all genuinely resolved: the raw-specifier cross-match, the canonical-spelling miss, the consumed-builder second-import failure, the nested-import base (for source registrations), the `ModuleRequest.Specifier` null deref, and the `_builders.Remove` key are each either structurally impossible now or pinned by a test that fails under the reverting mutation.

The remaining issues are all in the new code, they're all small, and none of them needs another redesign. Five I'd fix before merge, then advisory notes.

## Before merge

**1. Two registrations resolving to the same key: silent, and not actually last-wins.**

```csharp
engine.Modules.Add("http://localhost", srcA);
engine.Modules.Add("HTTP://localhost", srcB);   // Uri-canonicalizing loader: both → "http://localhost/"
```

The index pass overwrites `_builderKeys["http://localhost/"]`, so one registration becomes unreachable under *every* name — its own raw name also resolves to the shared key, which now maps to the winner. No exception anywhere, while `Add` of the identical spelling twice throws `ArgumentException`. That's host-supplied source silently discarded, the exact failure mode this PR exists to eliminate. And the `// Last registration wins` comment overstates what the code does: `pending` is built from `_builders.Keys` in dictionary enumeration order, which stops matching insertion order once consumption has removed entries and freed slots get reused. Detecting the collision during indexing and throwing (or at minimum picking a winner deterministically and documenting it) is a few lines.

**2. The catch is too broad, and the failure it swallows is permanent.**

Two halves:

- The bare `catch` also swallows `OperationCanceledException`. A loader whose `Resolve` honors a `CancellationToken` mid-pass has its cancellation silently discarded — the pass continues and the triggering import proceeds to load and evaluate. `catch (OperationCanceledException) { throw; }` before the swallow (or an equivalent filter) fixes it.
- Because the name is stamped into `_indexedBuilders` *before* `Resolve` runs and nothing ever removes it, one transient failure — a loader whose resolution depends on host state that isn't ready at first import — excludes that registration from the index for the engine's lifetime, with no retry, no diagnostic, and no recovery via re-`Add`. Every later import of the canonical spelling silently falls through to the loader. If attempt-once is the intended semantics for transient failures too (not just deliberate refusals), the `Add` doc should say so; the line it replaced had no swallowed error path at all.

**3. The `_builderKeys` write sits outside the catch it exists next to.**

```csharp
resolvedKey = ModuleLoader.Resolve(...).Key;   // inside try
...
(_builderKeys ??= ...)[resolvedKey] = specifier;   // outside
```

A host loader compiled without NRT that returns `new ResolvedSpecifier(request, key: null, null, SpecifierType.Bare)` instead of throwing makes the dictionary indexer throw `ArgumentNullException` out of whatever unrelated `Import` triggered the pass — precisely the cross-contamination the CA1031-suppressed catch was added to prevent. Guard the key or move the write inside.

**4. `Parse(location)` never uses `location` on two of its three branches.**

```csharp
internal Prepared<AstModule> Parse(string location)
{
    if (_module != null) return _module.Value;              // AddModule(prepared): location ignored
    ...
    if (_sourceRaw.Count <= 0)
    {
        return new Prepared<AstModule>(new AstModule(...));  // exports-only: location ignored
    }
```

The first one matters: `Add("lib", b => b.AddModule(Engine.PrepareModule("export { value } from './util.js';")))` with your own virtual-tree loader shape is found under its resolved identity, but its `Location` stays the `PrepareModule`-time source name (`"<anonymous>"` by default), so its nested `./util.js` resolves against the wrong base — the same bug class the PR fixes, unfixed on a public sibling path of the very function that was changed. Fix (re-key isn't possible for a host-prepared AST, so this one may be document-the-limitation) or at least say so in the `AddModule` doc. The exports-only branch is near-harmless — no source means no nested imports — but the `LoadFromBuilder` comment and the `Parse` doc state the naming unconditionally, and an exports-only module ends up with `Location == null`.

**5. Two XML-doc claims on `Add` are false as written.**

- *"A registration the loader refuses to resolve stays reachable only under the exact string it was registered with"* — it isn't reachable under that string either: `Import(thatString)` puts it through the same `Resolve` first, which throws the same refusal. With a deterministic loader the registration is reachable under no name at all. That's identical to pre-PR behaviour, so the code is fine — the sentence just needs to say "is not made *less* reachable" rather than promising reachability.
- *"the same specifier may be registered again afterwards"* — the re-`Add` succeeds but the builder can never load: `Load` consults the `_modules` cache under `(resolvedKey, attributes)` before any builder lookup and there's no eviction, so every import returns the cached first module forever, and a third `Add` of the name then throws. The test comment at `HostModuleBuilderIdentityTests.cs:129` ("The re-registration is reachable through the index too") asserts reachability the test never exercises.

## Advisory — noting, not blocking

- **The index pass is observable on the loader surface.** Registering three names and importing one now produces `Resolve` calls for the other two (verified on the head), interleaved inside the first raw-miss import. A loader that audit-logs `Resolve` as "script requested this" records phantom requests; one doing expensive per-resolve work adds N blocking calls to that import's latency. Inherent to the design and a fair trade — but worth one sentence in the `Add` doc, since it's a behaviour change on the embedder-facing `IModuleLoader` surface.
- **`_indexedBuildersVersion` is stamped before the loop**, so a loader whose `Resolve` re-enters the engine mid-pass early-outs on equal versions, misses a still-unindexed registration, and caches the loader-loaded module under its key — permanently shadowing the registration. Pathological host, fiddly to fix without double-resolving; fine to leave with a comment.
- **The raw fast path is technically a second identity.** `_builders.TryGetValue(key)` matches when an *import's* resolved key equals a registration string, without checking that the registration resolves to itself — so registration `"lib"` (→ `/vfs/lib.js`) is also consumable by an import that resolves to the literal key `"lib"`. Since this branch is also what keeps `Add` working without `EnableModules`, the fix is tempering the "exactly one identity" wording (raw name remains a match) rather than removing it.
- **`_indexedBuilders` and `_builderKeys` grow monotonically** — a pooled engine doing `Add($"virtual://{guid}", code)` + import per request accumulates entries with no bound. Trap for whoever fixes it: pruning the `_builderKeys` entry at consumption breaks the pinned re-`Add`-through-the-index behaviour, because the name stays in `_indexedBuilders` and is never re-resolved — cleanup has to drop both or neither.
- The parse-error-names-`_specifier` / runtime-stack-names-resolved-key split is a call you made deliberately and documented; fine by me, flagging only that log correlation by module name sees two identities for one module on the previously-broken path.
- `HostModuleBuilderIdentityTests.cs` lines 3–5 and 8: `System.Collections.Generic`/`System.IO` are covered by implicit usings and `Jint.Runtime` plus the `Module` alias are unreferenced.

On CI: agreed the red windows job is not yours — the three `intl402/DateTimeFormat` failures are runner ICU data, and the branch is green on the suites that matter including the host-contract leg.

lahma:
@gkurt no reply for a few days, so to keep this moving I used the maintainer-edit permission and pushed the round-two fixes to the branch myself — one commit on top of yours, plus a rebase onto current `main`, which had moved 49 commits ahead. Shout if you'd rather have done any of it differently; nothing here is design, it's all the small items from the last review.

Keyed to that review:

**1. Colliding registrations throw.** Two live registrations resolving to one key now throw `InvalidOperationException` from the import whose indexing pass discovers the pair, naming both registrations and the shared key. Both collision shapes are covered — two indexed names, and a name colliding with a registration filed under the canonical spelling itself. The `// Last registration wins` comment is gone along with the behaviour it overstated.

**2. The catch is narrower, and attempt-once is documented.** The indexing pass now catches `when (ex is not OperationCanceledException)`: a refusal is the loader rejecting one name, cancellation is the host calling off the whole operation, and the latter propagates. The `Add` doc says a failed resolution — deliberate refusal or transient failure — is attempted once and not retried.

**3. The null-key hole is closed.** A loader compiled without nullable annotations that returns a null `Key` instead of throwing is treated as the refusal it is, so the `_builderKeys` write can no longer throw `ArgumentNullException` out of an unrelated import.

**4. `Parse(location)`'s other branches.** The exports-only branch now stamps the resolved key as the module's location (`Location` is init-set via `WithSourceFile` on the synthesized AST), so no module ends up with `Location == null`. The `AddModule` branch genuinely cannot be re-keyed — the AST is the host's — so that limitation is now documented on `AddModule` together with the workaround (prepare under the key the loader resolves the registration to), the `Parse` doc and the `LoadFromBuilder` comment call out the exception, and a new test demonstrates the workaround shape actually works against the virtual-tree loader.

**5. The two false doc claims.** The refusal sentence now says the registration is made no *less* reachable — importing the name puts the same string through the same loader and surfaces the same refusal. The re-registration sentence now says the never-evicted cache keeps answering every import that resolves to that key, so a re-registered builder is only consulted by a request that misses the cache, i.e. one carrying different import attributes. The test comment in `ARegistrationIsConsumedUnderTheNameItWasFiledUnder` no longer asserts reachability the test doesn't exercise.

Advisory items I picked up while there: the `Add` doc mentions the resolve calls a loader observes from the indexing pass, the version-stamp-before-loop trade-off has a comment, and the unused usings in the test file are gone. The remaining advisory notes (raw fast path as a technical second identity, monotonic index growth, the parse-error/runtime name split) stay notes — none blocks.

Each behavioural fix is pinned by a new test in `HostModuleBuilderIdentityTests`, and I mutation-checked the pins the way you did: reverting the collision throw fails both collision tests, restoring the bare catch fails the cancellation test, restoring the pre-fix `ArgumentNullException` fails the null-key test, and nothing else moves.

Tests: `Jint.Tests.PublicInterface` fully green on net10.0 and net472, including under `JINT_HOST_CONTRACT_VERIFICATION=1`; full `Jint.Tests` green on both targets. From my side this is ready to land once CI agrees.

PR #2871, opened 2026-07-30
Let a module keep the whole url it was loaded from

gkurt:
`ModuleFactory` named every loaded module `resolved.Uri?.LocalPath`, which is the right shape for a `file:` url and lossy for every other one — `LocalPath` keeps only the path, so a module loaded from `http://localhost/lib/entry.js` knew itself as `/lib/entry.js`.

That name is not decoration. It is the `referencingModuleLocation` the module's own imports are resolved against, so a loader cannot resolve a relative specifier the way a browser does:

```js
// entry.js, loaded from http://localhost/lib/entry.js
export { value } from './dep.js';
// the loader is asked to resolve './dep.js' against '/lib/entry.js',
// which has no origin to resolve against at all
```

It is also the only name a module has to report through `import.meta.url`, where a bare path breaks anything doing `new URL(import.meta.url)`.

## The fix

```csharp
private static string? LocationOf(ResolvedSpecifier resolved)
{
    var uri = resolved.Uri;
    if (uri is null || !uri.IsAbsoluteUri) return resolved.Key;

    return uri.IsFile ? uri.LocalPath : uri.AbsoluteUri;
}
```

Applied at all five sites that named a module.

Three things came out of this that are worth a look independently of the main change:

- **A relative `Uri` threw.** `Uri.LocalPath` raises `InvalidOperationException` on a relative uri, so a loader returning one produced an exception instead of a module. The `IsAbsoluteUri` check covers that.
- **Four of the five sites could name a module `null`.** Only `BuildSourceTextModule` had the `?? resolved.Key` fallback; `BuildJsonModule`, `BuildBytesModule` and `BuildTextModule` did not. They now agree with it.
- **The doc comments were already stale.** They state that a null `Uri` gives a null location, while `BuildSourceTextModule` has been falling back to the key. Updated to describe the behaviour.

## Compatibility

Hosts loading modules from disk see no change — that is every host on `DefaultModuleLoader`, which resolves against a `file:` base path, so `IsFile` holds and the `LocalPath` is preserved. `LoadedFileModuleKnowsItselfByPath` is in the suite as the guard for exactly that, and passes with and without the change.

test262 is unaffected for a different reason: `Test262ModuleLoader.Resolve` returns no uri at all, so its modules were already named by their key.

## Tests

Three, using a loader that serves modules from a dictionary keyed by absolute url. The two http ones fail on `main`; the `file:` one is the compatibility guard described above.

They install a small `ImportMetaUrlHost`, because Jint leaves `import.meta` to the host and `Module.Location` is otherwise not observable from script. If you would rather assert on the location without going through a host, the alternative is to record the `referencingModuleLocation` the loader is handed — happy to switch.

`Jint.Tests` 4661 passed; test262 99484 passed, 0 failed, unchanged from `main`. The two failures on my machine are `ToStringUsesDaylightNameWhenInDst` and `DateToStringMethodsShouldUseCurrentTimeZoneAndCulture`, which assert OS-localised time zone display names and fail here regardless of this change.


Jint/Runtime/Modules/ModuleFactory.cs
```diff
@@ -8,20 +8,35 @@ namespace Jint.Runtime.Modules;
...
+    /// as <c>/lib/a.js</c> has no origin left for a relative import of its own to resolve against, nor
+    /// one to report through <c>import.meta.url</c>. A relative or absent uri leaves the resolved key,
+    /// which is the only name such a module has.
+    /// </summary>
+    private static string? LocationOf(ResolvedSpecifier resolved)
+    {
+        var uri = resolved.Uri;
+        if (uri is null || !uri.IsAbsoluteUri) return resolved.Key;
+
+        return uri.IsFile ? uri.LocalPath : uri.AbsoluteUri;
```
lahma: **`AbsoluteUri` is not the url the host gave you — and `Location` is fed back into `Resolve`.**

`Uri.AbsoluteUri` canonicalizes: default port stripped, host lowercased, dot segments collapsed, characters percent-encoded. `ResolvedSpecifier.Key` doesn't — it's the string the host chose, and it's already the identity the engine keys the module map on (`ModuleCacheKey.From` uses `resolved.Key`, and `Engine.Modules.cs:87` looks up builders by it). After this change the two can differ, and `Location` is exactly what comes back as `referencingModuleLocation` on the next `Resolve`.

Reproduced on this branch, with a loader in the shape of the existing `TestModuleLoader` (`ModuleTests.cs:917`) — `Key` = the url the host chose, `Uri` = that url parsed:

```
registry: http://localhost:80/lib/entry.js   (export { value } from './dep.js';)
          http://localhost:80/lib/dep.js

import.meta.url  ->  http://localhost/lib/entry.js     // :80 gone
resolve './dep.js' against that
          ->  http://localhost/lib/dep.js             // misses the registry
JavaScriptException: Could not load module ./dep.js
```

And with a space in the path: `Key='http://localhost/lib/my report.js'` → `Location='http://localhost/lib/my%20report.js'`.

`resolved.Key` has none of this: it round-trips by construction, it's already non-null, and using it would make the `?? resolved.Key` fallback the single rule instead of a special case:

```csharp
return uri.IsFile ? uri.LocalPath : resolved.Key;
```

Two more reasons to prefer it:

- **`AbsoluteUri` is target-framework-dependent.** Jint ships `net462;netstandard2.0;netstandard2.1;net8.0;net10.0` from this source, and the .NET Framework and .NET Core `Uri` parsers differ on unescaping and IRI normalization. For `http://host/lib/a%2Fb.js` or `http://host/lib/ä.js` the same embedder code would get a different `referencingModuleLocation` on net462 than on net10.0. `resolved.Key` is a host string and never had that property.
- **A path-treating loader breaks silently.** A shipped loader doing `Path.GetDirectoryName(referencingModuleLocation)` over `mem://pkg/entry.js` used to receive `/pkg/entry.js` and get `/pkg`; now it receives the full url and gets `mem:` (or `null` off Windows). No signature change, no compile error — it surfaces when the embedder's app runs.

On the PR description's compatibility argument ("every host on `DefaultModuleLoader` resolves against a `file:` base path"): nothing enforces that. `DefaultModuleLoader`'s constructor takes any absolute uri as `basePath` with no scheme check (`DefaultModuleLoader.cs:17-30`) and `Resolve` adopts any absolute specifier uri (`:52-54`), so `new DefaultModuleLoader("https://cdn.local/lib/")` is legal today and lands in this branch. Either the claim wants narrowing to "a `file:` base path", or the loader wants a scheme check.

lahma: **The `file:` branch keeps the reduction the doc comment above calls the bug.**

The `<summary>` you added says it exactly: "a module that knows itself as `/lib/a.js` has no origin left for a relative import of its own to resolve against". That is still true for `file:` after this change — the branch preserves `LocalPath` unchanged, so the one scheme the fix doesn't reach is the one most hosts are on.

It bites the store documented in `Jint.Tests.PublicInterface/ModuleLoaderTests.cs`, which uses `file:///./` as its namespace protocol. A module at `file:///lib/a.js` gets `Location = "/lib/a.js"` (`new Uri("file:///lib/a.js").LocalPath == "/lib/a.js"`, and `Uri.TryCreate("/lib/a.js", UriKind.Absolute, out _)` is `false`), so its `import './b.js'` fails the store's absolute-referrer check and falls back to `file:///./` + `./b.js` = `file:///b.js` — the wrong sibling, or `Module not found`. That is precisely what `LoadedModuleResolvesARelativeImportAgainstItsOwnUrl` fixes for `http:`.

I'm not saying flip it — reporting a filesystem path for a filesystem module is a real compatibility property and `LoadedFileModuleKnowsItselfByPath` is right to guard it. But the PR then only half-fixes the stated problem, and the test pins the other half in place. Worth saying so out loud in the doc comment ("`file:` keeps the path form for compatibility, at the cost of X") so the next person doesn't read the summary as describing the whole method.

lahma: **This makes `Location` script-readable in a way `LocalPath` wasn't.**

`LocalPath` discarded scheme, host, userinfo and query. `AbsoluteUri` keeps all of them — and this string becomes `SourceLocation.SourceFile`, which `JintCallStack.BuildCallStackString` (`JintCallStack.cs:176`) appends verbatim into `error.stack`, and which `BuildJsonModule` interpolates into a script-catchable `SyntaxError`.

So for a host loading from an authenticated internal endpoint:

```csharp
new Uri("https://svc:s3cr3t@internal.corp/lib/a.js?apikey=ABCDEF")
```

before, untrusted script calling `new Error().stack` saw `at /lib/a.js:3:5`; after, it sees the credential and the api key. A malformed JSON module reaches the same string through `catch (e) { e.message }`.

This is a consequence of the change rather than an argument against it — a browser's `import.meta.url` carries the query too. But it's a new disclosure path with no redaction and no way for a host to opt out short of not returning a `Uri` at all, so it deserves a line in the PR description at minimum. Using `resolved.Key` doesn't fix it either (a host that put a secret in the url put it in the key too) — it just stops Jint from being the thing that introduced it.

Jint/Runtime/Modules/ModuleFactory.cs
```diff
@@ -8,20 +8,35 @@ namespace Jint.Runtime.Modules;
...
+    /// The name a module knows itself by. A <c>file:</c> url reads better as a filesystem path, but
+    /// reducing any other url to one drops its scheme, host and query - and a module that knows itself
+    /// as <c>/lib/a.js</c> has no origin left for a relative import of its own to resolve against, nor
+    /// one to report through <c>import.meta.url</c>. A relative or absent uri leaves the resolved key,
+    /// which is the only name such a module has.
+    /// </summary>
+    private static string? LocationOf(ResolvedSpecifier resolved)
+    {
+        var uri = resolved.Uri;
+        if (uri is null || !uri.IsAbsoluteUri) return resolved.Key;
```
lahma: This guard is a genuine improvement — `resolved.Uri?.LocalPath` throws `InvalidOperationException: This operation is not supported for a relative URI` for a relative `Uri`, and `?.` only guarded null. Worth knowing what it trades, though.

A loader that mistakenly returns `new Uri("lib/dep.js", UriKind.Relative)` used to fail loudly at module construction, naming the mistake. Now the module is silently built with `Location = resolved.Key` (`"dep"`), and its own `import './util.js'` resolves against the bare specifier instead of `lib/` — so the host either loads a different file with the same name or gets a `Module Not Found` from far away from the actual defect. Nothing else validates `ResolvedSpecifier.Uri`, so the check isn't re-established anywhere.

Probably still the right trade (a relative `Uri` genuinely has no absolute name, and the key is the best available), but consider whether a relative `Uri` deserves to keep throwing — it's unambiguously a host bug rather than a configuration the engine should paper over.

Jint/Runtime/Modules/ModuleFactory.cs
```diff
@@ -8,20 +8,35 @@ namespace Jint.Runtime.Modules;
...
+    /// The returned modules location (see <see cref="Module.Location"/>) is the whole url of
+    /// <see cref="ResolvedSpecifier.Uri"/>, or its <see cref="Uri.LocalPath"/> for a <c>file:</c> url,
+    /// falling back to <see cref="ResolvedSpecifier.Key"/> when there is no absolute uri.
     /// </remarks>
     /// <exception cref="ParseErrorException">Is thrown if the provided <paramref name="code"/> can not be parsed.</exception>
     /// <exception cref="JavaScriptException">Is thrown if an error occured when parsing <paramref name="code"/>.</exception>
     public static Module BuildSourceTextModule(Engine engine, ResolvedSpecifier resolved, string code, ModuleParsingOptions? parsingOptions = null)
     {
-        var source = resolved.Uri?.LocalPath ?? resolved.Key;
+        var source = LocationOf(resolved);
```
lahma, line 68: Worth calling out in the PR description: this `source` is not only `import.meta.url`. It flows into `ParseModuleGuarded` → `Program.Location.SourceFile`, which is the module's identity for **the debugger and stack traces**.

`BreakPointCollection` keys on `BreakLocation.Source` (`Jint/Runtime/Debugger/BreakLocation.cs:26`), which is this string. A host driving the debugger over `http://cdn.local/lib/entry.js` registers `new BreakPoint("/lib/entry.js", 3, 0)` today; after this change the engine names that module `http://cdn.local/lib/entry.js`, the breakpoint silently never hits, and the debugger just looks dead. The same string change alters every `error.stack` / `JavaScriptException.ToString()` frame, so host log-parsing and assertions on frame text move with it.

No test covers a non-`file:` scheme under the debugger, so nothing here would catch a mistake in either direction.

Jint/Runtime/Modules/ModuleFactory.cs
```diff
@@ -53,14 +68,14 @@ public static Module BuildSourceTextModule(Engine engine, in Prepared<AstModule>
...
-    /// <see cref="ResolvedSpecifier.Uri"/> is null, the modules location source will be null as well.
+    /// The returned modules location (see <see cref="Module.Location"/>) is the whole url of
+    /// <see cref="ResolvedSpecifier.Uri"/>, or its <see cref="Uri.LocalPath"/> for a <c>file:</c> url,
+    /// falling back to <see cref="ResolvedSpecifier.Key"/> when there is no absolute uri.
     /// </remarks>
     /// <exception cref="JavaScriptException">Is thrown if an error occured when parsing <paramref name="jsonString"/>.</exception>
     public static Module BuildJsonModule(Engine engine, ResolvedSpecifier resolved, string jsonString)
     {
-        var source = resolved.Uri?.LocalPath;
+        var source = LocationOf(resolved);
```
lahma, line 107: **This changes a documented contract for uri-less loaders, and for three factories rather than one.**

On main these were `resolved.Uri?.LocalPath` with no `?? resolved.Key` — so for a loader returning `Uri: null`, `Module.Location` was `null`, and the `<remarks>` you're replacing promised exactly that: *"If `ResolvedSpecifier.Uri` is null, the modules location source will be null as well."* Now it's the bare key.

That's the shape of `Jint.Tests.Test262/Test262ModuleLoader.Resolve` and of `ModuleScript` in `Jint.Tests.PublicInterface/ModuleLoaderTests.cs`, so it's not hypothetical. A host with `if (module.Location != null) File.ReadAllText(module.Location)` — reasonable code against that promise — now enters the branch with `"config.json"` and throws `FileNotFoundException` where it used to take the null fallback.

It may well be the better behaviour (a name is more useful than no name). But it's a behaviour change for every non-source-text module type, it deletes a stated guarantee, and it isn't in the PR description or the new remarks. `BuildBytesModule` (`:120`) and `BuildTextModule` (`:132`) changed the same way and say nothing about it — their `<remarks>` are spec links, so neither documented the old behaviour nor documents the new one. Worth either documenting all four consistently or scoping the PR to `BuildSourceTextModule`.

Jint.Tests/Runtime/ModuleTests.cs
```diff
@@ -693,6 +693,107 @@ public void ModuleBuilderPassesReferencingModuleLocationToModuleLoader()
...
+        var ns = engine.Modules.Import("http://localhost/lib/entry.js");
+
+        ns.Get("value").AsString().Should().Be("from the sibling");
+    }
+
+    [Fact]
+    public void LoadedFileModuleKnowsItselfByPath()
+    {
+        // A file url is still reported as a filesystem path, which is what every existing host sees.
+        var path = Path.Combine(Path.GetTempPath(), "jint-location-test");
```
lahma: **This directory name is machine-wide and the `finally` recursive-deletes it.**

```csharp
var path = Path.Combine(Path.GetTempPath(), "jint-location-test");
```

Reproduced: three concurrent `dotnet test --filter "...LoadedFileModuleKnowsItselfByPath"` runs on this branch — the first passed, the other two failed with `System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\Users\...\Temp\jint-location-test'`, because the first run's `finally` deleted the shared directory out from under them. A single multi-TFM run is safe (VSTest runs net10.0 and net472 sequentially); the exposure is two overlapping invocations — a re-run overlapping a cancelled one, or a shared/self-hosted runner.

Separately, `Directory.CreateDirectory` succeeds silently on an existing directory, so on a machine that already has `%TEMP%\jint-location-test` for any reason, running the suite recursively deletes it.

One line:

```csharp
var path = Path.Combine(Path.GetTempPath(), "jint-location-test-" + Guid.NewGuid().ToString("N"));
```

Minor, same block: `File.WriteAllText` on line 777 is outside the `try`, so if it throws the directory leaks. Moving `Directory.CreateDirectory` + `File.WriteAllText` inside the `try` closes that.

Jint.Tests/Runtime/ModuleTests.cs
```diff
@@ -693,6 +693,107 @@ public void ModuleBuilderPassesReferencingModuleLocationToModuleLoader()
...
+            _modules = modules;
+        }
+
+        public override ResolvedSpecifier Resolve(string referencingModuleLocation, ModuleRequest moduleRequest)
+        {
+            var uri = Uri.TryCreate(referencingModuleLocation, UriKind.Absolute, out var referrer)
+                ? new Uri(referrer, moduleRequest.Specifier)
+                : new Uri(moduleRequest.Specifier, UriKind.Absolute);
+
+            return new ResolvedSpecifier(moduleRequest, uri.AbsoluteUri, uri, SpecifierType.RelativeOrAbsolute);
```
lahma: **These tests can't distinguish the policy the PR chose from the one it didn't.**

`Key` is set to `uri.AbsoluteUri` here — byte-for-byte the string `LocationOf` produces. So replace the whole non-file branch with `return resolved.Key;` (never consult the `Uri` at all) and both http tests still pass; `LoadedFileModuleKnowsItselfByPath` is unaffected because it takes the `IsFile` branch. The single contentious decision in the PR is unpinned by its own coverage.

Those two implementations differ for every host whose `Key` isn't the canonicalized form — including both in-repo url loaders (`ModuleTests.cs:595` and `Jint.Tests.PublicInterface/ModuleLoaderTests.cs:109`, which key on `uri.ToString()`).

Setting `Key` to something deliberately *not* equal to `AbsoluteUri` is what would pin the intent, whichever way you decide it:

```csharp
return new ResolvedSpecifier(moduleRequest, uri.ToString(), uri, SpecifierType.RelativeOrAbsolute);
```

With a `http://localhost:80/...` entry in the dictionary, that test states the answer to the `AbsoluteUri`-vs-`Key` question in one assertion.

Jint.Tests/Runtime/ModuleTests.cs
```diff
@@ -693,6 +693,107 @@ public void ModuleBuilderPassesReferencingModuleLocationToModuleLoader()
...
+        private readonly IReadOnlyDictionary<string, string> _modules;
+
+        public UrlModuleLoader(IReadOnlyDictionary<string, string> modules)
+        {
+            _modules = modules;
+        }
+
+        public override ResolvedSpecifier Resolve(string referencingModuleLocation, ModuleRequest moduleRequest)
+        {
+            var uri = Uri.TryCreate(referencingModuleLocation, UriKind.Absolute, out var referrer)
```
lahma: When this test's own regression fires, the failure won't read as one.

If `Location` stops being an absolute uri, `Uri.TryCreate` returns false and the fallback runs `new Uri("./dep.js", UriKind.Absolute)`, which throws `UriFormatException: Invalid URI: The format of the URI could not be determined`. `ModuleLoader.LoadModule` wraps only `LoadModuleContents`, not `Resolve`, so that escapes `Engine.Modules.Import` unconverted and CI reports an opaque URI-parsing crash that mentions neither the module, the referrer, nor the contract being guarded — it reads like a bug in the test helper.

Throwing something that names the invariant would make the regression self-describing:

```csharp
if (!Uri.TryCreate(referencingModuleLocation, UriKind.Absolute, out var referrer))
{
    throw new InvalidOperationException(
        $"referrer is not an absolute url: '{referencingModuleLocation}'");
}
var uri = new Uri(referrer, moduleRequest.Specifier);
```

(The entry-point call passes the specifier itself as the referrer, so that path needs to stay reachable — happy to be told this is over-engineering for a test helper.)

Jint.Tests/Runtime/ModuleTests.cs
```diff
@@ -693,6 +693,107 @@ public void ModuleBuilderPassesReferencingModuleLocationToModuleLoader()
...
+            }));
+            o.UseHostFactory(_ => new ImportMetaUrlHost());
+        });
+
+        var ns = engine.Modules.Import("http://localhost/lib/entry.js");
+
+        ns.Get("url").AsString().Should().Be("http://localhost/lib/entry.js");
+    }
+
+    [Fact]
```
lahma: All three new tests exercise `BuildSourceTextModule` with an absolute uri — but four of the five call sites changed, and the two branches that actually changed behaviour for them have no coverage:

- **the null/relative-uri fallback.** `BuildJsonModule`/`BuildBytesModule`/`BuildTextModule` used to call `resolved.Uri?.LocalPath`, which throws `InvalidOperationException` for a relative `Uri` and returns `null` for no `Uri`. The new guard fixes both — and a later refactor that inlines `resolved.Uri?.LocalPath` back, or drops the `!uri.IsAbsoluteUri` check, ships green through the whole suite. test262's loader always returns `Uri: null`, and every in-repo custom-loader test either overrides `LoadModule` or goes through the `Prepared<AstModule>` overload, so nothing reaches it.
- **the json/bytes/text factories themselves** — see the comment on `ModuleFactory.cs:78`.

Two small tests close it: one loader returning `Uri: null` asserting `Location == Key` through a json/text/bytes import, and one returning a relative `Uri` asserting it doesn't throw.

Jint.Tests/Runtime/ModuleTests.cs
```diff
@@ -693,6 +693,107 @@ public void ModuleBuilderPassesReferencingModuleLocationToModuleLoader()
             s => s.Should().Be("entry_point_module"));
     }
 
+    /// <summary>
```
lahma: Repo convention (`AGENTS.md`, *"Where integrator-facing tests belong"*): these belong in `Jint.Tests.PublicInterface`.

> `Jint.Tests.PublicInterface` is the only test project **without** `InternalsVisibleTo` […] so a test there actually proves the surface is reachable by a third party. Put new integrator-facing tests there, in **generically named files** describing the capability

`UrlModuleLoader`, `ImportMetaUrlHost` and all three facts use only public API — `ModuleLoader`, `ResolvedSpecifier`, `Jint.Runtime.Host.GetImportMetaProperties`, `Options.UseHostFactory<T>`, `Engine.Modules.Import` — and "the name a module knows itself by" is exactly the embedder contract that section governs. `Jint.Tests` *does* have `InternalsVisibleTo`, so these would keep compiling if any of those members later stopped being reachable from outside the assembly, and the regression that matters most here would ship unnoticed.

lahma (commented):
Thanks for this — the premise is right, and the second test states it better than any prose could: a module reduced to `/lib/entry.js` has no origin left for `./dep.js` to resolve against, and `import.meta.url` had nothing useful to report. That's a real gap and worth closing.

What makes it delicate is that `Module.Location` is three contracts wearing one field:

1. `import.meta.url` — what this PR is about.
2. The parser's `SourceFile`, which becomes `BreakLocation.Source` for the debugger and the frame text in `error.stack`.
3. The `referencingModuleLocation` handed back into every third-party `IModuleLoader.Resolve` (`Host.cs:128`/`:142` → `Engine.Modules.cs:79`).

One line moves all three at once, for every non-`file:` scheme, with no opt-out — so I went through it fairly carefully. Comments inline; the three things I'd most like your read on:

**1. Why `uri.AbsoluteUri` rather than `resolved.Key`?** This is the one decision in the PR I can't derive from the diff, and it's the one the tests can't settle either. `AbsoluteUri` is a *re-derivation* that canonicalizes; `Key` is the string the host actually chose and the identity the engine already keys the module map on. Reproduced on this branch with a loader in the shape of the existing `TestModuleLoader`: `Key = "http://localhost:80/lib/entry.js"` gives `Location = "http://localhost/lib/entry.js"` — the `:80` is gone, so the host's own registry lookup for `./dep.js` misses and the import throws. A space in the path percent-encodes the same way.

**2. The `file:` branch keeps the exact reduction the doc comment calls the bug.** A `file:`-scheme module still can't resolve its own relative imports, and the new test now pins that in place.

**3. `BuildJsonModule`/`BuildBytesModule`/`BuildTextModule` changed too**, from `null` to the bare key for a uri-less loader — deleting a promise the removed `<remarks>` made in words, and two of the three aren't mentioned anywhere.

One thing outside the diff hunks so I couldn't anchor it: `ModuleFactory.cs:84`'s `catch (Exception)` around `new JsonParser(engine).Parse(...)` converts *everything* — `OutOfMemoryException`, any `JintException` raised beneath it — into a script-catchable `SyntaxError`, so a guest `try { await import('./x.json', ...) } catch {}` can swallow a host-boundary failure and the embedder sees a successful evaluation. Pre-existing on main, and you're rewriting the block anyway, so it may be worth narrowing to the parser's own error type while you're here — or leaving it for a separate PR, your call.

One concrete bug independent of all the above: the new file-backed test uses a fixed `%TEMP%\jint-location-test` and recursive-deletes it in `finally`. I reproduced two overlapping suite runs destroying each other (`DirectoryNotFoundException`), and it will also delete a pre-existing directory of that name. One-line fix, noted inline.

Leaving this as a comment rather than a block — most of it is one design question with a couple of concrete fixes hanging off it, and you may well have a reason for `AbsoluteUri` that I'm not seeing.

gkurt:
Thanks — this changed the design, not just the tests. All eleven points are addressed; pushed as an amend.

## The one that mattered: `AbsoluteUri` → `resolved.Key`

You convinced me. `LocationOf` is now the key, with one exception:

```csharp
private static string LocationOf(ResolvedSpecifier resolved)
{
    var uri = resolved.Uri;
    if (uri is not null && uri.IsAbsoluteUri && uri.IsFile)
    {
        return uri.LocalPath;
    }

    return resolved.Key;
}
```

Each of your three reasons stands on its own, and the third — a target-framework-dependent `referencingModuleLocation` from one build of this source — is the one I had no answer to. The round-trip argument is the deeper one though: the key is already the identity `ModuleCacheKey.From` and the builder lookup use, so with `AbsoluteUri` a module's own name and the name it is cached under can drift, and your `:80` reproduction is that drift finding a sibling that is right there and failing to load it. Using the key also collapses the fallback into the single rule instead of a special case, and leaves the uri answering exactly one question.

## The `file:` half-fix

Fair, and it is now said out loud rather than left to be inferred from a `<summary>` that describes the other branch. The doc comment has a paragraph saying the exception *keeps* the reduction, that a `file:` module therefore still has to reconstruct its origin from its loader's base path, and that the trade is not renaming every module every host already loads from disk. I also took your note on the compatibility claim: the PR description no longer says every `DefaultModuleLoader` host is on a `file:` base path, because as you point out nothing enforces that — the constructor takes any absolute uri with no scheme check, so `new DefaultModuleLoader("https://cdn.local/lib/")` is legal and lands in the other branch.

## The disclosure path

In the commit message now, in your framing: a consequence of the fix rather than an argument against it, a browser's `import.meta.url` carries the query too, and switching to the key does not fix it — a host that put a secret in the url put it in the key as well. What it does change is that Jint is no longer the thing that introduced it. Recorded with the `error.stack` and `catch (e) { e.message }` routes named.

## The relative-`Uri` guard

I went with the fallback rather than restoring a throw, and wrote the reasoning down instead of leaving it implicit. Once the location is the key, the uri is only ever asked *is this a file?*, and a relative uri answers that no better than an absent one — so the two behaving alike is the consistent rule, and the module that comes out is the same shape a null-uri loader has always produced. The diagnostic you describe is genuinely given up, and the doc comment says so, including that the mistake now surfaces far from its cause. Happy to flip it if you'd rather it stayed loud; it is a two-line change either way.

## The debugger and stack-trace consequences

Both in the commit message, and both now have tests, since you noted nothing covered a non-`file:` scheme in either direction:

- `ADebuggerCanBreakInAModuleLoadedFromAUrl` sets `new BreakPoint("http://localhost/lib/entry.js", 2, 0)` and asserts it hits with that `SourceFile`.
- `LoadedModuleNamesItselfByItsKeyInAStackTrace` asserts the url reaches `JavaScriptStackTrace`.

## The uri-less contract, across four factories not one

Documented on all four identically, including that the location is never null and what it was before, and called out in the commit message with the `if (module.Location != null) File.ReadAllText(...)` case named as the deliberate cost. Worth adding: `Module.Location` is declared `string`, not `string?`, so the old null also disagreed with its own signature.

## Tests

Moved to `Jint.Tests.PublicInterface/HostModuleLocationTests.cs`, per the AGENTS.md section you quoted — you were right that `InternalsVisibleTo` is exactly what makes the difference here.

Your point that the tests couldn't distinguish the two policies was the sharpest one, and it is fixed the way you suggested: the loader keys an absolute specifier **verbatim** rather than through `Uri`, so the registry holds `http://localhost:80/lib/entry.js` and `http://localhost/lib/my report.js` — two keys `AbsoluteUri` would rewrite, in both of the ways you listed. Reversing the decision now fails a test instead of passing both. I checked that directly: with `LocationOf` put back to `AbsoluteUri`, `LoadedModuleKnowsItselfByTheNameItsLoaderChose` is the only failure; with it put back to `resolved.Uri?.LocalPath`, five of the eight fail.

The rest:

- The loader throws a message naming the invariant when a referrer is not an absolute url, so a regression reads as the contract it broke rather than as a `UriFormatException` escaping `Resolve` unwrapped. Not over-engineering — it is two lines and it is the difference between a CI failure that explains itself and one that looks like a broken helper.
- The temp directory carries a guid, and `CreateDirectory` + `WriteAllText` moved inside the `try`. Thank you for actually running three concurrent invocations to prove it.
- Null-uri and relative-uri fallbacks are covered for all four factories, via a `[Theory]` over both unusable-uri shapes calling `BuildSourceTextModule` / `BuildJsonModule` / `BuildTextModule` / `BuildBytesModule` directly and asserting `Location == Key`. Plus a `file:` case through the three synthetic factories, so the exception isn't only pinned for source text.

lahma (approved):
This changed the design rather than the tests, and it's better for it. `LocationOf` returning `resolved.Key` makes the uri answer exactly one question, collapses the fallback into the single rule, and removes the target-framework dependence entirely. Thanks for taking the round-trip argument seriously rather than papering the reproduction.

I verified the two claims that carry the weight, rather than taking them on trust — mutating `LocationOf` in a scratch worktree against `HostModuleLocationTests`:

| `LocationOf` reverted to | Failures |
| --- | --- |
| `uri.AbsoluteUri` (the first revision) | **1** — `LoadedModuleKnowsItselfByTheNameItsLoaderChose` |
| `resolved.Uri?.LocalPath` (main) | **5** of 8 |

Both match your numbers exactly. Keying the loader's absolute specifiers *verbatim* is what earns that: the registry holds `http://localhost:80/lib/entry.js` and `http://localhost/lib/my report.js`, two strings `AbsoluteUri` rewrites in the two different ways, so reversing the decision now fails a test instead of passing every one. That was the point I was least sure would land cleanly, and it is the one that came out strongest.

The `[Theory]` over both unusable-uri shapes across all four factories is more than I asked for and covers the branch that had no test at all. The debugger and stack-trace tests close the other gap — nothing covered a non-`file:` scheme in either direction before.

Also checked before merging, since CI here ran against a merge base that is now 45 commits behind `main`: merged into current `main` locally, build clean with zero warnings, `Jint.Tests.PublicInterface` 1306 passed and `Jint.Tests` 4767 passed. No semantic conflict with the module and public-interface work that landed in between.

One non-blocking nit, for a follow-up or for whoever reads it next. The new `<remarks>` justifies the key on the grounds that it is

> already the identity the module map and the module-builder registry are keyed on, so a module's own name and the name it is cached under cannot drift apart.

That holds for the key branch, but the `file:` exception gives it up — and that is the branch every `DefaultModuleLoader` host takes, since `Resolve` sets `Key = resolved.AbsoluteUri` while `LocationOf` returns `LocalPath`:

```
KEY      = [file:///C:/Users/.../Temp/jm-probe-.../entry.js]
LOCATION = [C:\Users\...\Temp\jm-probe-...\entry.js]
EQUAL    = False
```

Nothing breaks — that is how `main` behaves too and the loader copes with a path referrer — but the sentence reads as an invariant of the method rather than of one of its two branches, and the `file:` paragraph right below it is the natural place to say the compatibility trade also costs the no-drift property. Happy to send that one-line follow-up myself.

PR #2872, opened 2026-07-30
Let a host load modules asynchronously, as ECMA-262 already specifies

gkurt:
`IModuleLoader.LoadModule` must hand back a parsed module synchronously, so a host that fetches module source over I/O — HTTP, a dev server, an asset pipeline — can only comply by blocking the calling thread, which is not viable on a game loop or a UI thread. ECMA-262 has allowed better since ES2023: HostLoadImportedModule may complete whenever it likes and call FinishLoadingImportedModule later.

This implements the spec's load phase as written: `Module.LoadRequestedModules` returns a promise that settles when the transitive module graph is present (InnerModuleLoading / GraphLoadingState / ContinueModuleLoading, with `ModuleStatus.New` as the state a module starts in), and `GetImportedModule` becomes a lookup over each record's [[LoadedModules]] rather than a load.

The host surface is opt-in — a loader that only implements `IModuleLoader` is driven exactly as before:

- **`IAsyncModuleLoader` + `ModuleLoadCompletion`**, settled from any thread. A settle only records the outcome and queues a job; the module is built, registered and handed to the waiting importers on an engine turn. A settle arriving after `RestoreGlobalSnapshot` is discarded rather than written into the restored engine. The one exception is a settle that happens before `LoadModuleAsync` has returned — the engine's own stack is waiting for exactly that answer, so the load finishes on it directly, and a loader whose answers are already at hand (a warm cache) keeps the blocking `Import` fully synchronous, exactly like a synchronous `IModuleLoader`.
- **`AsyncModuleLoader`**, the template class — override `LoadModuleContentsAsync`; returning an already-completed task takes the synchronous path above.
- **`Engine.Modules.ImportAsync`** to await, and `StartImport` → `ModuleImportOperation` to drive from a game loop with `ProcessTasks()`. The synchronous `Import` is unchanged for synchronous loaders.

Concurrent loads of one resolved specifier coalesce, so the ordinary diamond is one fetch rather than two.

Which failures become rejections is deliberately asymmetric. For a synchronous loader nothing changed: a `JavaScriptException` is a throw completion, everything else still propagates on the caller's thread. For an asynchronous loader every failure becomes a rejection, because there is no caller's thread left to throw on. Constraint exceptions (cancellation, timeout, memory/statement budgets) are never turned into rejections — a constraint that becomes a rejection no longer bounds anything.

test262 cannot cover this layer: its host contract still specifies the pre-ES2023 synchronous HostResolveImportedModule, and no async hook name appears in any test or harness file. As in other engines, it belongs to the embedder-facing suite — `AsyncModuleLoaderTests` in Jint.Tests.PublicInterface (the project without InternalsVisibleTo, proving the surface a real integrator can reach) and `ModuleLoadPhaseTests` in Jint.Tests.


lahma (changes requested):
Thanks for this — the core design (completion-based load pipeline, event-loop integration, generation checks on settle) is solid, and the test coverage for the happy paths is genuinely good. I ran a deep multi-agent review over the diff and it confirmed a number of correctness gaps, almost all in exception routing and realm capture, several of which end in a permanently stranded import promise. Requesting changes for the first section; the second section is non-blocking food for thought.

## Blocking

**1. Async-loaded ShadowRealm modules are built and evaluated in the principal realm** — `ModuleLoadCompletion.cs` (~185–190)
The deferred `Build` calls `ModuleFactory.BuildFromContents(Engine, …)`, which reads ambient `Engine.Realm` at dequeue time. `ShadowRealmImportValue` enters the shadow realm's context only around the synchronous `LoadImportedModule` call, so when the settle arrives on a later event-loop turn the module is constructed against the principal realm and its top-level code evaluates against the main realm's globals — `globalThis.leak = secret` inside the sandboxed module writes the host's principal globals. With a synchronous loader the module is built inline inside the entered shadow context, so this asymmetry is new to the async path. `SetError`'s `CreateError` has the same ambient-realm capture (error objects minted from the wrong realm's intrinsics). The realm needs to be captured when the load starts and used by the deferred build/error paths.

**2. `Resolve` failures that aren't `JavaScriptException` strand the load** — `Engine.Modules.cs` (~166–172)
The catch around `ModuleLoader.Resolve` handles only `JavaScriptException`, but `ModuleResolutionException` (`DefaultModuleLoader`'s documented failure type; the README's sample loader can also throw `UriFormatException`) is a sibling `JintException`. Thrown while resolving a static import discovered inside a queued `ModuleLoadCompletion` job, it erupts out of `ProcessTasks`, the graph state's pending count is never decremented, and the import promise never settles — the exact stranding class the builder-branch `catch … when IsRunningJob` at ~196 guards against, left open for `Resolve`. `ASyntaxErrorInARegisteredModuleReachedThroughAnAsyncGraphRejectsInsteadOfStrandingTheLoad` pins this for the builder branch; nothing covers `Resolve`.

**3. The `LoadModuleAsync` catch-all flattens engine constraint exceptions into rejections** — `Engine.Modules.cs` (~224)
`catch (Exception ex) when (!completion.IsCompleted)` converts `ExecutionCanceledException` / `TimeoutException` / an `OperationCanceledException` from the engine's own constraint token into an ordinary script-visible rejection: `RunLoadPhaseBlocking` then throws `JavaScriptException("The operation was canceled")` instead of `ExecutionCanceledException`, so a host's `catch (ExecutionCanceledException)` shutdown handling never runs — while the same cancelled engine with a synchronous loader surfaces `ExecutionCanceledException`. This contradicts the invariant documented elsewhere in the PR that constraint exceptions are never flattened into rejections; the filter should rethrow constraint/OOM exception types.

**4. One waiter's exception in `FinishWaiters` strands the remaining waiters** — `ModuleLoadCompletion.cs` (~230)
Waiters are dispatched in a plain loop with no per-waiter isolation. Waiter *i*'s `FinishLoadingImportedModule → Continue` can run Link/Evaluate or further `LoadImportedModule` calls, and any escaping exception (a constraint exception, or the `ModuleResolutionException` from item 2) unwinds before waiter *i+1* is served. The pending-load entry is already removed and the module registered, so nothing ever finishes the rest: their import promises hang silently even if the host catches the eruption and keeps pumping.

**5. Non-JS failures in the queued build/register path strand all waiters** — `ModuleLoadCompletion.cs` (~206–216)
`Build`'s generic `catch { RemovePendingLoad(…); throw; }` rethrows without finishing the waiters, and the `Debugger.OnBeforeEvaluate` host callback in `RegisterModule` sits outside the try entirely. A throwing `GetModuleSource` override or debugger callback erupts out of `ProcessTasks` with every waiter permanently pending, possibly leaving a half-registered module in `_modules`.

**6. A `Modules.Add` racing an in-flight async load of the same key produces two live module records** — `ModuleLoadCompletion.cs` (~216) + `Engine.Modules.cs` (~182)
`LoadImportedModule` consults builders before the pending-load table, so a mid-flight importer can take the builder branch and register the builder module; when the fetch settles, `Build` unconditionally overwrites the registry entry and hands its waiters the fetched module. Two module records for one key, separate top-level state, side effects run twice — and the `HostLoadImportedModule` consistency guard never fires because the referrers differ.

**7. `Import()`/`Link()` failures lose the original error location** — `Engine.Modules.cs` (~710, ~735), `CyclicModule.cs` (~146)
Failures are reduced to their error value and rethrown via `Throw.JavaScriptException(engine, error, DefaultLocation)` — the parse location of a syntax error (file/line/column), the original .NET stack and `Exception.Data` are gone. On main the original exception propagated intact, so this is an observable regression on the existing synchronous surface: error UIs that read `ex.Location` now get `(0,0)` with no source.

**8. Every failed blocking import now fires a phantom unhandled-rejection event** — `Engine.Modules.cs` (~735), `CyclicModule.cs` (~146)
The internal load-phase promise is rejected but no reaction is ever attached on the blocking paths, so `HostPromiseRejectionTracker` fires a `Reject` (never followed by `Handle`) for a promise the host never created and cannot observe. A host's unhandled-rejection telemetry now alarms on every failed `Modules.Import` even though the failure was also delivered as an exception — a new observable behavior on the pre-existing sync-loader surface.

**9. `StartImport` throws synchronously on `ModuleResolutionException`** — `Engine.Modules.cs` (~641)
The root-load guard catches only `JavaScriptException`, so the single most common failure — resolution — bypasses the operation's `IsFaulted`/`Error` channel and erupts from the `StartImport` call itself, while the same failure delivered through the loader's async settle arrives as `operation.Error`. One failure, two delivery channels, depending on where resolution happens; host code written to the documented poll-then-`GetResult` pattern crashes at the start call.

## Non-blocking / worth considering

**10. Nested blocking `Import` inside a job always burns the full `PromiseTimeout`** — `EventLoop.cs` (~201). Deliberate per the in-code comment, but the host gets a 10-second stall and a misleading "loader did not finish in time" for a structurally impossible drain (queued completion unrunnable behind the re-entrancy guard). Detecting `IsRunningJob` + a pending async load at drain start and failing fast with an actionable message would be much kinder to UI/game-loop hosts.

**11. `AsyncModuleLoader.LoadModuleContents`' guidance message is unreachable** — `AsyncModuleLoader.cs` (~90). `ModuleLoader.LoadModule`'s `catch (Exception)` swallows the carefully worded `NotSupportedException` and replaces it with the generic "Could not load module X", which reads as a missing file — the API-misuse guidance the message was written to deliver is dead code on every reachable route.

**12. `FinishWaiters` never clears `_waiters`** — `ModuleLoadCompletion.cs` (~230). A host naturally retains completions to correlate with its transport callbacks (the PR's own `DeferredModuleLoader` test type does), and each settled completion then pins its waiters' referrer modules, payloads and promise-capability graphs. A `_waiters.Clear()` after the loop drops the graph.

**13. `IAsyncModuleLoader` doc over-promises "at most once per resolved specifier per engine"** — `IAsyncModuleLoader.cs` (~17). A failed load is not recorded, and a snapshot restore discards pending loads, so the loader is re-asked in both cases (the PR's own `APendingLoadIsForgottenWhenTheEngineEndsItsEvaluationCycle` test asserts `Asked == ["late", "late"]`). Since this line lands in the public-contract table, it should say "at most one *in-flight* load per resolved specifier; re-asked after a failed load or a snapshot restore."

**14. The `[[LoadedModules]]` memo now bypasses `Resolve` on repeat imports** — `Engine.Modules.cs` (~157). Spec-correct per `HostLoadImportedModule`, but on main every `Import`/dynamic `import()` called `IModuleLoader.Resolve` before consulting the cache, so a loader could veto repeat imports (per-call access control on a pooled engine is a real embedder pattern — note `RestoreGlobalSnapshot` clears only `_pendingLoads`, not the memo). A denial that used to throw now silently succeeds. At minimum this deserves a release-note/doc callout on `IModuleLoader`.

**15. Warm re-import cost** — `Engine.Modules.cs` (~573). `Import` of an already-evaluated graph now allocates a fresh capability + `GraphLoadingState` and takes the host detour on every call, where the old code answered from one `_modules.TryGetValue`. A status check on the cached module could skip `RunLoadPhaseBlocking` and restore the near-zero re-import cost for the Import-per-request embedding shape.

---
*Review assisted by Claude Code multi-agent analysis (independent finders + adversarial verification pass over both sides of the diff); every item above was verified against the actual code before being reported.*

gkurt:
Thanks for the thorough review — all nine blocking items and all six non-blocking ones are addressed in 4d47e664b. Each blocking fix is pinned by a new test in `AsyncModuleLoaderTests`; I verified the new tests fail on the previous commit.

**Blocking**

1. **ShadowRealm realm capture** — `ModuleLoadCompletion` now captures `Engine.Realm` at registration (alongside the generation, for the same reason) and re-enters it around the deferred build, error minting and waiter continuations. `AShadowRealmModuleDeliveredAsynchronouslyEvaluatesInTheShadowRealm` pins both directions: the export arrives, and `globalThis.leak` lands in the shadow realm, not the principal one.
2. **Non-JS `Resolve` failures** — the `Resolve` call now has the same treatment as the builder branch: on a queued turn a `ModuleResolutionException` (or any non-constraint exception) becomes the load's failure; on a synchronous stack it propagates as before. `AResolutionFailureInsideAnAsyncGraphRejectsEveryWaitingImportInsteadOfStrandingThem` covers it with two waiters on the shared load.
3. **Constraint flattening** — the `LoadModuleAsync` catch-all now rethrows constraint/abort exception types (shared `MustPropagate` predicate: `ExecutionCanceledException`, `MemoryLimitExceededException`, `StatementsCountOverflowException`, `TimeoutException`, `OperationCanceledException`, `OutOfMemoryException`) and removes the unsettled pending-load entry so the specifier is not poisoned for the next import.
4. **Waiter isolation** — `FinishWaiters` serves every waiter, capturing the first escaping exception via `ExceptionDispatchInfo` and rethrowing it after the loop.
5. **Queued build/register failures** — `GetModuleSource` and registration are inside the guarded region, and a non-JS failure on a queued turn settles the waiters instead of erupting (`AThrowingModuleSourceHookRejectsTheImportInsteadOfStrandingIt`). The debugger callback also moved before the registry commit, so a throwing callback can no longer leave a half-registered module.
6. **Builder/fetch race** — pending loads are consulted before builders, so a mid-flight importer attaches to the fetch, and the deferred build hands waiters an already-registered record instead of overwriting it. One key, one record (`ABuilderRegisteredWhileAFetchOfTheSameKeyIsInFlightDoesNotForkTheModule`).
7. **Lost error identity** — the original exception is remembered alongside the error value it was reduced to (matched by reference identity, so a stale entry can never fire for an unrelated failure) and rethrown with its stack on the blocking `Import`/`Link` paths. A syntax error keeps its file/line/column (`ASyntaxErrorReachesTheBlockingImportWithItsOriginalLocation`) and a `ModuleResolutionException` from two loads deep surfaces as itself, matching the synchronous-loader surface.
8. **Phantom unhandled rejections** — the load-phase capability's promise is marked handled at creation; a failed blocking import now fires zero tracker events (`AFailedBlockingImportRaisesNoUnhandledRejectionEvent`).
9. **`StartImport` resolution failures** — delivered through `IsFaulted`/`Error` like every other failure; constraint exceptions still propagate.

**Non-blocking** — all taken:

10. A blocking import that cannot progress inside an event-loop job now fails immediately with an `InvalidOperationException` naming the actual problem and pointing at `StartImport`/`ImportAsync`, instead of burning the whole `PromiseTimeout`.
11. `ModuleLoader.LoadModule` lets `NotSupportedException` propagate, so `AsyncModuleLoader`'s synchronous-entry guidance is reachable rather than reading as a missing file.
12. `FinishWaiters` clears `_waiters` after serving, so a retained completion no longer pins referrer modules and promise-capability graphs.
13. The `IAsyncModuleLoader` doc now says "at most one in-flight load per resolved specifier" and spells out the re-ask after a failed load or a snapshot restore.
14. `IModuleLoader.Resolve` documents the `[[LoadedModules]]` memo — a loader is a mapping, not a checkpoint, and per-call veto of a repeat import cannot be implemented there.
15. `Import` of a module past `[[Status]] new` skips `RunLoadPhaseBlocking` entirely, restoring the near-zero warm re-import for the Import-per-request shape.

lahma (approved):
Thank you!

PR #2875, opened 2026-07-31
Resume an awaiting for-let loop into the environment it suspended in

HermanusMuellerEU:
## Summary

Resume a suspended `for` and `with` into the state they suspended in.

## Details

An environment-owning statement that can be suspended by `await` or `yield` must resume into the
environment it suspended in. `JintAwaitExpression.SuspendForAwait` re-captures `_savedContext` at
**every** await, so the environment current at the top of a replayed body is the one that was live
*at the await* — never the statement's own outer environment. Two statements did not account for
that: the C-style `for` loop, and `with`.

Two commits: the first is the original `for` fix, the second answers review and closes three more
resume positions plus `with`.

### 1 — `for (let …)` restored the wrong environment on exit (`28fd3bd8f`)

The loop read its outer environment from the execution context on every entry, including a
re-entry that is an async *replay*. Resuming into the body, what it read was the body block's
environment whenever the body declares a `let`/`const`. The loop took that as its own outer
environment and reinstated it on exit.

```js
const OUTER = 'outer-visible';
(async () => {
    for (let i = 0; i < 1; i++) {
        await Promise.resolve();
        const x = i + 1;
        void x;
    }
    return OUTER;          // throws
})()
```

```
InvalidCastException: Unable to cast object of type 'DeclarativeEnvironment'
                      to type 'GlobalEnvironment'
```

**Four failures follow from that one line**, two loud and two silent:

- `InvalidCastException` from `JintEnvironment.TryGetIdentifierEnvironmentWithBindingValue`, which
  casts to `GlobalEnvironment` on the assumption that only the global environment has a null
  `_outerEnv` — but a slot-backed block environment parked into `JintBlockStatement._cachedEnv` is
  detached at park time and has exactly that shape.
- A hang in `ExecutionContext.GetThisEnvironment()`, whose `while (true)` walk relies on the chain
  ending at the global environment. Its own source comment states that invariant; this broke it.
- A silent scope leak — the block's `let`/`const` stayed visible after the loop, and on wider
  shapes even globals became unresolvable.
- A silent lost write, and a quadratic slowdown — a closure created before the await was stranded
  when the replay rebuilt the iteration environment, so writes after the await went to an abandoned
  environment. The rebuild also parented each new environment under the previous one, growing the
  scope chain by one level per await.

`ForLoopSuspendData` gains **`OuterEnv`** (restored on resume instead of re-read from the execution
context) and **`IterationEnv`** (resumed *into* rather than rebuilt, so a closure captured before
the await is not stranded), and `CreatePerIterationEnvironment` no longer runs on a body or update
resume: that lands mid-iteration, where `ForBodyEvaluation` step 2 creates no new per-iteration
environment.

`JintForInForOfStatement.BodyEvaluation` already does the first two, for the same reason — see
f089071fc, whose message describes this defect exactly, but which fixed it for `for-of` and only
for `for-of`.

### 2 — the other two resume positions, and `with` (`8e4e03b43`)

Review found that a `for` loop has three resume positions and the first commit had only handled
one and a half. It also found a statement rebuilding the environment everything inside it had
saved.

**An `await` in the init.** The range test that decides whether a re-entry is a replay
(`IsNodeInsideForStatementExcludingInit`) deliberately excludes the init, because an init resume
must re-run the init to complete its pending awaits — and that exclusion gated the `OuterEnv`
restore too. So an init resume went on taking `oldEnv` from the ambient context, which on a replay
is the *abandoned first-attempt loop environment*, whose header bindings never got past TDZ.
Restoring that on exit makes a later read of a name the header shadows throw `ReferenceError` —
including through `typeof`, which must never throw for an out-of-scope name. `OuterEnv` is repaired
at the read rather than merely tolerated, because it is what the `finally` writes back: leaving it
poisoned chains another dead environment onto the next suspension's outer link. `IterationEnv` is
deliberately *not* extended to an init resume, which re-runs the init and needs its bindings
uninitialized.

**An `await` in the test.** A test resume is the one resume that legitimately re-runs the test from
the top, so neither `skipTestOnce` nor `resumeUpdateOnce` is set and the entry
`CreatePerIterationEnvironment` ran anyway — forking away from the iteration environment the resume
had just reinstated, and stranding every closure the test made before the await:

```js
for (let i = 0; (fns.push(() => i), await p(i < 2)); ) { i++; }
```

gave `0,1,2` where Node 24 gives `1,2,2`. The guard now keys on *the restored environment* rather
than on the resume position, which covers all three positions at once: whenever that environment
was restored there is by definition an iteration already in progress to stay inside.

**A truthy suspension in the test** (pre-existing). The suspension check sat inside the
`if (!testValue)` branch, so a suspension that leaves a *truthy* value behind went unnoticed —
which is exactly what the trip after a resume does, since the resume's memo is not cleared while
the loop is still resuming. The loop fell through into the body and ran one extra iteration per
resume; a `yield` in the test showed it as yields `0, 1, 3` instead of `0, 1, 2`. Only observable
with an `i++`-shaped body, which is why it outlived the first fix.

**`with` rebuilt its environment on every replay.** Every environment-owning statement inside the
body saves that object environment as its own outer one, so a rebuilt one turned those saves into
references to an environment no longer on the chain. The chain then no longer terminated in the
global environment and the next identifier walk failed its `GlobalEnvironment` cast with
`InvalidCastException` — a .NET exception, so neither a script `try`/`catch` nor an embedder
catching `JavaScriptException` could see it, and an embedder took an unhandled crash out of
`Evaluate` where it expected a rejected promise.

```js
var obj = { v: 'V' };
(async () => {
    var seen = [];
    with (obj) {
        for (let i = 0; i < 2; i++) { let b = i; await Promise.resolve(); seen.push(v + '/' + b); }
        seen.push('post:' + v);
    }
    return seen.join(',');
})()
// Node 24: "V/0,V/1,post:V"   main: InvalidCastException out of Evaluate
```

It now carries `WithSuspendData` and resumes into the same environment, mirroring
`JintBlockStatement` — which fixes the case rather than just its symptom, since the shape now
returns Node's answer. `with` was the last environment-owning statement without resume awareness;
block, switch, catch and for-of already had it. Its object expression can itself suspend
(`with (await p) { … }`), so that path unwinds and lets the replay redo the whole statement rather
than building an environment over the placeholder value.

**`ForLoopSuspendData.BoundValues` is gone.** Every resume with suspend data takes the
`IterationEnv` branch, and a resume without it has no saved values either, so the save was paying
`_boundNames.Count` `GetBindingValue` calls plus dictionary stores per await for state nothing
read — confirmed by instrumenting the restore arm and getting zero hits across all of Test262 and
`Jint.Tests`. Removing the save also removes the read that threw a TDZ `ReferenceError` out of the
`finally` on an init suspension, so
`function* g() { for (let i = yield 0; i < 1; i++) { } return 'done'; }` completes instead of
throwing.

### Generators are affected, and only by half of this

Worth stating precisely, because the obvious summary is wrong in one direction.

`OuterEnv` is genuinely an async-only concern: a generator captures its context once at
`GeneratorStart` and re-enters that same function-level context on every resume, so the environment
it resumes with really is the loop's outer one.

**That argument does not extend to `IterationEnv`, which a generator needs exactly as much as an
async function does.** The context a generator re-enters is the *function's*, not the loop's; the
per-iteration environment is rebuilt on a generator's replay exactly as on an async function's:

```js
function* g() { let get; for (let i = 0; i < 1; i++) { get = () => i; yield i; i = 42; } return get(); }
```

returns `42` with the field — the write after the `yield` lands in the environment the closure
captured, as the spec requires — and `0` without it, the closure left reading an abandoned
environment that never saw the write. So do **not** guard the save on the suspendable being an
`AsyncFunctionInstance`. The same holds for the init TDZ throw above, which a generator hits
without ever reaching the environment-restore path at all. This is written up on
`ForLoopSuspendData.IterationEnv`, with the script inline, so the next reader does not re-derive
the wrong half of it.

### Worth reviewers' attention

**Why the `for` fix is not in `JintBlockStatement`.** The block is correct. Its saved
`OuterEnvironment` genuinely is the environment its block env was created under, and restoring it
is exactly what `for-of` does. The fault is one frame up, in the for statement.

**Why a closure in the body masked the crash.** Not by disabling the iteration-environment reuse
optimisation — forcing `_canReuseIterationEnvironment` false changes nothing, which is what made
this hard to find. It is that `BlockState` only takes slot-backed storage when
`EnvironmentEscapeAstVisitor.MayEscape` is false, and only a slot-backed block environment is ever
parked and detached. The wrong environment was restored either way, so the closure hid the crash
while leaving the scope leak and the lost write in place. The tests assert the bindings are gone
rather than merely that nothing threw.

**Why `OuterEnv` and `IterationEnv` are rewritten on every suspension.** That is load-bearing
rather than idempotent: unless `_canReuseIterationEnvironment`, `CreatePerIterationEnvironment`
installs a fresh environment at the end of every iteration, so the environment saved by the second
suspension is not the one saved by the first. Guarding the writes as redundant work silently
returns iterations 2+ to the stranded-closure behaviour — and no single-iteration test can see it.
The comment in the `finally` says so.

**The structural point, deliberately out of scope.** `SuspendForAwait` capturing the live
environment is why six statements have each had to independently re-save their own outer
environment (`JintBlockStatement`, `JintSwitchStatement`, the catch clause,
`JintForInForOfStatement` twice, and now `JintForStatement` and `JintWithStatement`), and every one
of them was found the same way — an `InvalidCastException` to `GlobalEnvironment`, or an
unbreakable spin, in production. Restoring the statement's own outer environment once in the resume
path would make the next environment-owning statement correct by construction. Happy to file that
separately; it is a bigger change than a bug fix should carry.

**Scope.** Splitting the two commits is possible — the first is the crash fix and stands alone —
but they are one mechanism and the second is what makes the first complete. Say the word.

**No new retention.** `Data.Clear(this)` sits in the `finally`'s not-suspended arm for both
statements, so it fires on normal completion, `break`, `return` and `throw` alike. The new
references live only while the statement is genuinely suspended, bounded by the async call's
lifetime.

**Cost on unaffected paths.** The pooled, flattened and tight-body lanes are all gated on
`suspendable is null` and cannot engage in an async frame; the new branches short-circuit on
`resumingInLoop` / `resumingInInit`, false for every synchronous loop. `with` reads
`ExecutionContext.Suspendable` once and takes the existing path when it is null. The two benchmark
controls below measure exactly this.

## Linked issue

Fixes #

## Test plan

- [ ] For interop changes: covered in `Jint.Tests/Runtime/Interop` — n/a, no interop surface touched
- [x] For perf changes: included before/after numbers from `Jint.Benchmark`

`Jint.Tests/Runtime/ForLoopIterationEnvironmentAwaitTests.cs` — **61 tests**:

| group | tests |
| --- | ---: |
| body/test/update resume: the four manifestations, the trigger matrix, and controls a naive fix would break (per-iteration bindings must stay distinct; closures must still capture per-iteration) | 24 |
| `await` in the init: header shapes (simple, destructured, multi-declarator, multi-await, `const`) crossed with whether the body awaits, plus the shadowing and loop-exit cases | 16 |
| the init-suspension TDZ throw and the three symptoms it caused (infinite re-suspension, function-locals reset, generators) | 10 |
| `await`/`yield` in the test: stranded closures, and the extra iteration from a truthy suspension | 5 |
| generator shapes for `IterationEnv` and the init | 2 |
| `with`: same environment on resume, nested, a `for-of` body, and an `await` in the object expression | 4 |

Every evaluation runs on a `DedicatedThread` with a join timeout, and the class has its own
non-parallel collection, because one manifestation is an uninterruptible spin: a `TimeoutInterval`
constraint cannot stop it, since Jint does not evaluate constraints for event-loop jobs. Against
the unfixed engine a large fraction of these shapes leave a spinning thread behind, so that load is
a real quantity, not a hypothetical one.

| suite | result |
| --- | --- |
| `Jint.Tests` | 4721 passed, 0 failed (4 skipped) |
| `Jint.Tests.PublicInterface` | 1240 passed, 0 failed (9 skipped) |
| both under `JINT_HOST_CONTRACT_VERIFICATION=1` | 4721 / 1244 passed, 0 failed |
| `Jint.Tests.CommonScripts` | 28 passed, 0 failed |
| `Jint.Tests.Test262` | 99484 passed, 0 failed (157 skipped) |

Every new shape is pinned against **Node v24**, not against "differs from before" — on several of
these the branch and its parent were each wrong in *different* ways, and on one the expected
polarity is counter-intuitive (the closure must observe the post-`yield` write, so `42` is correct
and `main`'s `0` was the bug). Also differentially tested against Node for the minimal repro, the
lost-write pair, init-await shapes, loop-exit edge cases (`break` / labelled `continue` / `throw` /
nested / sequential / header awaits) and sibling constructs (`while`, `do-while`, `switch`,
`try`/`catch`/`finally`, labelled block, `for-in`, `for-of`) — all matching.

### Performance

Removing the per-resume rebuild takes an awaiting `for (let …)` loop from quadratic to linear in
its trip count. Default BenchmarkDotNet job, idle machine.

| benchmark | before | after | |
| --- | ---: | ---: | --- |
| 250 trips | 589.9 µs | 172.0 µs | 3.4× |
| 1000 trips | 7,540.0 µs | 647.5 µs | 11.6× |
| 2000 trips | 31,098.9 µs | 1,298.6 µs | **24.0×** |
| `var` header (control — changed code unreachable) | 578.9 µs | 560.9 µs | unchanged |
| synchronous loop (control) | 24.78 µs | 24.96 µs | unchanged |

Allocation falls ~43% at every size (1000 trips: 522,622 B → 298,520 B).

> Measured at the first commit against its base. The second commit only *removes* per-await work
> (the `BoundValues` save: `_boundNames.Count` `GetBindingValue` calls plus dictionary stores on
> every suspension), so these are a lower bound on the current branch. Happy to re-run the pair if
> you want the figures restated against today's `main`.

The benchmark is not part of the commit, since it exists only to measure this change. Source to
reproduce:

<details>
<summary><code>Jint.Benchmark/ForLetAwaitBenchmark.cs</code></summary>

```csharp
using BenchmarkDotNet.Attributes;

namespace Jint.Benchmark;

[MemoryDiagnoser]
[HideColumns("Error", "Gen0", "Gen1", "Gen2")]
public class ForLetAwaitBenchmark
{
    private Engine _engine = null!;
    private Prepared<Script> _awaitLetLoop250, _awaitLetLoop, _awaitLetLoop2000;
    private Prepared<Script> _awaitVarLoop, _syncLetLoop;

    private static Prepared<Script> LetLoop(int n) => Engine.PrepareScript($$"""
        async function f() {
            var s = 0;
            for (let i = 0; i < {{n}}; i++) { s += await Promise.resolve(1); }
            return s;
        }
        f();
        """);

    [GlobalSetup]
    public void Setup()
    {
        _engine = new Engine();
        _awaitLetLoop250 = LetLoop(250);
        _awaitLetLoop = LetLoop(1000);
        _awaitLetLoop2000 = LetLoop(2000);

        // control: `var` header, so _boundNames is null and the changed code is unreachable
        _awaitVarLoop = Engine.PrepareScript("""
            async function f() {
                var s = 0;
                for (var i = 0; i < 1000; i++) { s += await Promise.resolve(1); }
                return s;
            }
            f();
            """);

        // control: no async at all
        _syncLetLoop = Engine.PrepareScript("""
            function f() {
                var s = 0;
                for (let i = 0; i < 1000; i++) { s += 1; }
                return s;
            }
            f();
            """);
    }

    [Benchmark] public void AwaitLetLoop250() => _engine.Evaluate(_awaitLetLoop250);
    [Benchmark] public void AwaitLetLoop1000() => _engine.Evaluate(_awaitLetLoop);
    [Benchmark] public void AwaitLetLoop2000() => _engine.Evaluate(_awaitLetLoop2000);
    [Benchmark] public void AwaitVarLoop1000() => _engine.Evaluate(_awaitVarLoop);
    [Benchmark] public void SyncLetLoop1000() => _engine.Evaluate(_syncLetLoop);
}
```

The bodies deliberately declare no `let`/`const`, because that is the only `for-let` + `await`
shape that also runs on the pre-fix engine, so patched and unpatched are comparable.

</details>

## Breaking change?

No. `ForLoopSuspendData`, `WithSuspendData`, `JintForStatement` and `JintWithStatement` are all
`internal`; there are no public API changes. The behaviour change is bug-fix only — the affected
shapes previously threw, hung, or produced results that disagree with the spec and with Node 24.

### DIFF
--- a/Jint.Tests/DedicatedThread.cs
+++ b/Jint.Tests/DedicatedThread.cs
@@ -1,0 +1,76 @@
+#nullable enable
+
+using System.Runtime.ExceptionServices;
+using Xunit.Sdk;
+
+namespace Jint.Tests;
+
+/// <summary>
+/// Runs a test body on a dedicated thread, rather than on the xUnit worker thread.
+/// <para>
+/// Two unrelated reasons to want one, both served here so there is a single implementation. Deep
+/// recursion needs a larger stack than the default thread provides
+/// (<see cref="LargeStackSize"/>). And a test whose failure mode is a <em>non-terminating</em>
+/// engine — an uninterruptible walk over a corrupt lexical environment chain, say — needs
+/// <paramref name="joinTimeout"/>: an execution constraint cannot stop one of those, because Jint
+/// does not evaluate constraints for event-loop jobs and so cannot interrupt a continuation.
+/// Without a timeout a single regression wedges the whole class (xUnit runs a class's tests
+/// sequentially) and CI hangs instead of reporting.
+/// </para>
+/// <para>
+/// The exception is re-thrown through <see cref="ExceptionDispatchInfo"/>, which preserves its
+/// type, its original stack trace and its <see cref="Exception.Data"/> — a regression surfaces in
+/// CI as a navigable failure rather than as a flattened string.
+/// </para>
+/// </summary>
+internal static class DedicatedThread
+{
+    /// <summary>
+    /// Enough stack for the deep-recursion tests; the platform default is far smaller.
+    /// </summary>
+    public const int LargeStackSize = 16 * 1024 * 1024;
+
+    public static void Run(Action action, TimeSpan? joinTimeout = null, string? timeoutMessage = null, int maxStackSize = LargeStackSize)
+    {
+        ExceptionDispatchInfo? exception = null;
+
+        var thread = new Thread(
+            () =>
+            {
+                try
+                {
+                    action();
+                }
+                catch (Exception e)
+                {
+                    exception = ExceptionDispatchInfo.Capture(e);
+                }
+            },
+            maxStackSize)
+        {
+            // A thread that outlives its join must not keep the process alive...
+            IsBackground = true,
+
+            // ...and must not compete with the rest of the run for a core while it does. Background
+            // only governs process exit; a spinning thread goes on burning CPU until then, which is
+            // exactly the condition the suite's known wall-clock flakes fail under.
+            Priority = ThreadPriority.Lowest,
+        };
+
+        thread.Start();
+
+        if (joinTimeout is { } timeout)
+        {
+            if (!thread.Join(timeout))
+            {
+                throw new XunitException(timeoutMessage ?? $"the test body did not complete within {timeout}");
+            }
+        }
+        else
+        {
+            thread.Join();
+        }
+
+        exception?.Throw();
+    }
+}
--- a/Jint.Tests/Runtime/EngineLimitTests.cs
+++ b/Jint.Tests/Runtime/EngineLimitTests.cs
@@ -27,7 +27,7 @@ public class EngineLimitTests
         // is deliberately generous: StackOverflowException is uncatchable and would kill the whole test
         // process, so this is a functional smoke test of deep call chains, not a per-frame native-stack
         // budget test (platforms differ too much in frame size for a tight budget to be safe).
-        RunOnDedicatedThread(static () =>
+        DedicatedThread.Run(static () =>
         {
             var script = GenerateCallTree(FunctionNestingCount);
 
@@ -36,25 +36,6 @@ public class EngineLimitTests
             engine.Evaluate("func1(123);").AsNumber().Should().Be(123);
             engine.Evaluate("x").AsNumber().Should().Be(FunctionNestingCount);
         });
-    }
-
-    private static void RunOnDedicatedThread(Action action)
-    {
-        ExceptionDispatchInfo? exception = null;
-        var thread = new Thread(() =>
-        {
-            try
-            {
-                action();
-            }
-            catch (Exception e)
-            {
-                exception = ExceptionDispatchInfo.Capture(e);
-            }
-        }, maxStackSize: 16 * 1024 * 1024);
-        thread.Start();
-        thread.Join();
-        exception?.Throw();
     }
 
     [Fact]
--- a/Jint.Tests/Runtime/ForLoopIterationEnvironmentAwaitTests.cs
+++ b/Jint.Tests/Runtime/ForLoopIterationEnvironmentAwaitTests.cs
@@ -1,0 +1,572 @@
+using Jint.Native;
+using Jint.Runtime;
+
+namespace Jint.Tests.Runtime;
+
+/// <summary>
+/// A C-style <c>for</c> loop restores, on exit, the environment that was current when it was
+/// entered. It used to read that environment from the execution context on every entry —
+/// including a re-entry that is an async <em>replay</em>.
+///
+/// That is unsound for an async function, because <c>JintAwaitExpression.SuspendForAwait</c>
+/// re-captures <c>_savedContext</c> at <em>every</em> await, so the environment current at the top
+/// of a replayed body is the one that was live at the await. When the body declares a
+/// <c>let</c>/<c>const</c>, that is the body block's environment, and the loop took it as its own
+/// outer environment — then restored it on exit.
+///
+/// Three observable failures followed, the first two loud and the third silent:
+/// reading an outer binding after the loop threw <see cref="InvalidCastException"/> from
+/// <c>JintEnvironment.TryGetIdentifierEnvironmentWithBindingValue</c>, which casts to
+/// <c>GlobalEnvironment</c> on the assumption that only the global environment has a null
+/// <c>_outerEnv</c> — but a block environment parked into <c>JintBlockStatement._cachedEnv</c> is
+/// detached (<c>_outerEnv = null</c>) at park time, and has exactly that shape; reading
+/// <c>this</c> hung forever in <c>ExecutionContext.GetThisEnvironment()</c>, whose
+/// <c>while (true)</c> walk relies on the chain ending at the global environment; and the block's
+/// own bindings stayed visible after the loop.
+///
+/// A closure anywhere in the loop body masked the first two entirely — not by disabling the
+/// iteration-environment reuse optimisation (forcing that off changes nothing), but because
+/// <c>BlockState</c> only takes slot-backed storage when
+/// <c>EnvironmentEscapeAstVisitor.MayEscape</c> is false, and only a slot-backed block
+/// environment is ever parked and detached. The wrong environment was restored either way, which
+/// is why the closure shapes below assert the block's bindings are gone rather than merely that
+/// nothing threw.
+///
+/// The fix mirrors <c>JintForInForOfStatement.BodyEvaluation</c>, which restores its
+/// <c>oldEnv</c> from <c>ForOfSuspendData.OuterEnv</c> for this same reason (commit f089071fc).
+/// Generators need no such thing: they capture their context once at GeneratorStart and re-enter
+/// that same function-level context on every resume.
+/// </summary>
+public class ForLoopIterationEnvironmentAwaitTests
+{
+    /// <summary>
+    /// Evaluates <paramref name="script"/> on a dedicated background thread and fails on a join
+    /// timeout rather than hanging the run.
+    /// <para>
+    /// This is not defensive padding. One manifestation of the defect these tests cover is an
+    /// infinite walk in <c>ExecutionContext.GetThisEnvironment()</c>, and a <c>TimeoutInterval</c>
+    /// constraint cannot stop it — Jint does not evaluate constraints for event-loop jobs, so it
+    /// cannot interrupt a continuation. Without this, a single regression wedges the whole test
+    /// class (xUnit runs a class's tests sequentially) and CI hangs instead of reporting.
+    /// </para>
+    /// </summary>
+    private static JsValue RunAsync(string script)
+    {
+        JsValue result = null;
+        Exception failure = null;
+
+        var worker = new Thread(() =>
+        {
+            try
+            {
+                result = new Engine().Evaluate(script).UnwrapIfPromise();
+            }
+            catch (Exception ex)
+            {
+                failure = ex;
+            }
+        })
+        {
+            // If it is spinning, it must not keep the test process alive.
+            IsBackground = true,
+        };
+
+        worker.Start();
+
+        if (!worker.Join(TimeSpan.FromSeconds(30)))
+        {
+            throw new Xunit.Sdk.XunitException(
+                "evaluation did not complete within 30s — the lexical environment chain is most "
+                + "likely corrupt, leaving GetThisEnvironment() or an identifier walk spinning");
+        }
+
+        if (failure is not null)
+        {
+            throw new Xunit.Sdk.XunitException($"evaluation threw: {failure}");
+        }
+
+        return result;
+    }
+
+    [Fact]
+    public void ShouldSeeOuterBindingAfterAwaitingLoopWithBlockScopedDeclaration()
+    {
+        var result = RunAsync("""
+            const OUTER = 'outer-visible';
+            (async () => {
+                for (let i = 0; i < 1; i++) {
+                    await Promise.resolve();
+                    const x = i + 1;
+                    void x;
+                }
+                return OUTER;
+            })()
+            """);
+
+        result.Should().Be("outer-visible");
+    }
+
+    [Fact]
+    public void ShouldSeeOuterBindingWhenDeclarationPrecedesTheAwait()
+    {
+        var result = RunAsync("""
+            const OUTER = 'outer-visible';
+            (async () => {
+                for (let i = 0; i < 1; i++) {
+                    const x = i + 1;
+                    void x;
+                    await Promise.resolve();
+                }
+                return OUTER;
+            })()
+            """);
+
+        result.Should().Be("outer-visible");
+    }
+
+    [Fact]
+    public void ShouldSeeOuterBindingWithLetDeclarationInBody()
+    {
+        var result = RunAsync("""
+            const OUTER = 'outer-visible';
+            (async () => {
+                for (let i = 0; i < 1; i++) {
+                    await Promise.resolve();
+                    let x = i + 1;
+                    void x;
+                }
+                return OUTER;
+            })()
+            """);
+
+        result.Should().Be("outer-visible");
+    }
+
+    [Fact]
+    public void ShouldSeeOuterBindingWhenAwaitIsNestedDeeperThanTheDeclaration()
+    {
+        var result = RunAsync("""
+            const OUTER = 'outer-visible';
+            (async () => {
+                for (let i = 0; i < 1; i++) {
+                    const x = i + 1;
+                    if (x > 0) { if (x > 0) { await Promise.resolve(); } }
+                }
+                return OUTER;
+            })()
+            """);
+
+        result.Should().Be("outer-visible");
+    }
+
+    [Fact]
+    public void ShouldSeeOuterBindingWhenDeclarationAndAwaitShareANestedBlock()
+    {
+        var result = RunAsync("""
+            const OUTER = 'outer-visible';
+            (async () => {
+                for (let i = 0; i < 1; i++) {
+                    { const x = i + 1; void x; await Promise.resolve(); }
+                }
+                return OUTER;
+            })()
+            """);
+
+        result.Should().Be("outer-visible");
+    }
+
+    /// <summary>
+    /// The <c>this</c> manifestation. Before the fix this did not throw — it spun forever in
+    /// <c>ExecutionContext.GetThisEnvironment()</c>. A <c>TimeoutInterval</c> constraint does NOT
+    /// help: Jint does not evaluate constraints for event-loop jobs, so it cannot interrupt a
+    /// continuation, which is precisely the case here. The evaluation therefore runs on its own
+    /// background thread and the test fails on a join timeout instead of hanging the run.
+    /// </summary>
+    [Fact]
+    public void ShouldReadThisAfterAwaitingLoopWithBlockScopedDeclaration()
+    {
+        var result = RunAsync("""
+            class K {
+                async leaf() { return 1; }
+                async run() {
+                    for (let i = 0; i < 1; i++) {
+                        const x = i;
+                        void x;
+                        await this.leaf();
+                    }
+                    return await this.leaf();
+                }
+            }
+            new K().run()
+            """);
+
+        result.Should().Be(1);
+    }
+
+    [Fact]
+    public void ShouldPreserveLoopVariableSemanticsAcrossAwait()
+    {
+        // The per-iteration environment must still be per-iteration: each iteration's `i` is a
+        // distinct binding. Guards against "fix it by never creating the environment".
+        var result = RunAsync("""
+            (async () => {
+                const seen = [];
+                for (let i = 0; i < 3; i++) {
+                    await Promise.resolve();
+                    const doubled = i * 2;
+                    seen.push(doubled);
+                }
+                return seen.join(',');
+            })()
+            """);
+
+        result.Should().Be("0,2,4");
+    }
+
+    [Fact]
+    public void ShouldStillCaptureDistinctBindingsPerIterationWhenBodyAwaits()
+    {
+        // The reuse optimisation exists to avoid allocating an environment per iteration when
+        // nothing captures it. Where something DOES capture it, each closure must still see its
+        // own `i` — the classic per-iteration-binding guarantee — even with an await in the body.
+        var result = RunAsync("""
+            (async () => {
+                const fns = [];
+                for (let i = 0; i < 3; i++) {
+                    await Promise.resolve();
+                    const captured = i;
+                    fns.push(() => `${i}:${captured}`);
+                }
+                return fns.map(f => f()).join(',');
+            })()
+            """);
+
+        result.Should().Be("0:0,1:1,2:2");
+    }
+
+    [Fact]
+    public void ShouldHandleNestedForLoopsWhereOnlyTheInnerOneAwaits()
+    {
+        var result = RunAsync("""
+            const OUTER = 'outer-visible';
+            (async () => {
+                for (let i = 0; i < 2; i++) {
+                    for (let j = 0; j < 2; j++) {
+                        await Promise.resolve();
+                        const x = j;
+                        void x;
+                    }
+                }
+                return OUTER;
+            })()
+            """);
+
+        result.Should().Be("outer-visible");
+    }
+
+    [Fact]
+    public void ShouldSeeOuterBindingAfterAForAwaitOfInsideAForLoopBody()
+    {
+        // `for await...of` carries an implicit await, so it suspends the body the same way.
+        var result = RunAsync("""
+            const OUTER = 'outer-visible';
+            async function* gen() { yield 1; }
+            (async () => {
+                for (let i = 0; i < 1; i++) {
+                    const x = i;
+                    void x;
+                    for await (const v of gen()) { void v; }
+                }
+                return OUTER;
+            })()
+            """);
+
+        result.Should().Be("outer-visible");
+    }
+
+    /// <summary>
+    /// The silent manifestation. A closure in the body keeps the block environment out of the
+    /// park-and-detach path, so nothing throws — but the loop still restored the block's
+    /// environment on exit, leaving the block's own bindings resolvable afterwards. Asserting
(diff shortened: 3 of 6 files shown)


lahma (changes requested):
Thanks for this — the diagnosis is right and the core of the fix is right. Reusing the suspension's iteration environment instead of rebuilding it is exactly what strands the closure today, and `ForOfSuspendData.IterationEnv` is the right precedent to follow. The test file is unusually good: it runs on a dedicated thread precisely because the failure mode is an uninterruptible spin, and it says so.

I reviewed it at depth against `main` (A/B repros with `Jint.Repl -c Release` in two worktrees, `28fd3bd8f` vs `0af0a25d8`, cross-checked against Node 24). Three correctness items need fixing before this goes in — the first is a regression against `main`, not a pre-existing gap. The rest are documentation and test-hygiene notes that matter mostly because they actively invite this bug back.

---

## 1. Blocking — an `await` in the for-**init** turns a silently-wrong read into a hard `ReferenceError`

`JintForStatement.cs:279` gates the `OuterEnv` restore on `resumingInLoop`, which by design excludes the init (`IsNodeInsideForStatementExcludingInit`). So on a resume *from the init*, line 272 still takes `oldEnv` from the ambient execution context — and that is the abandoned first-attempt loop environment, whose loop variable is still in TDZ. The next suspension (in the body) then latches it at `:402`, and every later resume reinstalls it at `:281-282`.

```js
async function f() {
  for (let i = await P(0); i < 3; i++) { await P(1); }
  return typeof i;
}
```

| | result |
| --- | --- |
| `main` (0af0a25d8) | `"number"` (leaked value — wrong, but quiet) |
| this PR (28fd3bd8f) | **throws** `ReferenceError: Cannot access 'i' before initialization` |
| Node 24 | `"undefined"` |

A leaked-but-readable value becoming a hard throw is the part that worries me: `typeof` on an out-of-scope name must never throw, so an embedder's script that ran before this change now aborts. I generated 180 for-loop shapes; 30 regressed, and all 30 — and only those — have an `await` in the for-init. A destructuring head does it too:

```js
for (let [a, b] = await P([0, 10]); a < 2; a++, b++) { await P(1); }
// then: typeof a   →   throws on this PR, leaks a value on main
```

The comment at `:399-401` is where the reasoning goes wrong:

> A suspension in the *init* also lands here and its oldEnv is not the loop's outer environment, but nothing reads it as one

Something does read it as one — just not on the pass that wrote it. The poisoned `oldEnv` isn't consumed by the init suspension; it's the value the **next** (body) suspension copies into `data.OuterEnv`, and *that* one is read as the loop's outer environment on every subsequent resume. The `finally` then leaves it current after the loop.

None of the 24 new tests uses a `for (let x = await …)` head, which is why this is green.

## 2. Blocking — a resume into the loop **test** still strands closures

`JintForStatement.cs:616`: the new `!skipTestOnce` guard suppresses the entry `CreatePerIterationEnvironment` for a resume into the body or the update, but not for a resume into the *test*. So a test-await forks away from the `IterationEnv` that `:292` just installed — the same stranded-closure failure this PR fixes on the body path.

```js
(async () => {
  const fns = [];
  for (let i = 0; (fns.push(() => i), await Promise.resolve(i < 2)); ) { i++; }
  return fns.map(f => f()).join(',');
})()
```

`0,1,2` on this branch, `1,2,2` in Node 24. Making the guard key on the restored environment rather than on the resume position fixes it:

```csharp
if (!(resumingInLoop && suspendData?.IterationEnv is not null)
    && !skipTestOnce && !resumeUpdateOnce
    && _shouldCreatePerIterationEnvironment && !_canReuseIterationEnvironment)
```

With that change Jint returns `1,2,2`, and ~40 other async/generator loop shapes I re-ran (closures on both sides of the await, nested loops, labelled break, recursion, interleaved concurrent invocations, update-await, for-await-of) all still match Node.

## 3. Blocking — inside `with`, the snapshot restore escapes as an uncatchable `InvalidCastException`

Swapping the live read for `suspendData.OuterEnv` at `:281` drops the guarantee that the restored environment exists in the *current* replay. `JintWithStatement` has no resume awareness and builds a fresh `ObjectEnvironment` on every replay, so the snapshot points at a dead one.

```js
var obj = { v: 'V' };
async function f() {
  var seen = [];
  with (obj) {
    for (let i = 0; i < 2; i++) { let b = i; await tick(); seen.push(v + '/' + b); }
    seen.push('post:' + v);
  }
  return seen.join(',');
}
```

| | result |
| --- | --- |
| `main` | promise rejects with JS `ReferenceError: seen is not defined` |
| this PR | `System.InvalidCastException: Unable to cast object of type 'DeclarativeEnvironment' to type 'GlobalEnvironment'` out of `Engine.Evaluate()` |

That's a change in kind, not degree: an `InvalidCastException` is not a `JavaScriptException`, so script-level `try/catch` can't catch it, no rejection handler runs, and an embedder catching `JavaScriptException` around `Evaluate` takes an unhandled crash instead of a rejected promise. Worth noting your own test-file doc names this exact exception as the loud symptom the PR eliminates.

---

## Non-blocking, but please fix the two comments — they invite the bug back

**`JintForStatement.cs:396` — the "no-op" claim is false.** The comment says the per-suspension rewrite of `OuterEnv`/`IterationEnv` is redundant "since such a resume installed exactly these two and the loop went on running in them". For any loop where `_canReuseIterationEnvironment` is false, `CreatePerIterationEnvironment` installs a fresh environment at the end of *every* iteration, so the environment saved at suspension 2 is not the one saved at suspension 1 — the rewrite is load-bearing. Acting on the comment as written (hoisting the writes to the first suspension, or guarding them with `if (data.IterationEnv is null)` as a "remove redundant work" cleanup) silently reverts iterations 2+ to the stranded-closure behaviour, and every single-iteration test in the new file still passes.

**`SuspendData.cs:118` — "a generator needs neither" is falsified by this PR's own generator behaviour change.** The argument (generators re-enter one function-level context) holds for `OuterEnv`. It does not hold for `IterationEnv`:

```js
function* g() { let get; for (let i = 0; i < 1; i++) { get = () => i; yield i; i = 42; } return get(); }
```

returns `42` on this branch and `0` on `main` — that's `IterationEnv` doing work for a generator. As written the doc invites exactly the wrong optimisation (guard the `IterationEnv` save on `suspendable is not AsyncFunctionInstance`), with no test to catch it.

**`SuspendData.cs:131` — the rationale is stated backwards.** "assignments … land in the new environment while the closure — and the loop's own test and update — go on reading the old one". The test and update run in the environment the loop just installed, i.e. the new one; only the closure is stranded. Your own test comment at `ForLoopIterationEnvironmentAwaitTests.cs:440` states it correctly and oppositely.

## Smaller items

- **`JintForStatement.cs:405` — `BoundValues` is now dead on the resume path.** All three fields are written in the same guarded block, so a resume *with* suspend data always takes the new `IterationEnv` branch at `:285`, and a resume *without* it has no `BoundValues` either. The `else` arm's `InitializeBinding` loop (`:339-345`) can no longer fire while `resumingInLoop`. Meanwhile every `await` in a `for (let …)` loop still pays `_boundNames.Count` `GetBindingValue` calls plus dictionary stores at `:405-411` for state nothing reads — on the same hot async-loop path the PR's allocation numbers are about. The `IterationEnv` XML doc's closing line, "BoundValues remains the fallback for a resume that has no saved environment", is never true: such a resume has no saved `BoundValues` either.
- **`JintForStatement.cs:282` — `engine.UpdateLexicalEnvironment(oldEnv)` here is dead.** Reaching it requires `resumingInLoop`, hence a non-null `suspendable`, so the pooled branch at `:295` is unreachable from here; the other two branches both end in `UpdateLexicalEnvironment(loopEnv)` (`:293`, `:336`) and nothing between reads the environment. It's copy-pasted from `JintForInForOfStatement.BodyEvaluation:391`, where it *is* load-bearing because the iterator step runs in `oldEnv`.
- **`JintForStatement.cs:409` (pre-existing, but the new comment claims otherwise).** `GetBindingValue` reads the loop's own binding unconditionally, and after an init suspension that binding is in TDZ, so it throws out of the `finally`. `function* g() { for (let i = yield 0; i < 1; i++) { } return 'done'; }` → `it.next(); it.next(0)` throws `ReferenceError` on both branches (Node returns `done`). Reproduces for every generator head shape I tried. Not introduced here — but `:395-401` now asserts this path is harmless, and it isn't.
- **`ForLoopIterationEnvironmentAwaitTests.cs:76` — a failing test abandons a core-pegging thread.** `IsBackground = true` keeps a spinning worker from blocking process exit, not from burning a core until then, and the failure this class targets is exactly an uninterruptible `while (true)` in `GetThisEnvironment()`. Running the class against `main` (16 of 24 fail) left two threads spinning at 29 s and 30 s, still running after their tests reported. A future regression here wouldn't just turn these red — it'd CPU-saturate the rest of the `Jint.Tests` run, which is the condition the repo's known wall-clock flakes fail under. Its own xUnit collection would contain the blast radius.
- **`…Tests.cs:53` — `RunAsync` duplicates `EngineLimitTests.RunOnDedicatedThread` and loses the exception.** That helper already rethrows via `ExceptionDispatchInfo`, preserving type, stack and `Data`; this one flattens everything into `new XunitException($"evaluation threw: {failure}")`, so a regression shows up in CI as a string with no type and no navigable stack. Adding a join timeout to the existing helper gets both behaviours from one implementation.
- **`…Tests.cs:223` — the guard comment names the wrong test.** `ShouldPreserveLoopVariableSemanticsAcrossAwait` says it "Guards against 'fix it by never creating the environment'", but every read of `i` happens in the iteration that wrote it, so one shared environment still yields `"0,2,4"` and the test passes. `ShouldStillCaptureDistinctBindingsPerIterationWhenBodyAwaits` (`:227`) is the one that actually pins it.

---

One structural thought, take it or leave it: this is the sixth place that has had to independently re-save its own outer environment, because `JintAwaitExpression.SuspendForAwait` captures the environment live at the await (`JintBlockStatement`, `JintSwitchStatement`, the catch clause, `JintForInForOfStatement` twice, now this). Each one is found the same way — an `InvalidCastException` to `GlobalEnvironment` or an unbreakable spin, in production. Restoring the statement's own outer environment once in the resume path would make the next environment-owning statement correct by construction. Well outside this PR's scope; worth an issue.

Happy to help land items 1–3 if that's useful.

lahma (approved):
Everything I raised is addressed, and the two I flagged as "fix the comment" you fixed by fixing the *code* the comment was wrong about. Verified against `8e4e03b43`, not taken on trust.

## The three blocking items

Each now matches V8 exactly, A/B'd with `Jint.Repl -c Release` against `0af0a25d8` and Node 24:

| | `main` | first revision | **8e4e03b43** | Node 24 |
| `for (let i = await P(0); …)` → `typeof i` | `"number"` | **ReferenceError** | `undefined` | `undefined` |
| closures created in an awaiting test | — | `0,1,2` | `1,2,2` | `1,2,2` |
| `with` + awaiting for-let | JS `ReferenceError` | **InvalidCastException** | `V/0,V/1,post:V` | same |
| `with (await p) { … }` | crash | crash | `W` | `W` |
| `function* g(){ for (let i = yield 0; …) }` | ReferenceError | ReferenceError | `done` | `done` |

Repairing the init case at the *read* rather than tolerating it at the write is the right call, and the comment now explains why the poisoned value is self-compounding — which is exactly the part that made it non-obvious.

Keying the per-iteration gate on `restoredIterationEnv` instead of the resume position is better than the fix I suggested: it derives the condition from the thing that actually matters rather than enumerating three positions, so the test case falls out rather than being special-cased.

And you found one I missed — the suspension check sitting inside the falsy branch, so a resumed trip with a memoized truthy value fell through and ran an extra iteration. Good catch, and `ShouldNotRunAnExtraIterationWhenTheTestSuspendsWhileTruthy` pins it.

Making `JintWithStatement` resume-aware at the root rather than working around it in the for statement is the right depth of fix, and it follows the established `IsResuming` + suspend-data shape. Handling `with (await p)` while you were in there was more than I asked for.

## Differential testing

I generated **107** async/generator loop shapes — init-await, destructured and const heads, closures on both sides of a test-await, nested loops, labelled break, try/finally, `with`-nesting, no-update heads — and diffed every one against Node 24:

- **this branch: 0 of 107 disagree**
- `main`: 80 of 107 disagree (73 silent wrong values, 4 hard `InvalidCastException`, 3 spurious throws)

Plus a 20-case adversarial `with` corpus (with-in-with, with-in-loop, `with (await p)` twice, generators, recursion, concurrent invocations, `switch`/`catch`/`for-of` nesting, assignment through the with-object, closures spanning the await): **0 mismatches**, fixing 3 `main` failures and breaking none of the other 17.

Suites: `Jint.Tests` 4721/0 (9 s — no wedged threads), `Test262` **99,484/0**, `PublicInterface` 1240/0, `CommonScripts` 28/0.

## The smaller items

All addressed: `BoundValues` deleted outright, which also removed the TDZ throw in the `finally` for free; the dead `UpdateLexicalEnvironment(oldEnv)` gone, with a note on why for-of needs it and this doesn't; `DedicatedThread` unifies both helpers with `ExceptionDispatchInfo` **and** `ThreadPriority.Lowest`, which is a better answer than the test-isolation I suggested; the guard comment now says outright that it does not pin the per-iteration environment and names the test that does. 24 → 61 tests.

## Performance — reproduced

Ryzen 9 5950X, default BDN job, serial, idle machine. I ran your benchmark and a re-cut giving each row its own engine (numbers below are the isolated cut):

| row | `main` | this branch | | allocated |
| --- | ---: | ---: | --- | --- |
| 250 trips | 469.9 µs | 135.2 µs | 3.5× | −42.5% |
| 1000 trips | 6,157.9 µs | 490.7 µs | 12.6× | −42.9% |
| 2000 trips | 27,097.3 µs | 948.2 µs | **28.6×** | −43.0% |
| `var` header (control) | 454.7 µs | 444.7 µs | flat | +8 B |
| synchronous (control) | 20.41 µs | 20.42 µs | flat | 0 |

Your claims hold, and the 2000-trip row is better than the 24.0× you reported.

One note on the harness, since the file isn't committed and this is advisory: `ForLetAwaitBenchmark` shares a single `Engine` across all five rows, and every row's script declares `function f` and `var s` on that one global object — the shape AGENTS.md's "never warm one engine with more than one row's workload" rule exists to prevent. It isn't academic here: the shared cut reports the `var` control **5% slower**, the isolated cut **2% faster**. A 7-point swing on a control row is precisely the failure mode, and the control rows are the ones a reviewer leans on. `IsolatedScript` in `Jint.Benchmark/` does this in one line per row.

## SunSpider / Dromaeo

I ran the full gate twice in both orderings, and the honest answer is that it **cannot resolve this change** — not that it's clean:

- Within-run StdDev is tiny (median 0.33%) but between-window deltas reach 10%, so the spread is drift between measurement windows, not row noise.
- Correlation between the two pairs' per-row deltas is **r = +0.28** (SunSpider) and **r = +0.08** (Dromaeo). A real effect gives r near +1.
- Median delta across all 50 rows: −0.14% and −0.20%. Centred on zero, both directions, no bias.
- 14 rows agree in sign at >1% across both orderings; ~12 are expected by chance.

So: no evidence of a regression, and no evidence the gate could have found one under 5%. The targeted controls are the better evidence anyway, and they're mechanistically airtight — the new branches all short-circuit on `resumeNode is not null`, which requires a suspendable, and **no benchmark script in the repo contains a `with` statement**, so `JintWithStatement` is never executed by either suite. `SyncLetLoop1000` at 20.41 → 20.42 µs (StdDev 0.1) pins the synchronous for-loop path far tighter than any SunSpider row could.

## One last thing before merge

The PR description is now stale in three places: it says "24 tests", it doesn't mention the `with` fix or `JintWithStatement`/`WithSuspendData` at all, and it still says "Generators were never affected" — which this branch's own `SuspendData` doc now correctly contradicts for `IterationEnv`, with the falsifying script inline. Worth updating, since that sentence is the one a future reader would use to justify guarding the save on `AsyncFunctionInstance`.

Approving. Nice work — the init-resume repair and the extra-iteration bug were both beyond what I flagged.

HermanusMuellerEU:
I've update the PR description.  Thanks for the good work that you're doing.  We're using this to run some serious JavaScript code on.

lahma:
> I've update the PR description.  Thanks for the good work that you're doing.  We're using this to run some serious JavaScript code on.

Great to hear, hope Jint serves you well. Don't forget to sponsor 😉

PR #2968, opened 2026-08-10
Let parseInt and parseFloat take the frameless fast-call lane

lahma:
> **Draft-quality until its gate lands**: the benchmark tables (`NumberParseBenchmark` targeted + full SunSpider/Dromaeo) are being measured now and will be posted here before this should merge.

`parseInt(s, 10)` at a warmed call site was eligible for the register lane on argument shape but always declined: `ClrFunction` does not — and cannot — override `GetFastCallShape`, since `JsCallDelegate` takes a `JsValue[]` and offers no arity-specialized entry point. Every call paid the pooled argument array, the call-stack frame and the delegate hop. Giving `ClrFunction` an optional fast hook was rejected: +8 bytes on a type allocated in the promise/iterator/class-definition hot paths (137 `new ClrFunction` sites) to speed up two functions.

**What actually blocked migrating them to `[JsFunction]` was an identity requirement, and it was being satisfied by a leak.** `Number.parseInt === parseInt` held through `ClrFunction.Equals` comparing the underlying delegate across two *distinct* objects — and the distinctness showed: `parseInt.length` was 2 while `Number.parseInt.length` was 0, contradicting the very sameness the equality asserts. They are now one realm intrinsic each (`Intrinsics.ParseInt`/`ParseFloat`, installed by both owners via `[JsIntrinsicReference]`, exactly as `eval` already is), so the sameness is reference identity and the lengths agree at their spec values.

Guards per the `FastCallLaneBenchmarks` precedent: `Arg0 = String` so `ToString` is the identity; `Arg1 = Number | Undefined` on `parseInt` so `ToInt32` is a no-op or the `0` an omitted radix produces; `Arg1 = Any` on `parseFloat`, which never reads it. Neither body raises a JS error. An object argument fails the guard, keeps its frame, and `error.stack` still contains `at parseInt` — both directions pinned by new theories, and the Debug leg (where `LeafCallGuard` is armed) passes.

Two fixes ride along: `NumberConstructor` no longer eagerly allocates two `ClrFunction`s per realm, and the global copy no longer takes its `%Function.prototype%` from whichever realm happened to be running when the lazy factory first ran — a `ShadowRealm` assertion pins it.

Behaviour changes to note: `Number.parseInt.length` 0 → 2 and `Number.parseFloat.length` 0 → 1 (spec-correct; test262 has no length test for either), and the two are no longer `ClrFunction` instances.

Full solution green on both TFMs; test262 exactly 99,738 / 0 / 157.

### DIFF
--- a/Jint.Tests/Runtime/FastCallLaneTests.cs
+++ b/Jint.Tests/Runtime/FastCallLaneTests.cs
@@ -154,6 +154,7 @@ public class FastCallLaneTests
     [InlineData("\"abc\".substring(0, p)", "substring")]
     [InlineData("\"abc\".slice(p)", "slice")]
     [InlineData("\"abc\".slice(0, p)", "slice")]
+    [InlineData("parseInt(\"10\", p)", "parseInt")]
     public void AWarmedSiteKeepsTheBuiltinFrameWhenAnArgumentCoercesThroughUserCode(string call, string builtin)
     {
         var engine = new Engine();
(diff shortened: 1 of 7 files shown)


lahma:
## Benchmark gate

**Targeted — `NumberParseBenchmark`, two pairs, opposite orderings, medians:**

| Row | base | cand | time Δ | alloc Δ |
| **ParseIntLoop** | 18.40 ms | 16.02 ms | **−12.9%** | 0.00% |
| ParseFloatLoop | 22.91 ms | 22.28 ms | −2.8% | 0.00% |
| NumberCoerce (control) | 22.10 ms | 21.76 ms | −1.6% | 0.00% |
| ToFixedLoop (control) | 31.45 ms | 30.45 ms | −3.2% | 0.00% |
| ToStringRadix (control) | 18.88 ms | 19.72 ms | +4.5% | 0.00% |

Per-pair: ParseIntLoop −16.9% then −8.7% — the win is robust across both orderings. **ParseFloatLoop is smaller than designed** (−0.4% / −5.1%): `parseFloat`'s double-parsing body dominates the call overhead the lane removes. `ToStringRadix` flagged +7.1% in pair 1 and did **not** reproduce (+1.8% in the reversed pair); its candidate-side draws of identical code differ by 5.6% between runs, so the movement sits inside the row's own demonstrated envelope. Allocation byte-identical on every row of every run.

**Wide — SunSpider (26) + Dromaeo (24) vs the merge-base, single pair:** allocation flat on all 50 rows. The time column is noise-bound and reported as such rather than cherry-picked: the largest movers (`string-base64` −15.6%, `bitops-nsieve-bits` −9.9%, `ObjectRegExp F/F` +11.2%) **contain no `parseInt` call** (grep-verified), and the same untouched rows moved in *contradictory* directions under the unrelated #2969 candidate measured against the same baseline in the same session — the fingerprint of single-draw baseline noise, not of either diff. The one parseInt-using SunSpider script among the moved rows, `date-format-xparb`, improved −4.7%.

Machine idle, serial, default BDN job, both sides run from their own worktree CWD, baseline pinned at the merge-base `20bbf8725`.

PR #2969, opened 2026-08-10
Answer a wrapped dictionary's existence questions from ContainsKey

lahma:
> **Draft-quality until its gate lands**: the benchmark tables (new `InteropDictionaryProbeBenchmark` + controls + full SunSpider/Dromaeo) are being measured now and will be posted here before this should merge.

`ObjectWrapper` had no `ProbeOwnProperty` override, so every existence question — `in`, `hasOwnProperty`, and the per-key filter behind `Object.keys`/`for-in`/spread/`JSON.stringify` — fell to the base implementation, whose answer is the *read*: `TryGetValue`, `FromObject` on the value (for an object value, a whole nested `ObjectWrapper`), a `PropertyDescriptor` around it — then both discarded. Dictionary members are the one thing the wrapper never caches, so this repeated on every call: `Object.keys` over a wrapped JSON document paid it once per key. #2839 compiled `ContainsKey`, but only `HasProperty`'s non-string-keyed arm could reach it — the one shape whose keys never enumerate.

The override answers `Enumerable` on a `ContainsKey` hit and otherwise falls through to `base` — it never decides a key is *missing*, because a wrapped dictionary still resolves CLR members for names the dictionary does not carry. The only possible disagreement is a target whose own `ContainsKey` and `TryGetValue` disagree, which the compiled lanes already trust. Precedence mirrors `GetOwnProperty`'s own order. `TypeDescriptor.CanTestDictionaryKey` exists because `ContainsDictionaryKey` returns `false` both for "absent" and for "there was no `ContainsKey` to ask" — and reading the second as the first is exactly the silent key-dropping the probe contract warns against.

Proof it engages: a counting `IDictionary` shows exactly **one `ContainsKey` and zero `TryGetValue`** for `'a' in doc`. The in-box agreement checker covers plain keys, object-valued keys, a key shadowing a CLR member, a CLR member name, an absent name, an index-shaped name, a symbol, and a script-redefined non-enumerable key; the embedder-side half lives in `Jint.Tests.PublicInterface/HostDictionaryEnumerationTests.cs` with the host dictionary as oracle. Green under `JINT_HOST_CONTRACT_VERIFICATION=1` on both projects — which verifies every consumed probe against `GetOwnProperty` — and AGENTS.md's dictionary gotcha records the new lane.

The new benchmark's own `InMiss` row is the honest control: a miss can only get *slower* (one extra compiled `ContainsKey` on top of the resolver lookup it already paid), and by how much is a number the gate reports rather than hides. `ObjectValues` and `ReadMember` are value-needing controls no existence answer can serve.

Full solution green on both TFMs; test262 exactly 99,738 / 0 / 157.


lahma:
## Benchmark gate

**Targeted — `InteropDictionaryProbeBenchmark` (new in this PR) vs the merge-base `20bbf8725`:**

| Row | base | cand | time Δ | alloc |
| --- | ---: | ---: | ---: | --- |
| **InHit** | 74.2 µs | 49.8 µs | **−32.8%** | 31.86 KB → **624 B** (−98.1%) |
| **HasOwnProperty** | 191.6 µs | 153.6 µs | **−19.8%** | 31.86 KB → **624 B** (−98.1%) |
| **ObjectKeys** | 824.1 µs | 520.5 µs | **−36.8%** | −31.8% |
| **ForIn** | 1.656 ms | 1.320 ms | **−20.3%** | −37.2% |
| ObjectValues | 1.228 ms | 1.020 ms | −16.9% | −29.5% |
| InMiss (cost-of-asking control) | 77.5 µs | 83.1 µs | **+7.2%** | flat |
| ReadMember (control) | 111.8 µs | 109.3 µs | −2.2% | flat |

The allocation column is the byte-deterministic channel and confirms every designed row: the discarded per-probe descriptor and nested wrapper are gone. `ObjectValues` improves despite being listed as a value-needing control, mechanically: its enumerability *filter* no longer materializes a descriptor per key before the values are read. **`InMiss` +7.2% is the predicted, disclosed cost** — one extra compiled `ContainsKey` ahead of the resolver lookup a miss already paid, ≈ 5.6 ns per miss.

**Neighbours:** `InteropNestedDictionaryBenchmark` and `ImmutableCrossingBenchmark` (whose read paths this diff does not enter) showed one +11.0% flag in pair 1; the order-reversed second pair exonerated it — every row within ±2.4%, allocation byte-flat in every run.

**Wide — SunSpider (26) + Dromaeo (24), single pair:** allocation flat on all 50 rows. No SunSpider or Dromaeo script wraps a CLR dictionary, so no row executes the new code; the time swings (both directions, up to ±12%) appear on exactly those unreachable rows and contradict the movements the unrelated #2968 candidate showed on the same rows against the same baseline — single-draw baseline noise, reported as such.

Machine idle, serial, default BDN job, both sides from their own worktree CWD; the new benchmark class was ported byte-identical to the baseline side so both ran the same harness.

PR #2980, opened 2026-08-11
Let the cheapest builtin calls run frameless

lahma:
> Draft until its benchmark gate lands — the release measurement block runs the new `FastCallLaneBenchmarks` rows plus the wide tables against this.

Extends the leaf fast-call lane to the Tier 1 candidates from the release campaign's scouting note. Every premise was re-verified in source; one was refuted **as written** and fixed properly, one was refuted in the other direction.

**The missing mechanism first.** "Unconditional `Leaf` with `Any` guards" was unimplementable: the generator refuses `Leaf` for a plain `JsValue` parameter, and `FastCallGuard.Any` cannot un-refuse it because `Any` *is* the absence of a declaration. The new spelling `FastCallGuard.Unguarded` is declaration-only — the generator lowers it to `Any` in the emitted shape, byte-identical to an unguarded position (snapshot-pinned), zero runtime cost.

**Shipped:**
- `Number.isFinite/isInteger/isNaN/isSafeInteger` — type test first, no `ToNumber`, receiver unread; `Leaf` + `Unguarded`, so even the object-argument calls take the lane (they are validation calls, which is what these functions are for).
- `String.prototype.indexOf/startsWith/endsWith/includes/at/substr` — the `charAt`/`substring` guard shape. The `IsRegExp` unreachability claim for `startsWith`-family was **verified in source, not trusted**: `IsRegExp` answers false for a non-`ObjectInstance` before any `@@match` read, and a guard-passing `JsString` is never one; a `String.prototype[Symbol.match]` cannot change that because the lookup would be on the *argument*, which is a primitive.
- Global `isNaN`/`isFinite` — `Number | String | Undefined` (Symbol/BigInt TypeErrors and object `valueOf` excluded by the guard). The scouting note's registration worry was refuted: these are generated dispatchers, never `ClrFunction`, no identity twin, no `length` change.
- `Array.isArray` — the note called it marginal; the old comment said `Leaf` was impossible. "Not a revoked proxy" is indeed inexpressible, but "one of our own arrays" is exact: `InternalTypes.Array` + the sealed `IsArray()` override make `LeafArg0 = Array` a flag test and a return, and a proxy fails the guard into the frame its TypeError needs.

**Leaf claims were proven by falsification, not by green Debug runs.** A green Debug leg only says no assert fired; each claim was temporarily falsified (guard widened, or a `ToNumber` probe inserted) and the `LeafCallGuard` assert shown to fire across all four call-site shapes, then reverted — the failure transcripts are in the commit history of this PR's preparation. Final Debug legs green everywhere including a 79,489-test Debug test262 with zero leaf-audit failures.

**Behaviour changes: none of #2968's kind.** No identity, `length`, `name` or attribute moves. The one inherent observable — a guard-passing call no longer charges `LimitRecursion` — is invisible (such a call provably runs no user code and raises no JS error), and recursion through user code still charges, pinned by the pre-existing test.

New pins include `valueOf`-count-zero for the Number predicates, the revoked-proxy frame for `isArray`, 38 warm-vs-cold agreement rows, and a warmed `indexOf` handed a throwing `toString` still showing `at indexOf` in `error.stack`.

Release build 0 warnings; test262 byte-identical to the `cf3c048de` baseline (same 8 pre-existing load-flake timeouts on this box, identical on both sides). Gate rows and expectations are listed in the PR preparation notes; >1% on any wide-table row blocks, with the caveat that `indexOf`/`includes`/`isNaN`/`isArray` appear inside the suite scripts themselves, so small improvements there are results, not drift.

### DIFF
--- a/Jint.Benchmark/FastCallLaneBenchmarks.cs
+++ b/Jint.Benchmark/FastCallLaneBenchmarks.cs
@@ -1,4 +1,4 @@
-using BenchmarkDotNet.Attributes;
+﻿using BenchmarkDotNet.Attributes;
 using Jint.Native;
 
 namespace Jint.Benchmark;
@@ -59,6 +59,18 @@ namespace Jint.Benchmark;
 /// <c>Substring_Object</c> is flat to better. Conforming arguments win roughly −5%, and the variadic
 /// rows −12..−22%. If a future change makes an object argument materially worse than this, that is a
 /// regression; the single-digit standing cost is the trade.</para>
+///
+/// <para><b>Predicates and searches.</b> The third group covers the built-ins whose bodies are a type
+/// test or a search rather than a coercion. <c>Number.isInteger</c> and its siblings inspect the
+/// argument and never coerce it, so they are declared <c>FastCallGuard.AnyValue</c> and take the
+/// frameless branch for every value — which is why <c>NumberIsInteger_Object</c> is the one
+/// <c>*_Object</c> row in this class that is <em>not</em> a control: it is expected to improve with
+/// its guarded sibling, and it is here to show that a body which only inspects its argument costs an
+/// object nothing extra. Every other <c>*_Object</c> row is a control in the usual sense.
+/// <c>IsNaN_String</c> is included because the global pair's guard admits strings, where the
+/// String.prototype rows admit numbers, so it is the only row that exercises that arm.
+/// <c>ArrayIsArray_Object</c> is a true control twice over: the guard names our own arrays, so a
+/// plain object both fails it and is the answer-is-false case the frame still has to cover.</para>
 ///
 /// Every row runs its call 1000 times inside one prepared script on an engine that has already
 /// evaluated it, so what is measured is dispatch plus body rather than parse or first-call warmup —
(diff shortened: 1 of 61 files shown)


lahma:
## Benchmark gate — PASS

Two order-reversed pairs vs the merge-base, medians. Every designed row improves with tight candidate spreads; the object-argument controls sit inside the class's documented envelope; allocation byte-identical everywhere.

| Row | Δ (median of 2 pairs) |
| --- | ---: |
| StringAt_Guarded | −19% (pair 1) / consistent pair 2 |
| StringIncludes_Guarded | **−12.0%** |
| NumberIsInteger_Object | **−11.7%** |
| StringStartsWith_Guarded | **−10.8%** |
| IsNaN_String | **−10.5%** |
| NumberIsInteger_Number | **−9.6%** |
| ArrayIsArray_Array | **−7.3%** |
| IsNaN_Number | −3.0% |
| controls: StringIncludes/Substring/StringAt _Object | +1.4% to +3.9% (envelope ≈ +5%) |

Honest context: the class's pre-existing rows showed base-side spreads up to 26% between the two pairs (`CharCodeAt_Guarded`), which is the known sub-200 ns scatter — the designed rows' candidate spreads are 0.1–2.9%, and their direction agrees in both orderings, which is what the two-pair rule requires. Machine idle, serial, default job, node reuse disabled.

PR #2984, opened 2026-08-11
Let Map and Set lookups run frameless

lahma:
> Draft until its benchmark gate lands — the release measurement block runs the new rows, `MapSetLookupBenchmark`, and the wide tables against this.

`Map.prototype.get/has/set/delete` and `Set.prototype.has/add/delete` claim `Leaf`, so a warm call site with a real Map/Set receiver skips the built-in's call-stack frame. All seven already had the register half of the lane; `map.get(k)` is a hot path in real embeddings.

**A bit, not a type test.** `InternalTypes.Map`/`Set` bits, aliased by `FastCallGuard.Map`/`Set`, keep `Satisfies` at one mask test for passing *and* failing receivers. The Date precedent (spare bit + `is JsDate` type test) was rejected deliberately: two more type-tested kinds lengthen the fallthrough tail every failing guard of every kind pays. Both classes are `public sealed` and the bit is set in the single constructor every instance — subclasses included — reaches, so the bit is exactly `is JsMap`/`is JsSet` and a host cannot forge it. Bit headroom is now documented; one free int bit remains.

**One bit each, not a shared collection bit** — verified that `Map.prototype.get.call(aSet, k)` throws today, so a shared bit would have made it throw *frameless*; pinned both by a wrong-brand theory and by `(Map & Set) == Empty` in the guard/flags agreement test.

**A second declaration-only guard rode along.** A Map key is only hashed and compared (`SameValueZero`), never converted, so no key reaches user code — but the generator blocks `Leaf` for a plain `JsValue` parameter and `Any` is also the "undeclared" sentinel. `FastCallGuard.AnyValue` expresses the author's claim; the generator resolves it to `Any` so no shape ever carries it, with a `Debug.Assert` making sure. Without it the guard would have been `String | Number`, sending the *object keys a Map exists for* back to the framed path.

**Mutating leaves verified safe**: frame elision skips only push/recursion-charge/pop; no version counter or iteration state hangs off the frame. Pinned by a test that mutates through warmed frameless sites and reads back via `size`, `forEach`, a live iterator and spread. Honest residual, documented in-code: a wrapped CLR key runs the host's `Equals`/`GetHashCode` — host code, identical on the framed path.

**Declined, with reasons in-code:** WeakMap/WeakSet (`set`/`add` TypeError on a non-weakly-holdable *argument* no guard expresses; two more bits on a colder path); a wrong-receiver benchmark row (every guard failure is a call that throws, so the row would time exception construction — replaced by `WeakMapHas_Framed`/`WeakSetHas_Framed` as same-shape framed controls); `getOrInsert` (not even `FastCall` today; own change).

**Leaf audit by falsification**: dropping only the receiver guard turns the wrong-brand theories red with the `LeafCallGuard` assert firing from the frameless branch; a temporary in-body probe proved the frameless branch is entered for guarded receivers including subclass instances. Both reverted; Debug legs green (5179/5098), verification legs green, test262 byte-identical to baseline including the same 8 pre-existing load-flake timeouts.

Flagged for someone to chase, unrelated to this change: three full-suite runs died with a test-host crash that `--blame-crash` pins on `Atomics_waitAsync_bigint` `no-spurious-wakeup-on-exchange.js` (a `$262.agent` worker test); 6/6 clean filtered repeats afterwards, and a concurrent test262 run from another worktree was contaminating the blame directory — environmental here, but a pre-existing flake.


lahma:
## Benchmark gate — PASS

Single clean pair vs the merge-base (node reuse disabled, idle machine). The verdict rests on the millisecond-scale `MapSetLookupBenchmark` rows — far above the ns-scatter floor — plus the deterministic allocation column:

| Row | base | cand | Δ |
| --- | ---: | ---: | ---: |
| **MapGetHit** | 16.23 ms | 14.08 ms | **−13.3%** |
| **MapSetDeleteChurn** | 31.21 ms | 28.09 ms | **−10.0%** |
| **MemoizePattern** | 19.24 ms | 17.74 ms | **−7.8%** |
| SetHasMixed / MapGetMixed / MapHasMixed | | | −2.0% / −1.6% / −1.4% |
| new ns rows: MapSet_StringKey / SetAdd_Churn / SetHas_ObjectValue / SetHas_NumberValue / MapDelete_Churn / MapGet_Miss | | | −13.1% / −12.4% / −11.9% / −9.3% / −6.7% / −4.9% |
| framed controls: WeakMapHas_Framed / WeakSetHas_Framed | | | +0.2% / −3.4% |

Allocation byte-identical on every row of the class and its neighbours.

One unresolved ns-scale cell, disclosed: `MapGet_ObjectKey` read +11.7% in this pair — but untouched neighbour rows in the same class swung up to +20% (`Substring_Guarded`, which read **−15%** in a different session's pair), so a single ns-scale reading cannot be attributed. The ms-scale `MapGetMixed` row, which is half object keys, improved −1.6%, which bounds any real object-key effect to well under the single-row reading.

lahma:
Rebased over #2975 (proper tail calls). The rebase was a real conflict, not a textual one: **PTC took `InternalTypes` bit `1 << 26` for `TailCall` — the bit this PR had assigned to `Map`.** `Map`/`Set` moved to `1 << 27`/`1 << 28`, consuming the last free flag bit (the headroom comment now says so), and the renumbering propagated through all three coordinated copies: the runtime enum, the source generator's mirror — whose JINT023 drift diagnostic caught the mismatch exactly as designed — and the generator test fixture, caught by its Verify snapshot.

Full post-rebase gate: solution green on both TFMs (5,525/5,443 Jint.Tests), generator snapshots 52/52, and test262 at the **new post-PTC baseline: 99,779 passed / 0 failed / 122 skipped** — #2975 un-parked the 35 tail-call exclusions and they pass.

PR #2985, opened 2026-08-11
Stop allocating a JsString for every inherited-global resolution

lahma:
> Draft until its gate lands — the two new benchmark rows plus `GlobalAccessBenchmarks` flat-check run in the release measurement block.

`GlobalEnvironment.HasBinding(Key)` and `GetBindingValueUnlikely` called `JsString.Create(name.Name)` on the prototype-resolution path — a fresh 32-byte `JsString` per miss. The review flagged it; **instrumentation then corrected the review's own probe**: `typeof undeclared` never reaches either site (it resolves through the `BindingName` overloads into an unresolvable reference). The shapes that actually pay, measured with counters over 1000 iterations:

| workload | `HasBinding(Key)` | `GetBindingValueUnlikely` |
|---|---|---|
| `typeof valueOf` (no host setup — `Object.prototype` is the global's default prototype) | 0 | **1000** |
| call of an inherited global | 0 | **1000** |
| destructure into an inherited accessor / non-writable target | **1000** | 0 |

Both callers already hold the name as a `JsString` — `Engine.GetValue` in the reference it is reading, `GetIdentifierReference` in the reference it was about to allocate anyway. Threaded through via `GlobalEnvironment`-only internal overloads plus one `is GlobalEnvironment` test at the two call sites: no new vtable slot, `Key`-only entry points unchanged.

Alloc probe (`#if NET`, per the in-repo precedent), bytes per operation over 20,000 iterations: `typeof` inherited **32.03 → 0.03**, call of inherited **32.03 → 0.03**, destructure into accessor **200.05 → 168.05** — exactly one `JsString` per operation, gone.

`GlobalAccessBenchmarks` had **no row exercising the miss path** (all five resolve names the global owns); two are added with their own prototype-bearing fixture — `InheritedGlobalReadLoop` and `InheritedGlobalDestructureLoop` — smoke-checked through the REPL for genuine per-iteration misses, not run. The gate also checks the pre-existing five rows for the cost of the new type test on the hit path.

test262 at baseline (the 8-test S7.8.5 load flake identical on unmodified `cf3c048de`).

### DIFF
--- a/Jint.Benchmark/GlobalAccessBenchmarks.cs
+++ b/Jint.Benchmark/GlobalAccessBenchmarks.cs
@@ -27,12 +27,28 @@ public class GlobalAccessBenchmarks
     private IsolatedScript _globalUpdateLoop;
     private IsolatedScript _nestedGlobalReadLoop;
     private IsolatedScript _nestedGlobalWriteLoop;
+    private IsolatedScript _inheritedGlobalReadLoop;
+    private IsolatedScript _inheritedGlobalDestructureLoop;
 
     /// <summary>Builds a fresh engine carrying the fixture every row needs, and nothing else.</summary>
     private static Engine CreateEngine()
     {
         var engine = new Engine(static options => options.Strict());
         engine.Execute("var gx = 0; var gy = 0; var gz = 0; var gobj = null; var gsum = 0; var gval = 0;");
+        return engine;
+    }
+
+    /// <summary>
+    /// The fixture for the two inherited-name rows, and for them alone: a prototype on the global object
+    /// carrying names that resolve as bare identifiers without being own properties of the global — the
+    /// shape a host gets from projecting an object as the global's [[Prototype]]. Installing it in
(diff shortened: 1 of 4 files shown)


lahma:
## Benchmark gate — PASS

Single clean pair vs the merge-base. The designed claim is allocation, and allocation answers:

| Row | base alloc | cand alloc | alloc Δ | time Δ |
| **InheritedGlobalReadLoop** | 6.10 MB | **568 B** | **−99.99%** | −3.5% |
| **InheritedGlobalDestructureLoop** | 9.54 MB | 8.01 MB | **−16.0%** | −4.0% |
| GlobalVarLoop / LocalVarLoop / NestedGlobalRead / NestedGlobalWrite / GlobalUpdate (hit-path controls) | | | 0.00% | −0.1% to −13.8% |

**Zero rows regressed in time** (0 above +1%), so the one-per-hop `is GlobalEnvironment` type test the fix adds to the hit path is invisible — the hit-path controls actually drew faster this session. Machine idle, serial, node reuse disabled.

PR #2994, opened 2026-08-11
Let a constraint exception unwind deep recursion without overflowing

lahma:
> The call-path gate rows (`RecursionBenchmark`, `ClosureCallBenchmarks`, `MethodCallBenchmark`, SunSpider `controlflow-recursive`, full wide tables) run in the release measurement block — merging should wait for them.

**`Options.LimitRecursion(N)` killed the host process for most useful values of N.** The constraint fired correctly at depth N — and then the unwind itself overflowed the stack: `JintStatementList.Execute` and `JintBlockStatement.ExecuteSingle` caught `Exception`, mapped four types to a throw completion, and **rethrew everything else from inside the handler at every interpreter level**. A rethrow from a handler restarts two-pass dispatch, so the runtime cannot collapse frames it has already searched — ~10 KB per level, live for the rest of the unwind. A JS `throw` unwound 1,153 frames; a CLR rethrow managed **158**. So `LimitRecursion(100)` worked and **200 through ~1040 crashed the process** — precisely the range a host sandboxing untrusted script picks. CI never saw it because the test helper defaults to 16 MB stacks. Every constraint that unwinds as a CLR exception shared the defect: cancellation and a CLR host exception at depth measured the same ceilings.

The fix is exception **filters** on exactly the two maybe-rethrow sites (the only callers of `HandleException`), so unmappable exceptions pass through uncaught and the runtime keeps one collapsible first-pass search. The filter is four `isinst` tests and a `JintException` check — no allocation, no engine access; filters run one at a time and each funclet returns before the next, so peak extra stack is one funclet frame, not N. Error decoration deliberately stays a *handler*: the innermost frame catches once, decorates, rethrows exactly once for the whole unwind — `BuildCallStackString` never runs in the first pass, and the CLR decoration is byte-identical before/after.

Measured (bisected, out-of-process):

| unwindable JS frames | 1.5 MB apphost | 1 MB thread |
| --- | --- | --- |
| `LimitRecursion` | 154 → **1082** | 102 → **716** |
| cancellation | 154 → **1138** | 102 → **750** |
| CLR host exception | 163 → **1120** | 102 → **735** |
| JS throw (control) | ~unchanged | ~unchanged |

`LimitRecursion(200)`/`(1000)` on both stacks: exit `0xC00000FD` before, clean exit with a catchable `RecursionDepthOverflowException` after, engine reusable (empty call stack, correct `fib(20)` three rounds, a 20,000-call workload exact).

**Codegen, measured not guessed:** `JitDisasm` shows the prolog grow `sub rsp, 152 → 168` — +16 bytes per frame, and a variant whose filter reads nothing measures the same, so it is the cost of having a filter clause at all. The non-throwing instruction sequence is otherwise unchanged. That is what the gate rows adjudicate.

Sites audited and deliberately left: the per-level `catch { bookkeep; throw; }` sites (`for-in/of`, destructuring, await/yield) always rethrow and their bookkeeping must run — converting `for-of`'s would mean assigning the `finally`-read completion from a filter, a separate riskier change; measured improvement for for-of-driven recursion is real but partial (51 → 121 frames) and recorded. `Engine.cs`'s entry catch runs once per entry and keeps `ResetCallStack`.

Seven new tests on 1 MB `DedicatedThread`s; the red run **crashes the test runner** (`0xC00000FD`). Verification legs green; test262 idle-machine standalone 99,738 / 0 / 157.

### DIFF
--- a/Jint.Tests/Runtime/ExecutionConstraintTests.cs
+++ b/Jint.Tests/Runtime/ExecutionConstraintTests.cs
@@ -483,6 +483,148 @@ myarr[0](0);
         Invoking(() => engine.Evaluate(code)).Should().ThrowExactly<RecursionDepthOverflowException>();
     }
 
+    // A constraint that fires deep in a recursion has to survive its own unwind. Every one of these
+    // runs on a small dedicated stack, because that is the only place the failure shows: an exception
+    // caught and rethrown at each interpreter level costs a nested exception dispatch per level, and
+    // enough of those overflow the stack on the way out and take the host process down with them
+    // (0xC00000FD, uncatchable). The 16 MB DedicatedThread default hides all of it, which is why CI
+    // never saw it. Every depth below sits well above the pre-fix ceiling (~100 JavaScript frames on
+    // a 1 MB stack) and well below the plain forward-recursion ceiling (~700), so a regression fails
+    // the run rather than merely reporting a smaller number.
+    private const int SmallStack = 1024 * 1024;
+
+    [Fact]
+    public void RecursionLimitFiringAtTwoHundredUnwindsToTheHost()
+    {
+        DedicatedThread.Run(
+            () =>
+            {
+                var engine = new Engine(o => o.LimitRecursion(200));
+                Invoking(() => engine.Execute("function f() { return f(); } f();"))
+                    .Should().ThrowExactly<RecursionDepthOverflowException>();
+            },
+            maxStackSize: SmallStack);
+    }
+
+    [Fact]
+    public void RecursionLimitFiringAtAThousandUnwindsToTheHost()
+    {
+        // 4 MB: a thousand frames do not fit forward on 1 MB. The pre-fix unwind ceiling scales with
+        // the stack too (~400 frames here), so a thousand still fails without the fix.
+        DedicatedThread.Run(
+            () =>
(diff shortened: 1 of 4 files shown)


lahma:
## Benchmark gate — PASS

Two instruments, agreeing:

**BDN, two order-rotated pairs vs the merge-base, medians:** the recursion rows — the shape where the filter clause's +16 bytes/frame is paid most often — sit inside the pair-to-pair spread (`Fib`/`DeepSum`/`Tak` read +1.7–2.3% with base-side spreads of 1.4–4.6%, and the same class shows untouched rows at ±8% both directions, e.g. `UserPrototypeMethod` −8.8%). Allocation byte-identical on all 18 rows.

**Alternating single-script REPL driver, 5 reps, per-rep pairing** (immune to the cross-process scatter): deep-recursion median **+0.7%**, hot-leaf-call median **−1.0%**.

So the non-throwing cost is at or below one percent on the worst shape, for a change that turns `LimitRecursion(200)` from a process kill into a catchable exception. The wide SunSpider+Dromaeo tables from the same session showed allocation flat everywhere and time swings attributable to the shared single-draw baseline (two unrelated candidates moved the same untouched rows in contradictory directions).

PR #3003, opened 2026-08-11
Ask the array again when join meets a hole

lahma:
> `ArrayHoleTraversalBenchmark.JoinHoley` is the one row that can move; it runs in the release measurement block with `JoinDense` and the class's other rows as controls.

`[0, funky, , ,].join("")` where `funky.toString()` sets `Object.prototype[3]` answered `"0funky"` — node and [the spec](https://tc39.es/ecma262/#sec-array.prototype.join) (`Get(O, ToString(k))` per element, every iteration) answer `"0funkyfnord"`. `ArrayReadOperations.Get`, the read-only fast lane, snapshotted `_dense` at construction and turned a hole into `undefined` without ever asking the array.

A `null` slot (or an index past the snapshot) now routes to `JsArray.Get(uint)`, which still short-circuits to `Undefined` while `CanUseFastAccess` holds and walks the chain only after the realm's array prototype chain has actually gained an index property. **A slot the snapshot holds is untouched**: one bounds compare, one load, one null test — identical to before. The red matrix needed one engine per case to expose, because the first prototype index write sets a sticky realm-wide flag that clears fast access — that mechanism is exactly why the pristine-realm fast path is safe.

**Siblings disclosed, deliberately not fixed here:** `toLocaleString` reads via `TryGetValue` and still short-circuits (spec says `Get`); `ArrayReadOperations.HasProperty` answers existence from the same snapshot, so `forEach`/`every`/`some`/`filter`/`indexOf` miss an index the prototype gained mid-iteration; and a side effect that *reallocates* `_dense` leaves the snapshot itself stale — fixing that one means a per-element field load on the hot path and needs its own benchmark gate.

12 new tests, red on exactly the 4 join/toString rows, both TFMs. test262 standalone 0 / 99,744 / 157, identical.

### DIFF
--- a/Jint.Tests/Runtime/ArrayTests.cs
+++ b/Jint.Tests/Runtime/ArrayTests.cs
@@ -43,6 +43,76 @@ public class ArrayTests
         result.Should().Be("[[null,false,null,false],[\"ap\",true,\"op\",true]]");
     }
 
+    /// <summary>
+    /// https://tc39.es/ecma262/#sec-array.prototype.join reads each element with <c>Get(O, ToString(k))</c> on
+    /// its own iteration, so a side effect from one element's <c>ToString</c> is visible to every later one.
+    /// The read-only array lane snapshots the backing store, and used to answer a hole from that snapshot as
+    /// <c>undefined</c> without ever asking the array — so an index the prototype chain gained mid-join was
+    /// silently skipped. Expectations read off node 24.
+    /// <para>
+    /// Each case needs its own engine: the first index property written to Array.prototype or Object.prototype
+    /// clears the pristine-prototype invariant for the whole realm, permanently, which routes every later array
+    /// operation down a different lane.
+    /// </para>
+    /// </summary>
+    [Theory]
+    // the reported case: Object.prototype gains index 3 while element 1 is being coerced
+    [InlineData("Object.prototype[3] = 'fnord'", "0funkyfnord")]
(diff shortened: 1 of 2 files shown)


lahma:
## Benchmark gate — measured, reworked once, now PASS

The first measurement caught a real cost: `ArrayHoleTraversalBenchmark.JoinHoley` at **+54.9%** — every hole paid a non-inlined `JsArray.Get` call even when `CanUseFastAccess` proves nothing can shadow it. The fix now answers a hole inline while that flag holds (re-read per hole, because a side effect during the very join is exactly what can clear it) and only falls back to the spec's chain-walking read once it is gone.

Re-gated after the rework:

| Row | base | cand | Δ |
| --- | ---: | ---: | ---: |
| **JoinHoley** | 536.2 µs | 554.7 µs | **+3.4%** |
| JoinDense | 945.2 µs | 949.3 µs | +0.4% |
| SumHoley / SumDense / InOperatorHoley / IndexOfMissHoley | | | −0.7% to −0.2% |
| IndexOfMissDense (untouched) | | | −6.9% (scatter) |

The residual +3.4% on the hole-heavy row is the honest cost of the correctness fix — per-hole flag read and branch where there used to be nothing — disclosed rather than hidden. Allocation byte-identical everywhere. ArrayTests 167/167 on both TFMs after the rework; the amended commit is force-pushed.

PR #3015, opened 2026-08-14
Give JavaScript strings a maximum length instead of a wrapped array rent

### LINKED ISSUE #3011: A string past the implementation limit overflows the builder instead of throwing RangeError
A string operation whose result exceeds what we can allocate does not fail as JavaScript, it fails as an arithmetic overflow inside the builder.

Found running `staging/sm/String/replace-math.js` (see the staging issue for why that file is not in our normal run): a `String.prototype.replace` substitution producing more than about 2³¹ characters overflows the length arithmetic in `ValueStringBuilder.Grow`, which then calls `ArrayPool<char>.Rent(-2147483648)` and throws an `ArgumentOutOfRangeException` straight out of `engine.Evaluate`, past every `catch` in the script.

V8 answers `RangeError: Invalid string length`, which is what a script can actually handle.

### What the fix needs

The natural place to raise the error is where the overflow is detected, but `GetSubstitution` is `static` and has no realm to raise a `RangeError` from — so this is either a signature change to thread the realm through, or a sentinel the caller turns into a `RangeError`. Worth deciding once and applying to every builder path that can overflow, not just this one: concatenation, `repeat`, `join` and `padStart`/`padEnd` all reach the same arithmetic.

While in there, it is worth deciding whether we want an explicit maximum string length at all (V8 has one, and it is well below `int.MaxValue`), since "as much as the allocator happens to give us" is not a limit a host can reason about.

_Recorded during the 4.16.0 pre-release review._

lahma:
`ValueStringBuilder.Grow` computes the size it needs as `_pos + additionalCapacityBeyondPos`, in unchecked `int`. Past 2³¹ characters that sum wraps negative; the `(uint)` → `Math.Max` → `(int)` round-trip carries the wrapped value straight through, and `ArrayPool<char>.Shared.Rent(-2147483648)` throws. The `ClrLimits.MaxArrayLength` cap sitting in that very expression bounds only the *doubling* term — never the required-size term — so nothing stopped it.

The result is a CLR exception escaping `engine.Evaluate` past every `catch` in the running script:

```
System.ArgumentOutOfRangeException : minimumLength ('-2147483648') must be a non-negative value. (Parameter 'minimumLength')
   at System.Buffers.SharedArrayPool`1.Rent(Int32 minimumLength)
   at System.Text.ValueStringBuilder.Grow(Int32 additionalCapacityBeyondPos) in Jint/Pooling/ValueStringBuilder.cs:line 361
   at Jint.Native.RegExp.RegExpPrototype.GetSubstitution(...) in Jint/Native/RegExp/RegExpPrototype.cs:line 538
   at Jint.Native.String.StringPrototype.Replace(...) in Jint/Native/String/StringPrototype.cs:line 1328
```

That is the new `ReplaceMathTargetsMoreCharactersThanAStringCanHold` test run against the unfixed engine — the script wraps the call in `try`/`catch` and the exception goes straight past it. Three sibling rows reported `System.OutOfMemoryException` the same way; `repeat`/`padStart`/`padEnd` did not throw at all and quietly built ~1 GB strings.

V8 answers `RangeError: Invalid string length`, which a script can handle.

## The limit

`JsString.MaxLength` is V8's `String::kMaxLength`, `(1 << 29) - 24` = **536,870,888** characters, so a script that builds too large a string now fails identically on both engines.

> **This is a breaking change above 512M characters.** `repeat` and `padStart`/`padEnd` used to be capped at `ClrLimits.MaxArrayLength` (2,147,483,591) and every other path was capped at nothing; anything between the two limits used to succeed (or die of an out-of-memory) and now raises a catchable `RangeError`. That is a deliberate trade for cross-engine-identical failures. `ClrLimits.MaxArrayLength` stays where it is — it is the *CLR* allocation ceiling and still bounds non-string builders.

## Three layers

**1. `Grow` can no longer wrap.** The required size is computed as a `long`, and a size no array could hold reports as the `OverflowException` it is. No JavaScript `RangeError` at this level: the builder also backs the JSON serializer, URI encode/decode and the Intl/Temporal formatters (~40 call sites), none of which has a realm to throw into. Its job here is only to stop lying about the size.

**2. The JavaScript-semantic paths** throw `RangeError: Invalid string length` — a message already in use — checked *before* the piece that would take the result past the limit is appended, so a rejected build never first allocates what it is about to refuse:

| Path | Site | Before |
| --- | --- | --- |
| `String.prototype.repeat` | `StringPrototype.Repeat` | capped at `ClrLimits.MaxArrayLength` → retargeted |
| `padStart` / `padEnd` | `StringPrototype.StringPad` | capped at `ClrLimits.MaxArrayLength` → retargeted |
| `GetSubstitution` | `RegExpPrototype.GetSubstitution` | none |
| `@@replace` accumulator | `RegExpPrototype.Replace` | none (plain `string` concat → OOM) |
| `String.prototype.replace` tail | `StringPrototype.Replace` | none |
| `String.prototype.replaceAll` | `StringPrototype.ReplaceAll` | none |
| `Array.prototype.join` | `ArrayPrototype.Join` | none |
| `Array.prototype.toLocaleString` | `ArrayPrototype.ToLocaleString` | none |
| `+` / `+=` | `JintBinaryExpression.ApplyAdditionToPrimitives`, `AdditionChainExpression.EvaluateWithoutSuspension` / `ConcatAll`, `JintAssignmentExpression.ComputeCompound` | none |
| `String.prototype.concat` | `StringPrototype.Concat` | none |
| template literals | `JintTemplateLiteralExpression` | none |
| `String.raw` | `StringConstructor.Raw` | none |

**3. `GetSubstitution` also gets an up-front estimate.** A replacement pattern expands without bound (`"$&$&$&…"` repeats the match once per token), so a per-append check alone would only fire after the builder already held half a billion characters — a gigabyte, allocated purely to be discarded, which is exactly what the reported repro does. `MinimumSubstitutionLength` sums the token contributions that are knowable without running user code (`$&`, the two context tokens, `$n`/`$nn`) and counts everything else as zero. That is an exact *lower* bound, so it can never over-count and never refuse a substitution that fits; it just lets the pathological case be refused while the builder is still empty, which is also what V8 does. `$<name>` is the one token deliberately skipped, because resolving it runs a `Get` that must not run twice; the per-append checks remain the exact enforcement and cover it. This estimate is what makes the tests below cost a few megabytes rather than a gigabyte.

## The realm parameter on `GetSubstitution`

`GetSubstitution` is `internal static` with exactly three callers — `RegExpPrototype.Replace`, `StringPrototype.Replace`, `StringPrototype.ReplaceAll` — and all three are instance methods holding `_realm`, so it simply takes a `Realm` now. No `Throw.CreateRangeError` / `ErrorDispatchInfo` sentinel: that exists for genuinely realm-less callers and would be indirection for nothing here.

`String.prototype.concat` was `static` and needed the same thing, so it became an instance method (the source generator handles both; `Repeat` next door is already instance + `FastCall`).

## Hot paths touched, and what to watch

The check is a widened add and a compare against a constant. On the interpreter paths the realm is resolved **only inside the cold branch** (`context.Engine.Realm` walks `_realmInConstruction ?? ExecutionContext.Realm`), so the warm path never touches it.

Methods touched that are genuinely hot:

- `JintBinaryExpression.ApplyAdditionToPrimitives` — the string branch materializes both operands into locals before `string.Concat` (it built the same two strings before, via `+`), plus one add and one compare. Gains an `EvaluationContext` parameter, unused on the numeric paths.
- `AdditionChainExpression.EvaluateWithoutSuspension` and `ConcatAll` — one add per operand, one compare per chain; the ≥5-operand paths accumulate the total in the loop that was already filling the `string[]`.
- `JintAssignmentExpression.ComputeCompound`, `+=` string branch — the coercion moved out of `JsString.Append` to the call site (same coercion, same order), plus one virtual `Length` read, one add and one compare.
- `JsString.Append` / `JsString.ConcatenatedString.Append` — signature changed from `Append(JsValue)` to `Append(string)`; the callers coerce, because they are the ones that need the length before the append.
- `JintTemplateLiteralExpression.EvaluateInternal` — one add and one compare per quasi and per interpolation.
- `ArrayPrototype.Join` — one add and one compare per element; the element read moved ahead of the separator append so a single check covers both (appending the separator is not observable, so the reordering is not either).
- `RegExpPrototype.GetSubstitution` — one extra pass over the *replacement pattern* (typically a handful of characters), on the `$`-containing path only.
- `StringPrototype.Concat` — static → instance.

Benchmark rows worth watching in the SunSpider + Dromaeo gate: **string-base64**, **string-fasta**, **string-tagcloud**, **string-unpack-code**, **string-validate-input**, **3d-raytrace** and **access-nbody** (heavy `+`), plus Dromaeo's **string-base** and **object-string** rows. I have deliberately run **no** benchmarks — that gate is yours, serially, on an idle machine.

## Tests

New `Jint.Tests/Runtime/StringLengthLimitTests.cs`, including a port of `staging/sm/String/replace-math.js` (`staging/` is not part of Jint's generated test262 projection, so it lives here — the same precedent as `DataViewBoundsTests`). Every row asserts the throw is something the *script* catches: the `try`/`catch` is written in JavaScript, so a CLR exception escaping the engine fails the test as an escaping exception rather than being mistaken for a handled error. Every row is also a path whose limit is decided from small inputs, so nothing allocates 512 MB — the whole class runs in well under a second.

A theory of ten substitution patterns that *fit* pins that the up-front estimate never over-counts (`$$`, `$&`, both context tokens, `$1`/`$12`/`$0`, `$<name>` with and without named groups, and a `$&` spelled inside a group name).

Updated to pin what is actually enforced: `ExecutionConstraintTests.ShouldThrowRangeErrorWhenPadStartExceedsMaxStringLength`, `ShouldLimitStringSizeForPadEnd` (its 536870911 target is now past the cap, so it uses `JsString.MaxLength` — still allowed, still built incrementally, still interrupted by the memory limit) and `StringTests.RepeatRejectsCountsThatExceedTheMaximumStringLength`.

All green in Release:

| Suite | net10.0 | net472 |
| --- | --- | --- |
| `Jint.Tests` | 5678 passed, 4 skipped | 5594 passed, 4 skipped |
| `Jint.Tests.PublicInterface` | 1486 passed, 9 skipped | 1485 passed, 9 skipped |
| `Jint.Tests.CommonScripts` | 28 passed | 28 passed |
| `Jint.Tests.Test262` | 99779 passed, 122 skipped | — |

Closes #3011

### DIFF
--- a/Jint.Tests.Test262/Test262Harness.settings.json
+++ b/Jint.Tests.Test262/Test262Harness.settings.json
@@ -60,11 +60,6 @@
     // Removed by the fix for #3010: five Temporal members the proposal dropped are still present.
     "staging/Temporal/removed-methods.js",
 
-    // Removed by the fix for #3011: String.prototype.replace has to raise a RangeError when the
-    // result would exceed the maximum string length; today the length overflows to int.MinValue and
-    // surfaces as an ArgumentOutOfRangeException out of the BCL.
-    "staging/sm/String/replace-math.js",
-
     // Cross-realm error identity. $262.createRealm() works, but an error Jint raises while running
     // a function belonging to another realm is constructed from the *current* realm's intrinsics,
     // so `e instanceof otherGlobal.TypeError` is false. The realm's global also does not carry its
--- a/Jint.Tests/Runtime/ExecutionConstraintTests.cs
+++ b/Jint.Tests/Runtime/ExecutionConstraintTests.cs
@@ -685,7 +685,7 @@ myarr[0](0);
     [Fact]
     public void ShouldThrowRangeErrorWhenPadStartExceedsMaxStringLength()
     {
-        // The result length (2147483647) exceeds ClrLimits.MaxArrayLength, so the size cap converts
+        // The result length (2147483647) exceeds JsString.MaxLength, so the size cap converts
         // a would-be OutOfMemoryException into a catchable RangeError without any constraints set.
         var engine = new Engine();
         var ex = Invoking(() => engine.Evaluate("'x'.padStart(2147483647)")).Should().ThrowExactly<JavaScriptException>().Which;
@@ -696,10 +696,10 @@ myarr[0](0);
     [Fact]
     public void ShouldLimitStringSizeForPadEnd()
     {
-        // The result (536870911) is below the size cap, so it must be built incrementally and the
-        // memory limit must be able to interrupt it instead of allocating ~1 GB up front.
+        // The result is exactly at the size cap rather than past it, so it must be built incrementally
+        // and the memory limit must be able to interrupt it instead of allocating ~1 GB up front.
         var engine = new Engine(o => o.LimitMemory(4_000_000));
-        Invoking(() => engine.Evaluate("'x'.padEnd(536870911, 'ab')")).Should().ThrowExactly<MemoryLimitExceededException>();
+        Invoking(() => engine.Evaluate($"'x'.padEnd({JsString.MaxLength}, 'ab')")).Should().ThrowExactly<MemoryLimitExceededException>();
     }
 
     // https://github.com/sebastienros/jint/issues/2486
--- a/Jint.Tests/Runtime/StringLengthLimitTests.cs
+++ b/Jint.Tests/Runtime/StringLengthLimitTests.cs
@@ -1,0 +1,147 @@
+using Jint.Native;
+
+namespace Jint.Tests.Runtime;
+
+/// <summary>
+/// A JavaScript string has a maximum length, and every way of building one past it has to report the
+/// same catchable <c>RangeError: Invalid string length</c>. Jint had no such limit: the size caps that
+/// existed were the CLR's <c>Array.MaxLength</c>, they covered only <c>repeat</c> and the two pad
+/// methods, and past <c>2^31</c> characters <c>ValueStringBuilder.Grow</c> computed its required size
+/// as an unchecked <c>int</c>, wrapped negative and handed <c>ArrayPool&lt;char&gt;.Shared.Rent</c> an
+/// invalid length. The <c>ArgumentOutOfRangeException</c> that produced escaped <c>Evaluate</c> past
+/// every <c>catch</c> in the script.
+/// <para>
+/// The limit is now V8's own, <c>(1 &lt;&lt; 29) - 24</c>, so a script that builds too large a string
+/// fails identically on both engines. The suite's own coverage of the reported shape is
+/// <c>staging/sm/String/replace-math.js</c> — a 2^20-character subject replaced with 2^15 <c>$1</c>
+/// tokens, so the substitution targets 2^35 characters — and <c>staging/</c> is not part of the
+/// generated test262 projection at all, so the port lives here.
+/// </para>
+/// <para>
+/// Every row below is a path whose limit is decided from small inputs — a repeat count, a pad length,
+/// a replacement pattern's expansion factor — so the throw happens before anything of that size is
+/// allocated and the test costs a few megabytes. The remaining guarded paths (<c>+</c>/<c>+=</c>,
+/// template literals, <c>concat</c>, <c>join</c>, <c>toLocaleString</c>, <c>String.raw</c> and the
+/// accumulators inside <c>replace</c>/<c>replaceAll</c>) accumulate their result one piece at a time,
+/// so their guard can only fire once roughly half a billion characters are already in hand; a test for
+/// those would have to allocate the gigabyte the guard exists to prevent, and there is no cheaper
+/// input that reaches them.
+/// </para>
+/// </summary>
+public class StringLengthLimitTests
+{
+    private readonly Engine _engine = new();
+
+    private const string InvalidStringLength = "RangeError: Invalid string length";
+
+    /// <summary>
+    /// What <paramref name="source"/> throws, as the running script sees it. The <c>try</c>/<c>catch</c>
+    /// is written in JavaScript rather than with <c>Assert.Throws</c> precisely because a CLR exception
+    /// escaping the engine is the bug: it would fail the test as an escaping exception instead of being
+    /// reported here as a handled JavaScript error.
+    /// </summary>
+    private string Caught(string source) => _engine
+        .Evaluate($"(function () {{ try {{ {source}; return 'did not throw'; }} catch (e) {{ return (e instanceof RangeError ? 'RangeError' : e.constructor.name) + ': ' + e.message; }} }})()")
+        .AsString();
+
+    [Fact]
+    public void TheLimitIsTheOneV8Uses()
+    {
+        JsString.MaxLength.Should().Be(536_870_888);
+        JsString.MaxLength.Should().Be((1 << 29) - 24);
+    }
+
+    [Fact]
+    public void RepeatPastTheLimitIsACatchableRangeError()
+    {
+        Caught($"'xx'.repeat({JsString.MaxLength / 2 + 1})").Should().Be(InvalidStringLength);
+        _engine.Evaluate("'xx'.repeat(3)").AsString().Should().Be("xxxxxx");
+    }
+
+    [Fact]
+    public void PadStartPastTheLimitIsACatchableRangeError()
+    {
+        Caught($"'x'.padStart({JsString.MaxLength + 1L})").Should().Be(InvalidStringLength);
+        Caught("'x'.padStart(2147483647)").Should().Be(InvalidStringLength);
+        _engine.Evaluate("'x'.padStart(3, 'ab')").AsString().Should().Be("abx");
+    }
+
+    [Fact]
+    public void PadEndPastTheLimitIsACatchableRangeError()
+    {
+        Caught($"'x'.padEnd({JsString.MaxLength + 1L}, 'ab')").Should().Be(InvalidStringLength);
+        _engine.Evaluate("'x'.padEnd(3, 'ab')").AsString().Should().Be("xab");
+    }
+
+    /// <summary>
+    /// <c>String.prototype.replace</c> with a string search value: the expansion is decided by how many
+    /// <c>$&amp;</c> tokens the replacement holds, so a 2 KB pattern against a 2 MB subject is enough.
+    /// </summary>
+    [Fact]
+    public void ReplaceWhoseSubstitutionPastTheLimitIsACatchableRangeError()
+    {
+        _engine.Execute("var subject = 'x'.repeat(1 << 20); var pattern = '$&'.repeat(1024);");
+
+        Caught("subject.replace(subject, pattern)").Should().Be(InvalidStringLength);
+    }
+
+    [Fact]
+    public void ReplaceAllWhoseSubstitutionPastTheLimitIsACatchableRangeError()
+    {
+        _engine.Execute("var subject = 'x'.repeat(1 << 20); var pattern = '$&'.repeat(1024);");
+
+        Caught("subject.replaceAll(subject, pattern)").Should().Be(InvalidStringLength);
+    }
+
+    [Fact]
+    public void RegExpReplaceWhoseSubstitutionPastTheLimitIsACatchableRangeError()
+    {
+        _engine.Execute("var subject = 'x'.repeat(1 << 20); var pattern = '$&'.repeat(1024);");
+
+        Caught("subject.replace(/x+/g, pattern)").Should().Be(InvalidStringLength);
+    }
+
+    /// <summary>
+    /// The port of <c>staging/sm/String/replace-math.js</c>. Upstream accepts an out-of-memory failure
+    /// as well as a success, because the point there was the arithmetic; here the point is that whatever
+    /// happens is something the script can catch.
+    /// </summary>
+    [Fact]
+    public void ReplaceMathTargetsMoreCharactersThanAStringCanHold()
+    {
+        _engine.Execute("""
+            function puff(x, n)
+            {
+              while (x.length < n)
+                x += x;
+              return x.substring(0, n);
+            }
+
+            var x = puff("1", 1 << 20);
+            var rep = puff("$1", 1 << 16);
+            """);
+
+        Caught("x.replace(/(.+)/g, rep)").Should().Be(InvalidStringLength);
+    }
+
+    /// <summary>
+    /// The up-front expansion estimate that makes the rows above cheap counts only the tokens whose
+    /// contribution is knowable without running user code, so it can never over-count and can never
+    /// refuse a substitution that fits. These are the shapes where over-counting would show up first.
+    /// </summary>
+    [Theory]
+    [InlineData(@"'abcabc'.replace(/b/g, ""[$`|$&|$']"")", "a[a|b|cabc]ca[abca|b|c]c")]
+    [InlineData(@"'abc'.replace(/(a)(b)(c)/, '$3$2$1$12$0$')", "cbaa2$0$")]
+    [InlineData(@"'John Smith'.replace(/(?<first>\w+) (?<last>\w+)/, '$<last>, $<first>')", "Smith, John")]
+    [InlineData(@"'abc'.replace(/b/, '$<x>')", "a$<x>c")]
+    [InlineData(@"'abc'.replace('b', '$<x>')", "a$<x>c")]
+    [InlineData(@"'abc'.replace(/(b)/, '$<a$&b>')", "a$<abb>c")]
+    [InlineData(@"'John'.replace(/(?<n>\w+)/, '$<a$&b>')", "")]
+    [InlineData(@"'aaa'.replaceAll('a', '$&$&')", "aaaaaa")]
+    [InlineData(@"'abc'.replace('b', '$$')", "a$c")]
+    [InlineData(@"'xyz'.replace('y', ""$`-$'-$&"")", "xx-z-yz")]
+    public void SubstitutionPatternsThatFitAreUnaffected(string source, string expected)
+    {
+        _engine.Evaluate(source).AsString().Should().Be(expected);
+    }
+}
--- a/Jint.Tests/Runtime/StringTests.cs
+++ b/Jint.Tests/Runtime/StringTests.cs
@@ -1,4 +1,5 @@
-﻿using Jint.Runtime;
+﻿using Jint.Native;
+using Jint.Runtime;
 
 namespace Jint.Tests.Runtime;
 
@@ -84,10 +85,10 @@ bar += 'bar';
     }
 
     [Fact]
-    public void RepeatRejectsCountsThatCannotFitInClrStringCapacity()
+    public void RepeatRejectsCountsThatExceedTheMaximumStringLength()
     {
         var engine = new Engine();
-        var repeatCount = ClrLimits.MaxArrayLength / 2 + 1UL;
+        var repeatCount = JsString.MaxLength / 2 + 1;
 
         var exception = Invoking(() => engine.Evaluate($"'xx'.repeat({repeatCount});")).Should().ThrowExactly<JavaScriptException>().Which;
 
--- a/Jint/Native/Array/ArrayPrototype.cs
+++ b/Jint/Native/Array/ArrayPrototype.cs
@@ -1846,6 +1846,12 @@ public sealed partial class ArrayPrototype : ArrayInstance
                     _engine.Constraints.Check();
                 }
 
+                // The element is read (and coerced, which can run user code) before the separator is
+                // appended so a single check covers both, and covers them before either is written:
+                // appending the separator is not observable, so the reordering is not either.
+                var element = StringFromJsValue(o.Get(k));
+                JsString.ThrowIfLengthExceeded(_realm, (long) sb.Length + separatorLength + element.Length);
+
                 if (separatorLength == 1)
                 {
                     sb.Append(singleCharSeparator);
@@ -1855,7 +1861,7 @@ public sealed partial class ArrayPrototype : ArrayInstance
                     sb.Append(sep);
                 }
 
-                sb.Append(StringFromJsValue(o.Get(k)));
+                sb.Append(element);
             }
 
             return sb.ToString();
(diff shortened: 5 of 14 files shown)


lahma:
**Not ready to merge despite green CI — the benchmark gate has not been run.**

This PR adds a length check to paths that are hot: `JintBinaryExpression.ApplyAdditionToPrimitives` and `AdditionChainExpression.ConcatAll` (both now also take an `EvaluationContext`), `JintAssignmentExpression.ComputeCompound`'s `+=` string branch, `JsString.Append` / `ConcatenatedString.Append`, and `JintTemplateLiteralExpression.EvaluateInternal`. On the warm path the cost is one widened add and one compare, with the realm resolved only inside the cold throw branch — but that is a design intention, not a measurement.

Per the repo's standing rule, this needs full SunSpider + Dromaeo tables at the **default** job on an idle machine, with >1% on any row blocking. Rows to watch: SunSpider `string-*` (`string-base64`, `string-fasta`, `string-tagcloud`, `string-unpack-code`, `string-validate-input`) and the Dromaeo string rows, against `3d-*`/`math-*` as controls that this diff cannot reach.

Holding until the maintainer schedules the run; the tables will be posted here.

lahma:
## Benchmark gate - BLOCKED

Two full pairs, default job, serial on an otherwise idle machine. Base pinned to `ff980b7bc` (this PR's merge-base, a fixed commit rather than a moving ref).

A row counts as a regression only if it exceeds +1% in **both** pairs. This machine's run-to-run envelope reached about +/-10% on the Dromaeo rows and 23 of 50 rows changed sign between the two pairs, so neither pair proves anything on its own.

| row | pair 1 | pair 2 | verdict | alloc delta |
|---|---:|---:|---|---:|
| `DromaeoBenchmark.CoreEval [Modern=False&Prepared=False]` | -31.43% | +1.22% | envelope | +0 |
| `DromaeoBenchmark.CoreEval [Modern=False&Prepared=True]` | +0.35% | +1.15% | envelope | +0 |
| `DromaeoBenchmark.CoreEval [Modern=True&Prepared=False]` | -0.97% | -1.24% |  | +0 |
| `DromaeoBenchmark.CoreEval [Modern=True&Prepared=True]` | -1.31% | +0.67% |  | +0 |
| `DromaeoBenchmark.Cube [Modern=False&Prepared=False]` | -2.49% | +1.54% | envelope | +0 |
| `DromaeoBenchmark.Cube [Modern=False&Prepared=True]` | -2.89% | +1.17% | envelope | +0 |
| `DromaeoBenchmark.Cube [Modern=True&Prepared=False]` | -1.43% | -1.74% |  | +0 |
| `DromaeoBenchmark.Cube [Modern=True&Prepared=True]` | -0.42% | +1.85% | envelope | +0 |
| `DromaeoBenchmark.ObjectArray [Modern=False&Prepared=False]` | -2.74% | +6.32% | envelope | +0 |
| `DromaeoBenchmark.ObjectArray [Modern=False&Prepared=True]` | -21.03% | +1.50% | envelope | +0 |
| `DromaeoBenchmark.ObjectArray [Modern=True&Prepared=False]` | -5.78% | +4.13% | envelope | +0 |
| `DromaeoBenchmark.ObjectArray [Modern=True&Prepared=True]` | -0.33% | +2.95% | envelope | +0 |
| `DromaeoBenchmark.ObjectRegExp [Modern=False&Prepared=False]` | +0.92% | -0.48% |  | -571,773 |
| `DromaeoBenchmark.ObjectRegExp [Modern=False&Prepared=True]` | -25.28% | +9.84% | envelope | +1,237,287 |
| `DromaeoBenchmark.ObjectRegExp [Modern=True&Prepared=False]` | -1.96% | -3.58% |  | +567,323 |
| `DromaeoBenchmark.ObjectRegExp [Modern=True&Prepared=True]` | +1.29% | -6.98% | envelope | +69,528 |
| `DromaeoBenchmark.ObjectString [Modern=False&Prepared=False]` | -0.18% | -1.28% |  | -5,962 |
| `DromaeoBenchmark.ObjectString [Modern=False&Prepared=True]` | -15.76% | +1.60% | envelope | +12,519 |
| `DromaeoBenchmark.ObjectString [Modern=True&Prepared=False]` | +0.66% | -11.27% |  | +58,079 |
| `DromaeoBenchmark.ObjectString [Modern=True&Prepared=True]` | +7.50% | +1.80% | **REGRESSION** | -11,678 |
| `DromaeoBenchmark.StringBase64 [Modern=False&Prepared=False]` | -6.51% | +2.12% | envelope | -32 |
| `DromaeoBenchmark.StringBase64 [Modern=False&Prepared=True]` | -2.40% | -3.96% |  | +0 |
| `DromaeoBenchmark.StringBase64 [Modern=True&Prepared=False]` | +0.97% | -6.92% |  | +0 |
| `DromaeoBenchmark.StringBase64 [Modern=True&Prepared=True]` | -3.56% | +5.75% | envelope | +0 |
| `SunSpiderBenchmark.Run [FileName=3d-cube]` | +0.42% | +1.71% | envelope | +0 |
| `SunSpiderBenchmark.Run [FileName=3d-morph]` | +2.99% | -0.44% | envelope | -2 |
| `SunSpiderBenchmark.Run [FileName=3d-raytrace]` | +2.88% | +1.18% | **REGRESSION** | +0 |
| `SunSpiderBenchmark.Run [FileName=access-binary-trees]` | -0.19% | -2.77% |  | +0 |
| `SunSpiderBenchmark.Run [FileName=access-fannkuch]` | -2.33% | +2.03% | envelope | +0 |
| `SunSpiderBenchmark.Run [FileName=access-nbody]` | -2.25% | +2.04% | envelope | +0 |
| `SunSpiderBenchmark.Run [FileName=access-nsieve]` | +0.89% | +0.34% |  | -385 |
| `SunSpiderBenchmark.Run [FileName=bitop(...)-byte [24]]` | +0.12% | +0.56% |  | +0 |
| `SunSpiderBenchmark.Run [FileName=bitops-bits-in-byte]` | -4.83% | -4.83% |  | +0 |
| `SunSpiderBenchmark.Run [FileName=bitops-bitwise-and]` | +1.71% | -4.98% | envelope | +0 |
| `SunSpiderBenchmark.Run [FileName=bitops-nsieve-bits]` | -2.96% | -0.56% |  | +0 |
| `SunSpiderBenchmark.Run [FileName=contr(...)rsive [21]]` | +2.31% | -3.93% | envelope | +0 |
| `SunSpiderBenchmark.Run [FileName=crypto-aes]` | -2.14% | -0.06% |  | +3,318 |
| `SunSpiderBenchmark.Run [FileName=crypto-md5]` | +0.28% | -1.56% |  | +0 |
| `SunSpiderBenchmark.Run [FileName=crypto-sha1]` | -0.11% | +0.78% |  | +0 |
| `SunSpiderBenchmark.Run [FileName=date-format-tofte]` | +5.79% | -5.48% | envelope | +0 |
| `SunSpiderBenchmark.Run [FileName=date-format-xparb]` | -2.07% | +4.23% | envelope | +0 |
| `SunSpiderBenchmark.Run [FileName=math-cordic]` | -3.04% | -0.74% |  | +0 |
| `SunSpiderBenchmark.Run [FileName=math-partial-sums]` | -4.44% | -2.79% |  | +0 |
| `SunSpiderBenchmark.Run [FileName=math-spectral-norm]` | -2.09% | +1.39% | envelope | +0 |
| `SunSpiderBenchmark.Run [FileName=regexp-dna]` | -0.53% | -2.37% |  | +2,394 |
| `SunSpiderBenchmark.Run [FileName=strin(...)input [21]]` | -0.72% | -2.83% |  | +0 |
| `SunSpiderBenchmark.Run [FileName=string-base64]` | +2.22% | +1.25% | **REGRESSION** | +0 |
| `SunSpiderBenchmark.Run [FileName=string-fasta]` | +1.31% | +2.64% | **REGRESSION** | +0 |
| `SunSpiderBenchmark.Run [FileName=string-tagcloud]` | -3.81% | -3.81% |  | +5,192 |
| `SunSpiderBenchmark.Run [FileName=string-unpack-code]` | +0.35% | -2.74% |  | +0 |

### Four rows regress in both pairs, and three of them are string rows

| row | pair 1 | pair 2 |
|---|---:|---:|
| `DromaeoBenchmark.ObjectString [Modern=True&Prepared=True]` | +7.50% | +1.80% |
| `SunSpider string-fasta` | +1.31% | +2.64% |
| `SunSpider string-base64` | +2.22% | +1.25% |
| `SunSpider 3d-raytrace` | +2.88% | +1.18% |

That is a coherent pattern rather than four survivors of a coin toss, and there is a mechanism for it.

### The mechanism: a virtual `Length` read added to every `+=`

`JintAssignmentExpression.ComputeCompound`'s string branch now reads `jsString.Length` before the append:

```csharp
var appended = TypeConverter.ToString(rprim);
if ((long) jsString.Length + appended.Length > JsString.MaxLength) { … }
return jsString.Append(appended);
```

`JsString.Length` is `public virtual` (`Jint/Native/JsString.cs:375`) and `ConcatenatedString` overrides it (`:628`) as `_stringBuilder?.Length ?? _value?.Length ?? 0`. Before this PR the `+=` path never read a length at all — it called `Append(rprim)` and let the override do the coercion internally. So each iteration of a `s += t` loop now pays a virtual dispatch, two null tests and a `StringBuilder.Length` read on top of the widened add and the compare. `string-base64` and `string-fasta` are precisely tight `+=` loops, and `ObjectString`'s hot rows are string building.

The allocation columns are byte-exact almost everywhere, which is the trustworthy half of the measurement and confirms the change adds no allocation — this is pure per-iteration overhead.

### What would fix it

The guard does not need the receiver's length through a virtual property at the call site. Both `Append` overrides already know their own length concretely and non-virtually: `JsString.Append` has `_value.Length` and `ConcatenatedString.Append` has `_stringBuilder`/`_value` directly in hand. Pushing the check down into the two `Append` implementations gets the same enforcement with no added dispatch on the receiver, and keeps the cold `Throw.RangeError` where it is. Worth confirming the same is not happening on the other lanes that now read a length before concatenating.

Happy to re-run the gate once that is addressed. Both pairs' raw artifacts are kept.

lahma:
## Benchmark gate - PASS (after 132b00b1e)

`132b00b1e` moves the length check inside the two `Append` overrides. Re-measured against the same pinned base `ff980b7bc`.

### The three regressions this PR had are gone

Five pairs for the SunSpider rows, two full pairs for Dromaeo. Median is across all pairs.

| row | before fix | after fix (median of 5) |
|---|---|---:|
| `SunSpider string-base64` | +2.22% / +1.25% | **+0.34%** |
| `SunSpider string-fasta` | +1.31% / +2.64% | **-0.64%** |
| `SunSpider 3d-raytrace` | +2.88% / +1.18% | **-0.72%** |
| `Dromaeo ObjectString [Modern=True&Prepared=True]` | +7.50% / +1.80% | **+0.94% / -10.50%** |

### Full SunSpider table, five pairs

| row | pair 1 | pair 2 | pair 3 | pair 4 | pair 5 | median | >+1% |
|---|---:|---:|---:|---:|---:|---:|---|
| `3d-cube` | +0.43% | -2.87% | +0.20% | +1.15% | +0.76% | **+0.43%** | 1/5 |
| `3d-morph` | -0.09% | +1.13% | -0.93% | +1.61% | +0.03% | **+0.03%** | 2/5 |
| `3d-raytrace` | +1.01% | -2.23% | -0.72% | +0.93% | -0.77% | **-0.72%** | 1/5 |
| `access-binary-trees` | +2.02% | -1.08% | +1.34% | +2.36% | +0.74% | **+1.34%** | 3/5 |
| `access-fannkuch` | +0.27% | -3.06% | +2.69% | -3.16% | +0.07% | **+0.07%** | 1/5 |
| `access-nbody` | -0.39% | -2.44% | -2.45% | +1.88% | +1.94% | **-0.39%** | 2/5 |
| `access-nsieve` | +2.01% | -6.04% | +1.77% | +0.34% | -0.60% | **+0.34%** | 2/5 |
| `bitop(...)-byte [24` | -1.47% | -4.33% | -1.64% | +3.86% | -0.91% | **-1.47%** | 1/5 |
| `bitops-bits-in-byte` | +4.17% | -5.14% | -0.17% | -0.67% | +0.03% | **-0.17%** | 1/5 |
| `bitops-bitwise-and` | -4.76% | +1.54% | -2.26% | +0.93% | +0.36% | **+0.36%** | 1/5 |
| `bitops-nsieve-bits` | +12.81% | +2.22% | +4.71% | -1.57% | +1.86% | **+2.22%** | 4/5 |
| `contr(...)rsive [21` | -2.14% | -3.26% | +1.46% | +0.89% | -2.28% | **-2.14%** | 1/5 |
| `crypto-aes` | -0.79% | -0.90% | +3.16% | +2.11% | +0.85% | **+0.85%** | 2/5 |
| `crypto-md5` | +0.12% | -2.39% | +2.97% | +1.59% | -0.53% | **+0.12%** | 2/5 |
| `crypto-sha1` | +0.90% | -4.74% | +3.13% | +0.16% | -1.30% | **+0.16%** | 1/5 |
| `date-format-tofte` | -1.17% | -1.50% | +3.20% | -0.94% | -1.49% | **-1.17%** | 1/5 |
| `date-format-xparb` | +0.11% | +0.10% | +1.56% | -2.57% | -4.48% | **+0.10%** | 1/5 |
| `math-cordic` | +0.89% | -3.20% | +0.24% | -1.77% | -0.19% | **-0.19%** | 0/5 |
| `math-partial-sums` | +0.39% | -4.59% | -7.07% | -1.39% | -3.45% | **-3.45%** | 0/5 |
| `math-spectral-norm` | -3.23% | +2.45% | +10.04% | +1.61% | -1.79% | **+1.61%** | 3/5 |
| `regexp-dna` | -0.21% | -3.33% | +1.25% | -1.15% | -1.71% | **-1.15%** | 1/5 |
| `strin(...)input [21` | -0.48% | -1.60% | -1.04% | +0.67% | -0.37% | **-0.48%** | 0/5 |
| `string-base64` | +3.41% | -2.47% | +0.34% | -0.67% | +1.24% | **+0.34%** | 2/5 |
| `string-fasta` | +4.16% | +1.21% | -0.64% | -2.79% | -2.13% | **-0.64%** | 2/5 |
| `string-tagcloud` | -12.37% | -7.30% | +4.21% | -0.35% | +4.57% | **-0.35%** | 2/5 |
| `string-unpack-code` | -1.97% | +0.20% | +0.08% | +1.50% | -0.56% | **+0.08%** | 1/5 |

### Full Dromaeo table, two pairs

| row | pair 1 | pair 2 |
|---|---:|---:|
| `DromaeoBenchmark.CoreEval [Modern=False&Prepared=False]` | -5.84% | -1.47% |
| `DromaeoBenchmark.CoreEval [Modern=False&Prepared=True]` | +4.29% | -3.07% |
| `DromaeoBenchmark.CoreEval [Modern=True&Prepared=False]` | +0.72% | +1.24% |
| `DromaeoBenchmark.CoreEval [Modern=True&Prepared=True]` | +1.09% | -0.98% |
| `DromaeoBenchmark.Cube [Modern=False&Prepared=False]` | -5.33% | -4.11% |
| `DromaeoBenchmark.Cube [Modern=False&Prepared=True]` | +0.99% | -1.55% |
| `DromaeoBenchmark.Cube [Modern=True&Prepared=False]` | -0.35% | -2.92% |
| `DromaeoBenchmark.Cube [Modern=True&Prepared=True]` | -0.90% | -4.66% |
| `DromaeoBenchmark.ObjectArray [Modern=False&Prepared=False]` | -2.10% | -5.87% |
| `DromaeoBenchmark.ObjectArray [Modern=False&Prepared=True]` | +0.52% | -4.11% |
| `DromaeoBenchmark.ObjectArray [Modern=True&Prepared=False]` | -0.66% | -2.21% |
| `DromaeoBenchmark.ObjectArray [Modern=True&Prepared=True]` | +6.80% | -0.86% |
| `DromaeoBenchmark.ObjectRegExp [Modern=False&Prepared=False]` | -3.85% | +3.42% |
| `DromaeoBenchmark.ObjectRegExp [Modern=False&Prepared=True]` | -0.07% | -2.97% |
| `DromaeoBenchmark.ObjectRegExp [Modern=True&Prepared=False]` | -12.98% | +6.34% |
| `DromaeoBenchmark.ObjectRegExp [Modern=True&Prepared=True]` | +0.26% | -7.38% |
| `DromaeoBenchmark.ObjectString [Modern=False&Prepared=False]` | +0.41% | -4.42% |
| `DromaeoBenchmark.ObjectString [Modern=False&Prepared=True]` | -0.02% | -0.18% |
| `DromaeoBenchmark.ObjectString [Modern=True&Prepared=False]` | -3.10% | +1.89% |
| `DromaeoBenchmark.ObjectString [Modern=True&Prepared=True]` | +0.94% | -10.50% |
| `DromaeoBenchmark.StringBase64 [Modern=False&Prepared=False]` | -1.34% | +1.69% |
| `DromaeoBenchmark.StringBase64 [Modern=False&Prepared=True]` | -1.56% | -3.35% |
| `DromaeoBenchmark.StringBase64 [Modern=True&Prepared=False]` | -0.33% | +5.66% |
| `DromaeoBenchmark.StringBase64 [Modern=True&Prepared=True]` | -8.24% | -0.35% |

### What still shows up, and why it is not this PR

Three rows exceed +1% in a majority of the five pairs: `bitops-nsieve-bits` (4/5, median +2.22%), `math-spectral-norm` (3/5, +1.61%) and `access-binary-trees` (3/5, +1.34%).

None of them touches a string. After `132b00b1e` this PR reaches `JsString.Append`, `StringPrototype.Concat`, the string branches of `JintAssignmentExpression`/`JintBinaryExpression`, `ValueStringBuilder.Grow`, `RegExpPrototype`, `ArrayPrototype.Join`/`ToLocaleString`, template literals and `String.raw`. An integer bit array, a float matrix and a binary-tree allocation loop are unreachable from every one of those, so there is no mechanism to attribute the movement to — where the three string rows above did have one, which is precisely why they were treated as real and fixed.

They are also what chance predicts at this machine's envelope: the run-to-run spread reached about ±10% on the Dromaeo rows, and across 26 SunSpider rows and five pairs, three rows clearing 3/5 is the expected count rather than a surprise. Worth stating plainly that a strict >1% single-pair gate is not enforceable on this hardware — a row is only trustworthy here when repeated pairs agree *and* a mechanism exists.

Allocation columns stayed byte-exact throughout, before and after the fix.

PR #3028, opened 2026-08-15
Refuse a JSON document too long to be a string while it is being built

### LINKED ISSUE #3024: JSON.stringify can build a string longer than the maximum string length the engine enforces everywhere else
#3015 gave JavaScript strings a maximum length — `JsString.MaxLength`, V8's `(1 << 29) - 24` = 536,870,888 — and made every string-building path in `String`, `Array.prototype.join` and the concatenation lanes raise `RangeError: Invalid string length` rather than fail as a CLR exception. `JSON.stringify` was not in that set, and it can still produce a string longer than the limit the engine now claims:

```js
var a = new Array(200000).fill("x".repeat(4000));
var s = JSON.stringify(a);
print(s.length);   // 800600001
```

`800,600,001 > 536,870,888`. So a value exists that, by the engine's own rule, cannot. Anything that later concatenates it correctly refuses, which makes the inconsistency observable rather than merely theoretical.

V8 answers `RangeError: Invalid string length` for the same input.

### Where it lives

`JsonSerializer` builds into a `ValueStringBuilder` (`Jint/Runtime/Interop/…`/`Jint/Native/Json/JsonSerializer.cs`, two `new ValueStringBuilder(...)` sites). #3015 deliberately did **not** put a JS `RangeError` inside `ValueStringBuilder` itself, because that type also backs URI encode/decode, Intl and Temporal ISO formatting, where a JavaScript error would be wrong. It hardened `Grow` so the length arithmetic can no longer wrap silently, and left each JS-semantic caller to guard its own result. `JSON.stringify` is the caller that never got its guard.

Past `Array.MaxLength` (0x7FFFFFC7) it now surfaces `OverflowException` instead of the old `ArgumentOutOfRangeException` — both CLR exceptions escaping `Evaluate`, so no regression, but neither is catchable as JavaScript.

### What the fix needs

The serializer holds a realm, so this is the same shape as the paths #3015 already fixed: check the accumulated length against `JsString.MaxLength` before growing, and raise `Throw.RangeError(realm, "Invalid string length")`. Worth doing the check where the builder is appended to rather than only at the end, so an over-long result is refused before ~1.6 GB has been allocated to build it.

`JSON.parse` is worth a look in the same pass — a `reviver` returning strings can concatenate too, though that path goes through the ordinary concatenation lanes which #3015 does cover.

Found while gating #3015; recorded rather than widened into that PR, which was already breaking behaviour for strings above 512 M characters.

lahma:
#3015 gave JavaScript strings a maximum length — `JsString.MaxLength = (1 << 29) - 24 = 536,870,888`, V8's `String::kMaxLength` — and guarded every path that builds one with a catchable `RangeError: Invalid string length`. `JSON.stringify` was not in that set, so the engine went on producing a string its own rule forbids.

## The inconsistency

```js
var a = new Array(200000).fill("x".repeat(4000));
var s = JSON.stringify(a);
print(s.length);   // 800600001, and 800600001 > 536870888
```

**Pre-fix, on `3bc449ec`: `800600001`, in 2.0 s.** Every path that would later concatenate that string correctly refuses it, so the inconsistency is observable from script; V8 answers `RangeError: Invalid string length` for the same input.

Larger inputs did not merely produce an over-long string. The new test's own row, run against the unfixed serializer, dies like this:

```
Jint.Tests.Runtime.StringLengthLimitTests.JsonStringifyOfAnArrayTooLongToSerializeIsACatchableRangeError [FAIL]
  System.OutOfMemoryException : Insufficient memory to continue the execution of the program.
     at System.String.Ctor(ReadOnlySpan`1 value)
     at System.Span`1.ToString()
     at System.Text.ValueStringBuilder.ToString() in Jint/Pooling/ValueStringBuilder.cs:line 92
     at Jint.Native.Json.JsonSerializer.Serialize(JsValue, JsValue, JsValue) in Jint/Native/Json/JsonSerializer.cs:line 132
```

Nine seconds and about seven gigabytes to reach a CLR exception escaping `Evaluate` past every `catch` in the script — the exact defect class #3015 was about, and thrown from the `finally` discussed below. With an engine timeout in place instead, the same two inputs simply never return: `-t 3` reports `Script execution timed out` for both `JSON.stringify(Object.assign([], { length: 4294967295 }))` and `JSON.stringify(new Array(268435445))`.

Post-fix all three are a `RangeError` the script catches, and the two cheap ones are immediate.

## Where the guard went, and where it did not

Not into `ValueStringBuilder`. #3015 deliberately left that type realm-free — it also backs URI encode/decode, Intl and Temporal ISO formatting, where a JavaScript error would be wrong — hardened `Grow` so the length arithmetic can no longer wrap, and left each JavaScript-semantic caller to guard its own result. This follows that split: one private `ThrowIfDocumentLengthExceeded(long)` on the serializer, raising `Throw.RangeError(_engine.Realm, "Invalid string length")` — the same message every other path already uses. The realm is read only on the throwing path, because `Engine.Realm` peeks the execution context stack and that is not something to do once per element.

It is checked where the walk **accumulates**, not on the finished string:

| Site | Cadence | What is checked |
| --- | --- | --- |
| `SerializeJSONArray` | per element | what is written **plus what the remaining indices must still contribute** |
| `SerializeJSONObject` | per key | what is written and already final |
| `TrySerializeShapedJSONObject` | per key | same |
| `SerializeJSONValue`, string branch | per string value | what is written plus the unquoted length |
| `SerializeJSONValue`, `JsRawJson` / interop `SerializeToJson` | per chunk | what is written plus the chunk's exact length |
| `Serialize` | once per call | the finished document, before it becomes a `JsString` |

Every estimate is a **lower bound** on the finished document, so the guard can never refuse one that would have fit.

The array bound is what makes a hopeless array fail immediately instead of after half a billion characters: an array announces its length, and each index contributes at least two characters — a separator (or the opening bracket) plus at least one character of value text, since an element with no JSON representation is written as `null`. So `JSON.stringify(new Array(268435445))` is refused before the first element is written. Object members carry no such bound — a member whose value has no representation rewinds the document to the position it started at — so those two sites check only the prefix that is already final, which is why the check sits *before* `position` is taken rather than after.

`Serialize`'s `finally` used to end in `result = json.ToString()`. On the throwing path that materialized the whole partial document, i.e. the very allocation the guard exists to avoid — it is literally where the `OutOfMemoryException` above was thrown. It disposes now; `ToString()` disposes on the way out, so the success path is unchanged.

## Hot methods touched — for the gate

`SerializeJSONArray`, `SerializeJSONObject`, `TrySerializeShapedJSONObject` and `SerializeJSONValue` are all hot. Each gains one predictable, not-taken compare per element / per key / per string value (the array site additionally a subtract and a shift-add, on a loop that already does a `%` per element), and the realm is not touched unless it throws. `JsonJsBenchmark` (`StringifyRecords`, `StringifyConfig`, `StringifyRecordsWithReplacer`, `RoundTripRecords`) and `JsonBenchmark.Stringify` are the rows that cover them. **I have not run any benchmark** — measurement is yours, serially and on an idle machine.

## The `IBufferWriter<byte>` overloads keep their old ceiling

Their result is bytes the host consumes and never a `JsString`, so the language's string limit has no business bounding them and a host streaming a gigabyte-sized document keeps working. The limit is per-call state (`_maxDocumentLength`), set from the entry point: `JsString.MaxLength` for the `JsValue`-returning overloads, `ClrLimits.MaxArrayLength` for the writer ones — where the check is all but inert, since `Grow` refuses first, except that a hopeless array now fails at once instead of after minutes. Documented in the type's `<remarks>`.

## `JSON.parse` with a reviver: nothing needed

A reviver that concatenates does it through the ordinary string-building paths, which #3015 already guards, and the error reaches the script's `catch` through the parse machinery unwrapped. Verified both shapes against this branch:

```js
JSON.parse('{"a":1}', (k, v) => typeof v === 'number' ? 'x'.repeat(536870889) : v);
// RangeError: Invalid string length

var big = 'x'.repeat(268435444);
JSON.parse('{"a":1}', (k, v) => typeof v === 'number' ? big + big + 'yy' : v);
// RangeError: Invalid string length
```

The parser itself needs no guard either: it builds each string with a `ValueStringBuilder` from the source text, and escapes only ever shrink, so a parsed string cannot outgrow the document it came from.

## Tests

New rows in `Jint.Tests/Runtime/StringLengthLimitTests.cs`, next to the other limit rows and asserting a **JS-catchable** `RangeError` through the file's existing `Caught` helper, which runs the `try`/`catch` inside the script precisely so a CLR exception escaping the engine fails as an escaping exception rather than being reported as a handled error.

- `JsonStringifyOfAnArrayTooLongToSerializeIsACatchableRangeError` — the two shapes above. They cost nothing: both arrays are empty, only their `length` is large, and the throw happens before the first element is written. `MaxLength / 2 + 1` is the smallest length that triggers it, and the just-under side is deliberately not tested — that one really would have to build the gigabyte.
- `JsonDocumentsThatFitAreUnaffected` — five shapes where an over-counting estimate would show up first: holes, `undefined`/function/symbol elements, indentation, a dropped object member, a 1 MB string value.
- `JsonParseReviverThatBuildsTooLongAStringIsACatchableRangeError` — the reviver finding, in the shape that costs nothing.

The class doc's paragraph about which paths have no cheap test is extended: `JSON.stringify`'s element-by-element half joins `+=`, `join`, `String.raw` and the rest there. Its guard can only fire once roughly half a billion characters are in hand, so a test for it would have to allocate the gigabyte the guard exists to prevent — the array-`length` row is the cheap route to the same guard.

Both non-array rows **pass against the unfixed serializer** (6/6 on net10.0 and net472), which is what makes them honest controls rather than the new coverage.

## Verification

Release throughout, no `--no-build`, zero warnings.

| Suite | Result |
| --- | --- |
| `Jint.Tests` | 5754 passed / 0 failed / 4 skipped (net10.0) · 5669 / 0 / 4 (net472) |
| `Jint.Tests.PublicInterface` | 1488 / 0 / 9 (net10.0) · 1487 / 0 / 9 (net472) |
| `Jint.Tests.Test262` | **102,324 passed / 0 failed** / 343 skipped, 3 m 07 s |
| `Jint.Tests.CommonScripts` | 28 / 0 / 0 on both frameworks |

Closes #3024

### DIFF
--- a/Jint.Tests/Runtime/StringLengthLimitTests.cs
+++ b/Jint.Tests/Runtime/StringLengthLimitTests.cs
@@ -19,13 +19,13 @@ namespace Jint.Tests.Runtime;
 /// </para>
 /// <para>
 /// Every row below is a path whose limit is decided from small inputs — a repeat count, a pad length,
-/// a replacement pattern's expansion factor — so the throw happens before anything of that size is
-/// allocated and the test costs a few megabytes. The remaining guarded paths (<c>+</c>/<c>+=</c>,
-/// template literals, <c>concat</c>, <c>join</c>, <c>toLocaleString</c>, <c>String.raw</c> and the
-/// accumulators inside <c>replace</c>/<c>replaceAll</c>) accumulate their result one piece at a time,
-/// so their guard can only fire once roughly half a billion characters are already in hand; a test for
-/// those would have to allocate the gigabyte the guard exists to prevent, and there is no cheaper
-/// input that reaches them.
+/// a replacement pattern's expansion factor, an array's <c>length</c> — so the throw happens before
+/// anything of that size is allocated and the test costs a few megabytes. The remaining guarded paths
+/// (<c>+</c>/<c>+=</c>, template literals, <c>concat</c>, <c>join</c>, <c>toLocaleString</c>,
+/// <c>String.raw</c>, the accumulators inside <c>replace</c>/<c>replaceAll</c> and the element-by-element
+/// half of <c>JSON.stringify</c>) accumulate their result one piece at a time, so their guard can only
+/// fire once roughly half a billion characters are already in hand; a test for those would have to
+/// allocate the gigabyte the guard exists to prevent, and there is no cheaper input that reaches them.
 /// </para>
 /// </summary>
 public class StringLengthLimitTests
@@ -122,6 +122,54 @@ public class StringLengthLimitTests
             """);
 
         Caught("x.replace(/(.+)/g, rep)").Should().Be(InvalidStringLength);
+    }
+
+    /// <summary>
+    /// <c>JSON.stringify</c> builds its document one element at a time, but an array announces up front
+    /// how many elements there are, and every index contributes at least two characters — a separator
+    /// plus at least one character of value text, an absent element being written as <c>null</c>. So an
+    /// array whose <c>length</c> alone puts the document past the limit is refused before the first
+    /// element is written, which is what makes this row cost nothing: the array itself is empty.
+    /// </summary>
+    [Fact]
+    public void JsonStringifyOfAnArrayTooLongToSerializeIsACatchableRangeError()
+    {
+        Caught("JSON.stringify(Object.assign([], { length: 4294967295 }))").Should().Be(InvalidStringLength);
+        Caught($"JSON.stringify(new Array({JsString.MaxLength / 2 + 1}))").Should().Be(InvalidStringLength);
+    }
+
+    /// <summary>
+    /// The estimate above is a lower bound on the finished document, so it can never refuse one that
+    /// would have fit. These are the shapes where over-counting would show up first: an element is
+    /// written as <c>null</c> however it came to have no representation, and indentation only ever
+    /// adds to the count.
+    /// </summary>
+    [Theory]
+    [InlineData("JSON.stringify(new Array(3))", "[null,null,null]")]
+    [InlineData("JSON.stringify([1, undefined, function () {}, Symbol()])", "[1,null,null,null]")]
+    [InlineData("JSON.stringify([1, 2], null, 4).replace(/\\n/g, '|')", "[|    1,|    2|]")]
+    [InlineData("JSON.stringify({ a: undefined, b: 1 })", "{\"b\":1}")]
+    [InlineData("JSON.stringify(['x'.repeat(1 << 20)]).length", "1048580")]
+    public void JsonDocumentsThatFitAreUnaffected(string source, string expected)
+    {
+        _engine.Evaluate(source).ToString().Should().Be(expected);
+    }
+
+    /// <summary>
+    /// <c>JSON.parse</c> needs no guard of its own — a reviver that builds too long a string does it
+    /// through the ordinary string-building paths, which have one — but the error has to reach the
+    /// script's <c>catch</c> through the parse machinery rather than being wrapped or swallowed on the
+    /// way out. Verified with <c>+</c> as well as <c>repeat</c>, in a run that could afford the half
+    /// gigabyte the concatenating shape costs.
+    /// </summary>
+    [Fact]
+    public void JsonParseReviverThatBuildsTooLongAStringIsACatchableRangeError()
+    {
+        Caught($"JSON.parse('{{\"a\":1}}', function (k, v) {{ return typeof v === 'number' ? 'x'.repeat({JsString.MaxLength + 1L}) : v; }})")
+            .Should().Be(InvalidStringLength);
+
+        _engine.Evaluate("JSON.stringify(JSON.parse('{\"a\":1}', function (k, v) { return typeof v === 'number' ? v + 1 : v; }))")
+            .AsString().Should().Be("{\"a\":2}");
     }
 
     /// <summary>
--- a/Jint/Native/Json/JsonSerializer.cs
+++ b/Jint/Native/Json/JsonSerializer.cs
@@ -41,6 +41,14 @@ namespace Jint.Native.Json;
 /// exists to avoid, and added a decode on top.
 /// </para>
 /// <para>
+/// <b>Document length.</b> The <see cref="JsValue"/>-returning overloads produce a JavaScript string and
+/// are bounded by the same maximum length as every other way of building one: a document that would
+/// exceed it raises <c>RangeError: Invalid string length</c>, which a script can catch, and raises it
+/// while the document is being built rather than after it has been paid for. The
+/// <see cref="IBufferWriter{T}"/> overloads produce bytes the host consumes and never a string, so the
+/// language's limit does not apply to them; they remain bounded only by what a CLR array can hold.
+/// </para>
+/// <para>
 /// <b>BigInt.</b> A <c>BigInt</c> that reaches the output throws a
 /// <see cref="Runtime.JavaScriptException"/> carrying a <c>TypeError</c> ("Do not know how to serialize a
 /// BigInt"), as the specification requires. A host using this type as a storage codec has one escape hatch:
@@ -62,6 +70,13 @@ public sealed class JsonSerializer
     private string _gap = string.Empty;
     private List<JsValue>? _propertyList;
     private bool _hasReplacerFunction;
+
+    // The largest document this call is allowed to build. JsString.MaxLength for the overloads whose
+    // result is a JavaScript string, so JSON.stringify can no longer hand back a string the engine's
+    // own limit forbids; the CLR array ceiling for the IBufferWriter overloads, whose result is bytes
+    // the host consumes and never a JsString, so the language's string limit has no business bounding
+    // them and a host writing a gigabyte-sized document to a stream keeps working.
+    private long _maxDocumentLength;
 
     // Declared last: this is the only field that is not read on the per-key hot path (only
     // UnwrapValueSlow touches it, and only when _hasReplacerFunction says there is one), and embedding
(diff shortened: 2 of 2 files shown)


lahma:
Reviewed the array lower bound independently. **It is sound.** At index `i` the contributions still to come are at least `(len - i)` characters of value text, plus `(len - i - 1)` separators when `i = 0` or `(len - i)` when `i > 0`, plus the closing bracket — so `2 * (len - i)` never exceeds what the finished document must contain, and indentation only widens the margin. It cannot refuse a document that would have fit.

**What the description does not say outright, and should:** the array bound is a deliberate, observable divergence from V8, not just an optimisation.

For an array whose `length` alone puts the result out of reach — `length > 268,435,444`, since `2 * length > MaxLength` at `i = 0` — Jint now throws before the first element is read. V8 throws the same `RangeError`, but only after invoking the replacer, any getters and any `toJSON` for every index it gets through first. So a script that counts replacer calls before catching the error sees a different number, and a getter with a side effect does not run.

I think it is right to keep, and the reason is stronger than the test convenience it also buys:

- Both engines reject the same inputs with the same error. Only the side effects along the way differ, and only for arrays above a quarter of a billion elements.
- The alternative is not "spec-correct behaviour", it is the measured pre-fix outcome: ~7 GB and nine seconds to reach an `OutOfMemoryException`, which is a CLR exception escaping `Evaluate` past every `catch` — uncatchable by the script that caused it. Trading side-effect parity on a pathological input for a catchable error and a bounded allocation is the right way round for an engine that gets embedded and fed untrusted script.
- AGENTS.md asks for the spec "as closely as practical", and no test262 test comes near these lengths.

Worth stating in the PR body so it is a recorded decision rather than something a future reader has to reconstruct from the bound.

Two other things in here are worth more attention than their size suggests:

- **The `finally { result = json.ToString(); }` fix is an independent defect**, not housekeeping. Every throwing path through `Serialize` was materialising the entire partial document — up to half a billion characters the caller never receives — on its way out. That is why the pre-fix repro reached ~7 GB rather than failing at the limit.
- **Leaving the `IBufferWriter<byte>` overloads uncapped is the right call.** Their result is bytes a host consumes, never a `JsString`, so the language's string limit has no claim on them, and capping them would break a host streaming a document larger than a gigabyte. Making the cap per-call state rather than a constant is what allows that split.

Benchmark gate to follow on `JsonJsBenchmark` + `JsonBenchmark`, with the `Parse*` rows as controls this diff cannot reach.

lahma:
## Benchmark gate - PASS

`JsonJsBenchmark` + `JsonBenchmark`, default job, serial on an idle machine, base pinned to the merge-base `3bc449ecd`. **Nine pairs: five with base measured first, four with the PR measured first.**

### Why the ordering matters more than the threshold

The first five pairs said this PR regressed five rows, worst of them `ParseRecords` at +5.58%. That reading was wrong, and the table said so itself: **`ParseRecords` and `ParseBigObject` are parse rows, and this PR does not touch the parser** - only `JsonSerializer`. Controls moving further than the treatment is not a result, it is a broken measurement.

It is not shared-engine contamination (`JsonJsBenchmark` uses `IsolatedScript.Warm`, one engine per row, exactly as AGENTS.md requires). It is run order:

| | median shift across all 12 rows |
|---|---:|
| five pairs, base measured first | **+0.90%** |
| four pairs, PR measured first | **-0.75%** |

The sign follows the order, not the code: whichever side runs second reads about 0.85%% slower, over a run of roughly twenty minutes. On this machine that bias is **larger than the 1%% gate threshold**, so any single-orientation A/B here measures the thermal ramp as much as the diff.

### Per row, both orientations

| row | base-first (5 pairs) | pr-first (4 pairs) | verdict |
|---|---:|---:|---|
| `JsonBenchmark.Parse [bestbuy_dataset.json]` | +0.29% | -0.63% | clean |
| `JsonBenchmark.Parse [twitter.json]` | -0.47% | +0.47% | clean |
| `JsonBenchmark.Stringify [bestbuy_dataset.json]` | +0.98% | +0.47% | clean |
| `JsonBenchmark.Stringify [twitter.json]` | -1.00% | -3.77% | clean |
| `JsonJsBenchmark.ParseBigObject` | +1.20% | -1.99% | sign follows order |
| `JsonJsBenchmark.ParseConfig` | +0.82% | -2.20% | clean |
| `JsonJsBenchmark.ParseRecords` | +5.58% | -0.88% | sign follows order |
| `JsonJsBenchmark.ParseRecordsWithReviver` | +0.01% | +1.20% | sign follows order |
| `JsonJsBenchmark.RoundTripRecords` | +3.46% | +0.78% | sign follows order |
| `JsonJsBenchmark.StringifyConfig` | +1.10% | -0.38% | sign follows order |
| `JsonJsBenchmark.StringifyRecords` | +1.55% | -0.89% | sign follows order |
| `JsonJsBenchmark.StringifyRecordsWithReplacer` | -0.94% | -2.01% | clean |

**No row is positive in both orientations.** Every row that looked like a regression in the first five pairs flips sign when the order is reversed, including all five that cleared +1%% there.

### Verdict

The guard costs nothing measurable. That is consistent with its shape — one predictable, not-taken compare per element, per key and per string value, with the realm read only on the throwing branch — but the shape was the hypothesis, and these are the measurements.

Allocation deltas were within a few bytes per row throughout, in both orientations.

### One thing worth keeping from this

The `+=` guard in #3015 was a *real* regression that this same procedure caught, and it was found because a mechanism existed for it — a virtual `Length` dispatch added to a tight loop. Here no mechanism could explain a parse row moving, and that mismatch is what exposed the bias. Neither pass nor block should rest on one orientation of pairs on this hardware.

PR #3036, opened 2026-08-16
Account memory across async continuations

sebastienros:
## Summary

- account managed allocations per engine operation across synchronous segments and thread-changing async continuations
- carry memory state through promise reactions, event-loop jobs, and asynchronous module completion without charging unrelated process allocations
- add explicit accuracy/usage diagnostics and `Begin`/`End` multi-entry scopes guarded by the engine ownership contract
- document guarantees and residual risks in the README and threat model

## Stack

This PR is stacked on #3035 (`sebros/reject-concurrent-engine-use`), which is stacked on #3030 (`initial-threat-model`). Review only the diff against `sebros/reject-concurrent-engine-use`.

## Validation

- `dotnet build Jint/Jint.csproj -c Release` across all target frameworks
- `dotnet test Jint.Tests/Jint.Tests.csproj -c Release` excluding the known unrelated timezone-sensitive year-10000 test: 5,760 passed, 4 skipped
- `dotnet test Jint.Tests.PublicInterface/Jint.Tests.PublicInterface.csproj -c Release`: 1,529 passed, 9 configuration skips
- host-contract verification public suite: 1,533 passed, 5 inverse-configuration skips
- combined memory/concurrency stress selection repeated 5 times: 41 passed per run
- `dotnet build Jint.Benchmark/Jint.Benchmark.csproj -c Release`
- `git diff --check origin/sebros/reject-concurrent-engine-use...HEAD`


lahma (approved):
Reviewed as the incremental diff against `sebros/reject-concurrent-engine-use`. The segment/operation-state design is sound: per-thread counters sampled only inside synchronous segments, the accumulated total carried by the reaction/job/module-completion records, and the three entry lifecycles (host entry, event-loop job, implicit execution-context) each bracketing cleanly. The 25 `HostMemoryLimitTests` hit exactly the edge cases I went looking for — the final-callback allocation with no later statement check, exceeded-state not replayed by an idle `Check()`, nested entries sharing the outermost budget, per-engine isolation on shared `Options`, and the cross-thread rejection surface. Docs (README, threat model, AGENTS.md contract table and the constraints gotcha) were all kept in step, and `LimitMemory`'s sentinel semantics are preserved while the now-public constructor validates its range. Approving.

Three non-blocking notes:

1. **Unconstrained-path cost.** `EnterExecutionContext`/`LeaveExecutionContext` are per-function-call hot paths and now carry a `_memoryLimitConstraint is not null` / `_implicitMemoryContextDepth != 0` branch even for engines with no memory limit, and `ExecuteWithConstraints` gained the `_hostEntryDepth` term. These should be branch-predictor noise, but since the suite already has the right instrument, it would be good to attach one `JINT_BENCH_MODE=gate` run of `ConstrainedExecutionBenchmark`'s all-false row (base vs. this branch) when you re-roll the stack on top of the #3035 fix.
2. **`isNested` semantics tightened for everyone, not just memory-limited engines**: `engine.Invoke(hostFn)` where the CLR function re-enters the engine used to re-arm budgets (no interpreter frame on the stack, so `_executionContexts.Count == 1`); with `_hostEntryDepth > 0` in the disjunction it no longer does. I read that as a bug fix in the safe direction, and the AGENTS.md gotcha was updated to match — just calling it out so it's a deliberate decision on record, since `HostCallLoopConstraintTests` pins this area.
3. `Modules.StartImport` now funnels through `ExecuteWithConstraints`, so it resets ordinary constraints as a top-level entry where it previously didn't. Consistent with treating it as an entry; worth a line in the PR body.

Merge is blocked beneath this by the #3035 admission-token leak; once that lands, this layer is good to go after the cascade rebase.

lahma:
#3035 is merged (`8b980dc71`), and this PR is now the bottom of the stack, auto-retargeted to `main`. The server-side cascade could not rewrite the branch, and after looking at it I don't think any mechanical rebase can — handing this back rather than guessing:

- The 2026-08-20 rebase onto current `main` covered only the #3035 branch. This branch (and every layer above it) still sits on the pre-rework #3035 commits (`7cda7429e`, `d158208a1`) and a `main` baseline from before roughly #3040–#3182, i.e. ~75 commits behind, including the whole opt-in web API series.
- Replaying just this PR's own four commits (`git rebase --onto main d158208a1`) conflicts **semantically**, not just textually, in the first commit already: `AsyncWaiter` here captures memory state into the `CancellationTokenSource` timeout design that #3095 has since replaced with pump-driven timeouts; `EventLoop.cs` and `ModuleLoadCompletion.cs` were both reworked since (timers/pump, async module loading); and `Engine.cs`'s ownership region is now the reworked protocol from `8a10ea1a6`, which "Integrate memory accounting with engine ownership" predates. How the accounting should hook into those redesigned seams is an author's call, so I stopped short of inventing it in a security stack.
- Separately, the upper layers still carry the net472 compile errors from the 08-16 windows legs (`ParsingLimitTests`' `string.Join(char, …)`, non-generic `TaskCompletionSource` and `ArrayBufferWriter<T>` in the test projects — #3060's current head is red on windows for exactly these).

The twelve approvals stand for the designs. Once the stack is cascaded onto post-#3035 `main` I'll re-review each layer's refreshed diff and merge bottom-up as each goes green.

sebastienros:
Thanks for stopping rather than guessing — the conflicts were semantic exactly as described.

I replayed the memory layer onto current `main`, including the merged #3035 ownership protocol, pump-driven Atomics/timers, the expanded web task sources, host-stream bridges, and the latest `QuotaExceededError` surface. During the refresh #3036 was merged server-side, so I preserved the post-merge fixes as follow-up **#3216**: https://github.com/sebastienros/jint/pull/3216

Key resolutions in #3216:
- finite continuations retain their originating allocation state (promises/modules/Atomics, fetch/streams, timers, idle callbacks and individual scheduler tasks);
- persistent external event sources start a fresh per-delivery budget rather than accumulating forever against object creation;
- transferred callbacks suspend/restore the original thread segment and account callback-thread allocations, including nested transfers;
- host Task failures preserve explicitly marked Jint constraint provenance through aggregates without treating ordinary host `TimeoutException` as fatal;
- a memory breach clears transient work, abandons stream bridges and advances the generation in `finally`, while retaining cleanup failures as `InnerException`;
- host-stream I/O ends its Busy state on the completion thread before any generation/memory-gated delivery;
- current-main `QuotaExceededError`, WPT and web API changes remain intact.

Validation: all-target Release build; 9,810 unit tests passed / 5 skipped (known local timezone test excluded); public interface 2,105 passed / 9 config skips; host-contract public interface 2,109 passed / 5 inverse skips; focused 112-test memory/concurrency/web selection passed three times; post-rebase focused suites passed 352 + 112. The benchmark restore is blocked only because the mandated proxy lacks `YantraJS.Core` 1.2.462, so I did not use another feed.

PR #3037, opened 2026-08-16
Bound parser source length and AST size

sebastienros:
## Summary

- add opt-in parser source-length and AST-node limits across scripts, modules, eval, Function constructors, ShadowRealm, debugger evaluation, and module builders/loaders
- preserve parser constraints through prepared code, execution contexts, async module thread hops, and nested dependency loading
- integrate fatal `ParsingLimitException` semantics with engine ownership and memory accounting, including cleanup of pending loads, waiters, async ownership, and memory operation state
- document the controls and their threat-model boundaries

## Stack

This PR is stacked on #3036 -> #3035 -> #3030 and targets `sebros/improve-memory-accounting`.

## Integration decisions

- async module completions capture both originating parsing constraints and memory-operation state at registration, then restore both for the queued engine turn
- parsing-limit failures remain fatal and non-script-catchable across module loading, debugger evaluation, CLR callbacks, and all CLR proxy traps
- deferred module failures remove pending-load entries and waiter graphs before propagating
- public ShadowRealm evaluation participates in the same host ownership guard as other engine entry points
- default parser settings such as source retention survive tighter active parsing constraints

## Validation

- `dotnet build Jint/Jint.csproj -c Release` (all target frameworks)
- full `Jint.Tests.PublicInterface`: 1,585 passed, 9 configuration skips
- host-contract focused security suite: 129 passed
- focused core debugger/module/constraint/interop suite: 739 passed, 1 existing skip
- combined parser/module/memory/concurrency suite: 231 passed
- fresh specialist review and confirmation re-review: clean after fixes
- `git diff --check`

The full core suite completed 5,760 tests with one unrelated environment-dependent Date failure (year 10000 rendered as Dec 31 9999 in Pacific time). The full solution build could not restore `Test262Harness.Console` 1.1.2 because the required package proxy does not currently carry it; no alternate feed was used.


lahma (approved):
Reviewed as the incremental diff against `sebros/improve-memory-accounting`. This layer is fully opt-in (both limits default to `null`), the composition rule is the right one (per-call/prepared options can only tighten, `Combine` takes the minimum), and the coverage in `ParsingLimitTests` is excellent — the cache-poisoning cases (eval cache and dynamic-function cache keyed on constraints), prepared-code inheritance into nested `eval`, the async module thread hop carrying constraints, and CLR non-catchability are exactly the places this could have gone quietly wrong. Nice touches along the way: the source-offset padding length is now checked (and computed in `long`) *before* the padding string is allocated, the Function-constructor wrapper length is checked before the join materializes, `DefaultModuleLoader` stops reading the file at limit+1 chars instead of slurping it, and `ClassDefinition` no longer constructs a throwaway `Engine` just to parse the generated default constructor. Approving.

Small notes, none blocking:

- `Module` implements `IScriptOrModule.ParserOptions` (`ParserOptions?`) with an `internal virtual ParserOptions ParserOptions => null;` — the virtual should probably be declared nullable to match the interface rather than relying on however this dodges the nullability warning.
- `DebugHandler.Evaluate(Prepared)` now pushes a dedicated execution context carrying a `ScriptRecord` where it previously executed in the paused frame without one. The environments/realm/strictness are copied from the active context so semantics should be unchanged, but it is a structural change to debugger evaluation beyond the limits feature — worth one sentence in the PR body.
- The `_parsingConstraintsOverride` save/restore sites all restore in `finally` — good; anyone adding a new override site later should keep that discipline, since a leaked override silently tightens (or loosens) every subsequent parse on the engine.

Merge remains blocked below by the #3035 fix; this layer itself is good to go once the stack is re-rolled.

lahma:
Follow-up to my approval: the `windows` CI leg on this branch (run 31961149930) fails to **compile** the test project on net472 — `ParsingLimitTests.cs` lines 77/191/207/226/632 use `string.Join(char, IEnumerable<string>)`, which does not exist on .NET Framework. The Jint assembly has that polyfill in `Polyfills.cs`, but `Jint.Tests.PublicInterface` doesn't see it. Switch those call sites to `string.Join(",", …)` (string separator) when re-rolling the stack — the windows leg is the only one that compiles net472, which is why the other legs stayed green.

The same run also has a single Test262 failure (an async `Atomics.waitAsync` test, 1 of 102,668 on net10.0) — that one looks like a timing flake on a loaded runner rather than anything this layer touches, but worth an eye on the re-run.

lahma:
Rebased the rest of the stack onto current `main` and started merging it bottom-up, as agreed. #3036 is in (`fd50151`); this is the note on everything that had to change on the way, because some of it is not what either of us approved back in August.

**Where the stack was.** The eleven layers above #3036 were 107 commits behind `main` and still carried the *pre-rework* #3035 commits (`7cda7429e`, `d158208a1`) that your squash superseded. Each layer is now collapsed to a single commit — the intra-layer history had no consumer once the layers get squash-merged, and one commit per layer makes the cascade after each merge trivial. That paid off immediately: after #3036 merged, GitHub retargeted #3037 to `main`, rebased it to one commit and dropped the duplicate by patch identity, with no local work at all.

Every layer builds on all five TFMs, and the tip is green locally: `Jint.Tests` 9904 (net10) / 7089 (net472), `Jint.Tests.PublicInterface` 2348 / 1862. Every layer also gets the full matrix in CI, since `pr.yml` has no `branches:` filter.

**Two decisions nobody approved, both because `main` moved:**

1. **The propagate list.** `main` extracted the job-boundary list into `ConstraintFailure.MustPropagate`; #3046 extracted an overlapping host-boundary one into `Throw.MustPropagateHostException`. Rather than keep two lists that will drift, the job list is now defined over the host list plus `TimeoutException` and `OperationCanceledException`. The two stay separate deliberately: a host method may raise either of those as its own ordinary failure, so an interop boundary must not read them as a bound, while on a queued turn they are only ever Jint's own.

2. **`AddToEventLoop`.** #3036 made the memory-state argument required, but `main` has since added around thirteen cross-thread enqueue sites — timers, `fetch`, streams, WebSocket, `EventSource`, `MessagePort`/`BroadcastChannel`, finalization-registry cleanup — none of which existed when the PR was written and none of which its stated scope covers. I added a documented *unaccounted* overload rather than quietly extending the budget into those subsystems without tests or review. Extending it is a real follow-up, not something to slip in here.

**A defect that is on `main` today, independent of this stack.** #3052's own test caught it and it is wider than the test. `TypeResolver.TryFindMemberAccessor`'s interface loop reflected with `iface.GetMethods()` — no binding flags — while the static lane passes `Public | Static | FlattenHierarchy`. On .NET Framework `System.Type` implements the COM mirrors `_Type` and `_MemberInfo`, which declare *instance* `GetType()`, `InvokeMember()`, `GetMethod()` and the rest, so nine of ten probed members were projected as **static** members of `System.Type` and sailed past the `MethodInfo { IsStatic: true }` guard — which never matched them, because they are not static. `System.Type.InvokeMember` was reachable from script on net472 with `AllowClr()`. The fix is to pass the lane's own flags into `iface.GetMethods(...)`; it is in #3052.

Methods only, on purpose: `ObjectWrapperReportedMethodBindingFlags` defaults to exactly what the parameterless overload uses, so the instance lane is provably untouched — but the *property* default is `Instance | Public` with no `Static`, so doing the same to the property loops would be a real behaviour change, and those loops are not reachable from the static lane anyway.

**A stale expectation, not a hole.** `ClrWriteConfigurationTests.ArrayGenericsCannotMutateClrListsByDefault`'s `@@species` case used an empty source, and #3043 confined `filter`'s `operations.SetLength(to)` to `a is JsArray` — correctly, since `filter` has no length step in the spec at all. Worth knowing what that removed: on the old branch, `[].filter` with a CLR-list species **truncated the host list to empty** whenever writes were permitted. Your default blocked it; #3043 removed the write outright. The theory case now uses `const source = [3, 4]` with a truthy callback so it exercises the `CreateDataPropertyOrThrow` that is the actual write. Every route where a write is genuinely attempted still throws and leaves the list intact.

**Three net472 compile breaks and one flake.** `Jint.Tests` and `Jint.Tests.PublicInterface` did not compile on net472 anywhere in the stack — `string.Join(char, IEnumerable<string>)`, the non-generic `TaskCompletionSource`, and `ArrayBufferWriter<T>` being internal on .NET Framework — so the windows leg had never reported a single *test* result for these layers. Fixing that is what surfaced the item above. Separately, `HostResultLimitsTests.OperationDeadlineSpansEvaluationAndConversion` failed roughly one run in three on net472: the 400 ms budget leaves 150 ms of headroom over the first 250 ms pause, and a cold first `Evaluate` on .NET Framework eats it, so the `TimeoutException` escaped from `Evaluate` instead of the conversion the test means to observe. It now warms parse, the interop call path and `ConvertResult` before `Begin`.

**Unchanged from the August review:** the five default flips — #3051 error redaction, #3054 `AllowClrWrite` false, #3056 the array `Copy` default (which reverses 4.14's `LiveView`), #3057 `StackOverflowGuard` on, #3058 `AgentCanSuspend` false — still want to ship as one major-release migration story rather than as five separate surprises. #3054 and #3056 in particular compose: a no-write `LiveView` would break script array mutation, and `Copy` is what restores it.

I will keep merging bottom-up as each layer goes green and re-verify parentage after every merge. Shout if you want any of the above undone.

PR #3045, opened 2026-08-16
Add module graph security limits and destination policy

sebastienros:
## Summary

- add cumulative module-count and source-byte limits plus per-top-level-load graph-depth and resolution-hop budgets
- add a fail-closed module destination allowlist for schemes, origins, hosts, and file roots
- preserve registration-time generation, realm, engine ownership, memory, parser, cancellation, and graph state across asynchronous module completion
- keep engine resource/control failures non-script-catchable while ordinary loader transport timeout/cancellation remains a load failure
- document the hardened module profile and threat-model guidance

## Stack

This PR targets #3037 (`sebros/parser-source-ast-limits`). The security stack is:

#3037 -> #3036 -> #3035 -> #3030

## Validation

- Release multi-target `Jint/Jint.csproj` build: 5/5 target frameworks
- full `Jint.Tests.PublicInterface`: 1643 passed, 9 skipped
- relevant core module, concurrency, constraint, and garbage-collection tests: 184 passed
- stacked security tests with host-contract verification enabled: 222 passed
- `git diff --check`
- specialist review followed by clean re-review

The approved NuGet proxy does not yet carry the target branch's newly bumped `Meziantou.Analyzer` 3.0.156 (nearest 3.0.141), so local post-rebase validation temporarily used the immediately preceding cached dependency versions without committing that override. The exact module source had already built and passed before the target's dependency-only update.

### DIFF
--- a/.github/THREAT_MODEL.md
+++ b/.github/THREAT_MODEL.md
@@ -81,6 +81,8 @@ Defaults are compatibility choices, not a hardened profile.
 | Writes through projected CLR objects | Enabled | A default engine can mutate host objects passed with `SetValue` |
 | CLR array conversion | Live view | Script writes can mutate a projected CLR array |
 | Modules | Disabled | The fail-fast loader rejects imports |
+| Module graph limits | Unlimited | Count, source bytes, graph depth, and resolution hops are unbounded unless configured |
+| Module load policy | None | A custom loader's resolved targets are unrestricted unless a policy is configured |
 | `require` | Disabled | No CommonJS-like loader global |
 | Web platform APIs (`Options.WebApi.Features`) | `None` | No `console`, timers, `URL`, `Blob`, … globals exist |
 | `fetch` | Disabled, and never part of `WebApiFeatures.Default` | Only `UseFetch()` grants script outbound HTTP |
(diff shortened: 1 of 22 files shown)


lahma (approved):
Reviewed as the incremental diff against `sebros/parser-source-ast-limits`. Everything here is opt-in (all four limits default to unlimited, `LoadPolicy` defaults to null), the single-choke-point discipline is right (`GuardedResolve` is the only import-driven resolution path, so the policy cannot be bypassed; registration indexing applies policy but deliberately doesn't consume hops), and the accounting is transactional — `RegisterModuleWithAccounting` reserves before the debugger callback can re-enter and rolls back on failure, and the registration-index pass now restores its version and in-flight marker when a policy throws mid-pass. The depth design is sensible: a real-path check during the walk for early abort, then the SCC/condensation longest-path validation for a deterministic answer independent of sibling and async completion order — I traced the Kosaraju + Kahn implementation and it holds up (root-seeded distances, the `> 0` guard keeping unreachable components inert, cycles contributing their full member count). Converting `InnerModuleLoading` to a worklist also quietly removes native-stack recursion for deep synchronous graphs, which is a win on its own.

Two details I particularly like: the `EngineAbortRegistry` CWT marker that lets the pipeline distinguish the engine's own `TimeoutException`/`OperationCanceledException` from a loader's transport timeout without inventing new exception types, and `ModuleLoadCompletion` capturing the cancellation token at registration so a background settle never has to enter the engine to rediscover it.

The 50+ `HostModuleGraphSecurityTests` cover exact boundaries, dedup/diamond/cycle counting, per-operation hop budgets on pooled engines, async pump propagation, the ordinary-vs-engine cancellation split from both loader kinds, and every allowlist dimension including the separator-boundary file-root case. Approving.

Non-blocking notes:

- `MaxModuleCount`/`MaxTotalModuleSourceBytes` reject non-positive values at engine construction, whereas the constraint helpers (`LimitMemory`, `MaxStatements`) treat non-positive as "remove the limit". Failing loud is arguably the better convention and the XML docs say so — just noting the deliberate divergence.
- `ModuleAllowlistPolicy`'s mutable `List<>` properties mean a policy shared across engines can be mutated mid-load; it's engine-thread-consulted so this is the usual Options-sharing caveat, and the symlink non-guarantee is documented. Fine as is.
- Depth is only *authoritatively* enforced when the graph finishes loading (the SCC pass), so a hostile graph inside the count/byte budgets still gets fully fetched before a depth rejection; the during-walk real-path check catches the linear-chain case early. That trade for determinism seems right — worth keeping the rationale in the XML doc as it is.

Stack merge remains gated on the #3035 fix below.

PR #3046, opened 2026-08-16
Bound result conversion, JSON serialization, and error rendering

sebastienros:
## Summary

- add `ResultLimits` and detached `Advanced.ConvertResult` handling for arrays, typed/binary values, maps, sets, and enumerable object graphs
- bound Jint JSON serialization and JavaScript error rendering by depth, structural work, strings, output characters, and bytes
- preserve one DAG-wide constraint cadence and count repeated shared references through `MaxPropertyCount`
- integrate result boundaries with Engine ownership and explicit deadline/memory operations spanning evaluation or import plus result handling
- keep result/resource failures fatal through CLR callbacks, proxy traps, reflection/conversion, debugger evaluation, and every async loader settlement path
- reject same-instance `JsonSerializer` reentry before nested callbacks can replace outer traversal state or limits
- document the cumulative untrusted-script deployment model in README and `.github/THREAT_MODEL.md`

## Stack

1. #3030 `initial-threat-model`
2. #3035 `sebros/reject-concurrent-engine-use`
3. #3036 `sebros/improve-memory-accounting`
4. #3037 `sebros/parser-source-ast-limits`
5. #3045 `sebros/stack-module-graph-limits`
6. this PR

## Validation

- Release multi-target `Jint/Jint.csproj` build
- full `Jint.Tests.PublicInterface` default and host-contract-verification runs
- full `Jint.Tests` default and host-contract-verification runs under UTC
- targeted result-limit, JSON, concurrency, memory, parser/module, proxy, and async-loader tests
- clean specialist security follow-up review
- `git diff --check`

### DIFF
--- a/.github/THREAT_MODEL.md
+++ b/.github/THREAT_MODEL.md
@@ -98,6 +98,7 @@ Defaults are compatibility choices, not a hardened profile.
 | Debugger | Disabled | Debug callbacks and debugger expressions are inactive |
 | Detailed CLR resolution errors | Disabled | Reduces script-visible CLR surface disclosure |
 | Function source retention | Disabled | Submitted function source is not retained solely for `toString()` |
+| Result conversion and JSON output limits | Unlimited | Compatibility default; hostile output is unbounded unless the host opts in |
 
 ## Threat inventory
 
@@ -341,9 +342,13 @@ when the outer deadline expires.
 
 ### TM-09: A sequence of engine entries escapes a per-entry budget
 
-**Threat.** `Execute`, `Evaluate`, `Invoke`, and `Call` are separate top-level runs. Built-in
+**Threat.** `Execute`, `Evaluate`, `Invoke`, `Call`, `Advanced.ConvertResult`, public
+`JsonSerializer.Serialize` calls, and bounded
+`JavaScriptException.GetJavaScriptErrorString` are separate top-level runs. Built-in
 constraints reset around each one, so a host loop that invokes a script once per record
-grants a fresh timeout, statement budget, and allocation baseline to every call.
+grants a fresh timeout, statement budget, and allocation baseline to every call. Evaluating
+a result and then converting or serializing it also grants each phase a fresh budget unless
+the host brackets both with one operation deadline.
 
 **Existing mitigations.**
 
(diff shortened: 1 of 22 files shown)


lahma (approved):
Reviewed as the incremental diff against `sebros/stack-module-graph-limits`. Approving.

What I checked closely:

- **`JSON.stringify` compatibility under defaults.** With `ResultLimits.Unlimited`, `_maxDocumentLength` still resolves to `JsString.MaxLength` and `_documentLengthResultLimit` goes null, so the script-visible overflow stays `RangeError: Invalid string length` exactly as before — the `ResultLimitExceededException` only ever wins when a host-configured bound is the tighter one. The added per-node work in the unlimited path is a couple of compares and a checked add, which is noise next to the property walk itself.
- **The serializer's new `Enter()` reentrancy rejection** turns what was silent state corruption (same instance re-entered from `toJSON`/replacer) into a deterministic `InvalidOperationException` — good, and the instance still resets per call so a host holding one per engine is unaffected.
- **`ResultConverter`**: path-based cycle rejection with reference identity, `IObjectWrapper` targets handed back without walking the CLR graph, typed-array/ArrayBuffer/DataView byte accounting before the copy, functions and symbols refused, and a constraint-check cadence inside the loops so a hostile getter can't hide from the time budget. The engine-affinity check on `ConvertResult` and routing through `ExecuteWithConstraints` (so getters/traps run under constraints) are both right.
- **`Throw.MustPropagateHostException` consolidation.** The three diverging ad-hoc exception lists (proxy traps, `ClrFunction`, `DefaultTypeConverter`, debugger, module pipeline) are now one predicate, and it closes a real hole: previously a host's `Interop.ExceptionHandler` could observe (and convert to a catchable JS error) a `MemoryLimitExceededException` or `ExecutionCanceledException` thrown while a `ClrFunction` was on the stack. The union also makes `RegexMatchTimeoutException`/`PlatformNotSupportedException` handling consistent between the two module-pipeline predicates that previously each had only one of them.
- **`GetJavaScriptErrorString(ResultLimits)`** running the render under constraints when the error is an object is the correct call, since a script-defined `stack` accessor executes JS.

Defaults are fully compatible (`ResultLimits.Unlimited`), the new surface is documented in AGENTS.md's contract table, and the test spread (HostResultLimitsTests, the JsonSerializer additions, the async-loader settlement paths, proxy traps) covers the boundaries and the fatality guarantees. Stack merge still gated on the #3035 fix below.

PR #3051, opened 2026-08-16
Sanitize host and module failures

sebastienros:
## Summary

- redact caught CLR exception messages, CLR resolution details, and module load failures by default while retaining explicit detailed-development opt-ins
- apply CLR error decorators exactly once across synchronous and asynchronous module boundaries, including parse, resolve, build, debugger, and custom-loader failures
- keep every resource/control failure fatal while treating ordinary loader transport timeout/cancellation as sanitized load failures
- scope module-policy provenance to the active engine and disclosure/decorator configuration, and preserve bounded host-only diagnostics through `TryGetClrException`

## Security integration

This PR is stacked on #3046 and preserves the ownership, memory, parser, module graph, result-limit, fatal-cleanup, and bounded-rendering layers below it. Module policy provenance uses an opaque active-policy token, so foreign/stale errors cannot bypass redaction or decoration. Blocking imports rethrow an original exception only when it carries the exact approved error; otherwise they throw the sanitized/decorated error and retain the original through the host-only diagnostic contract.

Stack: #3046 -> #3045 -> #3037 -> #3036 -> #3035 -> #3030.

## Conflict resolution

The reviewed redaction commits were transplanted from `330224cfc` onto `c7dac6a31`. Conflicts in `Engine.Modules`, `ModuleLoadCompletion`, `ModuleLoader`, `AsyncModuleLoader`, `AcornimaExtensions`, `Throw`, and `Options` retained the newer ownership, parsing, module-graph, result-limit, and engine timeout/cancellation classifications while layering sanitized disclosure and exactly-once decoration. `.github/THREAT_MODEL.md` preserves both graph-policy guidance and the sanitized-default/detailed-opt-in trust boundary.

## Review

A specialist security review found two issues: reusable boolean provenance across engines/configurations, and blocking imports rethrowing raw undecorated exceptions. Both were fixed with active-policy tokens and exact approved-error memoization. A fresh follow-up security review reported no high-confidence findings.

## Validation

- Release multi-target `Jint/Jint.csproj` build: 5 target frameworks, 0 warnings/errors
- full `Jint.Tests.PublicInterface`: 1691 passed, 9 skipped
- full `Jint.Tests.PublicInterface` with `JINT_HOST_CONTRACT_VERIFICATION=1`: 1695 passed, 5 skipped
- full UTC `Jint.Tests`: 5771 passed, 4 skipped
- focused disclosure/module policy: 127 passed
- focused result/parser/ownership/retention: 98 passed
- focused GC/debugger/interop: 518 passed, 1 skipped
- `git diff --check`

Validation used only `https://packagefeedproxy.microsoft.io/nuget/v3/index.json`. Because that proxy currently exposes Meziantou.Analyzer 3.0.141 rather than the pinned 3.0.156, the analyzer version was temporarily lowered only for local validation and restored exactly before commit; dependency manifests are unchanged.

### DIFF
--- a/.github/THREAT_MODEL.md
+++ b/.github/THREAT_MODEL.md
@@ -97,6 +97,8 @@ Defaults are compatibility choices, not a hardened profile.
 | `Atomics.wait` suspension | Enabled | Script may block the engine thread, including indefinitely |
 | Debugger | Disabled | Debug callbacks and debugger expressions are inactive |
 | Detailed CLR resolution errors | Disabled | Reduces script-visible CLR surface disclosure |
+| Detailed caught CLR exception messages | Disabled | Replaces host exception text with a generic script-visible error |
+| Detailed module load errors | Disabled | Hides loader messages, canonical paths, URLs, and parse source names from script |
 | Function source retention | Disabled | Submitted function source is not retained solely for `toString()` |
 | Result conversion and JSON output limits | Unlimited | Compatibility default; hostile output is unbounded unless the host opts in |
 
@@ -411,6 +413,9 @@ traces, and `import.meta.url` can disclose paths, hostnames, queries, or credent
 - `IModuleLoadPolicy` can inspect every final resolved target. `ModuleAllowlistPolicy`
   provides composable scheme, exact-host, exact-origin, canonical-file-root, and bare-
   specifier controls.
+- Loader exceptions and module parse locations are redacted from script by default. The
+  original exception remains available to the host through
+  `JintException.TryGetClrException`.
 
 **Missing or residual mitigation.**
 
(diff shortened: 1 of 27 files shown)


lahma (approved):
Reviewed as the incremental diff against `sebros/stack-bounded-result-rendering`. The mechanism is sound and consistently applied: every module/CLR failure funnels through one creation point (`Throw.CreateClrError` / `Modules.CreateLoadError`), the error carries an opaque policy token scoped to the `(ExposeDetailedLoadErrors, decorator)` configuration so decoration is exactly-once and a stale or foreign error can't ride through redaction, and the blocking-import rethrow only reuses an original exception when it carries the exact approved error value. `SetError(string)` staying un-redacted is the right call (the host chose that message), and preserving the constructor type through `GetErrorConstructor` keeps `instanceof SyntaxError` working for module parse failures under redaction. Host-side diagnostics remain complete through `TryGetClrException`. The 510-line `HostErrorDisclosureTests` sweep plus the updated existing suites cover both configurations. Approving.

Points to carry forward:

1. **This is the first layer in the stack that flips a default.** Scripts that today read a caught CLR exception's message (`CatchClrExceptions`) or a module load failure's message will see `"A host operation failed."` / `"Could not load module."` after this merges. That is the intended trust boundary and the opt-outs (`ExposeDetailedErrors()`, the two granular flags) are there — but embedders' tests and error-branching code *will* break, so this needs a prominent release-notes entry and, in my view, argues for the stack landing in a major (or at least clearly-flagged) release. Same applies to the later default-flipping layers above this one.
2. `Throw.FromClrException` now rethrows `MustPropagate` exceptions before consulting the interop `ExceptionHandler` — meaning a handler that used to convert an `OperationCanceledException` into a catchable JS error no longer can. Safe direction, worth a line in the notes.
3. Nit: `ModuleLoadCompletion.Abort(Exception)` is dead code — `SetConstraintError` inlines the same body and nothing calls `Abort`. Please drop it.
4. Nit: under redaction the module parse-error `SourceLocation` drops the source path (`ToLocation(ex, exposeDetails ? source : null)`), so `error.stack` hides the module name too. That's coherent with the policy, but since `Module.Location` is host-chosen (often a bare specifier the script itself wrote), consider whether hiding the *specifier* is worth the diagnostics loss — a script usually already knows what it imported.

PR #3054, opened 2026-08-16
Disable projected CLR writes by default

sebastienros:
## Summary

- disable direct writes through projected CLR objects by default
- add explicit `AllowClrWrite()` opt-in without making parameterless `AllowClr()` grant write authority
- preserve method and extension-method side effects as a separate documented capability boundary
- enforce write authority across fields, properties, statics, indexers, arrays, lists, dictionaries, dynamic objects, proxies, cached accessors, and borrowed/species array operations
- preserve ordinary JavaScript object semantics and write-enabled interop compatibility
- document the breaking default, migration, and residual risks in README and the threat model

## Security review

Specialist review found and this branch fixes:

- optimized borrowed/species array operations resizing CLR lists without write authority
- cached/detached setter and string-key routes bypassing frozen CLR wrappers
- frozen wrapper element deletion and partial `pop`/`shift`/`splice` mutation
- ordinary JavaScript non-extensibility bypasses in array `Set` and `CreateDataPropertyOrThrow` fast paths
- denied-delete semantics for absent array-like properties and numeric dictionary-shaped keys
- write-enabled `JsonArray` growth compatibility and .NET Framework base compatibility

A clean follow-up review reported no significant issues.

## Validation

- Release multi-target `Jint/Jint.csproj` build: net462, netstandard2.0, netstandard2.1, net8.0, net10.0
- PublicInterface default: 1,732 passed, 9 skipped
- PublicInterface host-contract verification: 1,736 passed, 5 skipped
- core default: 5,774 passed, 4 skipped
- core host-contract verification: 5,774 passed, 4 skipped
- `git diff --check`

Validation used the mandated NuGet proxy with the analyzer-only local workaround restored before publishing.

## Stack

#3052 -> #3051 -> #3046 -> #3045 -> #3037 -> #3036 -> #3035 -> #3030

Base: `sebros/stack-clr-containment-medium`

### DIFF
--- a/.github/THREAT_MODEL.md
+++ b/.github/THREAT_MODEL.md
@@ -79,8 +79,8 @@ Defaults are compatibility choices, not a hardened profile.
 | CLR namespace access (`Interop.Enabled`) | Disabled | `System`, `importNamespace`, and `clrHelper` are not installed |
 | `AllowGetType`, `AllowSystemReflection` | Disabled | Blocks instance type discovery, type widening, and reflection namespace/object access |
 | `AllowedAssemblies` | Empty until `AllowClr` adds entries | Closed allow-list for `System` / `importNamespace` type discovery, not for what admitted APIs can do |
-| Writes through projected CLR objects | Enabled | A default engine can mutate host objects passed with `SetValue` |
-| CLR array conversion | Live view | Script writes can mutate a projected CLR array |
+| Writes through projected CLR objects | Disabled | Fields, properties, indexers, dictionary/list entries, and live array elements require explicit `AllowClrWrite()` opt-in |
+| CLR array conversion | Live view | Script observes the CLR array directly; element writes remain blocked unless CLR writes are enabled |
 | Modules | Disabled | The fail-fast loader rejects imports |
 | Module graph limits | Unlimited | Count, source bytes, graph depth, and resolution hops are unbounded unless configured |
 | Module load policy | None | A custom loader's resolved targets are unrestricted unless a policy is configured |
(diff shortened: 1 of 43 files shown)


lahma (approved):
Reviewed as the incremental diff against `sebros/stack-clr-containment-medium`. The mechanics are thorough and the bypass-chasing shows: the borrowed/species array operations now funnel length and element mutation through the authority-checked `Set` paths instead of touching the CLR list directly, frozen-wrapper deletes answer honestly per key kind, the `CreateDataPropertyOrThrow` fast path respects non-extensibility, and `ReflectionDescriptor.DoSet` fails **loudly** with a `TypeError` — which matters enormously for migration, because a silently-ignored write would have embedders chasing ghost bugs for weeks. `AttachOwner` wiring so a frozen wrapper's members refuse writes is a nice closure of the detached-descriptor route. Approving the change itself.

The caveat, stated as plainly as I can: **flipping `Interop.AllowWrite` from `true` to `false` is the single most breaking change in this stack.** `engine.SetValue("obj", dto); engine.Execute("obj.Name = 'x'")` is the bread-and-butter embedding pattern, and every host doing it must now add `AllowClrWrite()`. I agree with the direction — script mutating host state should be an explicit grant — but this cannot ship in a patch/minor release. It needs a major version, a first-line release-notes entry, and ideally the migration one-liner in the exception message itself (the `TypeError` currently says "Cannot assign to read only property 'X'", which reads like a script bug; something like "…direct CLR writes are disabled by default, see Options.AllowClrWrite" would route embedders straight to the fix).

Two smaller notes:

- `ListOperations.EnsureCapacity` changed meaning from capacity reservation to observable length growth (routed through `SetLength`). Under `AllowClrWrite()` the end state of the array ops is the same, but a mid-operation failure can now leave default-filled elements where it previously left capacity. Worth a comment at the call sites that rely on fill-after-grow.
- Sloppy-mode assignments through the reflection descriptor now throw `TypeError` where the spec would silently ignore a failed write — a deliberate, contained interop divergence in the loud direction; worth one line in the README's breaking-change section so nobody files it as a bug.

PR #3056, opened 2026-08-16
Default CLR array projection to isolated copies

sebastienros:
## Summary
- make `ArrayConversionMode.Copy` the breaking security default while retaining explicit `LiveView` compatibility
- keep copied arrays as mutable engine-owned JS snapshots while requiring separate `AllowClrWrite()` authority for LiveView write-through
- preserve cyclic/deep graph identity with transactional cache publication, rollback, resource accounting, and weak converter-aware provenance
- document the migration and add public/core coverage plus a trustworthy cold-projection benchmark

## Security stack
#3054 -> #3052 -> #3051 -> #3046 -> #3045 -> #3037 -> #3036 -> #3035 -> #3030

## Validation
- Release Jint multi-target build (`net462`, `netstandard2.0`, `netstandard2.1`, `net8.0`, `net10.0`)
- PublicInterface: 1,745 tests default; 1,745 with host-contract verification
- Core under UTC: 5,800 tests
- Focused host-verification interop/memory/result/GC/concurrency: 439 tests
- Benchmark project Release build and `git diff --check`
- clean specialist follow-up review

## Benchmark
Default BenchmarkDotNet job, `ClrArrayProjectionBenchmark` (cold first crossing):

| Length | Copy | LiveView | Copy allocated | LiveView allocated |
| ---: | ---: | ---: | ---: | ---: |
| 0 | 445.98 ns | 53.29 ns | 458 B | 192 B |
| 10 | 574.63 ns | 53.50 ns | 774 B | 192 B |
| 100 | 2.104 us | 52.60 ns | 3,661 B | 192 B |
| 1,000 | 16.764 us | 53.58 ns | 32,465 B | 192 B |

### DIFF
--- a/.github/THREAT_MODEL.md
+++ b/.github/THREAT_MODEL.md
@@ -80,7 +80,7 @@ Defaults are compatibility choices, not a hardened profile.
 | `AllowGetType`, `AllowSystemReflection` | Disabled | Blocks instance type discovery, type widening, and reflection namespace/object access |
 | `AllowedAssemblies` | Empty until `AllowClr` adds entries | Closed allow-list for `System` / `importNamespace` type discovery, not for what admitted APIs can do |
 | Writes through projected CLR objects | Disabled | Fields, properties, indexers, dictionary/list entries, and live array elements require explicit `AllowClrWrite()` opt-in |
-| CLR array conversion | Live view | Script observes the CLR array directly; element writes remain blocked unless CLR writes are enabled |
+| CLR array conversion | Copy | Script receives a detached native JavaScript array snapshot |
 | Modules | Disabled | The fail-fast loader rejects imports |
 | Module graph limits | Unlimited | Count, source bytes, graph depth, and resolution hops are unbounded unless configured |
 | Module load policy | None | A custom loader's resolved targets are unrestricted unless a policy is configured |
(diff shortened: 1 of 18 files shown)


lahma (approved):
Reviewed as the incremental diff against `sebros/disable-clr-writes-default`. Approving.

The two defaults compose coherently: with `AllowWrite` now false (#3054), a LiveView array would refuse `push`/element writes, so making `Copy` the default actually *restores* script-side array mutability for the common case — the script gets a snapshot it can freely mutate, and the host graph is isolated. Good sequencing.

The transactional `ArrayCopyContext` earns its complexity: cyclic and mutually-referential CLR arrays reproduce their cycles in the JS snapshot, partially-built snapshots can never leak into `_objectWrapperCache`/the recent-wrapper ring when a constraint aborts mid-copy (publication is deferred to root success, with a real savepoint/rollback on the recent cache), and the generation-splicing hazard — a cached nested snapshot pointing back into an older copy of the same cycle — is explicitly guarded by `CanReuseCachedArray`. The iterative worklist keeps deep graphs off the native stack, and `CheckInteropProjectionConstraints` is a thoughtful touch: cancellation and an armed `OperationDeadlineConstraint` still bound a host-side projection that has no execution entry to charge.

Notes:

1. **Third default flip in the stack; same versioning point as #3054/#3051.** This one has an extra wrinkle: the LiveView default only shipped in 4.14, so hosts have now been asked to adapt twice in one year. The release notes should present #3054 + #3056 together as one migration story ("writes need `AllowClrWrite()`, live views need `ArrayConversionMode.LiveView`"), because the pair reads much more sensibly than either alone.
2. The benchmark table quotes Copy vs LiveView, but the interesting regression check is new-Copy (with the context bookkeeping) vs the old simple copy loop for the dominant flat, acyclic case — the CWT `Begin`/`Complete` per array isn't free. Non-blocking, but if you have the pre-context numbers handy, put them in the PR body; if flat-array copies got measurably slower, a fast path that skips the context when the element type can't contain an `Array` (primitives, strings, sealed non-array element types) would recover it cheaply.
3. Doc nit: the enum doc now says non-empty multidimensional arrays "throw during conversion" — worth stating the exception type so hosts can catch it deliberately.

PR #3057, opened 2026-08-16
Enable the stack-overflow guard by default

sebastienros:
## Summary

- enable `Options.Constraints.StackOverflowGuard` by default for every new `Engine`
- retain `options.Constraints.StackOverflowGuard = false` as the explicit compatibility opt-out for trusted, independently bounded scripts
- preserve the distinction between the native-stack backstop and `MaxRecursionDepth`, which counts repeated occurrences of one function definition
- update the README and threat model with the breaking default, migration guidance, residual risks, and hardened configuration

## Stack

This PR targets `sebros/default-clr-array-copy-498` (#3056), then stacks through #3054 -> #3052 -> #3051 -> #3046 -> #3045 -> #3037 -> #3036 -> #3035 -> #3030.

## Integration

The reviewed stack-guard commit was transplanted onto the live #3056 tip `39b68c1a398d8cea2e65026f5c202d7417311f77`. The only conflict was in `.github/THREAT_MODEL.md`; resolution retained both the cumulative parser-limit guidance from the stack and the default-on stack guard guidance. No prior security behavior was weakened.

A fresh security specialist review against `origin/sebros/default-clr-array-copy-498` specifically checked recursion-path coverage, thread/async state, exception flattening, cleanup/retry, `MaxRecursionDepth` interaction, and hot-path behavior. It reported no security findings.

## Validation

- Release multi-target `Jint/Jint.csproj` build: clean, 0 warnings
- Release `Jint.Benchmark` build: clean, 0 warnings
- full `Jint.Tests.PublicInterface`: 1,737 passed, 9 configuration-specific skips
- full `Jint.Tests.PublicInterface` with host-contract verification: 1,741 passed, 5 inverse-configuration skips
- full `Jint.Tests` under UTC: 5,796 passed, 4 existing skips
- host-verified focused stack/recursion/proxy/module/result/array/concurrency core tests under UTC: 280 passed
- focused stack-guard suites: 37 passed
- `git diff --check`: clean

The mandated NuGet proxy currently mirrors `Meziantou.Analyzer` through 3.0.141 while the repository pins 3.0.156, and does not mirror `Test262Harness.Console` 1.1.2. Validation used a temporary local analyzer downgrade that was restored exactly; test262 generation could not run through the mandated feed.

## Performance

Default-job BenchmarkDotNet `RecursionBenchmark`, Apple M4 Pro / .NET 10, compared directly with #3056:

| Row | #3056 | Default guard | Delta |
| --- | ---: | ---: | ---: |
| Fib | 270.931 ms | 275.915 ms | +1.8% |
| Tak | 8.202 ms | 8.419 ms | +2.6% |
| DeepSum | 133.758 ms | 135.700 ms | +1.5% |

Allocations were identical in all rows. These recursion-heavy deltas agree with the documented roughly 1.5-3% range; no claim is made below normal run-to-run noise.

### DIFF
--- a/.github/THREAT_MODEL.md
+++ b/.github/THREAT_MODEL.md
@@ -89,7 +89,7 @@ Defaults are compatibility choices, not a hardened profile.
 | `fetch` | Disabled, and never part of `WebApiFeatures.Default` | Only `UseFetch()` grants script outbound HTTP |
 | Timeout, statement, memory, recursion limits | None | Untrusted execution is unbounded unless the host opts in |
 | Parser source-length and AST-node limits | None | Hostile source and parser output are unbounded unless the host opts in |
-| Stack overflow guard | Disabled | Native stack exhaustion can terminate the process |
+| Stack overflow guard | Enabled | Interpreted function entry checks remaining native stack and turns exhaustion into a catchable `RangeError` |
 | Maximum array size | `uint.MaxValue` | Effectively unbounded for hostile input |
 | Regex timeout | 10 seconds | Bounds an individual regular expression operation |
 | Promise wait timeout | 10 seconds | Bounds host APIs that wait for promise or module settlement |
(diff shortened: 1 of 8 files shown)


lahma (approved):
Reviewed as the incremental diff against `sebros/default-clr-array-copy-498`. This is the cleanest of the default flips: a one-property change with the documentation, threat model, and both test suites (default-on pin plus explicit opt-out pin) updated coherently, and the `ScriptFunction`/`StackGuard` comments rewritten to match. The cost story is honest — ~1.5–3% on the recursion-heavy rows, shallow calls in the noise, allocations identical — and I agree the old ">1% blocks the default" verdict deserved to be revisited for this particular trade: a couple of percent on deep recursion against silent process termination is the right default for an engine whose README leads with running untrusted script. The opt-out is retained and pinned by `AnExplicitOptOutDoesNotProbeAtAll`. Approving.

Two small notes: the quoted numbers are a default-job run on an M4 Pro rather than a `JINT_BENCH_MODE=gate` run on the gating machine — they agree with the previously measured 1.7–2.3% so I don't doubt them, but the re-roll after the #3035 fix is a good moment to attach a gate-mode confirmation. And this is breaking-default number four; it belongs in the same major-release migration story as #3051/#3054/#3056.

PR #3058, opened 2026-08-16
Disable Atomics agent suspension by default

sebastienros:
## Summary

- disable synchronous `Atomics.wait` suspension by default and preserve an explicit `AgentCanSuspend = true` compatibility opt-in
- keep `Atomics.waitAsync` available and preserve blocking Test262 worker agents while honoring `CanBlockIsFalse`
- document the breaking security default and server-thread denial guidance in the README and threat model
- retain waiter cleanup and GC coverage, including rejected waits, timeout/notify races, repeated waits, and worker harness behavior

## Security behavior

A default engine now rejects `Atomics.wait` with a sanitized JavaScript `TypeError` before registering a waiter. Worker-like hosts may explicitly opt in, accepting that an indefinite synchronous wait cannot be preempted safely in process. `Atomics.waitAsync` remains nonblocking and unaffected.

The change was transplanted from the previously reviewed Atomics-only commit series without reimplementation. It applied conflict-free over the live `sebros/default-stack-guard` tip, preserving all prior security layers.

## Validation

- Release multi-target `Jint/Jint.csproj` build: passed (0 warnings/errors)
- focused Atomics/waiter/Test262-agent tests: 12 passed
- full `Jint.Tests` under UTC: 5,800 passed, 4 skipped
- full `Jint.Tests.PublicInterface` under UTC: 1,737 passed, 9 skipped
- full `Jint.Tests.PublicInterface` with host-contract verification under UTC: 1,741 passed, 5 skipped
- `git diff --check`: passed
- security specialist review against `origin/sebros/default-stack-guard`: clean, no confirmed findings

The mandated NuGet proxy does not contain `Meziantou.Analyzer 3.0.156` (nearest `3.0.141`), so validation used a temporary analyzer-only downgrade to `3.0.141`; `Directory.Packages.props` was restored exactly before publishing. The proxy also does not contain `Test262Harness.Console 1.1.2`, so the Test262 Atomics selection could not generate; the in-repo worker-agent harness paths were validated by the focused and full core suites.

## Stack

#3057 -> #3056 -> #3054 -> #3052 -> #3051 -> #3046 -> #3045 -> #3037 -> #3036 -> #3035 -> #3030

### DIFF
--- a/.github/THREAT_MODEL.md
+++ b/.github/THREAT_MODEL.md
@@ -95,7 +95,7 @@ Defaults are compatibility choices, not a hardened profile.
 | Promise wait timeout | 10 seconds | Bounds host APIs that wait for promise or module settlement |
 | Concurrent `Engine` use | Rejected | Public host entries throw instead of racing or silently serializing |
 | Dynamic `eval` / `Function` compilation | Enabled | Script can create and parse additional source at runtime |
-| `Atomics.wait` suspension | Enabled | Script may block the engine thread, including indefinitely |
+| `Atomics.wait` suspension | Disabled | A default engine rejects synchronous waits with `TypeError`; `Atomics.waitAsync` remains available |
 | Debugger | Disabled | Debug callbacks and debugger expressions are inactive |
 | Detailed CLR resolution errors | Disabled | Reduces script-visible CLR surface disclosure |
 | Detailed caught CLR exception messages | Disabled | Replaces host exception text with a generic script-visible error |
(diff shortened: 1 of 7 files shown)


lahma (approved):
Reviewed as the incremental diff against `sebros/default-stack-guard`. Straightforward and correct: flipping `AgentCanSuspend` to `false` makes a default engine behave like a browser main thread — `Atomics.wait` throws `TypeError` before a waiter is ever registered (per the spec's AgentCanSuspend check), `Atomics.waitAsync` is untouched, and the worker-like opt-in is one property. The Test262 harness change is right (worker agents suspend, `CanBlockIsFalse` tests don't, and spawned `$262.agent` engines opt in explicitly), and the new default-rejection test is unusually careful — the notify-loop cleanup path that prevents a *regressed* implementation from stranding an indefinite background waiter in the test process is a nice piece of defensive test engineering. Approving.

One note: breaking-default number five — same major-release migration-story point as the earlier flips; the README migration note added here is exactly the right shape for the combined release notes.

PR #3059, opened 2026-08-16
Add security configuration diagnostics

sebastienros:
## Summary

- add immutable, deterministic `JINTSEC001`-`JINTSEC057` diagnostics for untrusted server execution across constraints, parsing, modules, result bounds, CLR containment and errors, array/write policy, stack protection, and Atomics suspension
- add pre-construction `Options` validation plus post-construction effective-engine validation that never replays callbacks or constraint factories
- add unspoofable internal provenance for future first-party hardened-profile composition while preserving `JINTSEC031` for user `Configure`, converter, and host-factory callbacks
- document diagnostic guidance and host responsibilities in README and the threat model

## Security review

Specialist review completed clean after fixing:
- scheme-only module allowlists being treated as destination boundaries
- CLR extension methods missing from capability diagnostics
- effective reports reading mutable options instead of construction snapshots
- detailed CLR resolution and installed extension-method state not being snapshotted

## Validation

- Jint Release multi-target build: passed (`net462`, `netstandard2.0`, `netstandard2.1`, `net8.0`, `net10.0`)
- PublicInterface default: 1,758 passed, 9 skipped
- PublicInterface host-contract verification: 1,762 passed, 5 skipped
- core under UTC: 5,801 passed, 4 skipped
- focused core security/parsing/module/result/CLR/array/stack/Atomics/concurrency/GC: 1,550 passed, 1 skipped
- focused PublicInterface coverage: 662 passed, 2 skipped
- `git diff --check`: passed

The solution-level Test262 project build is blocked under the mandated NuGet proxy because it does not currently carry `test262harness.console` 1.1.2; all other solution projects built in Release, and the temporary proxy-compatible analyzer version used for local validation was restored.

## Stack

#3058 -> #3057 -> #3056 -> #3054 -> #3052 -> #3051 -> #3046 -> #3045 -> #3037 -> #3036 -> #3035 -> #3030

### DIFF
--- a/.github/THREAT_MODEL.md
+++ b/.github/THREAT_MODEL.md
@@ -103,6 +103,56 @@ Defaults are compatibility choices, not a hardened profile.
 | Function source retention | Disabled | Submitted function source is not retained solely for `toString()` |
 | Result conversion and JSON output limits | Unlimited | Compatibility default; hostile output is unbounded unless the host opts in |
 
+## Configuration diagnostics
+
+`Options.ValidateSecurityConfiguration(SecurityConfigurationPolicy.UntrustedScripts)` inspects the
+configured `Options` without invoking deferred callbacks or constraint factories. It returns an immutable report whose diagnostics
+have stable `JINTSECnnn` codes, an error or warning severity, a description, and remediation
+guidance. Results are sorted by code so deployment logs and policy tests are deterministic.
+`EnsureSecurityConfiguration` performs the same validation and throws
+`SecurityConfigurationException` when the report contains an error; ordinary engine construction
+does not validate or change defaults automatically.
+
+The untrusted-script policy reports:
+
+- missing per-entry and cumulative operation deadlines, statement, memory, cancellation, recursion,
+  native-stack, parser, result, and array limits;
+- non-positive and saturated statement, memory, and timeout values that remove a limit;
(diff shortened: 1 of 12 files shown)


lahma (approved):
Reviewed as the incremental diff against `sebros/default-atomics-suspension-off`. Purely additive and well-constructed: the report reads a construction-time `EngineSecurityConfigurationSnapshot` rather than live mutable options (closing the fixed TOCTOU finding from the specialist review), constraint factories are inspected through the instances the engine already built so user code is never replayed, the `Requested*` mirrors let the diagnostics distinguish "asked for a limit that saturated to nothing" from "never asked" — which is exactly the trap the saturated-sentinel gotcha documents — and callback provenance means a first-party hardened profile won't trip the user-callback warning while a host `Configure` still does. The default-delegate extraction (`_defaultWrapObjectHandler` etc.) is the right way to detect "configured" without comparing lambdas. No behavior or default changes anywhere in the layer; `ValidateSecurityConfiguration` on `Advanced` correctly takes the ownership guard. Approving.

One consideration for the future rather than this PR: `SecurityDiagnosticCodes` is now a stable machine-readable contract — 57 codes whose meanings hosts will encode in CI gates. That's the point, but it means codes can never be renumbered or repurposed, only added or retired-in-place; worth a sentence in the class doc saying so explicitly so a future refactor doesn't compact the list.

PR #3060, opened 2026-08-16
Add hardened untrusted-code profile

sebastienros:
## Summary

Adds the final hardened `ForUntrustedCode` profile and exact-profile operation scope on top of the complete security stack.

- records an immutable `UntrustedCodeLimits` declaration on shared `Options`, then constructs a private effective options snapshot per Engine without mutating shared state
- applies the complete profile before construction and reapplies it after user `Configure` callbacks, while preserving first-party provenance and reporting genuine host callbacks as `JINTSEC031`
- disables CLR namespace/reflection/GetType/write/operator/non-public-member access, extension methods, converters/factories, detailed error channels, live CLR arrays, debugger behavior, string compilation, programmatic/custom modules, blocking Atomics, and unsafe stack behavior
- composes finite parser, module graph, result, regex, promise, statement, recursion, array, memory, per-entry timeout, and cumulative operation controls
- makes `BeginOperation` exact-profile and ownership-safe, rejects overlap, requires a cancellable token, and spans cumulative deadline plus memory accounting across evaluation/import and bounded output handling; statement limits retain their existing per-entry reset contract
- preserves async ownership transfer and rejects scope disposal while async work still owns the Engine
- documents the profile as defense in depth, not an OS sandbox, with a shared-options server example and residual host/worker responsibilities

## Stack

This is the final/top PR in the stack:

#3059 -> #3058 -> #3057 -> #3056 -> #3054 -> #3052 -> #3051 -> #3046 -> #3045 -> #3037 -> #3036 -> #3035 -> #3030

Base: `sebros/security-configuration-diagnostics-79b` (`5c54305b8c0fc1debb2965f4770cd58213de0384`).

## Review

A max-scope security/API/concurrency review traced the complete composed branch. Confirmed findings fixed:

- shared `Options` mutation during concurrent Engine construction
- false first-party `JINTSEC031` provenance
- CLR operator and non-public binding-flag escape paths
- detailed CLR/module error disclosure left enabled
- non-atomic operation acquisition/disposal and cumulative-memory coordination
- profile-triggered constraint factory probing during diagnostics/construction
- effective diagnostics reading mutable binding flags instead of the Engine snapshot

The final follow-up review reported no confirmed security findings.

A correctly profiled Engine now has an empty expected security report. Genuine user construction callbacks remain the sole `JINTSEC031` finding and cannot reopen profile-controlled effective settings.

## Validation

- Release multi-target `Jint` build: passed (`net462`, `netstandard2.0`, `netstandard2.1`, `net8.0`, `net10.0`)
- `Jint.Tests.PublicInterface`: 1,785 passed, 9 expected verification-only skips
- `JINT_HOST_CONTRACT_VERIFICATION=1` PublicInterface: 1,789 passed, 5 expected non-verifying skips
- full `Jint.Tests` under UTC: 5,801 passed, 4 pre-existing skips
- combined 13-layer security selection: 380 passed
- concurrency/scope/async stress: 10 runs x 44 tests, all passed
- focused GC retention: 2 passed
- benchmark project build: passed
- `git diff --check`: clean

The mandated proxy does not currently contain `Meziantou.Analyzer` 3.0.156, so validation temporarily used proxy-available 3.0.141 and restored the repository pin before commit. `Test262Harness.Console` 1.1.2 is also unavailable from `https://packagefeedproxy.microsoft.io/nuget/v3/index.json`, so Test262 generation/conformance could not be run; no workaround is committed.

## Benchmarks

Default BenchmarkDotNet jobs (not `--job short`):

- fresh Engine: 518.0 ns, 8.5 KB
- hardened-profile Engine: 1,091.6 ns, 12.23 KB
- guarded execution control: 10.88 ms unconstrained; 11.12 ms statement-only; 11.90 ms timeout-only; 12.12 ms timeout+statement; exact memory accounting rows 28.96-30.41 ms; all rows allocated 960 B

The private effective snapshot adds about 0.57 microseconds and 3.73 KB per hardened Engine, with no shared `Options` mutation. Guard overhead remains consistent with the existing constraint cost model.

### DIFF
--- a/.github/THREAT_MODEL.md
+++ b/.github/THREAT_MODEL.md
@@ -74,6 +74,10 @@ All components above run with the hosting worker's process authority.
 
 Defaults are compatibility choices, not a hardened profile.
 
+`Options.ForUntrustedCode(UntrustedCodeLimits)` is the opt-in hardened profile. It leaves
+the defaults unchanged for compatibility, but closes the static capabilities below and
+requires the host to supply finite request-appropriate resource limits.
+
 | Control | Default | Security effect |
 | --- | --- | --- |
 | CLR namespace access (`Interop.Enabled`) | Disabled | `System`, `importNamespace`, and `clrHelper` are not installed |
@@ -176,6 +180,7 @@ interop, or arbitrary code execution with the worker's identity.
 - `AllowGetType` does not expose the static `System.Type.GetType(string)` family. Its
   `clrHelper` type-widening operations also enforce the namespace type policy.
 - `AllowSystemReflection` gates namespace discovery and wrapping for `System.Reflection`.
+- `ForUntrustedCode` resets all three controls and removes configured assemblies.
 - `ExposeDetailedResolutionErrors` is disabled by default.
 
 **Missing or residual mitigation.**
@@ -214,6 +219,8 @@ exfiltration, or resource-amplification paths.
 - The host chooses every projected value.
 - `TypeResolver.MemberFilter`, wrapper handlers, proxies, and narrow delegates can reduce
   the surface.
+- `ForUntrustedCode` removes registered CLR extension methods as well as disabling CLR
+  namespace access.
 
 **Missing or residual mitigation.**
 
@@ -242,6 +249,7 @@ state regardless of the direct-write option.
 - `ArrayConversionMode.Copy` is the default and disconnects the JavaScript array container
   from later CLR mutations and script writes.
 - Immutable DTOs and Jint-owned record layouts avoid live write-through.
+- `ForUntrustedCode` selects both read-only wrappers and copied CLR arrays.
 
 **Missing or residual mitigation.**
 
@@ -275,6 +283,7 @@ BigInt/string arithmetic, and adversarial algorithms can monopolize a worker.
 - Amortized timeout and cancellation checks are also made after CLR calls return.
 - The untrusted-script configuration report identifies missing limits and sentinel values that
   remove them.
+- `ForUntrustedCode` requires finite time and statement budgets.
 
 **Missing or residual mitigation.**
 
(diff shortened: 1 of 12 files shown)


lahma (approved):
Reviewed as the incremental diff against `sebros/security-configuration-diagnostics-79b`. This is a strong capstone. The parts I traced:

- **Shared-`Options` safety.** `CreateEngineOptions()` deep-clones every option group only when a profile is declared, hardens the clone, and `Options.Apply` re-hardens *the clone* after user `Configure` callbacks — so the documented "share one `Options` across engines" contract survives, concurrent construction included, and a user callback can't reopen a profile-controlled setting while still being visible to diagnostics as `JINTSEC031`. The ordering also holds up: the constraint partition exists before `Apply`, so `ReapplyUntrustedCodeOptions`'s `Find<…>()!` calls are safe by construction (a one-line comment saying the partition-before-Apply ordering is what those null-forgiving operators lean on would future-proof them).
- **`UntrustedCodeLimits` having no defaults for the per-request knobs** — and rejecting exactly the values the ordinary helpers treat as "unlimited" — is the right inversion of the saturated-sentinel trap. Requiring `BeginOperation` with a cancellable token for the cross-entry budget matches how `OperationDeadlineConstraint` was designed to be used.
- The static-surface shutdown list is comprehensive and each item maps to a layer below (writes, live arrays, namespace resolution, detailed errors, string compilation, Atomics, debugger, stack guard), and "a correctly profiled engine produces an empty security report" is a genuinely good invariant to pin — it keeps the profile and the diagnostics from drifting apart.

The 614-line profile test suite plus the 10×44 concurrency/scope stress runs and the construction benchmark (+0.57 µs / +3.7 KB per hardened engine, shared options untouched) close it out. Approving.

One design note, not blocking: `ApplyUntrustedCodeOptions` clears the host's own constraint factories and converters, so the profile cannot be composed with an additional host-defined constraint. I think that's the right default (the profile owns the constraint story, and an extra hook is an extra attack surface), but it's worth stating explicitly in the `ForUntrustedCode` docs that "profile + my own custom constraint" is intentionally unsupported, so it arrives as a documented decision rather than a surprise.

Whole-stack reminder: everything above #3035 is blocked on its admission-token fix, and the accumulated default flips (#3051, #3054, #3056, #3057, #3058) should ship together as one clearly-versioned migration story.

PR #3062, opened 2026-08-18
Recognise the global environment by its type, not by a null outer link

svenrog:
## Summary

Identifier resolution mistakes a released, pooled environment for the global one and throws on the cast.

## Details

Both identifier lookups in `JintEnvironment` short-circuit on "this environment has no outer environment, therefore it is the global one":

```csharp
if (env._outerEnv is null)
{
    return ((GlobalEnvironment) env).TryGetBinding(name, strict, out value);
}
```

That premise is not exclusive. Jint pools environments and **clears the outer link on release** — `JintBlockStatement.cs:160` and `:286`, `JintForInForOfStatement.cs:949`, `JintTryStatement.cs:236`, `ScriptFunction.cs:341`, `EvalFunction.cs:492` all do `env._outerEnv = null` before handing the record back. A released `DeclarativeEnvironment` that is still reachable — through a closure, a continuation, or a generator resuming after its block has been left — is indistinguishable from the global environment by that link alone, takes the branch, and the cast throws:

```
System.InvalidCastException: Unable to cast object of type
'Jint.Runtime.Environments.DeclarativeEnvironment' to type
'Jint.Runtime.Environments.GlobalEnvironment'
   at Jint.Runtime.Environments.JintEnvironment.TryGetIdentifierEnvironmentWithBindingValue(...)
```

The fix is to ask the type instead: `env is GlobalEnvironment`. The global environment still short-circuits exactly as before; anything else falls into the walk below, which ends at the cleared link and reports a miss — a `ReferenceError` for the script rather than a CLR exception through the embedder.

### Reproducible example

The mishandling reproduces directly, and `ReleasedEnvironmentLookupTests` is that repro:

```csharp
var engine = new Engine();

var env = new DeclarativeEnvironment(engine);
env.CreateMutableBindingAndInitialize("x", canBeDeleted: false, 42, DisposeHint.Normal);
env._outerEnv = null;                       // what every release site above does

JintEnvironment.TryGetIdentifierEnvironmentWithBindingValue(
    env, new Environment.BindingName("x"), strict: false, out var record, out var value);
// before: System.InvalidCastException: Unable to cast object of type 'DeclarativeEnvironment'
//         to type 'GlobalEnvironment'
// after:  true, record == env, value == 42
```

Both new tests fail on `main` with that exact exception and pass with the change.

### What this PR does *not* claim to fix

I have **no repro for the escape itself** — for what leaves a released environment reachable in the first place. That is a separate question about the pooling's escape analysis, and I am not proposing an answer to it here. What I can show is that when it does happen, the failure mode is a CLR cast exception escaping into embedder code, and that it need not be. I have seen it three times in production, in a host that runs untrusted page scripts, and never once from a script I could reduce.

That is also why the second lookup changed. `TryGetIdentifierEnvironmentWithBinding` takes the same branch but has no cast, and while a released record's outer link stays null it reaches the same answer either way — so this is not a behaviour fix there. It moves to the type test because the premise is what is wrong, and leaving one of the pair asserting it invites the next cast.

## Linked issue

No existing issue — the symptom is a production stack trace with no script-level reduction, so there was nothing useful to file separately. Happy to open one if you would prefer the PR linked.

## Test plan

- [x] Added or updated unit tests in `Jint.Tests` — `ReleasedEnvironmentLookupTests`: a released environment is not mistaken for the global one, a released environment missing the binding reports a miss rather than throwing, and the global environment is still answered directly. The first two fail on `main` with `InvalidCastException`.
- [x] Ran `dotnet test --configuration Release` locally — `Jint.Tests` 5761/5761 on net10.0 (excluding one pre-existing failure, below), 5677/5677 on net472.
- [x] For ECMAScript spec changes: ran `Jint.Tests.Test262` — 102,326 passed, 342 skipped, **0 failed**, no regressions.
- [ ] For interop changes — n/a.
- [ ] For perf changes — n/a. A type check replaces a null check on the same branch; the global environment still takes it.

Pre-existing failure, unrelated and identical on a clean `main` checkout on this machine: `DateTests.ToLocaleStringOutsideDateTimeRangeDoesNotThrowAClrException` — local ICU renders the year as `−271821` with U+2212 where the assertion expects an ASCII hyphen.

## Breaking change?

No. No public-API change, and no behaviour change for the global environment or for any environment that still has its outer link.

### DIFF
--- a/Jint.Tests/Runtime/ReleasedEnvironmentLookupTests.cs
+++ b/Jint.Tests/Runtime/ReleasedEnvironmentLookupTests.cs
@@ -1,0 +1,78 @@
+using Jint.Native;
+using Jint.Runtime;
+using Jint.Runtime.Environments;
+using Environment = Jint.Runtime.Environments.Environment;
+
+namespace Jint.Tests.Runtime;
+
+/// <summary>
+/// Identifier resolution must recognise the global environment by its type, not by having no outer
+/// environment. Environments are pooled and their outer link is cleared on release, so "no outer" is also
+/// true of any released environment that is still reachable — through a closure, a continuation, or a
+/// generator that resumes after the block it was declared in has been left.
+/// </summary>
+public class ReleasedEnvironmentLookupTests
+{
+    private static DeclarativeEnvironment ReleasedEnvironment(Engine engine, string name, JsValue value)
+    {
+        var env = new DeclarativeEnvironment(engine);
+        env.CreateMutableBindingAndInitialize(name, canBeDeleted: false, value, DisposeHint.Normal);
+
+        // What the pool does on release: JintBlockStatement, JintForInForOfStatement, JintTryStatement,
+        // ScriptFunction and EvalFunction all null the outer link before handing the record back.
+        env._outerEnv = null;
+        return env;
+    }
+
+    [Fact]
+    public void AReleasedEnvironmentIsNotMistakenForTheGlobalOne()
+    {
+        var engine = new Engine();
+        var env = ReleasedEnvironment(engine, "x", 42);
+        var name = new Environment.BindingName("x");
+
+        // The value-returning lookup used to take the "this is the global environment" branch here, cast to
+        // GlobalEnvironment and throw. Its sibling takes the same branch but has no cast, so it answers the
+        // same thing either way while a released environment's outer link stays null — the type test there
+        // is the premise being corrected, not a behaviour change.
+        JintEnvironment.TryGetIdentifierEnvironmentWithBindingValue(env, name, strict: false, out var record, out var value)
+            .Should().BeTrue();
+        record.Should().BeSameAs(env);
+        value.Should().Be((JsValue) 42);
+
+        JintEnvironment.TryGetIdentifierEnvironmentWithBinding(env, name, out var found).Should().BeTrue();
+        found.Should().BeSameAs(env);
+    }
+
+    [Fact]
+    public void AReleasedEnvironmentWithoutTheBindingReportsMissRatherThanThrowing()
+    {
+        var engine = new Engine();
+        var env = ReleasedEnvironment(engine, "x", 42);
+        var name = new Environment.BindingName("somethingElse");
+
+        // The walk ends at the cleared link. A miss is the honest answer — the outer chain this record
+        // used to sit in is gone — and it costs the caller a ReferenceError rather than a CLR exception.
+        JintEnvironment.TryGetIdentifierEnvironmentWithBindingValue(env, name, strict: false, out _, out _)
+            .Should().BeFalse();
+        JintEnvironment.TryGetIdentifierEnvironmentWithBinding(env, name, out _).Should().BeFalse();
+    }
+
+    [Fact]
+    public void TheGlobalEnvironmentIsStillAnsweredDirectly()
+    {
+        var engine = new Engine();
+        engine.Execute("var g = 7;");
+
+        var global = engine.Realm.GlobalEnv;
+        var name = new Environment.BindingName("g");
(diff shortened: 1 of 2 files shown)


svenrog:
Closing this: it does not fix a reachable defect.

I found the actual cause and sent it as #3063 — the Function-constructor environment park detached the outer link of an environment its own escape analysis lets a closure escape, so the escaped closure started its lookup at a record with no outer link. With #3063 applied to unmodified main, every shape I could reach answers what V8 answers.

Isolating this change on its own shows what it is worth: it turns that InvalidCastException into `ReferenceError: answer is not defined`, where the right answer is 42. So it replaces a crash with a wrong answer, and once #3063 is in it changes no answer at all. That is not worth a change to identifier resolution's hot path, and I would rather not carry a guard that makes the next such escape silent instead of loud.

Sorry for the noise.

lahma (commented):
The change itself checks out, and I could verify it end to end: `GlobalEnvironment` is sealed with a single construction site (`Host.cs` → `NewGlobalEnvironment`, always a null outer link, ShadowRealms included), so the type test never misclassifies a real global, and `env is GlobalEnvironment` is already the discrimination idiom elsewhere in the engine. Locally, both regression tests fail on `main` with the quoted `InvalidCastException` and pass with the fix; full `Jint.Tests` is green on net10.0 and net472, and test262 passes 102,326/0.

A few points, the first one substantive:

1. **The cast being removed has been this engine's loudest escape detector.** The doc comments in `ForLoopIterationEnvironmentAwaitTests` record that the two previously fixed chain-corruption bugs (iteration-environment replay, `with` replay) were found precisely because this `InvalidCastException` fired — and that the third symptom was silent. With the cast gone, the next escape of this class shows up as a stale read, a spurious miss the script can catch, or a poisoned identifier cache (`JintIdentifierExpression` caches environment references on the stated premise that pooled environments can never be cached) — none of which point anywhere near the pool. Since the escape has been seen in production but can't yet be reproduced, keeping detection seems worth more here, not less. Suggestion: keep the graceful Release behaviour exactly as in this PR, but add a check gated on `HostContractVerification.Enabled` (which covers Debug builds and the verification CI leg) that a null-outer environment reached during resolution is the `GlobalEnvironment`. The first two tests would then need to select on the verification switch like the existing verification-aware tests do.
2. Related: the first test pins the stale read as contract (`record == env`, `value == 42`). If a tripwire like the above ever lands, that assertion fires it in every verifying run. Asserting only "no CLR exception, and both lookups agree" — as the second test does — protects embedders equally without freezing the stale read into the suite.
3. In `TryGetIdentifierEnvironmentWithBinding`, capture the pattern variable and dispatch directly, as the value sibling now does: `if (env is GlobalEnvironment globalEnv) return globalEnv.HasBinding(name);` — devirtualizes on the sealed type and keeps the two deliberately mirrored methods in the same shape.
4. Optional: `if (env._outerEnv is null && env is GlobalEnvironment globalEnv)` is behaviourally identical (global: both true; released: falls to the walk) and keeps the old field-load-and-null-test codegen on the common non-global path, paying the type test only when the outer link is null.
5. Comment nit: "but the sibling casts on it, and throws" is present tense about code this same diff deletes.

Out of scope for this PR, but probably worth a follow-up issue: both `GetThisEnvironment` copies (`ExecutionContext.cs`, `JintCallStack.cs`) still `while (true)` on exactly the chain shape this PR tolerates — their comment ("the list of environments always ends with the global environment") is the same premise being corrected here. On a corrupt chain a script can swallow the `ReferenceError` and later wedge the host thread on a `this` read, uninterruptible by `TimeoutInterval`. A null-link bail-out there would complete the story this PR starts.

PR #3063, opened 2026-08-18
Keep the outer link on a parked Function-constructor environment

svenrog:
## Summary

A parked Function-constructor environment kept its outer link, so a closure it returned still resolves globals.

## Details

`ScriptFunction.ReturnEnvironment` parks a Function-constructor instance's call environment on its definition (`State._dynamicCachedEnv`) so the next one-shot instance rents it, and it detached the outer link first.

Every other park that detaches one pools an environment no closure can reach at all. This gate is weaker: `EnvironmentMayEscape` asks whether a closure reads one of the environment's *own* bindings, so a closure that reads none of them is allowed to escape the call. Such a closure still resolves its free identifiers **through** the record, and detaching left it starting its lookup at a record with no outer link — which `JintEnvironment` reads as "this is the global environment":

```js
var answer = 42;
var make = new Function('a', 'var b = a; return function () { return answer; };');
make(1)();
```

on `main`:

```
System.InvalidCastException: Unable to cast object of type 'Jint.Runtime.Environments.FunctionEnvironment'
to type 'Jint.Runtime.Environments.GlobalEnvironment'
```

The `var b = a;` matters: it is what gives the call fixed slots, so the slot-aware escape analysis runs and lets the closure — which names neither `a` nor `b` — escape a poolable environment. A closure that *does* read one of the slots (`return b;`) keeps the environment out of the pool, and that shape has always worked.

The fix is to leave the link attached. It roots nothing: `CreateDynamicFunction` closes a dynamic function over the realm's global environment, which the realm holds for its own lifetime anyway — and on rent, `Reset(f, newTarget, f._environment)` writes the same environment back. Reuse itself is unchanged; the park is one field write shorter.

Found with a temporary instrumented build that reports any park detaching the outer link of an environment a closure can still reach. Under it, test262, the repository's suites and a 75-site live-page render corpus flagged no other site — the block, catch, loop and eval parks are all guarded by their conservative analyses.

## Linked issue

None — found by instrumentation rather than reported. It reaches ordinary page code through anything that builds a function with `new Function` and hands back a closure (bundler loaders, templating).

## Test plan

- [x] Added or updated unit tests in `Jint.Tests` — `Jint.Tests/Runtime/DynamicFunctionEnvironmentReuseTests.cs`: the escaping closure before and after the parked environment is rented, an object the call returned, and the other half of the rule. Three of the four fail on `main` with the `InvalidCastException` above.
- [x] Ran `dotnet test --configuration Release` locally — `Jint.Tests` 5762 passed / net472 5678 passed, `Jint.Tests.CommonScripts` 28, `Jint.Tests.PublicInterface` 1488 (+1487 net472); also with `JINT_HOST_CONTRACT_VERIFICATION=1`. The one failure, `DateTests.ToLocaleStringOutsideDateTimeRangeDoesNotThrowAClrException`, fails the same way on unmodified `main` on this machine (ICU data), so it is not from this change.
- [x] For ECMAScript spec changes: ran `Jint.Tests.Test262` and confirmed no regressions — 102,326 passed, 0 failed.
- [ ] For interop changes: covered in `Jint.Tests/Runtime/Interop` — n/a.
- [ ] For perf changes: included before/after numbers from `Jint.Benchmark` — n/a. Not a perf change: the pooling behaviour is identical, with one field write removed from the park.

## Breaking change?

No. Behaviour only moves toward the spec — code that threw `InvalidCastException` now evaluates.

### DIFF
--- a/Jint.Tests/Runtime/DynamicFunctionEnvironmentReuseTests.cs
+++ b/Jint.Tests/Runtime/DynamicFunctionEnvironmentReuseTests.cs
@@ -1,0 +1,68 @@
+#nullable enable
+
+namespace Jint.Tests.Runtime;
+
+/// <summary>
+/// A Function-constructor instance's call environment is parked on its definition when the call returns,
+/// so the next instance built from the same source rents it instead of allocating. Escape analysis lets a
+/// closure the call handed out survive that park — <c>EnvironmentMayEscape</c> only asks whether a closure
+/// reads one of the environment's own bindings — and such a closure still resolves everything else
+/// <em>through</em> that record, so the park must leave its outer link attached.
+/// </summary>
+public class DynamicFunctionEnvironmentReuseTests
+{
+    private const string Make =
+        "var make = new Function('a', 'var b = a; return function () { return answer; };');";
+
+    [Fact]
+    public void AClosureReturnedByADynamicFunctionStillReachesTheGlobalScope()
+    {
+        var engine = new Engine();
+        engine.Execute("var answer = 42;");
+        engine.Execute(Make);
+
+        // The closure reads no binding of the call it came from, which is what makes that call's
+        // environment poolable; its scope chain runs through the parked record all the same.
+        engine.Evaluate("make(1)()").Should().Be(42);
+    }
+
+    [Fact]
+    public void TheClosureKeepsAnsweringAfterTheParkedEnvironmentIsRented()
+    {
+        var engine = new Engine();
+        engine.Execute("var answer = 42;");
+        engine.Execute(Make);
+
+        engine.Execute("var first = make(1);");
+
+        // The second call rents what the first parked, and re-binds it to that call. The first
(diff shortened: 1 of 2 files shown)


lahma (approved):
Verified end to end: the three new tests fail on main with exactly the `InvalidCastException` described (FunctionEnvironment → GlobalEnvironment) and pass with the fix; the full unit suite and test262 (102,326 tests) are green on both net10.0 and net472.

The fix is right by construction on both sides of the park: the rent path always re-links via `Reset(f, newTarget, f._environment)`, so the kept link is never observed by the next instance, and the kept link roots nothing new — a dynamic function's outer is the realm's own global environment, and the State holding the parked env lives in the realm-owned dynamic-function cache. This also brings the dynamic park in line with the instance-level `_envReuse` park, which has always kept its outer link; the pools that do clear theirs hold only environments no closure can reach, exactly as the new comment says. Nice control test for the slot-reading half of the escape rule, too.

PR #3095, opened 2026-08-20
Settle Atomics.waitAsync timeouts from the event-loop pump

lahma:
Fixes the CI flake that has been forcing re-runs across recent PRs: the `built-ins/Atomics/waitAsync/*` test262 family timing out sporadically on loaded runners (a different family member each time, incl. three occurrences across two CI runs of #3090 alone).

**Root cause.** Those tests give themselves ~1000 ms of wall clock while polling with `$262.agent.setTimeout(wait, 0)` — which keeps the engine thread hot — but a finite `Atomics.waitAsync` timeout settled through `Task.Run(… await Task.Delay(…) …)`: **two thread-pool dispatches**. On a saturated 2-core runner the pool injects roughly one thread per 500 ms once its baseline is busy, so a 1 ms timeout could miss a 1000 ms budget while the engine itself was behaving perfectly.

**The fix.** Finite `waitAsync` deadlines now live in an engine-owned min-heap (`AtomicsWaiterDeadlines`, all TFMs, lock-free because it is engine-thread-only) and are settled **from the event-loop pump**: the thread pool is no longer in the settlement path at all. The idle waits (`DrainEventLoopUntil`, `AwaitPromiseSettlementAsync`) bound themselves by the next deadline via `TimeUntilNextPumpScheduledWork`, and the `Task.Run`/`Task.Delay`/`CancellationTokenSource` machinery is deleted. The `Atomics.notify` race is unchanged — the settle-once CAS in `AsyncWaiter.Resolve` still decides, and the heap discards already-settled entries lazily exactly as a cleared timer is discarded.

One design point deserves its callout: the deadline check runs **once per pump iteration, not only at queue exhaustion**. Exhaustion-only would be wrong — the `$262.agent.setTimeout` polyfill is a promise chain, so a test polling for its own wait keeps the job queue permanently non-empty, and a deadline consulted only on exhaustion would never be consulted at all (mutating the check to exhaustion-only makes two of the new tests fail deterministically). The cost is held to **one predictable null test per job**: the registry field is null on every engine until the first finite-timeout wait and is nulled again the moment the last one leaves. On net462/netstandard the timer half of the pump hook folds away to nothing. No benchmarks were run for this PR per the maintainer's earlier call (the broad suites are covered by the pre-release regression pass); the field-load-plus-branch footprint is stated here explicitly so that pass knows what to look at.

Also in this PR: `TryPromoteDueTimerJob` moved into a new `Engine.Pump.cs` with the `#if` inside the body, so `EventLoop.RunAvailableContinuations` and both drain paths now carry no conditional compilation at all.

**Verification**: all suites green on net10.0 + net472 in both host-contract-verification configurations; the waitAsync-filtered test262 subset run three times serially (202/202 each); full test262 green (102,495/0). Ten new pump tests including the two mutation-backed pins above, restore-fence behavior, and blocking/async/manual-pump completion of short timeouts.

**Honest limitation**: the flaking tests check elapsed time *before* checking outcomes, so a full engine-thread stall longer than a test's budget still fails it — no engine change can prevent that. What this removes is the pool-starvation inversion, which is the mechanism actually observed in every recorded occurrence.


lahma:
CI on the first push demonstrated the residual mechanism this PR's engine half cannot reach: the harness's own `setTimeout` shim was still `Task.Run` + `Task.Delay` per poll iteration, so under a starved pool the *polling* missed the tests' wall-clock budget even with the wait itself settled. Second commit makes the shim pump-driven too (zero-delay → direct enqueue; positive delays promoted by the async wait loop on the test's own thread, thread-static pending list). Locally: Atomics-filtered subset 782/782, full test262 green modulo one unrelated staging test that took 33s under a deliberately loaded machine and passes in 1s isolated.

PR #3153, opened 2026-08-21
Engine memory report for pooled-engine retention hunting

### LINKED ISSUE #3107: Embedder diagnostics: an engine memory report for pooled-engine retention hunting
Campaign issue — engine diagnostics for production embedders, in the #3089 family. Hosts that pool engines have no way to answer "what is this engine retaining and why is it growing" short of a full CLR memory profiler — and the retention gotchas the integration docs describe (warmed call sites pinning receivers and closures, prototype-cache holders) are invisible until they hurt.

- [ ] **`Engine.Advanced.GetMemoryReport()`**: an on-demand walk (nothing recorded continuously) over engine-owned state: global bindings count, materialized intrinsics, event-loop queue depth, pending timers/waiters, handler-tree cache sizes (`_functionDefinitions`, `_scriptStatementLists`, evaluated scripts), interned keys, pooled buffers, per-category object counts reachable from the global surface under an explicit traversal bound (the report says *bounded at N*, never pretending completeness).
- [ ] The documented retention hot-list made visible: warmed call-site retained callees/receivers, module registry size, shape/layout counts.
- [ ] Shape: readonly records, forward-extensible, documented as diagnostic — numbers may change per release, the `ObjectRepresentation` non-contract precedent stated the same way.
- [ ] Tests: report reflects known constructions; the documented call-site retention scenario is visible in it; the traversal cap is honored; zero cost when never called.

Explicitly not: a heap dump, an object-graph export, or GC integration. It answers the operational question, not the forensic one.

lahma:
`Engine.Advanced.GetMemoryReport()` — the answer to "what is this engine retaining and why is it growing" for hosts that pool engines, without a CLR memory profiler. On-demand only: nothing is recorded continuously, and never calling it costs nothing.

The report is **counts, never bytes** (Jint tracks no allocation sizes and refuses to guess), and **non-perturbing by construction**: the walks read descriptor storage directly and gate on the `NonData`/`CustomJsValue` flags rather than calling `GetOwnProperty` or reading `Value`, because several descriptor kinds materialize on exactly those reads — a lazy global's factory must not run because a diagnostic looked at it (mutation-pinned: the naive key-driven walk fails 23 of 24 tests, including accessor/Proxy-trap/host-getter never-invoked pins). Shape-mode slots are read raw, so an unmaterialized lazy slot surfaces as unmaterialized rather than being forced.

What it reports: global property counts split into total/materialized/lexical (the gap between the first two is "how much of the language this engine has paid for"), event-loop depth, pending timers and atomics waiters (filtered of already-settled entries), module registry and pending loads, the three handler-tree cache sizes that are the documented retention roots for warmed call sites, the two per-engine interop caches that actually exist, pool counts with one exact slot total, and a bounded object census over the principal realm that stops mid-array the moment the bound is hit (a census that allocated an 80 MB list inside a memory diagnostic was found and fixed during development). Deliberately absent, and documented as such rather than omitted: per-call-site retention enumeration (handler nodes expose no child walk), byte figures, and anything on `ShadowRealm`s. Like `ObjectRepresentation`, the shape is documented as diagnostic rather than contract, recorded in AGENTS.md beside that precedent.

34 tests, all TFMs (`net462` through `net10.0`), both suites green with and without `JINT_HOST_CONTRACT_VERIFICATION=1`.

Closes #3107

https://claude.ai/code/session_011G7Ud48VAs1JicxRZxLDxg


lahma:
Supersedes #3151, which hit a merge conflict with #3150 at merge time (both grew Engine.cs) and whose head branch was then deleted by an over-eager cleanup. Rebased over #3150 — whose per-engine property-key handler cache is exactly the kind of retention root this report exists to surface, so it now appears as `HandlerTreeCaches.PropertyKeyExpressions`, with a pin.

PR #3183, opened 2026-08-21
Web Crypto: check the ECDH pair before the length ceiling, as every browser does

### LINKED ISSUE #3180: Web Crypto: ECDH deriveBits mismatched-curve error - spec prose (OperationError) vs browsers (InvalidAccessError)
Found by the WebCryptoAPI WPT corpus (#3178): the `P-384 mismatched curves` and `P-521 mismatched curves` rows of `derive_bits_keys/ecdh_bits` expect `InvalidAccessError` and get `OperationError` (the P-256 row passes only because its other curve is wider).

`EcAlgorithm.DeriveBits` follows the specification prose exactly: *maximumLength* is derived from the **public** key's domain parameters (step 4) and its `OperationError` for an over-long `length` is raised **before** the step that compares the two keys' curves (step 8) - so a P-521 base key handed a P-256 public key with `length: 528` is refused for the length rather than for the mismatch. Browsers answer the later step's `InvalidAccessError`, which is what the corpus asserts.

This is a prose-vs-browsers disagreement rather than a clear Jint bug. Options: (a) keep the prose and leave the two `NeedsTriage` exclusion rows as documentation, (b) match browsers by hoisting the curve comparison above the length check and delete the rows, (c) raise it at https://github.com/w3c/webcrypto and follow the resolution. My lean is (c) then (b) - browser consensus usually wins these.

lahma:
Resolves the ECDH `deriveBits` prose-vs-browsers disagreement by matching the browsers: the key-agreement checks (base key is private, algorithm names match, named curves match — steps 6/7/8) now run **before** the `maximumLength`/`length` ceiling (steps 4/5), so a mismatched pair earns `InvalidAccessError` whatever the length asked for. Every check body is byte-for-byte unchanged; only the ceiling moved, and once step 8 has proved the curves are one, measuring it off the public key's parameters is the same measurement the prose describes.

Closes #3180.

The evidence, verified rather than recalled: the two WPT rows ask for eight times the *base* key's field width against a P-256 public key (384 and 528 bits), which no implementation obeying step 5 where it is written can satisfy — and `ecdh_bits.https.any.html` is 40/40 in Chrome 152, Firefox 154 and Safari 26.6 on wpt.fyi, with Chromium/Gecko/WebKit/Node sources all validating the pair before any length reasoning. Nothing in w3c/webcrypto's issues covers this (searched); a ready-to-file upstream issue draft — including the analysis of why the P-256 row passes either way and why reordering beats re-deriving the ceiling from the base key — is attached below in a comment for the maintainer to file. If the spec reorders its steps, what goes is Jint's divergence comment, not the code.

One deliberate widening over the issue's ask, argued in the code: step 6 (base-key-is-private) hoists too, keeping 6/7/8 in their prose order among themselves — lifting only 7/8 would have left a public base key on a matched curve with an over-long length answering `OperationError` where Chromium answers `InvalidAccessError`, a *new* divergence no WPT row covers.

The two `NeedsTriage` exclusion rows are deleted (the driver enforces it — they would have gone stale the moment the code moved); `ecdh_keys` was audited and needed no change (its rows never reached the ceiling). The reordering is entirely Jint-level ordinal string comparisons decided before any BCL primitive is touched, so for once no platform can pick a different step — no `Platform:` scoping needed.

New pin `AMismatchedCurveIsAnInvalidAccessErrorWhateverTheLengthAsksFor` covers the two discriminating shapes plus a matched-curve over-long-length row that must stay `OperationError`; mutation-verified by reinstating the prose order (exactly that one test fails, on exactly the two discriminating rows). Full matrix green on both TFMs including host-contract-verification legs; all WPT suites green.

https://claude.ai/code/session_014n5uCpi2vvHWBSiewUwBiY

### DIFF
--- a/Jint.Tests/Runtime/WebApi/SubtleCryptoDeriveTests.cs
+++ b/Jint.Tests/Runtime/WebApi/SubtleCryptoDeriveTests.cs
@@ -459,7 +459,9 @@ public class SubtleCryptoDeriveTests
         // "Let maximumLength be the length in bits of the output of the field element to octet string
         // conversion … If length is not null and is greater than maximumLength, then throw an
         // OperationError." The conversion pads to whole octets, so P-521's maximum is 528 and not 521 — the
-        // one row of this table that cannot be read off the curve's name.
+        // one row of this table that cannot be read off the curve's name. Both keys are the same pair here,
+        // so the ceiling is the only thing that can refuse the request; where the two curves differ it is not
+        // reached at all, which is AMismatchedCurveIsAnInvalidAccessErrorWhateverTheLengthAsksFor's subject.
         Run($$"""
             const pair = await crypto.subtle.generateKey({ name: 'ECDH', namedCurve: '{{curve}}' }, false, ['deriveBits']);
             const params = { name: 'ECDH', public: pair.publicKey };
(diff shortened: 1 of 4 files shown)


lahma:
Ready-to-file draft for w3c/webcrypto, for whoever wants to raise it upstream:

**Title**

    ECDH deriveBits: step 5's OperationError precedes the curve-mismatch check, but no browser does that

**Body**

> ### The prose
>
> [ECDH "Derive Bits"](https://w3c.github.io/webcrypto/#ecdh-operations-derive-bits) orders
> its checks like this:
>
> 1. Let *publicKey* be the `public` member of *normalizedAlgorithm*.
> 2. If the `[[type]]` of *publicKey* is not `"public"` → **InvalidAccessError**.
> 3. If the `name` of *publicKey*'s `[[algorithm]]` ≠ *normalizedAlgorithm*'s `name` → **InvalidAccessError**.
> 4. Let *maximumLength* be the length in bits of the output of the field element to octet
>    string conversion … **for the EC domain parameters associated with *publicKey***.
> 5. If *length* is not null and is greater than *maximumLength* → **OperationError**.
> 6. If the `[[type]]` of *key* is not `"private"` → **InvalidAccessError**.
> 7. If the `name` of *publicKey*'s `[[algorithm]]` ≠ the `name` of *key*'s `[[algorithm]]` → **InvalidAccessError**.
> 8. If the `namedCurve` of *publicKey*'s `[[algorithm]]` ≠ the `namedCurve` of *key*'s `[[algorithm]]` → **InvalidAccessError**.
> 9.–11. the ECDH primitive, and the truncation of *secret*.
>
> Steps 4 and 5 measure the ceiling off the **public** key alone and raise before steps 7 and
> 8 have established that the two keys are a pair at all. So for a mismatched pair where the
> public key is on the *narrower* curve, the request is refused for its length rather than
> for the mismatch — a P-521 base key, a P-256 public key and `length: 528` is an
> `OperationError` under a literal reading, not an `InvalidAccessError`.
>
> ### The browsers
>
> All of them answer `InvalidAccessError`. Chromium
> (`components/webcrypto/algorithms/ecdh.cc`), Gecko (`dom/crypto/WebCryptoTask.cpp`), WebKit
> (`CryptoAlgorithmECDH::deriveBits`) and Node (`lib/internal/crypto/diffiehellman.js`) all
> validate the key roles and the curve equality first and only look at *length* once *secret*
> exists — several of them have no *maximumLength* step at all and rely on step 11's
> "if the length in bits of *secret* is less than *length*".
>
> ### The WPT rows that pin it
>
> `WebCryptoAPI/derive_bits_keys/ecdh_bits.https.any.js` builds its mismatched-curve cases in
> `derive_bits_keys/ecdh.js`, pairing each curve with `P-256` (or, for `P-256` itself, with
> `P-384`), and `derive.js`'s `derive()` defaults *length* to `8 * size` where *size* is the
> **base** key's field width. So:
>
> | row | base key | public key | length | step 5 says | asserted |
> | --- | --- | --- | --- | --- | --- |
> | `P-384 mismatched curves` | P-384 | P-256 | 384 | OperationError (384 > 256) | **InvalidAccessError** |
> | `P-521 mismatched curves` | P-521 | P-256 | 528 | OperationError (528 > 256) | **InvalidAccessError** |
> | `P-256 mismatched curves` | P-256 | P-384 | 256 | passes (256 ≤ 384) | InvalidAccessError |
>
> The first two are unsatisfiable by an implementation that runs step 5 where it is written.
> The third is green either way, which is why this is easy to miss. `ecdh_bits.https.any.html`
> is **40/40 in Chrome 152, Firefox 154 and Safari 26.6** on wpt.fyi (master/stable,
> 2026-08-21), so the tests are not aspirational — the spec is simply describing a different
> algorithm from the one that shipped.
>
> ### Suggested fix
>
> Move steps 4 and 5 to **after** step 8, so the order becomes 1, 2, 3, 6, 7, 8, then the
> domain-parameter/length steps, then the primitive. Nothing else needs to change:
>
> * Every one of steps 2, 3, 6, 7 and 8 is a comparison of already-validated key metadata, so
>   moving the ceiling past them costs nothing and cannot make a previously-succeeding call fail.
> * Once step 8 has run, "the EC domain parameters associated with *publicKey*" and those
>   associated with *key* are the same parameters, so the wording of step 4 need not change at
>   all — only its position. (If you would rather make that explicit, step 4 could equally read
>   *"associated with key"* after the move.)
> * The only observable difference is exactly the one the WPT rows assert: a mismatched pair
>   now reports the mismatch whatever *length* asked for, instead of reporting a length that
>   was never going to be derivable.
>
> A smaller alternative — leaving the steps where they are but deriving *maximumLength* from
> *key* rather than *publicKey* — would make the two rows above pass, but it leaves an
> `OperationError` reachable for a pair that step 8 is about to reject anyway (e.g. a P-256
> base key, a P-384 public key and `length: 264`), which is still not what browsers do.
> Reordering is the change that matches shipped behaviour in every case.
>
> Found while running the WebCryptoAPI corpus against
> [Jint](https://github.com/sebastienros/jint)'s Web Crypto implementation
> (sebastienros/jint#3180); Jint has taken the browsers' order and documented the divergence
> rather than keep two red WPT rows.

lahma:
The upstream issue is filed: https://github.com/w3c/webcrypto/issues/560 - the draft above went in as written, and the code comment now cites it directly.

PR #3214, opened 2026-08-21
Fix/nested entry callstack trim

jeske:
## Summary

Prevent a nested Engine.Execute() that throws an exception from destroying (clearing) the parent-call stack, by trimming only back to the evaluation entry depth on unhandled throw.

## Details

When a host callback re-enters the engine mid-run (e.g. `Engine.Execute` for a dynamically-inserted script) and that nested evaluation ends in an unhandled throw, `ScriptEvaluation` called `ResetCallStack()` — clearing ALL frames, including those of the still-live outer run. Consequences:

- The outer run's balanced pops desync (formerly a `"stack is empty"` crash in `Construct`; now silently tolerated by the `TryPop` guards — the existing `// if call stack was reset due to recursive call to engine or similar` comments acknowledge this).
- `Error`s constructed by the outer script afterwards capture truncated/empty `stack` traces.

Fix: capture `CallStack.Count` at `ScriptEvaluation` entry and, on the unhandled-throw paths, pop down to that depth via a new `JintCallStack.TrimTo(int)`. Popping (vs. clearing) keeps the recursion-depth statistics and the profiler's enter/exit stream balanced. `Advanced.ResetCallStack()` is unchanged.

Reviewers: the two `ResetCallStack()` → `TrimTo(...)` replacements in `ScriptEvaluation` are the behavioral change; at a top-level entry the entry depth is 0, so `TrimTo` is equivalent to the old `Clear()`.

## Linked issue

(issue not opened)

## Test plan

- Unit test added (failed on old code, passes on new)
- Validated downstream in a browser-style host: nested inserted-script failures during real-page loads (e.g. google.com) no longer corrupt the outer run, and a 314-test visual regression suite shows 0 changes.

## Breaking change?

No. Internal-only; public API unchanged. Observable effect: outer-run `error.stack` is no longer truncated after a nested evaluation's unhandled throw.

### DIFF
--- a/Jint.Tests.PublicInterface/CallStackTests.cs
+++ b/Jint.Tests.PublicInterface/CallStackTests.cs
@@ -107,4 +107,45 @@ b(7);
     at custom.ts:11:1".Replace("\r\n", "\n"));
     }
 
+    [Fact]
+    public void NestedEvaluationUnhandledThrowShouldNotClearOuterRunCallStack()
+    {
+        // A host callback invoked from a running script re-enters the engine with
+        // Engine.Execute (the browser-host "dynamically inserted <script>" shape) and
+        // contains that nested script's unhandled throw. The outer run is still live:
+        // its call-stack frames must survive, so stack traces captured by the outer
+        // script afterwards are complete and the outer run's pops stay balanced.
+        var engine = new Engine();
+
+        string nestedThrowMessageObservedByHost = null;
+        engine.SetValue("runNestedScriptContained", () =>
+        {
+            try
+            {
+                engine.Execute("throw new Error('nested failure');");
+            }
+            catch (JavaScriptException nestedScriptException)
+            {
+                nestedThrowMessageObservedByHost = nestedScriptException.Message;
+            }
+        });
+
+        engine.Execute("""
+            function makeProbeError() {
+                // 'new' exercises Construct's call-stack push/pop after the nested throw
+                return new Error('outer probe');
+            }
+            function outerFunction() {
+                runNestedScriptContained();
+                return makeProbeError().stack;
+            }
+            globalThis.outerStackAfterNestedFailure = outerFunction();
+            """);
+
+        nestedThrowMessageObservedByHost.Should().Be("nested failure");
+
+        var outerStack = engine.Evaluate("outerStackAfterNestedFailure").AsString();
+        outerStack.Should().Contain("at makeProbeError");
+        outerStack.Should().Contain("at outerFunction");
+    }
 }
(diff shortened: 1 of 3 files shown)


lahma (approved):
Nice find, and the diagnosis is exactly right: `ScriptEvaluation`'s `ResetCallStack()` was reaching past its own evaluation and clearing frames that belong to a still-live outer run. Trimming to the entry depth is the correct shape, and popping rather than clearing is the right call for the reasons you give — it keeps the recursion-depth multiset and the profiler's enter/exit stream balanced instead of abandoning both.

Verified locally on a rebase onto current `main`:

- The new test fails on the unfixed code on **both** TFMs (`at outerFunction` missing, outer stack truncated to `at <anonymous>:7:12`) and passes with the fix — so it pins the actual defect, not just the code path.
- `Jint.Tests` (net10.0 + net472), `Jint.Tests.PublicInterface` (both TFMs, and again with `JINT_HOST_CONTRACT_VERIFICATION=1`), `Jint.Tests.CommonScripts`, and test262 (102,495 passed / 0 failed) are all green.

A few things I checked that are worth recording, since they're the parts that could have gone wrong:

- **At a top-level entry the two are equivalent for the stack**, since the entry depth is 0. They are *not* equivalent for the bookkeeping: `Clear()` wiped `_statistics` outright and issued `RecordAbandon()`, where `TrimTo` decrements per frame and issues one `RecordExit` each. For the nested case that is the improvement — the outer run's frames really are still open, so leaving them open in the profile is the honest answer, and `RecordExit` already tolerates a frame that predates the session.
- **Tail-call retention is safe.** `ReplaceTop` retains an occurrence that only `ContinueTailCalls`' `finally` hands back through `ReleaseTailRetention`. Because `ScriptEvaluation`'s `catch` is outside every trampoline, those releases have already run by the time the trim happens, so popping cannot strand a retained occurrence in `_statistics`.
- **`SetJavaScriptCallstack` still runs before the trim** on the completion-throw path, so the thrown exception's own trace is unchanged.
- One consequence worth naming: preserving the outer frames means the outer run's `MaxRecursionDepth` budget is no longer silently reset by a nested evaluation's unhandled throw. That is the correct direction — the old clearing was a hole in the limit of the same family as the tail-call one fixed in #3022 — but it is a real behavioural difference, not only a stack-trace one.

Two non-blocking nits, neither worth another round-trip:

- `JintCallStack.cs`: the doc comment has a literal `\u2014` where an em dash was intended. C# doesn't process unicode escapes inside comments, so it stays as those six characters. I'll tidy it after merge.
- `DebugHandler` (around line 137) already open-codes this exact loop to restore a captured depth; it can now just call `TrimTo`. Also a follow-up.

Thanks for the careful write-up and for validating it downstream in a real host — that context made this much quicker to review.

lahma:
CI is green and the local run matched it. Merging — thank you, David! Careful diagnosis, a test that actually fails on the old code, and a fix that stays inside the evaluation that owns the frames.

PR #3259, opened 2026-08-22
Engine: give the pump the anonymous-wildcard identity, scoped to a top-level park

### LINKED ISSUE #3242: WaitForScheduledWork's reservation identity refuses callbacks authorized under an earlier entry
Found while fixing #3206 (#3240), which closed the two windows it named and deliberately left this one, because closing it widens the admission surface and #3206/#3220 do not need it.

## The gap

#3240 established that an **authorized** cross-thread callback must be able to wait for its turn wherever the engine thread has undertaken to yield — the contract README's thread-safety section and TM-13 both state. It closed two windows where `Engine._asyncOwner` was null: the tail of an `async` host call's body, and the per-iteration gaps in `DrainEventLoopUntil`.

`Engine.Advanced.WaitForScheduledWork` has the same class of gap with a **different shape**: its reservation is a fresh `new object()` rather than the anonymous wildcard. So the owner is non-null, and the admission check does not take the `owner is null` path — it takes the *wrong-owner* path instead. A callback authorized under an **earlier** host entry is refused while a host is parked on the pump, even though the pump is precisely a point where the engine thread has undertaken to yield.

This matters for the host loop shape `WaitForScheduledWork` exists to serve (added in #3217): a host parks its thread on the pump and expects callbacks it handed out earlier to be able to run turns. Under the current identity they cannot.

## Why it was not fixed in #3240

Closing it means giving the pump's reservation (and the async-await reservations that share the pattern) the **anonymous-wildcard identity**, so that any authorized callback matches rather than only one issued under that exact entry.

That is a real widening of the admission surface, and TM-13 is the mitigation it would be widening. The wildcard is what makes "any authorized callback may wait here" true, and the specific token is what makes "only *this* entry's callback may run here" true. Which one the pump should have is a deliberate security-relevant decision, not a mechanical fix — and #3206 and #3220 are both fully closed without it.

## What a fix should establish

1. **Whether the wildcard is correct for the pump specifically.** A host parked on `WaitForScheduledWork` has, by calling it, said "run scheduled work here" — which is arguably exactly the undertaking that justifies the wildcard, unlike a host blocked in an unrelated synchronous call.
2. **The residual wedge.** #3240 measured its own: a host that blocks its engine turn on a callback it was not itself given can wedge, and that hazard class already existed on `main`. Widening the pump's identity extends the same class further and the new exposure should be measured the same way, with probes, not reasoned about.
3. **Whether the async-await reservations (`AwaitPromiseSettlementAsync`) share the shape** and should move together, since a mixed answer would be harder to document than either uniform one.

## Not urgent

Nothing is failing on this. It is a known, bounded gap in a contract that #3240 otherwise makes true, recorded so it is closed deliberately rather than rediscovered from a host bug report.

Related: #3240 (the two windows that were closed, and the residual-exposure probes), #3206, #3220.

lahma:
Closes #3242. The analysis and verdict are in [that issue's comment](https://github.com/sebastienros/jint/issues/3242); this implements it, and both of its structural claims checked out exactly in the code.

#3240 established that an **authorized** cross-thread callback must be able to wait wherever the engine thread has undertaken to yield. The pump was the one frame left where it could not.

## Two forms, two different fixes

**`WaitForScheduledWorkAsync`** minted a fresh `new object()`. Every other reservation is minted by a frame under which script runs, so callbacks converted during it carry that token — but this frame provably runs none (`InspectScheduledWork` reads queue state and executes nothing), so its token was an **empty** match, not a narrow one. It now reserves under the anonymous wildcard. Nothing else is needed: the reservation is taken synchronously and refuses an engine any thread already owns, so this form can never be reached from inside a running evaluation.

**The synchronous `WaitForScheduledWork`** — which the issue's premise got wrong — took **no reservation at all**, so it refused from the *null-owner* branch, **and** held `_ownerThreadId` for the whole park. A reservation alone would have admitted a callback that then blocked in `AcquireHostCall` until the park ended. It needed the pair `DrainEventLoopUntil` already has: a window for the whole park, plus a per-iteration suspension around `WaitForWork` yielding only the thread.

`SuspendHostCallForCallbacks`, `DrainEventLoopUntil`, `AwaitPromiseSettlementAsync` and the `*Async` entries are unchanged — they hold an operation token, and under the uniform rule that is correct.

## The guard, and proof the suite cannot see it

A pump reached from inside a running evaluation has undertaken nothing, so it must keep refusing. The guard is `HostCallScope.IsEntryRoot` — #3240's claiming-scope rule, never `_ownerDepth == 0`, which #3240 got wrong first.

Evidence, by forcing the guard true, rebuilding, running, restoring:

| | with guard | guard removed |
| --- | --- | --- |
| `ANestedParkInsideARunningEvaluationAdmitsNothing` | admitted **0** / 12 | **`Expected admitted to be 0, but found 12`** |
| `HostPumpWaitTests` (2) | pass | pass |
| `WaitForScheduledWorkTests` (9) | pass | pass |

So the pre-existing suites genuinely cannot see it — the new pin is the only thing that catches its removal. Worth noting `HostPumpWaitTests.WaitForScheduledWorkWakesOnACrossThreadPost` is *itself* a nested park and stays green either way, which is exactly why it could not serve as the pin.

## Probes, before and after

| behaviour | before | after |
| --- | --- | --- |
| stale authorized callback at a **sync** park | **FAIL** — refused at `Engine.cs:202` | admitted, returns 42 |
| same at an **async** park | **FAIL** — same refusal | admitted, returns 42 |
| parked in an unrelated blocking host `Action` | refused | refused (unchanged) |
| unrelated **public** entry while parked | refused | refused (unchanged) |
| **nested** pump inside a running evaluation | 12/12 refused | 12/12 refused |
| admitted callback enqueues a microtask | **FAIL** — park burned its 10 s ceiling, reported no work | park wakes, reports `true` |
| admitted callback outruns the ceiling | **FAIL** — `hold()` never reached | park returns after the callback, elapsed > 2× ceiling |

Pre-fix `Failed: 4, Passed: 3`; post-fix `Passed: 7` — and the class dropped from 21 s to 2 s, the two 10 s burns being gone.

## Fail-fast is unchanged

Throughout both parks the reservation is non-null, so `EnterHostCall(asyncOwner: null)` from any other thread still throws. Pinned by two new tests, by `WaitUntilParked`'s own handshake (which *is* a refused `ProcessTasks()`), and by pre-existing untouched tests — `AParkedWaitOwnsTheEngineForItsWholeDuration`, `ASecondThreadWaitingIsRefused`, and **`CanonicalJsCallDelegateCannotBypassOwnership`**, all green and not edited.

## Documentation

TM-13's enumeration gains the park, and its scope claim is **corrected**: "carry operation-scoped authorization" was only ever true where an operation token is in force. The new bullet says what is actually true —

> How narrowly a window admits depends on whether an operation token is in force under it. A window opened under one admits only callbacks issued under that same operation. A window opened where none is in force is anonymous, and admits any callback this engine ever authorized, including one whose operation has long since ended: that is the case at a top-level blocking drain and at a top-level park. A pump reached from inside a running evaluation opens no window at all and refuses.

Two residual bullets added: the wedge shape the park extends (already reachable through the drain, outside the supported contract, and still unreachable by a raw `JsCallDelegate`), and **"`WaitForScheduledWork`'s timeout bounds the idle wait, not the call"** — the one genuinely new observable, which also replaces the now-false paragraph in the method's own remarks and updates README's window enumeration from three to four.

## Verification

Solution builds Release, 0 errors. `Jint.Tests` 10425 (net10.0) / 7206 (net472); `Jint.Tests.PublicInterface` 2492 / 1906; both repeated with `JINT_HOST_CONTRACT_VERIFICATION=1` (2496 / 1910). **test262 102,495 passed / 0 failed on both this branch and an `upstream/main` baseline taken in this same worktree — identical.**

Because this is threading: 10× the pump/concurrency/async/event-loop classes on net10.0 (65/65 and 400/400 every time), 5× on net472 (65/65 and 215/215). No flakes.

## One asymmetry this creates, worth knowing

**The drain does not have this guard.** `DrainEventLoopUntil` opens the wildcard window even when nested inside a running evaluation. That is pre-existing and outside the sanctioned scope here, so it is untouched — but it means the pump is now *stricter* than the drain about nesting rather than merely equal to it. If that should be closed it is a separate change with its own argument; filed separately.

Also naming debt, no behaviour: `SuspendHostCallForCallbacks(hasTransferredCallback:)` now reads oddly at two call sites — the drain already passes a literal `true` with no callback transferred to that frame, and the pump passes its guard bool. The parameter really means "yield the thread for callbacks".

### DIFF
--- a/.github/THREAT_MODEL.md
+++ b/.github/THREAT_MODEL.md
@@ -615,12 +615,19 @@ incorrect authorization context, cross-request disclosure, crashes, or hangs.
 - Same-thread synchronous callback re-entry is allowed. Async entry from inside an active
   engine call is rejected before work starts; top-level async APIs reserve ownership for the
   lifetime of their returned `Task` and transfer the active thread when a continuation resumes.
-- JavaScript callbacks converted to CLR delegates carry operation-scoped authorization. A host
-  may dispatch one to another thread while the engine holds an admission window — the engine
-  operation that made the call which received the callback has not returned, an async engine API is
-  outstanding, or the engine is draining its event loop for a blocking unwrap or import; Jint yields
-  and transfers the reserved engine one callback turn at a time without admitting unrelated public
-  callers.
+- JavaScript callbacks converted to CLR delegates carry an authorization recorded when the engine
+  handed the callback out. A host may dispatch one to another thread while the engine holds an
+  admission window — the engine operation that made the call which received the callback has not
+  returned, an async engine API is outstanding, the engine is draining its event loop for a blocking
+  unwrap or import, or a host thread is parked on `Engine.Advanced.WaitForScheduledWork` or its
+  asynchronous form; Jint yields and transfers the reserved engine one callback turn at a time
+  without admitting unrelated public callers.
+- How narrowly a window admits depends on whether an operation token is in force under it. A window
+  opened under one — a host call inside a running evaluation, an outstanding async engine API —
+  admits only callbacks issued under that same operation. A window opened where none is in force is
+  anonymous, and admits any callback this engine ever authorized, including one whose operation has
+  long since ended: that is the case at a top-level blocking drain and at a top-level park. A pump
+  reached from inside a running evaluation opens no window at all and refuses.
 - Background Task and module completions only enqueue work; an owning host turn drains it.
 
 **Missing or residual mitigation.**
@@ -636,6 +643,13 @@ incorrect authorization context, cross-request disclosure, crashes, or hangs.
   `InvalidOperationException` an unrelated caller receives. Waiting there would mean blocking on a
   thread that may itself be blocked on that callback, so the refusal is deliberate rather than a
   gap; a host must keep the engine inside a window for as long as its callbacks may arrive.
+- A host that blocks its own engine turn on a *second* authorized callback — one it was not itself
+  handed — can wedge the engine, because that callback can only run after the turn that is waiting
+  for it. The shape already exists at the blocking drain; the park extends its reach by one frame.
+  It is outside the supported contract, and a raw `JsCallDelegate` still cannot reach it at all.
+- `WaitForScheduledWork`'s timeout bounds the idle wait, not the call. An admitted callback holds
+  the engine and the park cannot return until it finishes, so a host budgeting a frame can be handed
+  control back arbitrarily later by a callback of its own making.
 - A host can still share projected CLR objects across otherwise separate engines.
 
 **Required host action.** Give an engine exclusive ownership for each complete operation and
--- a/Jint.Tests.PublicInterface/HostPumpAdmissionTests.cs
+++ b/Jint.Tests.PublicInterface/HostPumpAdmissionTests.cs
@@ -1,0 +1,463 @@
+#nullable enable
+
+using System.Diagnostics;
+using Xunit.Sdk;
+
+namespace Jint.Tests.PublicInterface;
+
+/// <summary>
+/// Which callbacks a host parked on <see cref="Engine.AdvancedOperations.WaitForScheduledWork"/> — or on its
+/// asynchronous sibling — may be interrupted by, and which it may not. The pump is a callback-admission
+/// window: a host that parks its engine's own thread there has said "run scheduled work here", which is the
+/// undertaking to yield that lets an authorized cross-thread callback wait for its turn instead of being
+/// refused (sebastienros/jint#3242, sebastienros/jint#3240).
+/// </summary>
+/// <remarks>
+/// <para>
+/// The window is deliberately scoped to a <em>top-level</em> park. A pump reached from inside a running
+/// evaluation has undertaken nothing — the thread is in the middle of somebody else's script — so it keeps
+/// refusing, which <see cref="ANestedParkInsideARunningEvaluationAdmitsNothing"/> is the pin for.
+/// </para>
+/// <para>
+/// Every handshake here is released by a thread the test owns. The one place a park is detected rather than
+/// signalled is <see cref="WaitUntilParked"/>: the pump runs no script, so the only thing it can announce
+/// itself with is the refusal it hands an unrelated public entry.
+/// </para>
+/// </remarks>
+public class HostPumpAdmissionTests
+{
+    private const string ConcurrentUseMessage =
+        "*already in use by another thread or has an asynchronous operation in progress*";
+
+    /// <summary>
+    /// Reached only by a genuine wedge — every wait below is released by a thread the test owns.
+    /// </summary>
+    private static readonly TimeSpan WedgeCeiling = TimeSpan.FromMinutes(2);
+
+    /// <summary>
+    /// How often a detection loop looks. It decides only when an observation is made, never what it has to be.
+    /// </summary>
+    private static readonly TimeSpan ProbeInterval = TimeSpan.FromMilliseconds(20);
+
+    /// <summary>
+    /// A park short enough that an admitted callback outlives it, for the one test about what a ceiling
+    /// actually bounds.
+    /// </summary>
+    private static readonly TimeSpan ShortParkCeiling = TimeSpan.FromMilliseconds(200);
+
+    /// <summary>
+    /// The bound on waiting for something an admitted callback is supposed to do. Nothing spends it on a
+    /// working engine — it is what turns "the callback was never admitted" into a report rather than into a
+    /// two-minute wedge.
+    /// </summary>
+    private static readonly TimeSpan AdmissionCeiling = TimeSpan.FromSeconds(10);
+
+    /// <summary>
+    /// How many times the nested-park test tries the callback, and how long it leaves between tries. Neither
+    /// number selects the state being asserted about — see that test's remarks — so they only decide how many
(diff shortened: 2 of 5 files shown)


lahma:
The macOS failure of `ANestedParkInsideARunningEvaluationAdmitsNothing` is resolved: **the test was racy, the guard has no hole.** The engine diff is unchanged — byte-identical to the previous push — and only the test's handshake moved (34 insertions, 14 deletions).

I held this PR unmerged rather than re-running CI, because a green re-roll would have told us nothing: if the guard *had* had a hole, that failure was the only thing that would ever have surfaced it.

## What the instrumentation shows

`EnterTransferredHostCallback` and the pump were hooked to record the state at the admission decision. Reproducing the CI shape by shrinking the ceiling — a slow runner, deterministically:

```
[   2ms] PARK-ENTER isEntryRoot=False asyncOwner=null ownerThreadId=6
[  52ms] PROBE  0 refused      (asyncOwner=null ownerThreadId=6)
[  99ms] PROBE  1 refused      (asyncOwner=null ownerThreadId=6)
[ 115ms] >>> EXECUTE RETURNED - the engine is now FREE <<<
[ 147ms] ENTER-CALLBACK asyncOwner=null ownerThreadId=0 ... PROBE 2 ADMITTED
   ... 10 admitted, every one strictly after EXECUTE RETURNED, all ownerThreadId=0 ...
```

Against a genuine top-level park:

```
[   2ms] PARK-ENTER isEntryRoot=True  asyncOwner=WILDCARD ownerThreadId=11
[ 211ms] ENTER-CALLBACK asyncOwner=WILDCARD ownerThreadId=0 reentry=11 -> ADMITTED
```

That settles it:

- **`IsEntryRoot` is `False` inside a running evaluation** and `asyncOwner` stays `null` for the entire nested park. No window is ever opened there.
- **Every refusal** reads `asyncOwner=null ownerThreadId=<evaluating thread>` — ordinary owned-engine fail-fast.
- **Every admission** reads `asyncOwner=null ownerThreadId=0` — a *free* engine, timestamped after `Execute` returned. That is `EnterTransferredHostCallback`'s `owner is null → EnterHostCall()` path claiming an idle engine, which is correct and has nothing to do with the pump.
- The two routes have **different signatures** and cannot be confused: a window admission reads `asyncOwner=WILDCARD` with `reentry` naming the parked thread; a free-engine admission reads `asyncOwner=null reentry=0`.

So the macOS 3-of-12 was the probing loop outliving the park. On a fast run the loop finished at 771 ms against a park ending at 1714 ms — only ~2.2× headroom, which a 3-core macOS VM running xUnit in parallel eats. The other candidate was checked and ruled out: `Execute`'s trailing drain is a bare `RunAvailableContinuations()`, not `DrainEventLoopUntil`, so it opens no window and yields no thread.

## The fix is by construction, not by retuning

Observing "the engine is parked" still leaves a gap between observing and acting. So the test now **forces the state to persist for the whole probing window**: the prober is armed by a host call the script makes immediately before the park, and **the park is ended by the prober**, which cancels only once its last attempt has returned.

```csharp
using var endPark = new CancellationTokenSource();
// prober: 12 attempts, then -> endPark.Cancel();
engine.SetValue("park", new Func<bool>(() =>
{
    try { return engine.Advanced.WaitForScheduledWork(WedgeCeiling, endPark.Token); }
    catch (OperationCanceledException) { return false; }
}));
engine.Execute("armProber(); globalThis.reported = park();");
```

The evaluation cannot finish while an attempt is outstanding, whatever the machine does. The interval and count now choose only how many observations are made and how far apart — never which state they are made in. The assertion is unchanged: `admitted == 0`, `refused == 12`, `attempts == 12`.

**Proof it is timing-proof rather than merely widened:** the prober was run **2.4× slower than the old ceiling allowed** — 12 × 300 ms against what used to be a 1500 ms park — and it passes. That is a larger overrun than the one that broke macOS.

## Mutation evidence re-verified

Forcing `isTopLevelPark = true`, rebuilding and running:

```
Failed ... ANestedParkInsideARunningEvaluationAdmitsNothing [597 ms]
  Expected admitted to be 0 ..., but found 12 (difference of 12)
```

12/12, no hang, the other six tests still green — then restored and confirmed `var isTopLevelPark = ownership.IsEntryRoot;` is back with the engine clean. **The guard was not narrowed to make the test pass.**

## On local reproduction, honestly

`DOTNET_ThreadPool_ForceMinWorkerThreads=1` could **not** reproduce it — six runs, all passed — and the reason is structural: `DedicatedThread.RunAsync` uses a real `Thread`, and the prober loop is `Thread.Sleep` plus a thrown exception, so nothing in it touches the thread pool. The macOS factor is raw per-iteration slowness, which shrinking the ceiling reproduces deterministically and with the exact CI signature rather than an approximation.

## Verification

`HostPumpAdmissionTests` **7/7 across 20 consecutive runs**. `Jint.Tests` 10498 (net10.0) / 7256 (net472); `Jint.Tests.PublicInterface` 2492 / 1906; both repeated with `JINT_HOST_CONTRACT_VERIFICATION=1` (2496 / 1910). **test262 102,495 passed on both this branch and an `upstream/main` baseline taken with the engine reverted — identical.**

Issue #3332, opened 2026-08-24
Value construction is outside the concurrency guard, and nothing says so


lahma:
Jint guards *operations* against concurrent use and does not guard *value construction*. That is a reasonable line to draw for performance, but it is drawn silently: nothing says where it is, the README's thread-safety paragraph reads as though it is not drawn at all, and the two APIs on the wrong side of it are the two the README tells hosts to use.

Found while answering [discussion #3317](https://github.com/sebastienros/jint/discussions/3317).

## Measured

`main` @ `8acbf6ff8`. A second thread builds a value while the engine thread is inside a host call:

| public API taking an `Engine` | second thread, engine busy |
| --- | --- |
| `engine.Evaluate` (control) | `InvalidOperationException` |
| `JsValue.FromObject` | `InvalidOperationException` |
| `new JsonSerializer(e).Serialize(value)` | `InvalidOperationException` |
| `new JsonParser(e).Parse(json)` | ran concurrently |
| **`JsObject.CreateFromEntries(e, entries)`** | **ran concurrently** |
| **`new JsArray(engine, 4)`** | **ran concurrently** |

`JsonParser` is a plain inconsistency and is #3330. The other two are the interesting case: `JsObject.Create` / `CreateFromEntries` and building a `JsArray` are what README's **Projecting host data** recommends, in preference to subclassing `ObjectInstance`, so they are per-object APIs on a bulk path. An always-on `EnterHostCall` there is a benchmark question, not an obvious fix — which is exactly why the answer should not be "guard everything".

## The right home is the verifier, not the hot path

`HostContractVerification` already exists for this class of problem — a contract the engine trusts, cannot afford to re-check on the hot path, and whose violation is otherwise silent. Its gate is a `static readonly bool` read once at type initialization, so `if (HostContractVerification.Enabled)` and the block it guards are **folded out entirely** in a Release process that never set the switch. A thread-affinity check there costs production nothing.

The check is the one `EnterHostCall` already performs, minus the claim: *another thread currently owns this engine and it is not me*. It cannot fire when the engine is idle, which is the legitimate case (a host building a value between turns), so it has no false positives to tune.

Proposed sites: `JsObject.Create` / `CreateFromEntries`, the `JsArray` constructors, and whatever else the audit turns up — `ObjectInstance`'s own constructor is the obvious candidate to consider, since it is where every engine-affine object begins.

[...]

PR #3361, opened 2026-08-25
StackGuard: the thread that continues the script owns the engine, so a host callback across the hop is not refused

### LINKED ISSUE #3343: StackGuard continues a script on another thread without transferring engine ownership, so a host callback across the hop is refused by the engine's own concurrency check
Once a script recurses deep enough for `StackGuard` to continue it on a fresh thread, **every public engine entry made from a host callback fails** with *"This Engine is already in use by another thread"* — on a single-threaded host, running a single-threaded script, using the engine exactly as documented. The engine refuses itself.

Found while building the thread-affinity verifier in #3332, which surfaced it as noise. Verification is **off** in everything below; this is the shipped behaviour.

## Reproduction

```csharp
var engine = new Engine(o => o.Constraints.MaxExecutionStackCount = 1_000_000);
var mainThread = Environment.CurrentManagedThreadId;

engine.SetValue("probe", new Func<int, int>(depth =>
{
    var onGuardThread = Environment.CurrentManagedThreadId != mainThread;
    try { engine.Evaluate("1 + 1"); Console.WriteLine($"{onGuardThread} -> OK"); }
    catch (Exception ex) { Console.WriteLine($"{onGuardThread} -> {ex.GetType().Name}: {ex.Message}"); }
    return depth;
}));

engine.Execute("function recurse(n) { if (n === 0) { return probe(n); } return recurse(n - 1) + 0; }");
engine.Evaluate("recurse(10)");      // shallow
engine.Evaluate("recurse(2000)");    // deep enough to cross the guard
```

```
depth 10:    main thread  -> Evaluate OK
depth 2000:  guard thread -> InvalidOperationException: This Engine is already in use by
                             another thread or has an asynchronous operation in progress.
depth 6000:  (same)
depth 12000: (same)
```

`JsValue.FromObject` fails the same way, and so does anything else behind the ownership check. The exception propagates out of the host callback and kills the evaluation.

## Cause

`StackGuard.RunOnEmptyStackCore` continues the evaluation on a **thread-pool thread** while the original thread blocks:

```csharp
Task<R> task = Task.Factory.StartNew((Func<object?, R>) action, state, CancellationToken.None,
    TaskCreationOptions.DenyChildAttach, TaskScheduler.Default);

if (!task.IsCompleted)
{
    ((IAsyncResult) task).AsyncWaitHandle.WaitOne();
}
```

Nothing transfers engine ownership across that hop. `Engine._ownerThreadId` goes on naming the original thread — which is parked in `WaitOne()` and cannot possibly be using the engine — so `EnterHostCall` on the thread that is *actually* running the script takes the wrong-owner path and throws.

Every other place in the engine where an operation resumes on a different thread already handles this: the async settlement loop re-claims through `EnterTransferredHostCall`, and the README's thread-safety section describes ownership transfer as a first-class concept. The stack guard is the one hop that does not participate.

## Scope

Needs `Options.Constraints.MaxExecutionStackCount` to be set — `StackGuard._enabled` is `_maxExecutionStackCount != Disabled`, and the default is `Disabled`. The default `Options.Constraints.StackOverflowGuard` is the *other* mechanism (`_backstopEnabled`), which measures remaining stack and throws without hopping threads, so a default-configured engine is unaffected.

That still leaves it reachable by any host that bounds recursion by frame count, which is an ordinary documented thing to configure, and Jint's own suite does it.

Two aggravating notes:

- **It is silent until it is not.** Script that never calls back into the host across the boundary works fine, so this surfaces as an intermittent failure that depends on recursion depth — the last place anyone would look for a threading error.
- **Recent guard work widens the exposure rather than causing it.** #3334 and #3330 put the ownership check in front of `JsValue.FromObject` and `JsonParser.Parse`, so a host callback doing either now hits this too. Those are correct changes; the hole they land on is this one.

## What a fix has to establish

The hop is fully synchronous — the original thread is blocked for the whole of it — so a transfer is safe in principle. What is not obvious is which primitive it should use and what has to travel with it:

1. `_ownerThreadId`, obviously, restored on the way back.
2. `_ownerDepth`, `_ownerToken` and the callback-admission state, or a callback authorized under the outer frame stops matching once execution crosses the hop.
3. The memory-limit operation state, which `EnterTransferredHostCall` carries for the async paths and which is charged per thread.
4. Whether a *third* thread may enter during the hop. It must not: the operation is still in progress, merely on a different thread. So this is a transfer, not a release-and-reclaim.

`EnterTransferredHostCall` exists for exactly this shape and is the obvious candidate, but it was written for an async resume where `_asyncOwner` is set, and here it usually is not.

## Blocks

#3332 — the thread-affinity verifier's premise is that `_ownerThreadId` names the thread actually inside the engine. While this is open it does not, and the verifier reports Jint's own guard hop as a host violation (three tests in `Jint.Tests`: `EngineTests.ShouldThrowErrorWhenMaxExecutionStackCountLimitExceeded`, `EngineLimitTests.ShouldNotStackoverflowWhenStackGuardEnable`, `StackOverflowGuardTests.MaxExecutionStackCountTakesPrecedenceOverTheGuard`). Fixing this first makes that verifier land clean; working around it would bake a known-wrong ownership record into a new API.

## Related

- #3035 — where the ownership guard came from.
- #3334, #3330 — guards that now sit in front of this hole.
- #3332 — blocked on this.

lahma:
Fixes #3343.

`Options.Constraints.MaxExecutionStackCount` selects a lane that continues a deep call chain on a fresh thread-pool thread while the calling thread blocks in `WaitOne()`. Nothing transferred engine ownership across that hop, so `Engine._ownerThreadId` went on naming the parked thread — which cannot possibly be using the engine — and every public entry made from the far side took the wrong-owner path and threw `InvalidOperationException`. One host thread, one script, no concurrency: the engine refused itself.

## It is broader than the report

`MemoryLimitConstraint.Check()` takes the ownership check too, and it is exact, so it runs per statement. `LimitMemory` together with `MaxExecutionStackCount` therefore killed a **plain recursion with no host callback anywhere near it**:

```csharp
var engine = new Engine(o =>
{
    o.Constraints.MaxExecutionStackCount = 1_000_000;
    o.LimitMemory(200_000_000);
});

engine.Execute("function recurse(n) { return n === 0 ? 0 : recurse(n - 1) + 0; }");
engine.Evaluate("recurse(10)");    // 0
engine.Evaluate("recurse(2000)");  // InvalidOperationException: This Engine is already in use by another thread
```

Two documented settings, no host code involved, and the second combination is unusable.

## The fix

`StackGuard.RunOnEmptyStack` becomes an instance method and brackets the hop with `Engine.BeginStackGuardHandoff` on the blocking thread, `ClaimStackGuardHandoff` on the continuation thread, `EndStackGuardHandoff` on the way back.

Four things the issue asked to be established, and how each is answered:

- **`_ownerThreadId`** travels, and is restored by the blocking thread rather than by the claim's disposal, so a hop that failed before it could claim — or after it claimed and before it released — still leaves ownership naming the thread about to resume.
- **`_ownerDepth`, `_ownerToken`, `_hostEntryDepth` and the callback-admission state** deliberately do *not* travel, because they do not change: the hop continues **one entry** rather than starting another. A callback authorized under the outer frame still matches, and the outer run's constraints still apply — which is also why this fix cannot become a new route into the "a single function call is a run" trap.
- **The memory-limit operation state** does travel, and has to. It is charged from the runtime's per-thread allocation counter, so a segment opened on one thread and accumulated on another would either mis-bill the allocations or trip `MemoryLimitConstraint`'s own changed-thread guard. `TrySuspendActiveSegment` on the way out, `BeginSegment`/`EndSegment` around the continuation, `EndSegment` on the way back — the same primitives `SuspendHostCallForCallbacks` already uses.
- **A third thread must not enter.** It cannot: `_ownerThreadId` is never written back to zero, so this is a transfer and not a release-and-reclaim. An unrelated thread reaching a public entry mid-hop is refused exactly as it was before the hop started, and there is a test that pins it.

`EnterTransferredHostCall` was the obvious candidate the issue named, and it is the wrong one here: it is keyed on `_asyncOwner`, which is normally null at this point, so it would fall through to `EnterHostCall` and throw. What the hop needs is narrower than an async resume — no reservation, no admission gate, no drain — and expressing that as its own three-call bracket keeps both readable.

## Tests

`Jint.Tests.PublicInterface/HostStackGuardThreadHopTests.cs`, four tests, all four failing against `main`:

```
Failed AMemoryLimitedScriptSurvivesTheHopWithNoHostCallbackAtAll [144 ms]
  System.InvalidOperationException : This Engine is already in use by another thread or has an
  asynchronous operation in progress. Engine instances may only be used by one host operation at a time.
Failed AnUnrelatedThreadIsStillRefusedWhileTheHopHoldsTheEngine [36 ms]
  System.InvalidOperationException : This Engine is already in use by another thread ...
Failed AHostCallbackReachedAcrossTheHopCanReEnterTheEngine [41 ms]
  System.InvalidOperationException : This Engine is already in use by another thread ...
Failed TheMemoryOperationIsChargedOnBothSidesOfTheHop [5 ms]
  System.InvalidOperationException : This Engine is already in use by another thread ...
Failed!  - Failed: 4, Passed: 0, Skipped: 0, Total: 4
```

**On determinism**, since these are threading tests and this suite already has a timing-flake problem. There is no sleep and no timing assertion anywhere in the file:

- Every test runs on `DedicatedThread.Run(..., maxStackSize: 1 MB)`, the same deliberately small stack `StackOverflowGuardTests.MaxExecutionStackCountTakesPrecedenceOverTheGuard` uses, so a 2000-deep recursion crosses the hop rather than hoping to.
- Every test then **asserts the hop happened** — that the probe ran on a thread other than the one that entered. A machine with a roomier stack does not silently pass them; it fails them. That is the difference between pinning this and pinning nothing.
- `AnUnrelatedThreadIsStillRefusedWhileTheHopHoldsTheEngine` uses an explicit two-way handshake (`ManualResetEventSlim`), not a window: the continuation signals that it is inside the hop, the unrelated thread answers, and only then does the continuation return. The waits are bounded by `TestBudgets.WedgeCeiling`, which is a wedge ceiling and never an assertion.

## Verification

Re-run after rebasing onto `main` at `63459e223`:

- `dotnet build -c Release` clean, five target frameworks, `TreatWarningsAsErrors` on.
- `Jint.Tests`: 10,677 passed on `net10.0` and `net8.0`, 7,316 on `net472`, 0 failed.
- `Jint.Tests.PublicInterface`: 2,988 passed on `net10.0`, 2,359 on `net472`, 0 failed.
- Both again with `JINT_HOST_CONTRACT_VERIFICATION=1`: 0 failed.
- `Jint.Tests.Test262`: 102,495 passed, 0 failed.
- No public surface change — everything added is `internal` — so the five API baselines and the documentation register are untouched.
- `docs/v5-migration.md` §4.26.

One note on the suite, since it will show up in CI eventually: `SchedulerTests.AHostsOwnPumpRunsTheTasks` fails roughly one full leg in five, on **pristine `main`** with nothing else in the tree. It is a race in that test's first assertion and has nothing to do with this change — filed as #3372 with the measurements.

This is what #3332 is blocked on: with it, `_ownerThreadId` names the thread actually inside the engine, which is the premise the thread-affinity verifier there needs.

### DIFF
--- a/Jint.Tests.PublicInterface/HostStackGuardThreadHopTests.cs
+++ b/Jint.Tests.PublicInterface/HostStackGuardThreadHopTests.cs
@@ -1,0 +1,196 @@
+#nullable enable
+
+using System.Collections.Generic;
+using Jint.Constraints;
+using Jint.Native;
+using Jint.Native.Json;
+
+namespace Jint.Tests.PublicInterface;
+
+/// <summary>
+/// <see cref="Options.ConstraintOptions.MaxExecutionStackCount"/> selects a lane that continues a deep call
+/// chain on a fresh thread-pool thread while the calling thread blocks. These pin that the hop is invisible
+/// to the host: engine ownership travels with it, so a host callback reached across it may re-enter the
+/// engine, and the memory-limit segment is charged on whichever thread is actually allocating.
+/// <para>
+/// Every test here runs on a deliberately small stack and asserts that the hop <em>happened</em> — that some
+/// engine work ran on a thread other than the one that entered. Without that assertion a machine with a
+/// roomier stack would pass them all without ever crossing the boundary they are about.
+/// </para>
+/// </summary>
+public class HostStackGuardThreadHopTests
+{
+    /// <summary>Smaller than the platform default, so the recursion below reliably crosses the hop.</summary>
+    private const int SmallStack = 1024 * 1024;
+
+    /// <summary>Deeper than a 1 MB stack holds, so the lane hops rather than throwing.</summary>
+    private const int DeepEnoughToHop = 2000;
+
+    /// <summary>High enough never to fire; this suite is about the hop, not about the limit.</summary>
+    private const int UnreachedStackCount = 1_000_000;
+
+    private const string RecurseThenProbe =
+        "function recurse(n) { return n === 0 ? probe(n) : recurse(n - 1) + 0; }";
+
+    [Fact]
+    public void AHostCallbackReachedAcrossTheHopCanReEnterTheEngine()
+    {
+        DedicatedThread.Run(
+            () =>
+            {
+                var engine = new Engine(options => options.Constraints.MaxExecutionStackCount = UnreachedStackCount);
+                var entryThread = Environment.CurrentManagedThreadId;
+                var probeThreads = new List<int>();
+                var failures = new List<string>();
+
+                engine.SetValue("probe", new Func<int, int>(depth =>
+                {
+                    probeThreads.Add(Environment.CurrentManagedThreadId);
+                    try
+                    {
+                        engine.Evaluate("1 + 1");
+                        JsValue.FromObject(engine, new { a = 1 });
+                        new JsonParser(engine).Parse("{\"a\":1}");
+                    }
+                    catch (Exception exception)
+                    {
+                        failures.Add($"{exception.GetType().Name}: {exception.Message}");
+                    }
+
+                    return depth;
+                }));
+
+                engine.Execute(RecurseThenProbe);
+                engine.Evaluate("recurse(0)");
+                engine.Evaluate($"recurse({DeepEnoughToHop})");
+
+                probeThreads.Should().HaveCount(2);
+                probeThreads[0].Should().Be(entryThread, "a shallow call never leaves the entry thread");
+                probeThreads[1].Should().NotBe(
+                    entryThread,
+                    "the deep call has to actually cross the guard's thread hop, or this test pins nothing");
+                failures.Should().BeEmpty("the host is single-threaded and the engine is running its script");
+            },
+            maxStackSize: SmallStack);
+    }
+
+    [Fact]
+    public void AMemoryLimitedScriptSurvivesTheHopWithNoHostCallbackAtAll()
+    {
+        DedicatedThread.Run(
+            () =>
+            {
+                var engine = new Engine(options =>
+                {
+                    options.Constraints.MaxExecutionStackCount = UnreachedStackCount;
+                    options.LimitMemory(200_000_000);
+                });
+
+                engine.Execute("function recurse(n) { return n === 0 ? 0 : recurse(n - 1) + 0; }");
+                engine.Evaluate("recurse(0)").AsNumber().Should().Be(0);
+                engine.Evaluate($"recurse({DeepEnoughToHop})").AsNumber().Should().Be(0);
+            },
+            maxStackSize: SmallStack);
+    }
+
+    [Fact]
+    public void TheMemoryOperationIsChargedOnBothSidesOfTheHop()
+    {
+        DedicatedThread.Run(
+            () =>
+            {
+                var engine = new Engine(options =>
+                {
+                    options.Constraints.MaxExecutionStackCount = UnreachedStackCount;
+                    options.LimitMemory(200_000_000);
+                });
+
+                var entryThread = Environment.CurrentManagedThreadId;
+                var probeThread = 0;
+                long observed = -1;
+                string? failure = null;
+
+                engine.SetValue("probe", new Func<int, int>(depth =>
+                {
+                    probeThread = Environment.CurrentManagedThreadId;
+                    try
+                    {
+                        observed = engine.Constraints.Find<MemoryLimitConstraint>()!.AllocatedBytes;
+                    }
+                    catch (Exception exception)
+                    {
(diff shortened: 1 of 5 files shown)


scgm0:
> `Options.Constraints.MaxExecutionStackCount` selects a lane that continues a deep call chain on a fresh thread-pool thread while the calling thread blocks in `WaitOne()`

Does this mean that the engine will execute code concurrently at certain times?

Issue #3442, opened 2026-08-27
A prepared script's regexes ignore Options.Constraints.RegexTimeout and use the 10-second prepare-time default


lahma:
`Options.Constraints.RegexTimeout` is the host's regex budget, and `ValidateSecurityConfiguration` reports an untimed engine as an error. A script parsed *by the engine* honours it. A script handed to `Engine.PrepareScript` / `Engine.PrepareModule` without an explicit `RegexTimeout` does not: it carries `Engine.DefaultRegexTimeout` (10 s), whatever the executing engine was configured with.

## Where

There is no engine at prepare time, so `ScriptPreparationOptions.GetParserOptions()` / `ModulePreparationOptions.GetParserOptions()` fall back to `Engine.DefaultRegexTimeout` rather than to a constraint they cannot see:

```csharp
static parsingOptions => parsingOptions.ApplyTo(_defaultParserOptions, RegexCompilation.Compiled, Engine.DefaultRegexTimeout);
```

That value is baked into the `RegexConversionOptions` instance hanging off `ParserOptions.OnRegExp`, and from there into the `Regex`'s `MatchTimeout` when a literal is first adapted.

## Reproduction

```csharp
var prepared = Engine.PrepareScript("'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa!'.match(/^(a+)+$/)");
var engine = new Engine(o => o.Constraints.RegexTimeout = TimeSpan.FromMilliseconds(400));

engine.Execute(prepared);
// RegexMatchTimeoutException, MatchTimeout == 00:00:10
```

Measured, not inferred. It applies to the regexes such a script builds at run time as well, after sebastienros/jint#3431 made those agree with its literals.

## Why it is worth a separate issue from #3431

It is documented — `IParsingOptions.RegexTimeout`'s XML doc says the constraint applies to "non-prepared scripts and modules ... otherwise the default value of 10 seconds is used", and README tells an embedder preparing untrusted code to pass the limit explicitly. So it is known, not accidental like #3431 was.

But it is the same class of gap seen from the embedder's side: a host that tightens `RegexTimeout` for security and then adopts `PrepareScript` — the path this repository recommends for production — silently gets 10 s and no diagnostic saying so. `ValidateSecurityConfiguration` inspects `Options`, which by then says 400 ms.

## Why the fix is not small

[...]

Issue #3426, opened 2026-08-26
Audit: process-wide caches whose value depends on per-engine configuration (three found by accident)


lahma:
Three defects found independently in this campaign share one shape: **process-wide state whose correctness depends on per-engine configuration.**

| | state | what its key omits | outcome |
|---|---|---|---|
| #3331 | `Engine._defaultEngineOptions` | that `new Engine()` shares one instance | a `WebApi.Enable` callback published one tenant's `HttpClient` to every default-built engine in the process |
| #3368 | `JsAccessibleRegistry.Generation` | which type was registered | one registration dropped every engine's whole accessor cache |
| #3424 | `JintBinaryExpression._knownOperators` | the engine's `ValueCoercion` and installed `ClrTypeConverter` | two engines with different conversion configuration over the same host types: whichever evaluates first decides for both |

The first two are fixed. The third is filed because re-keying a hot-lane cache wants its own measurement.

**Three of one shape is a pattern, and the useful artefact is an inventory rather than a fourth point fix.** Each was found by accident, while doing something else — which is the argument that more exist and that nothing will surface them on purpose.

**The scope:** every `static` cache in the interop and interpreter lanes, checked against the question *"does the value this stores depend on anything not in the key?"* Candidates noticed in passing during #3407, explicitly **not** audited and not to be read as a clean bill:

- `JintBinaryExpression._knownOperators` — known bad, #3424
- the shared `MethodInvoker` / `ConstructorInvoker` tables in `MethodDescriptor`
- `DefaultTypeConverter._knownCastOperators`
- `DefaultTypeConverter._knownFromResultGenerics`
- `TypeResolver`'s process-wide accessor cache — already partitioned by `InteropResolutionProfile`, so the question there is whether that profile still covers everything the resolution depends on after #3346 removed the converter from it

For each: does the key cover every input the value depends on? If not, is the divergence reachable by an embedder — two engines, one process, different `Options.Interop` — and what does it produce? A cache whose value is genuinely configuration-independent is a fine answer and should be recorded as one, so the next person does not re-audit it.

The deliverable is the inventory with a verdict per entry, not necessarily fixes. Where a fix is small and safe it can ride along; where it is a hot-lane re-keying it should be filed with the measurement it needs.

lahma:
## The inventory

[...]

Issue #3379, opened 2026-08-25
RegExpTests.PreparedScriptHonorsRegexTimeoutForCustomEngine bounds wall clock with a fixed 15s constant


lahma:
`RegExpTests.PreparedScriptHonorsRegexTimeoutForCustomEngine` asserts an upper bound on wall-clock time and fails on a contended runner:

```
[FAIL] PreparedScriptHonorsRegexTimeoutForCustomEngine(scriptTemplate: "/{1}/.test('{0}')")
Expected time to be less than 15s because Expected RegexMatchTimeoutException within 15s,
took 16.7s, but found 16s, 666ms and 537.2µs.
```

Same family as the assertions #3369 fixed, but in `Jint.Tests` rather than `Jint.Tests.PublicInterface`, so it was outside that PR's enumerated set. Seen on an unrelated PR (`[JsAccessible]` packaging), which is the usual cost — an author is told their change broke something it cannot reach.

The property under test is real and should not be deleted: a regex timeout that never fires would also fail, just later. But 15 s is a fixed constant against an engine-configured timeout, so the margin is not derived from the thing it is bounding.

#3369 established the pattern worth reusing here: express the bound as a ratio of the configured budget rather than as a constant, so a slow runner cannot invert it, and use `TestBudgets.WedgeCeiling` for any wait whose duration nothing asserts. Whether this one can take a `TimeProvider` is worth checking — the regex timeout path reads a `Stopwatch` deadline inline (there is a note about this in the campaign's memory), so a fake clock may not reach it, in which case the ratio is the answer.

Worth a sweep of `Jint.Tests` for the same shape while someone is in there: a fixed `TimeSpan` compared against elapsed wall clock.

lahma:
A second instance of the same shape in `Jint.Tests`, on macOS this time:

```
[FAIL] Jint.Tests.Runtime.AsyncTests.EngineReusableAfterEvaluateAsyncTimeout
Jint.Runtime.PromiseRejectedException : Promise was rejected with value Timeout of 00:00:01 reached
```

A **one-second** budget, which is tighter than any of the others and correspondingly easier to blow. Seen on an unrelated PR (the `[JsAccessible]` packaging work).

So the sweep this issue proposes has at least two confirmed members outside the set #3369 fixed:

| test | budget | seen on |
|---|---|---|
| `RegExpTests.PreparedScriptHonorsRegexTimeoutForCustomEngine` | 15 s wall-clock bound | Windows |
[...]

PR #3443, opened 2026-08-27
RegExp: a script-constructed pattern is bounded by the engine's configured RegexTimeout

### LINKED ISSUE #3431: new RegExp() ignores Options.Constraints.RegexTimeout and uses the parser's default instead
`Options.Constraints.RegexTimeout` is documented as "maximum time a Regex is allowed to run", is copied by `Options.CopySecurityPosture`, and `ValidateSecurityConfiguration` reports an untimed engine as an error. A regexp **literal** honours it. A regexp built at run time — `new RegExp(...)`, `RegExp.prototype.compile`, `"x".match("...")` — does not: it gets Acornima's `ParserOptions.RegexTimeout` default instead, which is the same value for every engine in the process.

## Where

`RegExpConstructor.CreateRegExpFromPattern` reads the *compilation mode* off the `Engine.RegexConversionOptions` that Jint hangs on `parserOptions.OnRegExp`, and then reads the *timeout* off Acornima's own property beside it:

```csharp
var compilation = (parserOptions.OnRegExp?.Target as Engine.RegexConversionOptions)?.Compilation == RegexCompilation.Interpreted
    ? RegexCompilation.Interpreted
    : RegexCompilation.Adaptive;

regExpParseResult = RegExpParseCache.GetOrAdapt(p, f, compilation, parserOptions.RegexTimeout);
```

Jint never assigns `ParserOptions.RegexTimeout`. `ScriptParsingOptions.ApplyTo` / `ModuleParsingOptions.ApplyTo` set `AllowReturnOutsideFunction`, `OnRegExp`, `OnNode` and `Tolerant`, and `Engine.BaseParserOptions` sets `EcmaVersion` and `ExperimentalESFeatures`. So the value that reaches `Regex.MatchTimeout` on this path is the package default, measured at 5 seconds.

The literal path is fine: `JintLiteralExpression` passes `conversionOptions.Timeout`, which `GetOnRegExpHandler` derived from `RegexTimeout ?? engineOptions.Constraints.RegexTimeout`.

## Reproduction

```csharp
var engine = new Engine(options => options.Constraints.RegexTimeout = TimeSpan.FromMilliseconds(50));

// dynamic
engine.Evaluate("'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa!'.match(new RegExp('^(a+)+$'))");
// RegexMatchTimeoutException after ~5049 ms

// literal, same engine configuration
engine.Evaluate("'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa!'.match(/^(a+)+$/)");
// RegexMatchTimeoutException after ~60 ms
```

Both directions are wrong, and only one of them is benign: an engine that *raises* the timeout to 30 s is cut off at 5 s, and an engine that *lowers* it to 50 ms — the security-relevant direction — runs for 5 s. An unconfigured engine (10 s default) is silently tightened to 5 s.

## Note on the cache

`RegExpParseCache`'s key is `(Pattern, Flags, Compiled, TimeoutTicks)`, so this is *not* one engine's timeout leaking into another's cache entry — the key is correct. The input handed to it is wrong, identically, for every engine. Worth saying because the two failure modes look the same from the outside.

## Suggested shape

The compilation mode is already read off `RegexConversionOptions`; the timeout is on the same object. Reading `Timeout` from there (falling back to `engine.Options.Constraints.RegexTimeout`) would make the dynamic path agree with the literal one. `RegExpPrototype.GetCustomEngineTimeout` already does exactly that fallback for the custom-engine path, so the two would then line up. `TryCompileWithCustomEngine` two lines below takes `parserOptions.RegexTimeout` as well and needs the same treatment.

Found while auditing process-wide caches for #3426.

lahma:
Fixes #3431.

`Options.Constraints.RegexTimeout` is documented as "maximum time a Regex is allowed to run", is copied by `CopySecurityPosture`, and `ValidateSecurityConfiguration` reports an untimed engine as an error. It reached only the regular expressions Jint's *parser* saw.

## The lanes, and which were unbounded

A regex literal is adapted from the `RegexConversionOptions` instance Jint binds its `OnRegExp` handler to, which resolved `ScriptParsingOptions.RegexTimeout ?? Options.Constraints.RegexTimeout` at parse time. Everything else is adapted at run time by `RegExpConstructor.RegExpInitialize`, which read the timeout off Acornima's own `ParserOptions.RegexTimeout` sitting beside it — a field Jint never assigns, so it was the parser package's default (5 s), identically, for every engine in the process.

| Lane | Before | After |
| --- | --- | --- |
| Regex literal, script | configured value | unchanged — **verified**, it is the control test |
| Regex literal, module | configured value | unchanged — **verified** |
| `new RegExp(...)` / `RegExp(...)` | 5 s | configured value |
| `RegExp.prototype.compile` | 5 s | configured value |
| `"x".match("...")` (string coerced to a pattern) | 5 s | configured value |
| `@@split` — rebuilds a sticky splitter **even from a literal** | 5 s | configured value |
| `search`, `replace`, `matchAll`, `test`, `exec` over any of the above | 5 s | configured value |
| Custom (QuickJS-port) engine, built at run time | compiled under 5 s, matched under the engine constraint | one value for both |
| Prepared script / module | its own prepare-time value | unchanged for literals; runtime-built patterns now agree with them |

`@@split` is the one worth pausing on: it is reachable without writing `new RegExp` at all, so a script containing nothing but literals still left the parser's lane.

## Proving it first

`Jint.Tests.PublicInterface/HostRegexTimeoutTests.cs` was written against unmodified `main`. Thirteen of its fourteen cases failed, and the one that passed is the literal control:

```
Failed AScriptBuiltRegexIsBoundedByTheConfiguredRegexTimeout("'{S}'.match(new RegExp('{P}'))") [4 s]
   Expected 400ms because the match must have run under the engine's configured
   Constraints.RegexTimeout rather than a default the host never chose, but found 5s.
...
Failed!  - Failed: 13, Passed: 1, Skipped: 0, Total: 14, Duration: 1 m 5 s
```

Post-fix the same fourteen pass in 6 s — the drop from 65 s is the second witness, every catastrophic match now aborting on the 400 ms budget instead of the 5 s one.

`Jint.Tests/Runtime/RegExpTests.PreparedScriptHonorsRegexTimeoutForRuntimeBuiltCustomEngineRegex` pins the custom-engine half and was likewise confirmed failing first, reporting the engine's 30 s after having actually waited it out:

```
Failed PreparedScriptHonorsRegexTimeoutForRuntimeBuiltCustomEngineRegex [30 s]
   Expected 1s because the match must have run under the prepared script's own regex
   timeout rather than falling through to the engine's 00:00:30, but found 30s.
```

**No test asserts a duration.** The witness is `RegexMatchTimeoutException.MatchTimeout`, which the matcher constructs from the very interval it was enforcing — so it names *which* budget fired, exactly, on a loaded runner as well as an idle one. That is the discriminator #3379 introduced for the prepared lane, for the same reason: two candidates whose overshoot scales together cannot be separated by a clock.

## The change

`RegExpInitialize` takes both halves of how a pattern is adapted — the compilation mode and the timeout — off the one conversion-options instance, falling back to `Options.Constraints.RegexTimeout` where no Jint handler is in force. That is the same expression `RegExpPrototype.GetCustomEngineTimeout` already resolved at match time, so the two now line up rather than only looking as though they do.

The custom-engine branch gets the same value and, additionally, carries the conversion options onto the parse result the way `JintLiteralExpression` does. A .NET `Regex` embeds `MatchTimeout` and so cannot be matched under a budget other than the one it was built with; the custom engine has nothing to embed it in and re-reads it per match, and without this the two ends disagreed for a prepared script — compiled under its own value, matched under the engine's.

## On the two things worth checking before merging this shape

**No `CancellationTokenSource` timer was introduced.** The .NET lane enforces the interval through `Regex.MatchTimeout`; the custom lane through the inline `Stopwatch` deadline `RegExpPrototype.ExecuteWithTimeout` already used, precisely so a saturated pool cannot starve the abort. The one pre-existing CTS is in `TryCompileWithCustomEngine`, where it bounds *compilation* rather than matching, and it is untouched — a starved timer there fires late, never early, so a tightened budget cannot turn a legitimate compile into a spurious failure.

**`RegExpParseCache`'s key already covers the timeout** (`(Pattern, Flags, Compiled, TimeoutTicks)`), so this was never one engine's budget leaking into another's cache entry — the input was wrong identically for all of them. That is now pinned from outside by `TwoEnginesWithDifferentBudgetsDoNotShareOneAdaptation`, which runs the same pattern on a 400 ms engine and then a 900 ms one and asserts each got its own. This is the shape being audited in #3426; here it is clean.

## Left out, filed separately

A prepared script or module with no explicit `RegexTimeout` runs under `Engine.DefaultRegexTimeout` (10 s) whatever the executing engine's constraint says — measured, not inferred: a default-prepared script on a 400 ms engine reports `MatchTimeout = 00:00:10`. It is documented behaviour (there is no engine at prepare time to read a constraint from) and it is unchanged here, but it is the same class of gap for a host that tightens the constraint and then uses `PrepareScript`, which is the recommended production path. It is not a small fix — the per-node `UserData` adaptation cache and the process-wide one both bake the timeout in, so making it engine-dependent would let one engine's budget reach another through a shared `Prepared<Script>` — so it is #3442 rather than part of this.

## Notes

- No public API change; all five baselines unchanged, and the `PublicApiTest` leg is green.
- `docs/v5-migration.md` **§4.41** (numbering assigned at authoring above the range in flight; renumber at merge). No §5 section — nothing new is exposed.
- One gotcha added to `Jint/Constraints/AGENTS.md` (13,671 bytes, well inside the 32 KiB budget), stating the rule any future construction path has to follow.
- The orphaned duplicate `<summary>` above `GetCaseFoldingFallbackEngine` is moved onto the member it describes.

## Verification

- `dotnet build -c Release` (solution): 0 errors, 0 compiler warnings.
- `Jint.Tests`: 10,754 passed / 0 failed / 5 skipped on net8.0 and net10.0; 7,378 / 0 / 4 on net472.
- `Jint.Tests.PublicInterface`: 3,099 / 3,109 / 2,480 passed, 0 failed on net8.0 / net10.0 / net472.
- `Jint.Tests.CommonScripts`: 28 / 28 passed on net10.0 and net472. `Jint.Tests.SourceGenerators`: 71 passed.
- **`Jint.Tests.Test262`: 102,495 passed / 0 failed / 189 skipped** — identical to the control.


lahma:
CI note: the first run had one red leg — `windows` / net8.0 — on `WebApiTimerTests.ATimerOutlivingTheUnwrapTimeoutTimesOutByDesign`, which asserts a 200 ms `UnwrapIfPromise` timeout against a 5 s `setTimeout`. It touches no regex at all, and the same suite passed on the net10.0 and net472 legs of that same job and on every other platform. A re-run of the failed job is green, so all seven legs now pass.

Worth flagging rather than burying, though: the failure was `Expected: <PromiseRejectedException> But was: null`, i.e. `UnwrapIfPromise(200 ms)` returned *without* throwing rather than overshooting its budget. That is not the usual loaded-runner overshoot — a 25x margin does not close from that direction — so it may be a real race in the unwrap-timeout path rather than a test that needs a wider bound. Not investigated here; noting it so it is on the record if it recurs.

PR #3454, opened 2026-08-27
Docs: the migration guide's numbering rule fails a test run instead of being stated

lahma:
The guide's **Keeping this document current** section already states the rule, and explains exactly why it is needed:

> **Numbering is assigned at merge, not at authoring.** Several pull requests are usually in flight at once, each picking what looked like the next free subsection number when it was written […] and because two of them edit different parts of this file, git merges them cleanly and the duplicate only shows up in the rendered document.

Nothing on the way in reports that. Not the compiler, not a reviewer reading a diff that shows one added section, not the merge itself. During the v5 campaign collisions have been caught by hand roughly **ten times**, and twice reached `main` — #3344 and #3347 exist only to clean up after them.

It happened again while I was writing this. #3443 was authored at §4.41; #3447 merged §4.41 first; rebasing #3443 onto that `main` succeeded **without a conflict** and produced a document with two §4.41s. A rule that is only prose does not survive that.

## The three assertions

- **`NoTwoSectionsShareANumber`** — the collision git merges silently.
- **`EverySectionSitsUnderItsOwnChapter`** — a section numbered `4.x` belongs under `## 4.`, because the chapter is what the number means; a renumbering that crosses chapters has to move the section too.
- **`SectionNumbersAscendInDocumentOrder`** — so resolving a collision cannot leave the new entry sitting above the one it was numbered after.

**Contiguity is deliberately not asserted.** Gaps are legitimate and constant during a campaign: a merged section sits at §4.37 while §4.36 is still in an open pull request. Asserting contiguity would fail on `main` today and would push authors toward reusing numbers, which is the defect.

## Verification

Each failure was injected into the real document and the matching assertion watched to fail, naming the offending line numbers and titles. The exact #3443/#3447 collision trips two of the three:

```
§4.41 is claimed by line 1794 ("A narrow date name is narrow, and a `timeStyle` day period is the locale's …")
   and line 1887 ("Two holes in the freeze, closed (#3360)")

line 1842: §4.35 follows §4.41 at line 1794
```

and a section moved across chapters:

```
line 1794: "### 5.9 A narrow date name is narrow, …" sits under "## 4."
```

Green on net472, net8.0 and net10.0 against the unmodified guide (3/3 each). Test-only; no production code, no baselines, no migration row.

### DIFF
--- a/Jint.Tests/MigrationGuideTests.cs
+++ b/Jint.Tests/MigrationGuideTests.cs
@@ -1,0 +1,136 @@
+#nullable enable
+
+using System.Globalization;
+using System.Text.RegularExpressions;
+
+namespace Jint.Tests;
+
+/// <summary>
+/// <c>docs/v5-migration.md</c> held to the numbering rule it states for itself.
+/// </summary>
+/// <remarks>
+/// <para>
+/// <b>Why a test.</b> The guide's own "Keeping this document current" section already says numbering is
+/// assigned at merge rather than at authoring, and explains exactly why: several pull requests are in flight
+/// at once, each picks what looked like the next free subsection when it was written, and because two of them
+/// append to different parts of one file <b>git merges them cleanly</b>. The duplicate exists only in the
+/// rendered document, so nothing on the way in reports it — not the compiler, not the reviewer reading a diff
+/// that shows one added section, not the merge. Stating the rule in prose was not enough: during the v5
+/// campaign collisions were caught by hand roughly ten times and twice reached <c>main</c>, needing
+/// <see href="https://github.com/sebastienros/jint/pull/3344">#3344</see> and
+/// <see href="https://github.com/sebastienros/jint/pull/3347">#3347</see> to clean up after them.
+/// </para>
+/// <para>
+/// <b>What it does not check.</b> Gaps are legitimate and common — a merged section can sit at 4.37 while
+/// 4.36 is still in an open pull request — so only duplication and ordering are asserted, never contiguity.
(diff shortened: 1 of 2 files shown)


lahma:
The three red legs were the test working, not the test being wrong.

Main today reads **4.35, 4.37, 4.40, 4.38, 4.36, 4.41**. No duplicates — that part held — but the numbers do not ascend, and a reader scanning for a section reads downwards.

The cause is mine, and worth recording because it will recur. Seven pull requests were in flight at once, and rather than renumbering each at merge as the guide instructs, I pre-assigned them distinct numbers 4.36–4.40 so they could not collide. That bought collision-freedom and paid for it in ordering: every section is appended at the end of chapter 4, so document order follows **merge** order, and the seven merged in a different order than they were numbered.

Fixed in place — heading text only, not the blocks:

| line | was | now | |
|---|---|---|---|
| 1887 | 4.40 | **4.38** | One list of calendars |
| 1920 | 4.38 | **4.39** | A host object's `Symbol.dispose` … |
| 1942 | 4.36 | **4.40** | An index on a host collection … |

Renumbering rather than moving the blocks is deliberate: six more pull requests are appending to the end of this same section right now, and moving text there would conflict with every one of them. Three headings conflict with none. The file's only intra-document anchor link points at 4.41, which does not move.

4.36 is now unused, which is fine — gaps are legitimate and the test does not assert contiguity, precisely so an open pull request can hold a number.

Green on net472, net8.0 and net10.0.

PR #3521, opened 2026-08-31
Interop: a JavaScript function converts to the delegate type the host asked for

### LINKED ISSUE #3434: Interop: the delegate a JS function converts to is cached process-wide without the target delegate type in the key
Found by the #3426 audit.

`DefaultTypeConverter` caches the CLR delegate it builds for a JS function in two process-wide
`ConditionalWeakTable`s:

```csharp
private static readonly ConditionalWeakTable<IFunction, Func<object, Delegate>> _targetBinderDelegateCache = new();
private static readonly ConditionalWeakTable<object, Delegate> _boundTargetDelegateCache = new();
```

`Jint/Runtime/Interop/DefaultTypeConverter.cs:104-105`, used at `:220-238`.

The outer key is the JS **function instance**; the inner key is that function's **AST node**. Neither
carries the one thing the value is built for — `type`, the CLR delegate type the conversion targets, which
`BuildTargetBinderDelegate(type, func)` bakes into the compiled binder.

So the second target type a given JS function is converted for is served the first one's delegate, and the
reflection invoke that follows rejects it:

```
System.ArgumentException : Object of type 'Notify' cannot be converted to type 'Transform'.
```

Three reachable shapes, in increasing order of surprise. The third is the #3426 family proper: the AST node
is process-wide state, and `Prepared<Script>` is documented as shareable across engines, so two engines with
different host types and different `Options` disagree according to which one evaluated first.

## Failing tests

Against unmodified `main` (`ec4b43506`), all three fail; `ControlEachHostAlone` passes.

```csharp
public class DelegateConversionTargetTypeTests
{
    public delegate void Notify(int value);
    public delegate string Transform(string value);

    public sealed class NotifyHost
    {
        public string Call(Notify f) { f(1); return "notify"; }
    }

    public sealed class TransformHost
    {
        public string Call(Transform f) => "transform:" + f("x");
    }

    public sealed class BothHost
    {
        public string A(Notify f) { f(1); return "notify"; }
        public string B(Transform f) => "transform:" + f("x");
    }

    /// <summary>Control: each host alone works. PASSES today.</summary>
    [Test]
    public void ControlEachHostAlone()
    {
        var one = new Engine();
        one.SetValue("host", new NotifyHost());
        Assert.That(one.Evaluate("host.call(function (x) { return String(x); })").AsString(), Is.EqualTo("notify"));

        var two = new Engine();
        two.SetValue("host", new TransformHost());
        Assert.That(two.Evaluate("host.call(function (x) { return String(x); })").AsString(), Is.EqualTo("transform:x"));
    }

    /// <summary>_boundTargetDelegateCache: one function instance, two delegate types, one engine.</summary>
    [Test]
    public void OneFunctionInstanceTwoDelegateTypes()
    {
        var engine = new Engine();
        engine.SetValue("host", new BothHost());

        var result = engine.Evaluate(@"
            var f = function (x) { return String(x); };
            host.a(f) + '|' + host.b(f);
        ").AsString();

        Assert.That(result, Is.EqualTo("notify|transform:x"));
    }

    /// <summary>_targetBinderDelegateCache: two instances of one AST node, two delegate types, one engine.</summary>
    [Test]
    public void TwoInstancesOfOneAstNodeTwoDelegateTypes()
    {
        var engine = new Engine();
        engine.SetValue("host", new BothHost());

        var result = engine.Evaluate(@"
            function make() { return function (x) { return String(x); }; }
            host.a(make()) + '|' + host.b(make());
        ").AsString();

        Assert.That(result, Is.EqualTo("notify|transform:x"));
    }

    /// <summary>
    /// The #3426 shape: two engines, one process, one shared Prepared&lt;Script&gt;. The AST node is the key
    /// of the process-wide binder cache, so whichever engine evaluates first decides for both.
    /// </summary>
    [Test]
    public void TwoEnginesSharingOnePreparedScript()
    {
        var prepared = Engine.PrepareScript("host.call(function (x) { return String(x); });");

        var first = new Engine();
        first.SetValue("host", new NotifyHost());
        Assert.That(first.Evaluate(prepared).AsString(), Is.EqualTo("notify"));

        var second = new Engine();
        second.SetValue("host", new TransformHost());
        Assert.That(second.Evaluate(prepared).AsString(), Is.EqualTo("transform:x"));
    }
}
```

```
Failed OneFunctionInstanceTwoDelegateTypes
Failed TwoInstancesOfOneAstNodeTwoDelegateTypes
Failed TwoEnginesSharingOnePreparedScript
  System.ArgumentException : Object of type '...+Notify' cannot be converted to type '...+Transform'.
```

## Why the cache exists, and what a fix has to keep

The comment above it says it: *"caching of .NET delegates per function instance is required to be able to
support unregistering event handlers"* — `host.on(f); host.off(f)` only works if both conversions produce
the **same** `Delegate` instance. Adding the target type to the key preserves that: it is still one delegate
per `(function instance, delegate type)`, which is exactly the identity `-=` needs.

Two shapes for the fix:

* a per-key map — `ConditionalWeakTable<object, TypeKeyedDelegates>` — which keeps the identity guarantee for
  every target type and costs a second lookup on the hit path;
* or the cheap guard — keep the single slot, and on a hit return it only when `type.IsInstanceOfType(d)`,
  building an uncached delegate otherwise. Zero extra allocation, one cast check per hit, the identity
  guarantee kept for the dominant one-type-per-function case, and the wrong-type throw gone in every case.
  The second target type loses the `-=` identity guarantee, which today throws instead.

The same argument applies one level up to `_targetBinderDelegateCache`, where the stored `Func<object, Delegate>`
would carry the `Type` it was built for and be compared by reference.

Both touch the interop conversion path, so which shape to take is a measurement question rather than an
obvious one — which is why this is an issue and not a pull request.

lahma:
Fixes #3434.

`DefaultTypeConverter` memoizes the CLR delegate a JavaScript function converts to, and it has to: `host.on(f)`
followed by `host.off(f)` only unsubscribes if both conversions hand the host the **same** `Delegate` instance.
The memo was keyed on the function *instance*, and one level below it on that function's *AST node*. Neither
carries the target delegate `Type`, which is the one thing `BuildTargetBinderDelegate(type, func)` bakes into
the compiled binder — so the second target type a given function was converted for was served the first one's
delegate, and the reflection invoke behind the call rejected it.

The premise still holds on today's `main` (`f6fa93bb7`).

## The three reproductions, as failing tests against unmodified `main`

`Jint.Tests/Runtime/DelegateConversionTargetTypeTests.cs` — four cases, of which `ControlEachHostAlone`
(each host alone, in its own engine) passes before and after and is there to say the conversion itself was
never broken:

```
Failed OneFunctionInstanceTwoDelegateTypes [14 ms]
   System.ArgumentException : Object of type 'Jint.Tests.Runtime.DelegateConversionTargetTypeTests+Notify'
   cannot be converted to type 'Jint.Tests.Runtime.DelegateConversionTargetTypeTests+Transform'.
Failed TwoInstancesOfOneAstNodeTwoDelegateTypes [4 ms]
   System.ArgumentException : Object of type '...+Notify' cannot be converted to type '...+Transform'.
Failed SameFunctionAndSameTargetTypeIsTheSameDelegateInstance [2 ms]
   System.ArgumentException : Object of type '...+Notify' cannot be converted to type '...+Transform'.
```

The first is the bound-delegate cache (one function instance, two delegate types, one engine); the second is
the binder cache below it (two instances of one AST node); the third is the reason the cache exists at all —
the same function converted twice to the same delegate type must be the same instance — asserted per target
type rather than for whichever type asked first.

And the sharp one, in `Jint.Tests.PublicInterface/HostDelegateConversionTests.cs`, because the AST node is
process-wide state and `Engine.PrepareScript` is documented as shareable across engines. It needs no second
delegate type in any one script: two engines, different host types, one cached `Prepared<Script>` — the shape
the README recommends for production — and whichever evaluated first decided for both.

```
Failed TwoEnginesSharingOnePreparedScript [2 ms]
   System.ArgumentException : Object of type '...HostDelegateConversionTests+Notify'
   cannot be converted to type '...HostDelegateConversionTests+Transform'.
Failed TwoEnginesSharingOnePreparedScriptInTheOtherOrder [1 ms]
   System.ArgumentException : Object of type '...+Transform' cannot be converted to type '...+Notify'.
Failed AlternatingEnginesOnOnePreparedScript [134 ms]
   System.ArgumentException : Object of type '...+Notify' cannot be converted to type '...+Transform'.
```

The other-order case is there so the fix cannot be a rule about which type wins, and the alternating one so it
cannot be a single slot that the last engine through overwrites.

## The key shape, and why this one

Three shapes were on the table. **A composite key is not available**: `ConditionalWeakTable` keys on reference
identity, so a `(function, Type)` key object would never match a later one and would root the function it
names — the weak-key semantics are the whole reason these are `ConditionalWeakTable`s.

**The #3501 value-keyed memo — store the `Type` the entry was built for and rebuild on a mismatch — is right
where the rebuild is cheap, and this rebuild is not.** There it re-adapts a regex through `RegExpParseCache`, a
dictionary lookup, so two engines alternating on one node is a non-event; here it is `Expression.Compile()` on
a freshly built expression tree. A single slot would recompile a binder on every alternation of the third
reproduction, which is exactly the shape this PR is about.

So: **a per-entry, type-keyed map, on both caches.** `TypeKeyedCache<T>` is an append-only singly-linked list
of `(Type, T)` nodes, read with a `Volatile.Read` and published with one `Interlocked.CompareExchange` — the
binder cache genuinely can be reached concurrently, since a shared `Prepared<Script>` is documented as
thread-safe across engines. Entries are added and never replaced, so the identity guarantee survives and is
merely finer-grained: one `Delegate` per *(function instance, delegate type)*, which is exactly the identity
`-=` needs, where one per function instance was only accidentally it.

### Cost shape on the hit path

No benchmark numbers here — this is a warm interop path and the machine is busy. **Gate tables to follow in a
comment before merge.** What the change does to the shape:

- **The lookup itself is unchanged in kind and one comparison longer.** Where a hit used to be one
  `ConditionalWeakTable` probe returning a `Delegate`, it is now the same probe returning the per-entry map
  plus a walk of a list that is **one node long** for the overwhelming majority of functions, which are
  converted to exactly one delegate type. That walk is one reference compare against the node's `Type`.
- **Two allocations leave the hit path.** `ConditionalWeakTable.GetValue(key, callback)` needs its callback
  constructed before the probe, and the old callback closed over `type`, `func` and `functionInstance` — a
  display class and a delegate allocated on *every* conversion, hit or miss. The callbacks are now cached
  static lambdas that capture nothing. The `_hostCallbackDelegates` registration beside it gets the same
  treatment: it probes first and only builds its closure when there is something to add.
- **On a miss** — once per *(function instance, delegate type)* rather than once per function instance —
  the extra cost is one small node object beside the `Expression.Compile()` that dominates it.
- **Memory.** The AST node holds one compiled binder per delegate type it is converted to instead of one in
  total, which is one for essentially every function and two for the deliberately mixed case. The function
  instance holds one delegate per target type, on the same terms.

### Lifetimes

Both tables stay `ConditionalWeakTable`, so entries still die with the node or the object they hang off, and
nothing new is rooted statically. What the AST node now carries — a `Type`, and a binder that takes its target
as a parameter — is engine-neutral, as `Jint/Runtime/Interpreter/AGENTS.md` requires of anything reachable
from a shared tree. `GarbageCollectionTests` gains
`SharedPreparedScriptConvertingFunctionsToDelegatesDoesNotRetainEngines`: twenty engines convert two functions
to two delegate types through one shared preparation, and every engine must be collectable while the
preparation stays rooted. It is a pin rather than a reproduction — it passes on `main` too, and would have to,
since the point is that the fix did not make the shared cache engine-affine. The existing
`SharedPreparedScriptDoesNotRetainEngines`, its parse-only siblings and `NestedTypeAccessDoesNotRetainEngines`
all pass (11/11 in that fixture).

## Test results

| Suite | Result |
| --- | --- |
| `Jint.Tests` net472 / net8.0 / net10.0 | 7788 / 11168 / 11168 passed, 0 failed |
| `Jint.Tests.PublicInterface` net472 / net8.0 / net10.0 | 2689 / 3312 / 3322 passed, 0 failed |
| `Jint.Tests.PublicInterface` net8.0, `JINT_HOST_CONTRACT_VERIFICATION=1` | 3325 passed, 0 failed |
| `Jint.Tests.CommonScripts`, `Jint.Tests.SourceGenerators` | 28 / 71 passed, 0 failed |
| `Jint.Tests.Test262` | **102,537 passed / 0 failed / 151 skipped** of 102,688 |

The suite total is the expected 102,688 and nothing fails. The skip count is four above the figure this branch
was measured against; all 151 skips resolve to entries in `Test262Harness.settings.json`, so they are
configuration rather than run-time timeouts, and neither that file nor the pinned SHA is touched here — the
difference is `main` drift since that reading, and this change reaches no test262 case, none of which converts
a JavaScript function to a CLR delegate.

`dotnet build -c Release` on the solution is clean with `TreatWarningsAsErrors`. No public signature changed,
so no API baselines were regenerated.

## Documentation

Migration guide: **§4.82** — the wrong-delegate `ArgumentException` disappearing is observable, even though
nothing that used to succeed changes. The shared-cache rule in `Jint/Runtime/Interpreter/AGENTS.md`, which
already says a shared interop cache may hold only engine-independent values, gains the other half of it: a
shared cache's key must carry everything its value was built for.


lahma:
Discharging the "gate tables to follow" promise with the measurement decision instead of tables, per the gating session's findings:

The gated suites cannot reach this lane. SunSpider and Dromaeo are pure-JS workloads; no case in either converts a JavaScript function to a CLR delegate, so a paired run would measure noise and report it as evidence. What can be said by inspection is stronger here: the hit path is **strictly cheaper** than before — the same `ConditionalWeakTable` probe, plus a walk of a list that is one node long for virtually every function, while the old `GetValue(key, callback)` shape allocated a display class and a delegate on **every** conversion, hit or miss, both now gone. The miss path pays one small node beside the `Expression.Compile()` that dominates it.

Merging on correctness (7 reproductions, both orders, shared-`Prepared<Script>` included) plus that inspection argument. The one benchmark-gated PR from this batch remains #3524, whose lane the suites *do* reach.

PR #3522, opened 2026-08-31
Interop: a converter also decides whether a declared member gets an indexer probe

### LINKED ISSUE #3436: Interop: a converter also decides whether a declared member gets an indexer probe, and IsConverterNeutral does not exclude that
Found by the #3426 audit. This is the one entry in that sweep that is the issue's own shape: a
process-wide cache serving one engine an entry another engine's `Options` produced.

## The claim that is one artefact short

`TypeResolver.IsConverterNeutral` decides whether a resolved accessor may be stored in, and served from,
the accessor cache that every engine on `TypeResolver.Default` shares. Its remark says:

> An `IndexerAccessor` bakes in the index key that the engine's converter produced from the member name …
> **That is the whole of the converter's influence on resolution**, which is why it is answered here rather
> than by partitioning the cache

and the implementation matches the claim exactly (`Jint/Runtime/Interop/TypeResolver.cs:384-395`):

```csharp
var filter = engine._typeConverterTargetFilter;
if (filter is null || accessor is not IndexerAccessor indexer)
{
    return true;
}
```

It is not the whole of it. `IndexerAccessor.TryFindIndexer` asks the engine's converter whether the member
*name* converts to the indexer's index type (`IndexerAccessor.cs:172`), and its answer also decides the
`indexerToTry` argument handed to the **declared-member** accessors a few lines later
(`TypeResolver.cs:877`, `:905`):

```csharp
var reflected = new PropertyAccessor(property, indexerToTry);
…
var reflected = new FieldAccessor(field, indexerToTry);
```

`ReflectionAccessor.GetValue` probes that indexer **before** the declared member — `// first check indexer so
we don't confuse inherited properties etc` (`ReflectionAccessor.cs:109`). So whether a `PropertyAccessor`
carries an indexer probe at all is converter-derived, and a `PropertyAccessor` is not an `IndexerAccessor`,
so it passes straight through the exclusion in both directions.

The divergence has one direction. The stock `DefaultTypeConverter` accepts `string → string` outright
(`type.IsInstanceOfType(value)`), so a stock engine always finds the indexer; an engine whose converter
declines does not. **A converter engine therefore poisons the shared cache for every stock engine that asks
afterwards** — and, being a cache, the outcome depends on which engine ran first rather than on anything
either host configured.

The trigger is a host `ClrTypeConverter` written directly against the public abstract base (rather than
derived from `DefaultTypeConverter`) that returns `false` for target types it does not handle. Both engines
below are otherwise default-configured, so their `InteropResolutionProfile`s are equal and both use
`TypeResolver.Default`.

## Failing test

Against unmodified `main` (`ec4b43506`) the two controls pass and the third fails.

```csharp
public class ConverterIndexerProbeCacheTests
{
    /// <summary>Declines every conversion, including string -> string, which the stock converter accepts.</summary>
    private sealed class NarrowConverter : ClrTypeConverter
    {
        public override object? Convert(object? value, Type type, IFormatProvider formatProvider)
            => throw new NotSupportedException();

        public override bool TryConvert(
            object? value,
            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors | DynamicallyAccessedMemberTypes.PublicFields)] Type type,
            IFormatProvider formatProvider,
            [NotNullWhen(true)] out object? converted)
        {
            converted = null;
            return false;
        }
    }

    // One host type per test: the cache is process-wide and never evicts, so sharing one type between
    // these three tests would make them order-dependent on each other.
    public sealed class BagA { /* as BagC */ }
    public sealed class BagB { /* as BagC */ }

    public sealed class BagC
    {
        private readonly Dictionary<string, string> _entries = new() { ["Name"] = "from-indexer" };
        public string Name => "from-property";
        public string? this[string key] => _entries.TryGetValue(key, out var v) ? v : null;
    }

    /// <summary>Baseline: a stock engine alone reads the indexer, which is probed before the member. PASSES.</summary>
    [Test]
    public void StockEngineAloneReadsTheIndexer()
    {
        var engine = new Engine();
        engine.SetValue("bag", new BagA());
        Assert.That(engine.Evaluate("bag.Name").AsString(), Is.EqualTo("from-indexer"));
    }

    /// <summary>Baseline: a declining converter loses the indexer probe, so the member answers. PASSES.</summary>
    [Test]
    public void ConverterEngineAloneReadsTheProperty()
    {
        var engine = new Engine(o => o.SetTypeConverter(_ => new NarrowConverter()));
        engine.SetValue("bag", new BagB());
        Assert.That(engine.Evaluate("bag.Name").AsString(), Is.EqualTo("from-property"));
    }

    /// <summary>
    /// The converter engine warms the shared accessor cache first, and the stock engine is then served an
    /// entry whose indexer probe the converter decided away. FAILS.
    /// </summary>
    [Test]
    public void ConverterEngineFirstThenStockEngine()
    {
        var withConverter = new Engine(o => o.SetTypeConverter(_ => new NarrowConverter()));
        withConverter.SetValue("bag", new BagC());
        Assert.That(withConverter.Evaluate("bag.Name").AsString(), Is.EqualTo("from-property"));

        var stock = new Engine();
        stock.SetValue("bag", new BagC());
        Assert.That(stock.Evaluate("bag.Name").AsString(), Is.EqualTo("from-indexer"));
    }
}
```

```
Failed ConverterEngineFirstThenStockEngine
  Assert.That(stock.Evaluate("bag.Name").AsString(), Is.EqualTo("from-indexer"))
  Expected: "from-indexer"
  But was:  "from-property"
```

Swap the two blocks and `stock` reads `"from-indexer"` — order-dependence is the tell.

## A second, milder consequence of the same fact

`maySubstituteGenerated` is gated on `indexerToTry is null` (`TypeResolver.cs:873`), so the same converter
answer also decides whether a `[JsAccessible]` type gets its generated accessor or the reflected one, and
that decision is likewise cached process-wide. Per `Jint/Runtime/Interop/AGENTS.md` the two lanes are
behaviourally equivalent, so this one costs a fast lane rather than a wrong answer — but it is the same
missing exclusion.

## Fix shape

Mirror what is already there rather than partition the cache: widen `IsConverterNeutral` so it also declines
an accessor whose indexer probe the converter could have decided — i.e. ask the target filter about
`indexerToTry`'s index type, not only about `IndexerAccessor.FirstIndexParameter`. That keeps the
"one artefact, excluded on the way in and on the way out" design and costs a converter engine only the
members whose type actually carries a string-keyed indexer.

The remark on `IsConverterNeutral` and the corresponding paragraph in
`Jint/Runtime/Interop/AGENTS.md` ("the only resolution artefact it steers is an `IndexerAccessor`'s baked-in
key") both need the same correction in whatever pull request closes this.

lahma:
`TypeResolver.IsConverterNeutral` decides whether a resolved accessor may enter, and be served from, the
accessor cache every engine on `TypeResolver.Default` shares. Its remark claimed that an `IndexerAccessor`'s
baked-in index key is "the whole of the converter's influence on resolution", and the implementation matched
the claim exactly. It is one artefact short.

`IndexerAccessor.TryFindIndexer` asks the engine's `ClrTypeConverter` whether the member *name* converts to an
indexer's index type, and that **one answer decides four things**:

1. whether an `IndexerAccessor` exists at all, and what index key it bakes in — the one that was excluded;
2. whether the declared `PropertyAccessor`/`FieldAccessor` is handed an `indexerToTry`, which
   `ReflectionAccessor.GetValue` probes **before** the declared member (`// first check indexer so we don't
   confuse inherited properties etc`);
3. whether a `[JsAccessible]` type may take its generated lane, gated on `indexerToTry is null`;
4. whether resolution ends at `ConstantValueAccessor.NullAccessor` rather than at the indexer.

Only the first is visible in the resolved artefact, and a `PropertyAccessor` is not an `IndexerAccessor`, so
the other three passed straight through the exclusion — **in both directions**.

## Failing tests, against unmodified `main` (`f6fa93bb7`)

`Jint.Tests.PublicInterface/ConverterIndexerProbeCacheTests.cs` — four controls pass and all three poisoning
tests fail. The host type carries a declared `Name` property *and* a string-keyed indexer answering the same
name differently; the converter declines every conversion, including the `string -> string` the stock one
accepts outright.

```
Failed AConverterEngineDoesNotDecideForAStockEngine
  Expected string to be the same string, but they differ at index 5:
       ↓ (actual)
  "from-property"
  "from-indexer"
       ↑ (expected).

Failed AStockEngineDoesNotDecideForAConverterEngine
  Expected string to be the same string, but they differ at index 5:
       ↓ (actual)
  "from-indexer"
  "from-property"
       ↑ (expected).
```

Same host type, same member name, and the answer depended on which engine ran first. A third case covers
path (4), where the indexer is an *explicitly implemented* `IEntries.this[string]` and there is no declared
member at all, so the converter decides between the indexer and "no such member" — a `ConstantValueAccessor`,
which is not an `IndexerAccessor` either:

```
Failed AConverterEngineDoesNotDecideForAStockEngineThroughAnInterface
  Expected string to be the same string, but they differ at index 0:
  ↓ (actual)
  ""
  "from-indexer"
  ↑ (expected).
```

The controls (`StockEngineAloneReadsTheIndexer` → `"from-indexer"`,
`ConverterEngineAloneReadsTheProperty` → `"from-property"`,
`StockEngineAloneReadsTheInterfaceIndexer` → `"from-indexer"`, `ConverterEngineAloneFindsNoMember` →
`undefined`) pass before and after, so the fix changes no answer an engine gives on its own.

## The fix

Ask the question of the **type** instead of the accessor. An entry is converter-neutral when the engine's
converter is not consulted about any index key type the type could offer: the index parameter types of its own
and its interfaces' single-parameter indexers, minus `int`, which `ComposeIndexerFactory` keys without asking
anyone. The interface half of that walk is what covers path (4) — an explicit implementation is reported by
the interface alone. Both sides of the cache ask it, exactly as before — an entry a stock engine stored must not be served
to an engine whose converter would have answered differently, nor the reverse.

That set is a property of the type alone, so it is memoized per resolver (not per process: it keeps reflected
types alive on exactly the terms the accessor cache already does) and reached only by engines that installed a
converter at all. It is computed without `MemberFilter`, which the probe itself applies — a superset, and the
exclusion it feeds errs toward not sharing.

`GetStaticAccessor` drops the converter check entirely. `ResolveStaticAccessor` goes through
`TryFindMemberAccessor` alone, passing no indexer to try, and that method never probes one, so the static lane
never consults the converter; the nested-type exclusion is inlined there and is all that lane needs.

## Cost shape

- **Stock engine, hit path: unchanged.** `IsConverterNeutral` still returns on `engine._typeConverterTargetFilter is null` — one field read and a branch, on the read side as on the write one. No wider key, no new field on the hit path, no allocation.
- **Converter-owning engine, hit path:** one memoized `ConcurrentDictionary<Type, Type[]>` lookup plus a walk of that array (empty for any type with no non-`int` indexer, one element for the ordinary string-keyed one), in place of the previous `is IndexerAccessor` type test and single `Claims` call.
- **What such an engine loses:** its share of the cache entry for *every* member of a type whose index key type its converter claims, rather than only for that type's indexer accessors. `SetTypeConverter(f, targetTypes)` / `ClrTypeConverter.HandledTargetTypes` narrows it exactly as before — declaring `TimeSpan` still costs nothing on a `string`-keyed dictionary — and that population already re-resolved the indexer lane on every access, which is the same types.
- **Cold path:** one reflection scan of a type's (and its interfaces') properties, once per resolver per type, and only for converter-owning engines.

gate tables to follow in a comment before merge

## Also in this PR

- `Jint.Tests/Runtime/InteropAccessorCacheSharingTests.cs` gains
  `ADeclaredMemberOfAnIndexerCarryingTypeIsConverterDerivedToo`, which pins the sharing decision through the
  counting `MemberFilter` engagement probe rather than through a value; the existing
  `ACustomTypeConverterOnlyCostsTheIndexerAccessorsItCouldHaveKeyedDifferently` and
  `AnEngineWithACustomTypeConverterStillSharesOrdinaryMembers` pass unchanged.
- The claim is corrected everywhere it was stated: `IsConverterNeutral`'s remark, `InteropResolutionProfile`'s,
  the comment at the profile's capture in `Engine.cs`, `Jint/Runtime/Interop/AGENTS.md` (the
  `SetTypeConverter(f)` row and the paragraph under it) and `Jint/Runtime/Interpreter/AGENTS.md`'s shared-cache
  rule.
- `docs/v5-migration.md` §4.85.

## Verification

`dotnet build -c Release` clean (`TreatWarningsAsErrors`). `Jint.Tests` and `Jint.Tests.PublicInterface` green
on net472 / net8.0 / net10.0, and both green again with `JINT_HOST_CONTRACT_VERIFICATION=1`. test262
**102,537 / 0 / 151** of 102,688, byte-identical to a control run of pristine `f6fa93bb7` in a second worktree
(the 102,541 / 0 / 147 figure predates main's recent staging-exclusion changes).

Fixes #3436


lahma:
Discharging the "gate tables to follow" promise with the measurement decision instead of tables:

The gated suites cannot reach this lane. SunSpider and Dromaeo install no `ClrTypeConverter` and project no host objects, so neither side of the changed code executes in them — a paired run would measure noise. By inspection, the stock-engine hit path is **unchanged** (`filter is null` → one field read, as before), and only a converter-owning engine pays the memoized per-type lookup, on a lane that population already re-resolves. Merging on the correctness evidence (both directions, both orders, the interface-only third case).

Issue #3407, opened 2026-08-26
Interop: a plain type mismatch scores 100 instead of being discarded, so 'text' + an overloaded CLR value throws


lahma:
### Version

main (a487ee067)

### Summary

`InteropHelper.CalculateMethodParameterScore` has no "impossible" verdict for a plain type mismatch. Every earlier branch returns a specific score, and the method ends:

```csharp
        // will rarely succeed
        return 100;
```

`FindBestMatch` discards a candidate only when its score is **negative** (`if (score < 0)`), and `AssignableResult.IsAssignable` is `Score >= 0`. So **100 is an accept.** A parameter that simply cannot take the argument still scores 100, the candidate survives, and when it is the only one `FirstOrDefault().Method` hands it straight to `method.Call(...)`.

### What an embedder sees

With `Interop.AllowOperatorOverloading` on, and any CLR type whose only `op_Addition` is `(T, T)`:

```csharp
private struct Vector2D
{
    public double X { get; set; }
    public double Y { get; set; }
    public Vector2D(double x, double y) { X = x; Y = y; }
    public static Vector2D operator +(Vector2D lhs, Vector2D rhs) => new(lhs.X + rhs.X, lhs.Y + rhs.Y);
    public override string ToString() => $"({X}, {Y})";
}

var engine = new Engine(cfg => cfg.Interop.AllowOperatorOverloading = true);
engine.SetValue("Vector2D", TypeReference.CreateTypeReference<Vector2D>(engine));
engine.Evaluate("let v1 = new Vector2D(1, 2);");

engine.Evaluate("'### ' + v1 + ' ###'");   // expected "### (1, 2) ###"
```

`JsString("### ")` is scored against parameter type `Vector2D`: not a `JsValue`, not `object`, not numeric, not an enum, not assignable, not a spec array, not a generic parameter; `CanChangeType` fails, there is no `op_Implicit`/`op_Explicit`, and `TryConvertViaTypeCoercion` fails. It falls through to `return 100`, the operator is selected, and converting the argument throws `InvalidCastException: Invalid cast from 'System.String' to 'Vector2D'`.

It is not confined to string concatenation. All of these throw instead of performing the ordinary JavaScript addition each is specified to perform:

| script | today | expected |
| --- | --- | --- |
| `'### ' + v1 + ' ###'` | throws | `"### (1, 2) ###"` |
| `'### ' + v1` | throws | `"### (1, 2)"` |
[...]

PR #3526, opened 2026-08-31
Interop: an operator-overload resolution is not shared by engines that would not reach it

### LINKED ISSUE #3424: Operator-overload resolution is cached process-wide, but the resolution depends on the engine
`JintBinaryExpression` caches operator-overload resolution in a **process-wide** dictionary:

```csharp
private readonly record struct OperatorKey(string? OperatorName, Type Left, Type Right);
private static readonly ConcurrentDictionary<OperatorKey, MethodDescriptor> _knownOperators = new();
```

The key is the operator name and the two operand CLR types. Nothing about the engine is in it — but the
value is produced by `InteropHelper.FindBestMatch(context.Engine, …)`, and that **does** depend on the
engine:

* `Options.Interop.ValueCoercion` steers `CalculateMethodParameterScore`'s `TryConvertViaTypeCoercion` rule
  (score 10 vs. falling through);
* since #3407, the installed `ClrTypeConverter` decides the last rule outright — a conversion it confirms
  scores 100, one it refuses scores −1, and −1 discards the candidate.

So two engines in one process, over the same host types, can disagree about which operator applies, and
whichever engine evaluates the expression **first** decides for both. Concretely, with a host type whose
only `op_Addition` is `(T, T)`:

| order | engine A (custom `ClrTypeConverter` that converts `string` → `T`) | engine B (default converter) |
| --- | --- | --- |
| A evaluates `'s' + v` first | operator call — correct for A | cache hit: operator call, then the conversion throws |
| B evaluates `'s' + v` first | cache hit: `null`, so string concatenation — **not** what A's converter asked for | concatenation — correct for B |

The bottom-left cell is the one #3407 introduces; the top-right one is what an engine gets today either way.
The `ValueCoercion` half predates both and can reorder *which* overload wins among several.

This needs two engines in one process with different `Options.Interop` conversion configuration and the same
host operator types, so it is narrow — but it is silent when it happens, and the answer depends on evaluation
order rather than on anything the host configured.

Fixing it means putting whatever the resolution actually depends on into the key (the converter instance and
`ValueCoercion`, or an interop-resolution identity like the one `TypeResolver` already partitions on), or
moving the cache off `static`. Either touches a lane on the hot path for every overloaded operator, so it
wants a measurement rather than a patch — which is why this is an issue and not a pull request.

Found while working on #3407; not a regression of it in the case that matters most (an engine getting a
*wrong* operator), only in the mirror case.

lahma:
`JintBinaryExpression._knownOperators` remembers which CLR operator a `(operator name, left CLR type, right
CLR type)` triple selects, in a **process-wide** dictionary. The selection runs `InteropHelper.FindBestMatch`,
which reads two things the embedder configures — and neither was in that key:

* `Options.Interop.ValueCoercion`, which `CalculateMethodParameterScore`'s gray-zone rule consults;
* the installed `ClrTypeConverter`, whose answer is the last rule outright since #3407 — a conversion it
  confirms keeps the candidate, one it refuses scores −1, which discards it.

So two engines in one process, over the same host types, could disagree about which operator applies, and
whichever evaluated first decided for both.

## Failing tests, against unmodified `main` (`f6fa93bb7`)

`Jint.Tests.PublicInterface/OperatorOverloadResolutionCacheTests.cs`. Eight tests, one host type each (the
table never evicts): four controls, which pass before and after, and four poisoning tests, which fail on
`main` — **both inputs, both orders, both directions**.

The converter half uses a host type whose only `+` is `(T, T)`, so `'s' + v` asks whether a string can become
a `T`, which only the converter can answer:

```
Failed AConverterEngineDoesNotDecideForAStockEngine
  Expected Add(StockEngine(), new MoneyC()) not to be "operator".

Failed AStockEngineDoesNotDecideForAConverterEngine
  Expected Add(ConverterEngine(), new MoneyD()) to be "operator", but was
  "sJint.Tests.PublicInterface.OperatorOverloadResolut…".
```

The coercion half needs no converter at all — two stock engines differing only in `ValueCoercion`, over a type
whose only `+` takes a `string` on the right, handed an opaque host object:

```
Failed ACoercingEngineDoesNotDecideForANonCoercingOne
  Expected AddOpaque(NonCoercingEngine(), new CoercedC()) not to be "operator".

Failed ANonCoercingEngineDoesNotDecideForACoercingOne
  Expected AddOpaque(CoercingEngine(), new CoercedD()) to be "operator", but was
  "Jint.Tests.PublicInterface.OperatorOverloadResolut…".
```

The four controls — each engine alone, in its own process-lifetime host type — pass throughout, so no answer
an engine reaches on its own changes.

## The mechanism, and why it is two different ones

The issue offered three shapes: put the inputs in the key, partition by `InteropResolutionProfile`, or an
`IsShareable`-style exclusion. Reading what the resolution actually consumes gives a different answer per
input, and **(b) is out for both**: `InteropResolutionProfile` carries binding flags, `AllowGetType` and the
extension-method lookup — none of which operator resolution reads — and deliberately carries neither
`ValueCoercion` nor the converter.

**`ValueCoercion` goes in the key (a).** It is a value: putting it there costs one enum field in the struct's
hash and equality, retains nothing, and lets every engine that coerces alike go on sharing its entries. It is
snapshotted onto the engine at construction beside `_operatorOverloadingAllowed`, the same way and for the
same reason.

**The converter selects the table (c), and must not go in the key.** `SetTypeConverter` hands its factory the
`Engine`, so a host converter routinely captures one; keying a process-lived static dictionary on the
converter instance — or on its `TypeConverterTargetFilter`, which holds it — would pin that converter and that
engine for the life of the process, which is exactly the retention `GarbageCollectionTests` exists to refuse.
An `IsConverterNeutral`-style exclusion does not fit either: unlike the accessor cache, where one artefact
carries the converter's influence, here the converter reorders and discards candidates, and what is stored is
a winner rather than an artefact.

So: an engine running the stock `DefaultTypeConverter` — `engine._typeConverterTargetFilter is null`, the test
every other interop lane already gates on — resolves what every other stock engine would, and uses the static
table. An engine with a converter of its own uses `Engine._engineOperatorOverloads`, created on first use,
cleared by `InstallTypeConverter` when the converter is replaced, and collected with the engine.

`JintUnaryExpression._knownOperators` is **clean and stays process-wide**: it takes the operand type's first
one-parameter method of that name and computes no score, so it consults neither input. It now carries a
comment saying so, and saying that anything giving it a scoring step inherits the binary lane's rules.

## Cost shape

- **Hit path:** one `ConcurrentDictionary.TryGetValue` as before, against a key one `ValueCoercionType` field wider (one more `int` mixed into the generated hash and one more compare in equality), preceded by one field read on the engine to choose the table. No allocation, no lock, no wider indirection.
- **What the hit path *loses*:** the `GetOrAdd` it replaces took a lambda capturing `clrName`, both `Type`s, `arguments` and `context`, so a display class was allocated on **every** evaluation, hit or miss. `TryGetValue` + a static resolve method allocates none.
- **Converter-owning engines:** one extra null-coalescing check to reach their own table, and each `(operator, left, right)` triple resolved once per engine rather than once per process. They keep a real cache — this is not a fall-through to re-reflecting per evaluation.
- **Cold path:** unchanged reflection and `MethodDescriptor.Build`, moved into a named method.

gate tables to follow in a comment before merge

## Also in this PR

- `Jint.Tests/Runtime/OperatorOverloadResolutionCacheTests.cs` pins the mechanism rather than the answers —
  which table each engine uses, that the engine-local one really caches (a second evaluation adds no entry),
  and that replacing the converter drops what the old one decided. "Both engines were right" is also what a
  cache that was never consulted looks like, so the public-interface tests alone would not have said this.
- `Jint/Runtime/Interpreter/AGENTS.md` gains the rule beside the shared-accessor-cache one it is the
  interpreter's copy of.
- `docs/v5-migration.md` §4.86.

## Verification

`dotnet build -c Release` clean (`TreatWarningsAsErrors`). `Jint.Tests` and `Jint.Tests.PublicInterface` green
on net472 / net8.0 / net10.0, and both green again with `JINT_HOST_CONTRACT_VERIFICATION=1`. test262
**102,537 / 0 / 151** of 102,688 — the run reported two `intl402/supportedLocalesOf-unicode-extensions-ignored`
failures at 32 s each, the known ~30 s timeout-under-load flake; both pass in 4 s in isolation. That figure is
byte-identical to a control run of pristine `f6fa93bb7` in a separate worktree (the 102,541 / 0 / 147 figure
predates main's recent staging-exclusion changes).

Fixes #3424


lahma:
Discharging the "gate tables to follow" promise with the measurement decision instead of tables:

The gated suites cannot reach this lane. `_knownOperators` is consulted only when operator-overload resolution runs, which requires CLR operands — SunSpider and Dromaeo are pure-JS and never enter it, so a paired run would measure noise. By inspection the hit path gained one field read to pick the table and one enum in the key's hash, and **lost** a per-evaluation display-class allocation (`GetOrAdd` with a capturing lambda → `TryGetValue`). Merging on the correctness evidence (4 poisoning tests, both inputs, both orders, both directions).

Issue #3577, opened 2026-09-01
Overload scoring rates `int` and `long` parameters with no range check, so an out-of-range number is a perfect match


lahma:
`InteropHelper.CalculateMethodParameterScore` rates a JavaScript number against a numeric parameter type by
*magnitude*, and every narrowing type it recognizes carries a range check — except the two that do not:

```csharp
var isInteger = num.IsInteger() || TypeConverter.IsIntegralNumber(num._value);
if (isInteger)
{
    if (paramType == typeof(int))
    {
        return 0;                                    // <- no range check
    }

    if (paramType == typeof(long))
    {
        return ScoreForDifferentTypeButFittingNumberRange;   // <- no range check
    }

    // check if we can narrow without exception throwing versions (CanChangeType)
    var integerValue = (int) num._value;
    if (paramType == typeof(short) && integerValue is <= short.MaxValue and >= short.MinValue)
    ...
```

`float`, `short`, `ushort`, `byte` and `sbyte` all gate on the value fitting. `int` scores a **perfect 0** and
`long` scores 2 for any integral double whatever its magnitude, so `3000000000` is a perfect match for an
`int` parameter and `1e300` is a good match for a `long` one. The conversion that follows then throws.

`FindBestMatch` returns on a perfect score without examining the remaining candidates, so for `int` the
better-fitting overload beside it is never even scored.

## Reproduction

Measured on `main` @ ef706d721, one process per block.

**Operator lane — a CLR `OverflowException` out of `Evaluate`.** The operator path takes the first match and
calls it, with no retry, so the failed conversion is what the host sees:

```csharp
public sealed class IntAndObject
{
    public static string operator +(IntAndObject left, int right)    => "int:" + right;
    public static string operator +(IntAndObject left, object right) => "object:" + right;
}

var engine = new Engine(o => o.Interop.AllowOperatorOverloading = true);
engine.SetValue("m", new IntAndObject());

engine.Evaluate("m + 5");            // "int:5"
engine.Evaluate("m + 3000000000");   // OverflowException: Value was either too large or too small for an Int32.
```

`long` fails the same way one octave up — `m + 1e300` against `operator +(T, long)` beside `operator +(T, object)`
is `OverflowException: Arithmetic operation resulted in an overflow.`

**Method lane — a spurious "no such method".** Here the perfect-score early return is what does the damage:
the `object` overload is not in the returned set, so there is nothing for `MethodInfoFunction.TryCall` to fall
back to.

```csharp
[...]

PR #3578, opened 2026-09-01
Interop: an operator overload is chosen by the arguments in hand

### LINKED ISSUE #3567: The operator-overload cache still lets one argument value decide another's overload
#3526 closed #3424 by putting `Options.Interop.ValueCoercion` into `OperatorKey` and giving an engine with its own `ClrTypeConverter` a table of its own. Both of those were real and both are fixed. What the key still omits is the thing selection *also* depends on: the **argument values**.

```csharp
internal readonly record struct OperatorKey(string? OperatorName, Type Left, Type Right, ValueCoercionType Coercion);
```

`ResolveOperatorOverload` hands the actual arguments to `InteropHelper.FindBestMatch`, and `CalculateMethodParameterScore` scores several parameter types by *magnitude* rather than by type:

```csharp
if (paramType == typeof(byte) && integerValue is <= byte.MaxValue and >= byte.MinValue)
{
    return ScoreForDifferentTypeButFittingNumberRange;
}
```

Every JavaScript number reaches the key as `System.Double` — `(5).ToObject()`, `(300).ToObject()` and `(5.5).ToObject()` are all `System.Double` — so two numbers that select different overloads share one key, and the first evaluation in the process decides for every one after it.

## Reproduction

```csharp
public sealed class Money
{
    public string Last = "";
    public static Money operator +(Money m, byte v)   { m.Last = "byte:" + v;   return m; }
    public static Money operator +(Money m, object v) { m.Last = "object:" + v; return m; }
}

// a fresh Engine per call, both with the stock converter and identical options
static string Run(string script)
{
    var engine = new Engine(o => { o.Interop.AllowOperatorOverloading = true; o.AllowClr(typeof(Money).Assembly); });
    var money = new Money();
    engine.SetValue("m", money);
    try { engine.Execute(script); } catch (Exception e) { return e.GetType().Name + ": " + e.Message; }
    return money.Last;
}
```

Measured, `main` @ 55a45cec5, one process, distinct host types so the two orders do not share a key:

```
small value first        large value first
  m + 5   -> byte:5        m + 300 -> object:300
  m + 300 -> OverflowException:    m + 5   -> object:5
             Value was either too
             large or too small
             for an unsigned byte.
```

Read the columns as the whole defect. In the first order, `m + 300` is resolved to `operator +(Money, byte)` because `m + 5` got there first, and the conversion of 300 to `byte` throws `OverflowException` **out of `Execute`** — a host exception for an expression that works fine when it runs first. In the second order, `m + 5` is resolved to `operator +(Money, object)` because `m + 300` got there first — no exception, just the wrong overload, silently.

Both engines are fresh, both are configured identically, and both use the stock converter, so this is `_knownOperators` — the process-wide table — not the per-engine one #3526 added. The same thing happens inside a single long-lived engine; separate engines just make it obvious that the state outlives them.

## Why the key cannot simply be widened

The argument values are not a bounded key. The fix is the other direction: cache the **candidate set** — `MethodDescriptor.Build(...)` over the two types' operator methods, which is a pure function of `(clrName, leftType, rightType)` and carries no engine and no values — and run `FindBestMatch` per evaluation. `ValueCoercion` then drops out of the key too, since it only ever fed the scoring.

That moves overload selection onto every operator evaluation, which is why it wants a measurement rather than a patch. Worth knowing before taking one: `TryOperatorOverloading` runs only when `AllowOperatorOverloading` is on, and arming it already disables six fast paths in `JintBinaryExpression` (`OperatorOverloadingAllowed` is consulted before any of them engages), so this is not on the default hot path at all — the affected population is engines that opted in, which are already paying `ToObject()` on both operands per evaluation. `Jint.Benchmark` with `JINT_BENCH_MODE=gate`, paired before/after, on a lane that actually has operator overloading armed.

## Note on the unary sibling

`JintUnaryExpression._knownOperators` looks like the same defect and is not: its factory takes the first one-parameter method with the matching name, so `(clrName, operandType)` fully determines the answer — no `FindBestMatch`, no engine, no values. Recorded so it does not get "fixed" alongside.

Found while auditing process-wide caches for #3426.

lahma:
`JintBinaryExpression` cached the **selected** `MethodDescriptor` under
`(operator name, left CLR type, right CLR type, ValueCoercion)`. Selection is
`InteropHelper.FindBestMatch`, and several of its scoring rules read the argument's *value* rather than its
type — a number is a perfect fit for a `byte` parameter inside that range and no match at all outside it.
Every JavaScript number reaches that key as `System.Double`, so `5` and `300` shared one entry and whichever
arrived first decided for the other.

Reproduced on `main` @ ef706d721, in a single process, with a distinct host type per column so the two orders
do not share a key:

| small value first | large value first |
| --- | --- |
| `m + 5` → `byte:5` | `m + 300` → `object:300` |
| `m + 300` → **`OverflowException`** out of `Evaluate` | `m + 5` → **`object:5`**, silently wrong |

Both engines are fresh and identically configured, so this is the process-wide table; the same collision
happens inside one long-lived engine, which is why #3526's per-engine table did not close it.

## The fix

What is cached is the **candidate set** — the two-parameter operator methods the two operand types declare
under that name. That is a reflection scan and nothing else: no engine, no configuration and no argument
value takes part in building one, so one process-wide table serves everybody. Choosing between them runs on
every evaluation.

That also retires the two embedder inputs #3526 had to key on. `Options.Interop.ValueCoercion` and the
installed `ClrTypeConverter` are both read while *scoring*, so with scoring per evaluation neither is in a
shared table any more: `OperatorKey` loses its `Coercion` component, and `Engine._engineOperatorOverloads`
(and the `InstallTypeConverter` clear that went with it) is gone. The public-interface tests those two shapes
were added with are unchanged and still pass — each engine now reaches its own answer by computing it rather
than by having a table of its own.

## Cost, and the three things that hold it down

Selection moving onto every operator evaluation is the trade. What is left on that path is scoring one or two
candidates; the reflection scan the table exists for is unchanged and still happens once per triple.

* A pair with **no operator at all** — the common case for a binary expression over host objects — leaves
  through the empty candidate set, and now without allocating the argument array either, which the old code
  built before the lookup.
* A candidate found on **both** operand types is listed once instead of twice. `T + T` put the same
  `MethodInfo` in the set twice (so did a derived operand whose `FlattenHierarchy` scan reports the base's
  operator), which cost a second scoring pass per evaluation and, being a second survivor, cost the
  single-candidate lane its allocation-free path.
* `FindBestMatch` holds the first surviving candidate in a local and builds its `List<MethodMatch>` only when
  a **second** one survives. A single-candidate resolution now allocates nothing at any score, where before
  only a *perfect* score avoided the list. That is a general improvement — `MethodInfoFunction` and
  `TypeReference`'s constructor selection take it too — and it is behaviour-neutral: a one-element `Sorted`
  and a `Single` are the same match to `FirstOrDefault` and to the enumerator.

None of this is on a default engine's path at all: `TryOperatorOverloading` runs only under
`Options.Interop.AllowOperatorOverloading`, and arming that already disables six fast paths in
`JintBinaryExpression` and pays `ToObject()` on both operands per evaluation. **I have not benchmarked it** —
a paired `JINT_BENCH_MODE=gate` run on a lane with operator overloading armed is the measurement this wants
before or after merge.

## Tests

`Jint.Tests.PublicInterface/OperatorOverloadResolutionCacheTests.cs` gains a `value-decided resolution`
region: each order alone, each order after the other, and the two orders inside one engine. Against unfixed
code, 3 of those 5 fail on every target framework — `net472`, `net8.0` and `net10.0`, 3 failed / 13 total per
leg — with exactly the two symptoms above.
`Jint.Tests/Runtime/OperatorOverloadResolutionCacheTests.cs` is rewritten to pin the mechanism: what the
table holds is the candidate set and not the selection, a candidate is listed once, a pair with no operator
is remembered as having none, and two differently-configured engines share one entry rather than one of them
keeping a table of its own.

Full `dotnet build -c Release` and `dotnet test -c Release` are green, including `Jint.Tests.PublicInterface`;
test262 holds at 102,537 passed / 0 failed / 151 skipped of 102,688.

`docs/v5-migration.md` gains §4.99, and §4.86's mechanism paragraphs — which described the key and the table
split this replaces — are repointed at it.

## Not fixed here

The `OverflowException` has a second, unrelated way out, and it is **not** this cache:
`CalculateMethodParameterScore` scores an `int` parameter a perfect 0 and a `long` parameter a 2 for any
integral double *whatever its magnitude*, where `float`, `short`, `ushort`, `byte` and `sbyte` all check the
range. So `m + 3000000000` against `operator +(T, int)` throws deterministically, in a fresh process, with the
`object` overload sitting unexamined beside it. Filed separately as #3577, since fixing it changes overload
*selection* rather than only an error.

Closes #3567


lahma:
On the flagged measurement: the standing SunSpider/Dromaeo gates cannot reach this lane — both suites run pure JS on default-configured engines, and no row arms `AllowOperatorOverloading`, so a paired run would measure zero executions of the changed code. The per-evaluation cost argument therefore stands on the code itself: the empty-set exit is earlier and allocation-free than before, the single-survivor path (the common case) allocates nothing where the old code allocated a list for any imperfect score, and scoring work per evaluation is the same scorer that already ran on every *miss* before. If an `AllowOperatorOverloading` benchmark row is ever added, this lane is the first thing to put under it. Merging on green CI.

PR #3586, opened 2026-09-01
DevTools: a new Jint.DevTools package answers the Chrome DevTools Protocol from a pinned, generated description

lahma:
Jint has no living tooling protocol. The two prior attempts — [Jither/Jint.DevToolsProtocol](https://github.com/Jither/Jint.DevToolsProtocol) and [Jither/Jint.DebugAdapter](https://github.com/Jither/Jint.DebugAdapter) — stopped in early 2024, were built on a private engine branch, and predate the debugger they wrapped. Every client that matters (Puppeteer, PuppeteerSharp, Playwright, chrome-remote-interface, and the Chrome DevTools front end itself) speaks the Chrome DevTools Protocol, so this implements CDP in-repo against the current debugger rather than reviving either.

This is the **skeleton**: the vendored protocol, the generator that turns it into code, the envelope and its error codes, an in-process transport, and two answered commands. Sessions, the WebSocket transport and the `Runtime` / `Debugger` / `Profiler` domains build on it.

It is engine-level and works for **every** embedder: an Orchard or Elsa host will be able to attach Chrome DevTools to its own scripts. The page-level domains belong to a later `Jint.Browser`, which is *AngleSharp plus Jint* and is never positioned as a DOM stack of its own.

## What is answered today

| method | answer |
| --- | --- |
| `Schema.getDomains` | the manifest's reported domains, with versions |
| `Browser.getVersion` | `protocolVersion` `1.3`, `product`/`userAgent` `Jint/<version>` — Jint, not an impersonated Chrome build |
| `Browser.close` | a host callback, or a successful no-op |

Everything else is `-32601 '<Domain.method>' wasn't found`, in Chrome's exact wording, which is what a client feature-detecting a domain reads.

## Three decisions worth reviewing

**The protocol description is vendored and pinned, and its generated output is checked in.** `tools/devtools-protocol/` holds `js_protocol.json` and `browser_protocol.json` verbatim from `ChromeDevTools/devtools-protocol@ea39a11d80de9a08ce2af03f52125ed2e462cf84` (npm `devtools-protocol@0.0.1687809`), under the 3-Clause BSD licence beside them. **A bump is a code change, not a pin change** — the same discipline `Test262Harness.settings.json`'s `SuiteGitSha` carries — because upstream renames methods, moves them between domains and turns optional parameters into required ones, and the diff of `Jint.DevTools/Protocol/Generated/` is that change stated in the vocabulary this repository compiles.

A Roslyn source generator cannot do this job: the `System.Text.Json` context has to be generated *over* the DTOs and generators do not chain, and `Jint.SourceGenerators` is `netstandard2.0` without `System.Text.Json`. `GeneratedProtocolIsCurrentTests` runs the emitter in memory and diffs byte for byte, so editing the manifest or the pin without regenerating fails the build, and so does hand-editing a `.g.cs`.

**The manifest is the boundary between what is described and what is answered, and it is load-bearing at run time.** Every command of a generated domain gets a `protected virtual`, so the whole surface is discoverable in the IDE; only the commands `manifest.json` lists get a `case` in the generated dispatch. So an unimplemented method is `-32601` — and, the part that is easy to get wrong, `-32601` **before its parameters are read**. A command a backend does not implement is not in Chrome's dispatch table at all, so answering `-32602` there tells a client it called a command wrongly when the command does not exist here. (This one is not theoretical: the first draft deserialized first, and `Browser.setWindowBounds` answered `-32602`. The test caught it and the dispatch was reshaped.)

`Schema.getDomains` reports only domains that have an implemented command — the generator refuses an entry that does not — and `ProtocolManifestTests` holds the manifest and the code to each other in both directions. The workflow for a new command is therefore exactly three steps, and skipping any of them fails: add the manifest entry, regenerate, override the virtual.

**There is deliberately no `InternalsVisibleTo` grant from `Jint`.** This package consumes the published engine API and nothing else, which is the point: a protocol server that could only be written from inside the engine assembly is one no third party could have written, and every seam it turns out to need is a seam the engine should expose, as public API with a migration row. Everything in the package is `internal` in turn, so its public API baseline is empty and the first member promoted out of that is a diff somebody reads; when one is, it carries `[Experimental("JINTDT001")]`, saying its shape follows a living upstream document rather than this repository's compatibility contract.

## What the generator produced

199 records, 32 `const string` classes for the protocol's string enumerations (never C# enums — `JsonStringEnumMemberName` is .NET 9+ and `net8.0` has to compile), 115 command virtuals and 27 event factories, across `Runtime`, `Debugger`, `Profiler`, `Console`, `Log`, `Schema`, `Target` and `Browser`. Every payload crosses through the source-generated `ProtocolJsonContext`, so no protocol type is reached by reflection.

## Tests

39 on `net10.0`, 37 on `net8.0` (the two public-API tests run on one leg, the way `Jint.Tests.PublicInterface`'s do and for the same reasons).

- **In-process protocol tests assert text, not objects** — a client library matches on `id`, `error.code` and `error.message`, so a test that called a domain method directly would pass with the envelope broken. Covered: both answered commands, the `id` and `sessionId` echo, `-32601` in six shapes, `-32700`, `-32600` in four, `-32602`, ignored unknown parameters, and one reply per message.
- **`GeneratedProtocolIsCurrentTests`** — the currency diff, plus that two runs produce the same bytes, which is what makes the diff mean "the inputs changed".
- **`ProtocolCitationTests`** — every `chromedevtools.github.io` URL under `Jint.DevTools/` resolves against the vendored JSON. Offline and register-free, deliberately unlike `Jint.Tests/SpecCitationTests.cs`: the authoritative document is already in the tree.
- **`DomainLifecycleTests`** — the enable/disable hooks and event emission, through a domain the suite declares itself. An untested extension point is a design nobody has tried.

Each new test was checked against code that did not satisfy it: the manifest was edited without regenerating (currency test fails), a method was claimed implemented without an override (manifest test fails), a citation anchor was mistyped (citation test fails), and the `-32602`/`-32601` ordering above.

## Also in this change

`Jint.slnx` gains the two projects and the generator; `build.yml` and `release.yml` pack `Jint.DevTools` beside `Jint`, at the same version — it references Jint by project and its surface is only meaningful against the engine it was built with. `Jint.DevTools/AGENTS.md` carries the thread rule, the pause-loop shape that P4 will need, the pin and the manifest discipline, with a row in the root index and `.claude/rules/devtools.md` on the other half of the routing map; `AgentInstructionFileTests` is green with 262 bytes of root headroom left.

No public API of `Jint` changed, so there is no `docs/v5-migration.md` row: the row belongs with the first *public* member of this package.


Jint.DevTools/Protocol/Generated/Console.g.cs
```diff
@@ -0,0 +1,156 @@
+// <auto-generated/>
```
sebastienros, line 1: Do these have to be included?

lahma: These usually aide ides and analyzers to skip as something not actionable and not worth spending time with.

lahma: Ah you mean in VCS, good point. 😄

lahma: They stay checked in on purpose, and it is written down in `docs/design/devtools-protocol.md` §4 and the generator README. Two reasons carry it: the System.Text.Json serializer context has to be generated *over* the DTOs, which a Roslyn generator cannot do over another generator's output (and reflection-based STJ would cost the AOT claim); and a protocol pin bump is a diff we want to read — the same rule as the test262 SHA. A currency test re-runs the emitter in memory and fails on any drift, so the files cannot go stale or be hand-edited.

What they should not do is fill a review: #3675 marks both generated directories `linguist-generated`, so GitHub collapses them in pull requests and drops them from the language statistics.

PR #3659, opened 2026-09-02
DevTools: a Profiler recording is the engine's sampler, and the exact profiler is the fallback

### LINKED ISSUE #3630: SampledProfile publishes only a Firefox Profiler document; a public read-only view of its tables would let Jint.DevTools serve the same samples
Found while landing #3629 (`Profiler` domain of `Jint.DevTools`), part of #3575.

#3608's sampling profiler keeps its tables (`_funcs`, `_frames`, `_stacks`, `_sampleStacks`, `_sampleTimes` on `SampledProfile`) `internal` and publishes only the Firefox Profiler document. `Jint.DevTools` deliberately consumes public API only (no `InternalsVisibleTo`), so CDP `Profiler.start`/`stop` cannot be backed by the sampler today: the domain samples on its own activations behind an internal `IProfileSource` seam, with `(root)`/`(program)` only. Consuming the sampler would mean writing the JSON document and parsing it back.

Ask: public read-only accessors on `SampledProfile` (the sample times, the stack table as parent links, the frame table with function name, source, line, column and `ProfileFrameCategory`), documented like the rest of the diagnostics facet (`[Experimental]` if the shape is still moving). With that, the sampler becomes a second `IProfileSource` in `Jint.DevTools` and the DevTools Performance panel shows the same samples the Firefox Profiler does. `docs/design/devtools-protocol.md` §5 names this gap.

lahma:
**Stacked on #3651, #3652 and #3654**; review the top commit only, and merge those three first.

A CDP `Profile` is what V8's *sampling* profiler produces and what `Profiler.setSamplingInterval` sets the rate of, so a sampler is the instrument a front end is asking for. This package could not use the engine's one, because `SampledProfile` published a Firefox Profiler document and nothing else — which is what #3654 fixes. With its tables public it becomes a second `IProfileSource`, and the domain does not change shape: that is what the seam was written for before there was a second implementation of it.

## Samples as a balanced stream

`SampledProfileSource.Stop` converts the sample table into the enters and leaves `RecordedProfile` speaks, which is lossless in that direction: two consecutive samples differ by a *suffix* of the stack, so the difference between them is a run of leaves and a run of enters at the later sample's time, and the interval between two samples belongs to the stack observed at the earlier one. `ProfileBuilder` then does exactly what it did.

Two decisions in it:

- **The sampler's synthetic `(program)` root is dropped.** `ProfileBuilder` has one of its own and gives it the time no script function was on the stack; carrying both would put the node in the tree twice. `(idle)` stays unemitted for the reason it always was — an engine target has no idle state to report, and a host that is not running script is doing something this package cannot see.
- **`DroppedSampleCount > 0` becomes `Truncated`**, which is what a client reads to know the profile describes the beginning of a run and nothing about the rest.

## Which instrument, and when

Chosen when a recording starts: the sampler, unless the engine is already sampling for the host — there is one sampling session per engine — in which case the client gets the exact profiler rather than a refusal. That costs the run more and misses nothing, which is the right way round for a fallback. `AHostAlreadySamplingLeavesTheClientTheOtherInstrument` pins it, including that the host's own session is untouched.

## What it costs the engine

Nothing an engine pays unless a client records. A sampling session is one `Stopwatch.GetTimestamp` and one comparison at the check points the amortized constraints already ride, where the exact profiler pays a frame lookup and a 16-byte event on *every call*; so the instrument this switches to is the cheaper of the two, and neither is armed until `Profiler.start`. No path in this package runs while script does.

## What this changes for a client, and for the tests

A sampler cannot show a call it never sampled: a function entered and left between two check points is not in the profile at all, where the exact profiler recorded every one of them. So `ProfilerSessionTests` now profiles a script that spends its time where it says it does, and asks for the fastest rate the protocol can name (`setSamplingInterval: 0`, "as fast as you can") because a workload short enough for a unit test has to be observed more than once. Ten consecutive runs of the file were green on both target frameworks.

The new one is the pin the issue asks for: `ACpuBoundScriptAttributesMostOfItsTimeToItsHotFunction` adds up the time deltas per node and requires the hottest to be the function the loop is in, with more than half the run. It asserts on the deltas rather than on the count of samples, because that is what the panel adds up per node.

`Jint.Tests.DevTools`: 376 on net10.0, 374 on net8.0, all green. No public surface changes here, so no baseline and no migration row; #3654 carries both.

`Domains/AGENTS.md`'s profiling section and `docs/design/devtools-protocol.md` now say the sampler is the source; the design doc moves #3630 out of the open-seams list and names it E7.

Fixes #3630

### DIFF
--- a/Jint.DevTools/Domains/AGENTS.md
+++ b/Jint.DevTools/Domains/AGENTS.md
@@ -219,24 +219,28 @@ snapshots, so a value changed after one was handed out is not reflected in it.
 
 The `Profiler` domain answers two questions with two instruments, and neither is the other's fallback.
 
-**A profile comes through `IProfileSource`, and the shipped source is the *exact* profiler.** The engine has
-two: `Engine.Diagnostics.StartProfiling`, which records every call at the call boundary, and the sampling
-profiler of [#3608](https://github.com/sebastienros/jint/pull/3608), which notes what the stack looks like at
-the engine's own check points. Only the first is usable from here, and the reason is a visibility one rather
-than a design one: `SampledProfile` keeps its sample, frame, stack and function tables as `internal` fields
-and publishes a Firefox Profiler document as its only output, so consuming it would mean serializing to JSON
-and parsing it back. **Making those tables public is an engine change**, and when it lands the sampler becomes
-a second `IProfileSource` rather than an edit to the domain — which is why the seam exists before there is a
-second implementation of it. The seam speaks a function table and a balanced stream of enters and leaves,
-deliberately not the engine's own `ScriptProfileFrame`: a seam that names one profiler's type is a seam only
-that profiler fits through.
+**A profile comes through `IProfileSource`, and both of the engine's profilers fit through it.**
+`SampledProfileSource` notes what the stack looks like at the engine's own check points
+([#3608](https://github.com/sebastienros/jint/pull/3608), readable since
+[#3630](https://github.com/sebastienros/jint/issues/3630)); `EventedProfileSource` records every call at the
+call boundary. The seam speaks a function table and a balanced stream of enters and leaves, deliberately not
+either profiler's own type: a seam that names one instrument is a seam only that instrument fits through. A
+sampler converts into that shape losslessly, because two consecutive samples differ by a suffix of the stack
+and that difference is a run of leaves and enters.
+
+**The sampler is what a recording uses, and the choice is made when it starts.** A CDP `Profile` is what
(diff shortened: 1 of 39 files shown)


lahma:
Folded into #3654, which is now the engine seam and this consumer in one commit — the stack was rebased onto main at 34113ca1c after #3651 and #3652 landed, and one pull request per seam is fewer rebase-and-CI cycles than four stacked ones. Nothing here is dropped: `SampledProfileSource`, the source choice at `Profiler.start`, and the reworked `ProfilerSessionTests` are all in #3654.

PR #3663, opened 2026-09-02
DevTools: the four pause-on-exception states are four engine modes, and nothing is filtered here

### LINKED ISSUE #3631: A Break handler that wants to decline a pause must return a StepMode, which cancels a step in flight; ExceptionPauseMode has no Caught
Found while landing #3625 (`Debugger.setPauseOnExceptions`), part of #3575.

CDP's `setPauseOnExceptions` has four states: `none`, `caught`, `uncaught`, `all`. E3 (#3623) gave `DebugHandler.PauseOnExceptions` three (`None`, `Uncaught`, `All`), so `caught` is implemented as `All` plus filtering `IsUncaught` in the `Break` handler — and the only way to *not* pause from a `Break` handler is to return a `StepMode`, whose value **is** the next step mode. Returning `StepMode.None` to skip an uncaught throw therefore cancels a step that was in flight. It is harmless in practice only because the frames a step was walking are about to be unwound, but it is a trap for every other host that filters pauses.

Two fixes, either is enough: an `ExceptionPauseMode.Caught` so the engine never raises the pause it would be asked to skip; and/or a way for a `Break`/`Step` handler to say "no change" (a nullable return or a `StepMode.Unchanged` sentinel that `HandleNewStepMode` treats as "keep the current mode"), which is the general form.

lahma:
**Stacked on #3651, #3652, #3654, #3659 and #3662**; review the top commit only, and merge those five first.

`setPauseOnExceptions`'s `caught` had no engine mode, so this domain asked the engine for `All` and dropped the uncaught half inside the `Break` handler — the one place in the package that declined a pause. Declining means returning a `StepMode`, and every one of those but `Unchanged` *sets* the mode, so the filter cancelled a step the client had in flight. With `ExceptionPauseMode.Caught` (#3662) the client's four states map one to one onto the engine's:

```csharp
SetPauseOnExceptionsRequestStateValues.Uncaught => ExceptionPauseMode.Uncaught,
SetPauseOnExceptionsRequestStateValues.Caught   => ExceptionPauseMode.Caught,
SetPauseOnExceptionsRequestStateValues.All      => ExceptionPauseMode.All,
_                                               => ExceptionPauseMode.None,
```

and the filter in `DebuggerDomain.Stop` is gone: a throw that reaches the handler is one the client asked to stop on.

What a client sees is unchanged, which is what the existing four state tests (each against a throw something catches and a throw nothing catches) say by still passing. The new one, `EachProtocolStateIsAnEngineModeOfItsOwn`, pins the mapping itself — it is the thing that changed — and `Domains/AGENTS.md` now says that deciding a pause in this domain is what must not come back, with `StepMode.Unchanged` named as the answer where a future filter genuinely has no engine mode behind it.

`Jint.Tests.DevTools`: 377 on net10.0, 375 on net8.0, all green. No public surface changes here, so no baseline and no migration row; #3662 carries both. `docs/design/devtools-protocol.md` names the seam E8 and its open-seams list is now empty.

Fixes #3631

### DIFF
--- a/Jint.DevTools/Domains/AGENTS.md
+++ b/Jint.DevTools/Domains/AGENTS.md
@@ -219,24 +219,28 @@ snapshots, so a value changed after one was handed out is not reflected in it.
 
 The `Profiler` domain answers two questions with two instruments, and neither is the other's fallback.
 
-**A profile comes through `IProfileSource`, and the shipped source is the *exact* profiler.** The engine has
-two: `Engine.Diagnostics.StartProfiling`, which records every call at the call boundary, and the sampling
-profiler of [#3608](https://github.com/sebastienros/jint/pull/3608), which notes what the stack looks like at
-the engine's own check points. Only the first is usable from here, and the reason is a visibility one rather
-than a design one: `SampledProfile` keeps its sample, frame, stack and function tables as `internal` fields
-and publishes a Firefox Profiler document as its only output, so consuming it would mean serializing to JSON
-and parsing it back. **Making those tables public is an engine change**, and when it lands the sampler becomes
-a second `IProfileSource` rather than an edit to the domain — which is why the seam exists before there is a
-second implementation of it. The seam speaks a function table and a balanced stream of enters and leaves,
-deliberately not the engine's own `ScriptProfileFrame`: a seam that names one profiler's type is a seam only
-that profiler fits through.
+**A profile comes through `IProfileSource`, and both of the engine's profilers fit through it.**
+`SampledProfileSource` notes what the stack looks like at the engine's own check points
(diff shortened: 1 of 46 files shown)


lahma:
Folded into #3662, which is now the engine seam and this consumer in one commit — the stack was rebased onto main at 34113ca1c after #3651 and #3652 landed, and one pull request per seam is fewer rebase-and-CI cycles than four stacked ones. Nothing here is dropped: the four-state mapping, the deleted filter, `EachProtocolStateIsAnEngineModeOfItsOwn` and the `Domains/AGENTS.md` rule are all in #3662.

Issue #3575, opened 2026-09-01
Headless browser and DevTools protocol: a campaign plan


lahma:
Sebastien's direction for what follows the web-API campaign: *"Next: headless browser. Check obscura, kitesurf. And the starling browser (c#)."* This issue is the authoritative statement of the campaign that answers it. The longer form — the mechanisms each decision rests on — is `docs/design/headless-browser.md` and `docs/design/devtools-protocol.md`; where those files and this issue disagree, this issue wins.

## What the landscape says (verified 2026-09-01)

- **Lightpanda** (Zig + V8, AGPL, 34k stars): DOM + JS + network, **no layout, no pixels**; 21 CDP domains plus WebDriver BiDi and an MCP server; a public daily WPT scoreboard. The product category "a browser that does not render" is proven. AGPL: read for design, never copy.
- **Obscura** (Rust + V8, Apache-2.0, 23k stars): adds layout and rendering; its CDP domain list (Target/Page/Runtime/DOM/Network/Fetch/IO/Storage/Input + a custom `LP.getMarkdown`) is the minimum surface that makes Puppeteer and Playwright work; its benchmark repo defines a WPT "Core tier" (DOM/HTML/URL/fetch) and a 33-fixture obstacle course.
- **Kitesurf** (Cloudflare, closed, obscura-inspired, runs in V8 isolates): "enough CDP" for Puppeteer, Playwright and the DevTools frontend; ~215k WPT subtests; 3–7× less CPU and memory than Chromium at 1.7× its wall-clock time — exactly an interpreter's profile. Agent-first: token count and snapshots matter, pixels do not.
- **Starling** (C#, 57 stars, single author, last push 2026-07-07): everything from scratch including its own JS engine; it explicitly refuses to embed Jint and uses Jint only as a benchmark baseline; AngleSharp as an optional HTML backend. Not a base and not competition on the automation lane; the one actionable question it poses is host-object cost, which the binding design answers.
- **AngleSharp / AngleSharp.Js** (MIT, active): the parser, DOM and CSSOM exist and are good; `.Js` binds Jint 4.15 to that DOM by reflection with a process-shared prototype layout, runs scripts, modules, import maps and workers. The binding is the slow part and lives outside this repo.
[...]

PR #3669, opened 2026-09-02
Browser: a page's observers fire, and its DOM views answer

lahma:
Campaign item **R4** of the headless-browser campaign (#3575): the window miscellany a real page depends on
and that answered nothing before — the three observers, and the DOM views that let the generated `Range`,
`TreeWalker` and `NodeIterator` be used at all.

A page that watches its own DOM, lazily loads a list, measures a component when it mounts or parses markup
does none of those things when the interface is simply absent. It does not fail either: it stops, with
nothing to read.

## The three observers, and the lane each delivers on

**`MutationObserver` delivers at the microtask checkpoint.** The *records* are AngleSharp's —
`DocumentExtensions.QueueMutation` already does the whole of DOM §4.3.4: inclusive ancestors, the registered
observer list, `subtree`, `attributeFilter`, and `oldValue` cleared for an observer that did not ask for it.
What AngleSharp has no answer for is **when**. Its `MutationHost` schedules through an `IEventLoop` service,
*nothing in AngleSharp implements one*, and `EventLoopExtensions.Enqueue` runs the action **inline** when the
loop is null — so out of the box the callback fires synchronously in the middle of `appendChild`. Registering
an event loop to fix that is precisely the thing that would make a step of the parse asynchronous and take
`PageDocument`'s parser-hop fallback.

So the inline call is used as the *arrival* of a record and nothing more: `JsMutationObserver` parks it, and
`MutationObserverLane` puts one job on the engine's queue per batch. The ordering then falls out of a plain
enqueue and is pinned:

```js
Promise.resolve().then(() => log.push('before'));
host.appendChild(p);            // queues "notify mutation observers" here
host.appendChild(p2);           // same batch
Promise.resolve().then(() => log.push('after'));
log.push('sync');
// → sync | before | mo:2 | after
```

**`IntersectionObserver` and `ResizeObserver` deliver as a *task*** (a zero-delay timer entry), because both
belong to update-the-rendering and a microtask would run before the promises of the same turn. Both are
**stubs, and the shape of the lie is stated rather than hidden**: every observed target is reported exactly
once — fully intersecting, or at zero size — and never again, because with no layout nothing can change.
"Never intersecting" would stop every lazy list and reveal-on-scroll animation dead; the initial resize
notification is the one a component uses to measure itself. `root`, `rootMargin` and `thresholds` are parsed,
validated and reflected exactly as the specification says and change nothing. Rectangles are zeros and are
**plain objects rather than `DOMRectReadOnly` instances** — there is no rectangle interface in the package
yet, and one whose every instance is zeros would be worse than none. C4's flat-box model turns them into real
numbers.

## The DOM views

`DOMParser` (HTML, and XML through **`AngleSharp.Xml`**), `XMLSerializer`, `Selection` + `getSelection()`,
`NodeFilter`, `MediaQueryListEvent`, and the members that make `Range` / `TreeWalker` / `NodeIterator` usable:
`createTreeWalker`, `createNodeIterator`, `Range.toString`, `slot.assignedNodes()`, `ShadowRoot.mode`.

- A **parsed document cannot run anything**: its parser gets a browsing context of its own with no scripting
  service and `IsScripting` false, which is what "not a browsing context" means in the standard. Pinned.
- A **failed XML parse answers the `parsererror` document** the standard prescribes, which is what a page
  tests for, rather than throwing.
- **`getComputedStyle` keeps R1's cascade and gains CSSOM's computed flag**: every write is refused with a
  real `NoModificationAllowedError` instead of landing silently in a detached copy nothing can read.
- **`matchMedia` gained its other half**: `PageRuntime.SetViewport` is the seam device emulation (C5) will
  drive, and every `MediaQueryList` the page holds fires `change` — with a real `MediaQueryListEvent`, because
  `e => e.matches` is how the listener is written — only if its own answer moved.

## AngleSharp findings

Recorded in `Jint.Browser/AGENTS.md`'s divergence table and **reported upstream rather than patched around**;
the two places a wrapper corrects one, it says so and says why.

| What | The standard | AngleSharp |
| --- | --- | --- |
| `observer.disconnect()` | empties the registered-node list | unregisters from the document but **leaves that list populated**, so the next `observe()` of any node silently re-registers every node it used to watch |
| `observer.observe(document, …)` | observes the document node | silently retargets a `Document` to its `documentElement` |
| `observe(…)` with bad options | `TypeError` | `DomException`, and the dictionary's optional members resolve differently (`{ attributeOldValue: false }` alone is valid in DOM, invalid there) |
| `record.addedNodes` on an attribute record | an empty `NodeList` | `null`, for both `Added` and `Removed` |
| `IShadowRoot.Mode` | `ShadowRoot.mode` | the one member of the interface carrying no `[DomName]` |
| `slot.assignedNodes()` | DOM §4.2.5 | named `getDistributedNodes` — the Shadow DOM v0 spelling |
| `getComputedStyle(span).display` | `inline` | the empty string: the cascade reports only what was *declared*, and `display: inline` is the initial value |
| the computed style is writable | `NoModificationAllowedError` | an ordinary writable — and detached — declaration |

`MutationObserver` also has **no `IEventLoop` implementation anywhere in AngleSharp**, so its own microtask
scheduling degenerates to a synchronous call inside the mutation. That is the finding the whole design above
turns on.

## The generator

Two changes, both in this area:

- **`overrides.json` gains an `additions` list** — a member the *standard* puts on a generated interface and
  AngleSharp's attributes cannot express: a callback parameter, a stringifier with no `[DomName]`, a name that
  moved. It is the only override list whose entries name something the pinned assemblies do **not** have, so
  what is checked is the opposite: the interface exists and the member does not. An addition whose member the
  generator *skipped* replaces that skip and removes it from the report.
- **A numeric enum crosses through a CLR type wide enough to hold it.** `FilterSettings` is a `ulong` whose
  `SHOW_ALL` is `0xFFFFFFFF`; the `(int)` cast made `TreeWalker.whatToShow` answer `-1`. Two generated lines
  change.

`AngleSharp.Xml` is referenced for `DOMParser`'s four XML content types and nothing else. It is MIT and from
the same project, and writing an XML parser here instead is the one thing this package is not for. **It is
deliberately not in `tools/dom-bindings/pin.json`**: the generator reads two assemblies and projects no
interface from this one, so the pin discipline is untouched. This is the one decision worth a second opinion
before merge.

## Verification

`Jint.Tests.Browser` goes from 107 to 179 tests, all green on both legs, and green again under
`JINT_HOST_CONTRACT_VERIFICATION=1`. Full solution green. No public API change, so
`PublicApiTest.verified.txt` is untouched.

`Views/HostInterfaceDisciplineTests` is new discipline rather than new coverage: `DomPrototypeTests` and
`WebIdlPropertyAttributeTests` walk the *generated* registry, and none of the eleven runtime-owned interfaces
is in it — so it holds every one of them to the same two rules (a shared-shape prototype, WebIDL's attributes
per member kind), which is exactly where a hand-written shape goes wrong.

## Left out, and why

- **`Element.getBoundingClientRect`** and a real `DOMRect` interface: C4 owns the box model, and zeros with no
  interface is more honest than an interface full of zeros.
- **Transient registered observers** are cleared at mutation time rather than at the checkpoint, because
  AngleSharp clears them inside the call this design uses as the record's arrival. A node removed from an
  observed subtree therefore stops being observed one mutation early.
- **`selectionchange`** is not fired and `Selection` has no direction: both need a user, and there is none.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -12,6 +12,11 @@
          these assemblies' [DomName] surface. Bump both together, regenerate, and read the diff. -->
     <PackageVersion Include="AngleSharp" Version="1.7.2" />
     <PackageVersion Include="AngleSharp.Css" Version="1.0.2" />
+    <!-- AngleSharp.Xml is deliberately NOT in the binding generator's pin: the generator reads AngleSharp and
+         AngleSharp.Css and nothing else, and this package contributes no [DomName] interface to the binding.
+         It is referenced for one thing only — DOMParser's XML content types, which need an XML parser and
+         which this package's own principle says must not be written here. -->
+    <PackageVersion Include="AngleSharp.Xml" Version="1.2.0" />
     <PackageVersion Include="BenchmarkDotNet" Version="0.15.8" />
     <PackageVersion Include="BenchmarkDotNet.TestAdapter" Version="0.15.8" />
     <PackageVersion Include="AwesomeAssertions" Version="9.6.0" />
--- a/Jint.Browser/AGENTS.md
+++ b/Jint.Browser/AGENTS.md
@@ -88,6 +88,7 @@ version of AngleSharp nobody references.
 | `manual` | An interface whose shape is hand-written in `DomManualShapes`. Today: `IHtmlCollection<T>`, whose generic invariance keeps a member body from naming its receiver. A manual interface contributes no members to any closure — its children inherit them through the prototype chain. |
 | `skip` | A member a later campaign item owns: navigation (`location.assign`, `location.href`'s setter), activation (`click`, `focus`, `blur`, `form.reset`), the parser (`document.open`/`close`/`load`), and `document.createEvent`, whose AngleSharp `Event` must never reach script. `half: "setter"` skips the write half only. |
 | `hooks` | A member routed through `DomHostHooks` so the parser driver can replace its body: the `innerHTML` and `outerHTML` setters, `insertAdjacentHTML`, `document.write`/`writeln`. The default implementations *are* the AngleSharp call, so the seam costs nothing until R3 uses it. The same class carries `WrapperCreated`, which is the other direction — a member the generator could not emit at all, added to one wrapper. |
+| `additions` | A member the **standard** puts on a generated interface and AngleSharp's metadata cannot express: a callback parameter (`createTreeWalker`, `createNodeIterator`), a stringifier with no `[DomName]` (`Range.toString`), a member AngleSharp spells by its Shadow DOM v0 name (`slot.assignedNodes`), a member whose `[DomName]` is simply missing (`ShadowRoot.mode`), and the two CSSOM View rectangles a page with no layout still calls. The body is one line calling `Dom/Views/DomViewMembers`, and it is the **only** list whose entries name something the pinned assemblies do not have — what is checked is the opposite, that the interface exists and that the member does not, so an addition can never quietly shadow a member AngleSharp grew. An addition whose member AngleSharp *skipped* replaces that skip and removes it from the report. |
 | `nullableStrings` | The members whose IDL type is `DOMString?` rather than `DOMString`. See the conversion table below. |
 | `stringEnums`, `constants` | The two enum decisions the heuristics above cannot make. |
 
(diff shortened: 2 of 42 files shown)


lahma:
## Rebased onto #3667 (`c54e961e9`), and the instruction-file split is no longer needed

Rebased onto R5's navigation merge. Three conflicts, all resolved in R5's favour where they had moved
something: `PageRuntime`'s constructor (both lanes kept), the README's browser section (R5's network half
kept, the observers and views written as their own paragraph, and the stale "no network / no
`MutationObserver`" claims dropped), and `Jint.Browser/AGENTS.md`, where R5 had replaced the page-loop,
window and parse sections with a pointer to the new `Jint.Browser/Runtime/AGENTS.md` — that pointer stands,
and my override-table row and divergence rows sit alongside R7's accessibility section.

**The split I was asked to make is not needed, because R5 already made the cut.** Measured after the rebase,
with everything of mine in it:

| File | Bytes (CRLF) | Headroom |
| --- | --- | --- |
| `AGENTS.md` (root) | 24,564 | **12** of 24,576 |
| `Jint.Browser/AGENTS.md` | 29,734 | 3,034 of 32,768 |
| `Jint.Browser/Runtime/AGENTS.md` | 12,232 | 20,536 of 32,768 |

Moving the DOM material into a `Jint.Browser/Dom/AGENTS.md` would buy a file that is already 3 KiB under its
cap, and it would cost the root index a row — about 380 bytes carved out of a file with 12 to spare. That is
the wrong trade today, so the two rows the root already has are unchanged and no `browser-dom.md` was added.
The prepared split is one command away if the headroom is wanted before the next section lands.

**The root file's 12 bytes are worth someone's attention on their own**, and they are main's rather than
this branch's: nothing here touches `AGENTS.md` or `.claude/`. The next contributor to add an index row has
to free the bytes first.

Two documentation corrections that my change makes necessary, both in R5's file: `getSelection` is no longer
an example of "an operation that needs nothing", and the observers' reference to "the parser hop above" is
now a checked link to the file that section moved into.

PR #3679, opened 2026-09-02
Jint.Browser: a page's turns are bounded, and one call hardens a browser for content nobody vouches for

lahma:
R6 of the headless-browser campaign (#3575, Phase 2). `docs/design/headless-browser.md` §7 is the contract this
makes true, and it is corrected here where the implementation differs from it.

## What

**A page's turns are bounded.** `BrowserOptions.MaxTaskDuration` (5 s) and `MemoryLimit` were declared as
"carried and not read". They are read now, and what they bracket is a **turn**: one mailbox request, one
`ProcessTasks` drain — every due timer callback, microtask, promise reaction and animation frame together —
and one inline `<script>`. `Runtime/PageBudget` arms an `OperationDeadlineConstraint` and, where a memory
budget is configured, a `MemoryLimitConstraint` around each of them.

A `Page` call that runs out of budget fails its own task with `TimeoutException`, because the bracket sits
*outside* `PostAsync`'s own `catch`. A drain and an inline script are recorded as a new
`PageErrorKind.BudgetExceeded` and the page goes on: a timer queued in front of a runaway still runs, a
document whose `<script>` never returns still loads, and the page still answers `Evaluate`. A worker's pump
takes the same bracket over the constraint factories its parent replayed. `Timeout.InfiniteTimeSpan`, zero and
any negative value disable the bracket.

**`BrowserOptions.ForUntrustedContent(limits = null)`** applies `Options.ForUntrustedCode` to every page
engine from inside the package's own construction callback — so it is expanded before the engine builds
anything and re-expanded over the host's `ConfigureEngine` callbacks — derives `MaxTaskDuration` and
`MemoryLimit` from the limits when the host has not set them, folds a host-chosen memory budget *into* the
limits so the bracket and the profile can never state two numbers, and turns `BlockPrivateNetwork` on for
every context of that browser that did not assign it itself.

**Page-sized names for three engine settings** — `MaxActiveTimers`, `MaxResponseBytes`, `FetchTimeout` —
applied to page *and worker* engines before any host callback. **`MaxDomNodes`** (0 = unlimited) is checked
after each parse (a larger document fails the navigation with `NavigationFailedException`) and in the wrapper
factory (the projection that would pass it throws a `RangeError` into the script that asked for it).

## Why it has to be a bracket

The root `AGENTS.md` constraints gotcha applies to a page twice over: a page is a host-driven sequence of
entries, so `LimitExecutionTime` arms a fresh deadline for each `Page` call and bounds no sequence of them;
and a page's event loop is *pumped*, so a job chain under `ProcessTasks` never reaches
`ExecuteWithConstraints` and a per-entry timeout there never fires even once. The two constraints whose
window the host owns are the only ones that survive both, and `PageBudget` *finds* them on the engine rather
than creating them — so one code path arms a turn whether or not the untrusted profile registered them.

Three details worth the reviewer's eye. `WaitForIdleAsync` is posted unbracketed and brackets each drain
itself, because charging a whole wait to one turn's budget would fail every wait longer than
`MaxTaskDuration`. `PageLoop.ReplaceEngine` closes the outgoing engine's turn and opens one on the incoming
engine, because a constraint belongs to one engine — and that also stops an engine construction from spending
the budget the new document's first script will meet. A nested turn (an inline script inside the navigation
request) re-arms the *deadline* and hands the enclosing turn a full budget back, so a document is not failed
for having several scripts; the *allocation* budget is not re-armed, because `MemoryLimitConstraint.Begin`
refuses to start while the engine is executing.

## The one engine commit

`Options.ForUntrustedCode` clears every constraint the host registered — right for a budget, wrong for the
one registration that is not one. `ObserveCancellation` went out with them, so an engine built from both was
deaf to the token while `CancellationConstraintRequested` said cancellation had been asked for: `fetch`,
`XMLHttpRequest`, `EventSource`, a WebSocket, the module loader and the engine's own blocking waits all ran
on `CancellationToken.None`. For a page that means a navigation would not abandon what the outgoing document
had in flight, which is the property `Jint.Browser/Runtime/AGENTS.md` calls "the token is what makes
abandonment real". The token is now recorded beside its registration and re-registered at the end of the
profile's expansion; withdrawing it still withdraws it. Its own commit, with `docs/v5-migration.md` §4.117
and three pins in `Jint.Tests.PublicInterface/UntrustedCodeProfileTests.cs`.

## Left out, and why

- **A page-level `import()` has no module loader in this version**, so the module leg of the untrusted-network
  test is a worker's module load — the one module graph a page fetches today. R3's parser driver brings the
  document's own.
- **`MaxDomNodes` is one number checked against two quantities**, so a page is bounded at roughly twice it
  rather than at it. Seeding the wrapper counter from the parsed document's size was tried and reverted: it
  makes merely *walking* a document of the permitted size a refusal, and a framework touching most of its own
  tree is the ordinary case rather than the abusive one. An `innerHTML` assignment materializes a subtree with
  no wrapper of its own and is bounded by `MemoryLimit` instead, which the docs say.
- **The deadline observes no cancellation token.** The engine already carries a `CancellationConstraint` over
  the page's token; handing the same token to the deadline as well would change what a closing page's pending
  work fails with.
- **No animation-frame bracket of its own.** The batch is a timer entry, so it is inside the `ProcessTasks`
  turn already; §7 said otherwise and is corrected.

## Verified

`dotnet test -c Release` over the solution, green. `Jint.Tests.Browser` 584 (net8.0) / 586 (net10.0) with
19 new cases in `Runtime/PageConstraintTests.cs`; `Jint.Tests.PublicInterface` +3. **Every new test was run against the
unwired code first**: eight fail outright, three (the runaway timer, the runaway inline script, the runaway
`Evaluate`) never return at all and take the test host with them, which is the point.
`AgentInstructionFileTests` green — `Jint.Browser/AGENTS.md` 30,531 B, `Jint.Browser/Runtime/AGENTS.md`
17,113 B and `Jint/Constraints/AGENTS.md` 27,082 B, all against the 32 KiB cap; the root index is untouched.
`Jint.Browser/AGENTS.md` is the one worth watching: 2.2 KiB of headroom left, most of it spent by #3669
rather than here.
`Jint.Tests.Browser/Verify/PublicApiTest.verified.txt` regenerated.


lahma:
Rebased onto main at af5d80ff9 (R2 landed first: the `PageHostHooks` line in the factory conflict is dropped, both README paragraphs kept) and one commit added on top: `Jint.Browser/AGENTS.md` was at its cap once R2 and R4 were in, so the five binding-layer sections move to `Jint.Browser/Dom/AGENTS.md` with a third root-index row and the rules file naming all three files. Nothing in the constraints work changed.

PR #3710, opened 2026-09-03
Browser: an obstacle course of real pages, driven by both .NET clients, and the seven defects it found

lahma:
## What

An **obstacle course** of eighteen offline fixtures under `Jint.Tests.Browser/Fixtures/`, each a directory of
vendored files served over a real socket by the loopback server and driven through the public `Page` API — and
the **two .NET automation clients** driving three of them over the protocol, PuppeteerSharp and (for the first
time) Playwright for .NET. Plus the seven defects they found, six of them fixed.

## Why

Everything that gates `Jint.Browser` today measures a mechanism: the web-platform-tests browser lane says the
DOM behaves, the protocol tests say the envelope is right, `PuppeteerSharpPageTests` says a client is satisfied
by a document written to satisfy it. None of them answers "does a page work". A fixture built out of React's
own published bundle, on a real origin, asserting a DOM end state **and** an empty `Page.Errors`, does — and it
found seven things nothing else had.

## The fixture inventory

Each case asserts the end state *and* that the page reported nothing at all: a framework that threw half way
through still renders something, so the error sink is what tells a half-working page from a working one.

| Fixture | What it exercises | Status |
| --- | --- | --- |
| `todomvc-react` | React 18 with a scheduler and synthetic events: controlled input, keyed list, hash-routed filters | **passes** |
| `todomvc-vue` | Vue 3 compiling the document's own markup: `v-model`, `v-for`, `:class`, `@keydown.enter`, mustaches | **passes** |
| `todomvc-preact` | Preact hooks writing to the DOM directly, with no scheduler between them | **passes** |
| `todomvc-svelte` | Svelte 5 compiled ahead of time: no framework runtime is loaded, only the component's own output | **passes** |
| `ssr-hydration` | React `hydrateRoot` over server markup — the nodes are adopted, not replaced, and `onRecoverableError` stays empty | **passes** |
| `jquery` | jQuery 3.7: `$.ajax({ async: false })` — a *synchronous* XMLHttpRequest — `$(fn)` readiness, delegated events on rows added later | **passes** |
| `htmx` | htmx 2: `hx-trigger="load"`, `hx-get` + `hx-swap`, `hx-boost` | **needs triage** |
| `alpine` | Alpine 3: `x-data`, `x-model`, `x-text`, `x-show` — attributes found by walking the document, kept live by a `MutationObserver` | **passes** |
| `spa-router` | An intercepted click, `history.pushState`, `popstate`, `back()`/`forward()`, and a link the router does not claim | **passes** |
| `custom-elements` | `customElements.define`, the upgrade of an element the parser already made, all four reactions | **needs triage** |
| `modules-importmap` | `<script type="importmap">`, a bare specifier, a mapped directory prefix, a dynamic `import()` | **passes** |
| `fetch-json` | `fetch` of JSON rendered into a list, a request header the script set, the page's own request log | **passes** |
| `form-redirect` | A form's entry list, a urlencoded `POST`, a `303`, and the `GET` that shows what the server read | **passes** |
| `cookie-login` | `Set-Cookie` on a redirect, the jar carrying it to a protected page, `HttpOnly` invisible to script | **passes** |
| `local-storage` | `localStorage` across two navigations and a second page of the same context; `sessionStorage` not leaving the page | **passes** |
| `intersection-observer` | A lazy list and a reveal panel — which load **everything at once**, the documented divergence | **passes** |
| `mutation-observer` | `childList`, `attributeFilter` with old values, `characterData`, one batch per turn | **passes** |
| `dialogs` | `alert`/`confirm`/`prompt` through `Page.DialogOpened`, and the default dismiss | **passes** |

`Fixtures/README.md` is the inventory, and `FixtureInventoryTests` holds it to the corpus in four ways — every
fixture has a row and every row a fixture, every fixture has an entry document, every vendored library has a
licence beside it, and **the set of `needs triage` rows is exactly the set of `[Explicit]` cases**. That last
one is what makes turning a failing fixture off quietly impossible, which is the discipline
`Jint.Tests/Wpt/AGENTS.md` puts the exclusion table under.

Vendored: React 18.3.1, Vue 3.5.22, Preact 10.29.8, Svelte 5.57.0 (compiled output, with its source and the
build command beside it), jQuery 3.7.1, htmx 2.0.10, Alpine 3.17.1 — each the library's own published bundle,
byte for byte, with its own `LICENSE`. All MIT except htmx's 0BSD. Nothing reaches the network: the page's
`UrlFilter` is pinned to the loopback server.

## The two client suites

`PuppeteerSharpCourseTests` drives the React TodoMVC, the router and the cookie login with the client's own
*trusted* input and none of the harness — `WaitForSelectorAsync`, `TypeAsync`, `Keyboard.PressAsync`,
`ClickAsync`, `EvaluateFunctionAsync`, `GetContentAsync`, `GetCookiesAsync`, `Accessibility.SnapshotAsync`,
`SetRequestInterceptionAsync` on the fetch fixture, and the screenshot refusal followed to the `Jint.getMarkdown`
it names.

`PlaywrightCourseTests` is new, and it is not a duplicate: Playwright discovers the browser over HTTP rather
than being handed a socket, insists that every page target name a browser context, scrolls into view and asks
for content quads before clicking, and re-resolves a locator on every use. `ConnectOverCDPAsync`, `GotoAsync`,
`Locator(...).ClickAsync()`, `FillAsync`, the keyboard, `TextContentAsync`, `RouteAsync` fulfilling a request,
`Context.CookiesAsync`, `EmulateMediaAsync(ColorScheme.Dark)` flipping the page's own `matchMedia`, and
`GetByRole` — which goes through Playwright's **injected script**, not the `Accessibility` domain, so the two
suites cover both paths.

**Playwright needs no `playwright install`.** `connectOverCDP` downloads no browser, and the `Microsoft.Playwright`
package carries its driver *and* a node runtime for the host's RID — verified by running it here with nothing
installed. What it does cost is a driver process per run, so the suite reads `JINT_BROWSER_CLIENTS` and a new
`browser-clients` CI leg (Linux) sets it; that leg then runs the suite again with the variable unset and requires
four skips, so the gate is a gate rather than a comment. PuppeteerSharp's half is not gated — it is pure .NET and
launches nothing — and the fixtures run on every leg as part of an ordinary `Jint.Tests.Browser` run.

Every method both clients send on these steps is answered. The recorded handshake pin needed no change (no
recording was made).

## The seven defects

Six fixed here. Every one was invisible to every existing gate.

| # | Found by | Defect | Fix |
| --- | --- | --- | --- |
| 1 | Vue, Preact | `insertBefore(node, null)` was a `TypeError`. WebIDL's second parameter is `Node?` and a null child appends — which is how **every** virtual DOM adds its last row. The emitter reads C# optionality, which cannot say it. | `overrides.json` grows a `nullableParameters` table for argument positions WebIDL declares nullable |
| 2 | jQuery | An unparseable selector reached the host as a **CLR exception**, so nothing in the page could catch it; DOM requires a `SyntaxError` `DOMException`. jQuery 3.7 asks for `:has(*,:jqfake)` inside a `try` during support detection, so jQuery did not load at all. | The five selector members are re-declared over the same AngleSharp calls, wrapped (`Dom/DomSelectorMembers`) |
| 3 | Playwright | `Target.getTargetInfo` refused an omitted `targetId`. The protocol says an omitted one names the session's own target, and on the browser session that is the browser. Playwright sends exactly that, with no parameters, before it will do anything else. | Answered, with a `browser` target info on the root session |
| 4 | Playwright | A page target in the **default** browser context named no context at all. Chrome's default context has an identifier and every page reports one; Playwright asserts on it and **kills its driver process** without it. | The default context gets an identifier, kept out of `Target.getBrowserContexts` the way Chrome keeps it out |
| 5 | Playwright | `Node.getRootNode` did not exist — AngleSharp has no `[DomName]` for it, so nothing could generate it. Playwright's injected script calls it on every element it touches. | Declared by hand (`Dom/DomNodeMembers`) |
| 6 | Playwright | `Storage.getCookies` answered on a **page** session only. Puppeteer asks a page for its cookies and Playwright asks the browser for a *context's*, so one client was answered and the other got `-32601`. | `ITargetHost` grows a seam for the domains a host answers about the browser, and `BrowserStorageDomain` uses it |

**The seventh is recorded rather than fixed**, and it is the one that matters most:
`getComputedStyle(el).visibility` is the **empty string** for every element, because AngleSharp reports the
declared cascade and `visible` is CSS's initial value. Playwright's actionability check ends in
`style.visibility !== "visible"`, so it reads every element of every page as hidden: `IsVisibleAsync` is false
for an element with a real 1280×16 box, an unforced `ClickAsync` waits out its timeout, and the ARIA role
engine drops the element. The suite drives past it with `Force` and `IncludeHidden`;
`PlaywrightCallsEveryElementHidden` pins the reason so that fixing it fails the test. It is **not** patched
here because the repository already decided (`Views/ComputedStyleTests`) to record this rather than keep an
initial-value table of its own — what is new is that the decision costs a supported client, which is a call for
the maintainer rather than for this PR.

## What was left out and why

- **`htmx`** does not pass and is a `needs triage` row. htmx 2 builds an `XPathEvaluator` expression at the
  **top level** of its bundle to find `hx-on:` attributes, so the library is a `ReferenceError` before any `hx-`
  attribute is read. This browser has no XPath at all — AngleSharp keeps it in `AngleSharp.XPath`, which nothing
  references — so nothing about `hx-get`, `hx-swap`, `hx-trigger` or `hx-boost` has been reached, let alone
  refuted. The fixture is checked in unchanged and works the day XPath exists.
- **`custom-elements`** is `[Explicit]` because `customElements` is campaign item C6 and had not merged. The
  fixture is vendored rather than stubbed, so C6 turns the case on by deleting one attribute.
- **Nothing was recorded.** The handshake pin covers `playwright-dotnet` already and needed no change.
- **`getComputedStyle`'s initial values**, above.

## Verification

`Jint.Tests.Browser`, gate off: 957 passed / 5 skipped of 962 (net10.0) and 955 / 5 of 960 (net8.0) — the
four Playwright cases plus one pre-existing skip. Gate on (`JINT_BROWSER_CLIENTS=1`): 961 / 1 of 962 and
959 / 1 of 960, both frameworks at once, so the two drivers do not contend. Also green under
`JINT_HOST_CONTRACT_VERIFICATION=1`.

`Jint.Tests.DevTools` 392, `Jint.Tests` 12 198 (net10.0 and net8.0) and 8 447 (net472),
`Jint.Tests.PublicInterface` 3 578 / 3 568 / 2 855, `Jint.Tests.SourceGenerators` 71,
`Jint.Tests.CommonScripts` 28, `Jint.Tests.Test262` 102 573 of 102 688 with **0 failures** on a quiet machine
(an earlier run under load timed two `intl402` cases out at 30 s; they take 3 s each). Full solution builds
clean.

A full Playwright run's protocol trace carries **no `-32601` at all**, and one `-32000`
(`Runtime.callFunctionOn` against a handle whose context a navigation replaced) that Chrome answers the same
way in the recording and that the client retries through.


sebastienros:
Amazing

PR #3736, opened 2026-09-03
Promise rejections: an unhandled one is reported at the microtask checkpoint, so a handler attached in the same turn is in time

### LINKED ISSUE #3711: An already-rejected promise fires unhandledrejection before a handler can attach: the tracker's cadence is not HTML's checkpoint
`HostPromiseRejectionTracker` fires at the tracker's own cadence rather than at HTML's microtask-checkpoint cadence, so a promise that is *already rejected* when it is handed to script fires `unhandledrejection` before the caller has had a chance to attach a handler. `Jint/WebApi/DiagnosticsSink` documents the cadence.

Where it showed: `custom-elements/CustomElementRegistry.html` in the WPT browser lane — the corpus's largest custom-elements file — becomes a whole-file harness error because `customElements.whenDefined("invalid name")` returns a rejected promise and the test attaches `.then(…, reject)` on the next line, which is HTML's normal shape for a rejected promise. It affects every Jint API that returns an already-rejected promise (`fetch` on a refused URL, `Response.json()` on a used body, `import()` of a missing module, …): under a browser's rules the rejection is reported only if it is still unhandled at the end of the current task's microtask checkpoint.

The fix is in the engine: defer the "unhandled" report to the end of the job (HTML: "notify about rejected promises" runs from the event loop, after the microtask checkpoint, over the "about-to-be-notified rejected promises" list), and clear the entry when a handler is attached in between. One `docs/v5-migration.md` row, since the timing is observable to a host's tracker.

Found by C6 (#3709); recorded rather than left as a TODO, per the campaign's rule.

lahma:
Fixes #3711

## What was wrong

`HostPromiseRejectionTracker` reported **where it was called**. `Engine.Tasks.PromiseRejectionTracker`, `Options.WebApi.Diagnostics.Sink` and the `unhandledrejection` / `rejectionhandled` global events all fired the instant a promise was rejected with nothing attached to it.

So a promise that is *already rejected* when script gets hold of it was announced before the caller could attach anything — and that is the normal shape for every API that returns one:

```js
const p = customElements.whenDefined("invalid name");   // already-rejected
p.then(resolve, reject);                                // one line later, too late
```

A browser raises nothing here: HTML's host hook only *records* the rejection, and [notify about rejected promises](https://html.spec.whatwg.org/multipage/webappapis.html#notify-about-rejected-promises) runs from the event loop over the promises that are **still** unhandled at a microtask checkpoint. Jint raised `unhandledrejection` and then `rejectionhandled`. Where it showed: `custom-elements/CustomElementRegistry.html` — the wpt corpus's largest custom-elements file — became a whole-file testharness ERROR, and the same held for `fetch()` on a refused URL, `Response.json()` on a used body and `import()` of a missing module.

## What changed

[`HostPromiseRejectionTracker`](https://html.spec.whatwg.org/multipage/webappapis.html#unhandled-promise-rejections) is split into recording and reporting, exactly as HTML has it:

- `Engine.OnPromiseRejectionTracker` moves the promise between HTML's two lists and reports nothing. `"reject"` adds it to *about-to-be-notified rejected promises*; `"handle"` takes it back out (and reports nothing at all), or, if it had already been announced, leaves the *outstanding* set owing a `rejectionhandled`.
- `Engine.NotifyAboutRejectedPromises` is the report. It copies and clears the list before firing anything — firing runs script, and a promise *that* script rejects belongs to the next checkpoint — and skips any promise whose `[[PromiseIsHandled]]` is true by then.
- The checkpoint is **the job queue running dry** in `EventLoop.RunAvailableContinuations`, which is already the point this engine treats as a task boundary (it is where a due timer is promoted), **plus the end of a host entry that completed**. The second one matters: `Invoke` and `Call` run script and drain nothing, so without it an embedder whose only entry is `predicate.Call(row)` would be told nothing at all until something else happened to pump the engine.
- `preventDefault()` on `unhandledrejection` is now HTML's *notHandled* (step 3, "the promise rejection is handled"), so it decides whether a handler attached later completes the pair. The `DiagnosticsSink` is still told either way — it is deliberately unsuppressible by script.
- Membership of the *outstanding rejected promises* set is a byte on `JsPromise` rather than a collection on the engine, because HTML's is a **weak** set: an engine-held set of every rejection ever announced would keep them all alive for the engine's lifetime. The about-to-be-notified list is a `List<JsPromise>` allocated on an engine's first unhandled rejection and cleared at every checkpoint, and `RestoreGlobalSnapshot` discards it beside the timer queue and the `Atomics` deadlines.

`Jint.DevTools` is fed by the same tracker and needed no change beyond its own documentation: Chrome reports neither `Runtime.exceptionThrown` nor `Runtime.exceptionRevoked` for a rejection handled on the same line, and now neither does this — so `exceptionRevoked` is the answer for a handler that arrives after the report rather than a correction issued a moment after it.

## Cost

Two additions sit on paths worth naming, and neither was benchmarked — flagging them rather than quoting a number:

- **`Engine.ExecuteWithConstraints`** and its memory-limit twin, which is the bracket every `Execute`, `Evaluate`, `Invoke` and `Call` goes through — so it is the hot path for the `foreach (var row in rows) predicate.Call(row);` shape. The addition is one `bool` local, and in the `finally` one field read plus a branch: `NotifyAboutRejectedPromises` returns on `_rejectionNotifications is null`, which is every engine whose script has never rejected without a handler.
- **`EventLoop.RunAvailableContinuations`**, at the exhaustion branch only — once per drain, not once per job, and the same field read.

Nothing is allocated until an engine's first unhandled rejection, and the reporting path itself is unchanged in cost; what moved is when it runs.

## Tests

New `Jint.Tests/Runtime/PromiseRejectionNotificationTests.cs` (engine-level, every target framework); the first three fail against unfixed code:

| test | before | after |
| --- | --- | --- |
| `ARejectionCaughtSynchronouslyIsNeverReported` | `Reject` + `Handle` | nothing |
| `ARejectionCaughtInTheSameCheckpointIsNeverReported` | `Reject` + `Handle` | nothing |
| `AnAwaitedAsyncFunctionThatThrowsIsNeverReported` | `Reject` + `Handle` | nothing |
| `ARejectionNothingHandlesIsReportedOnce` | `Reject` | `Reject`, at the checkpoint |
| `AHandlerAttachedInALaterTaskIsReportedAsHandled` | `Reject`, `Handle` | unchanged |
| `ARejectionDuringAHostInvokeIsReportedWhenTheEntryEnds` | `Reject` | unchanged |

plus `ARejectionInsideAJobIsReportedAtThatTurnsCheckpoint` and `APumpOnlyEngineIsToldAtTheTurnThatRejected` for the two pump shapes, `ARejectionDuringAHostInvokeIsReportedUnderAMemoryLimitToo` for the second copy of the host-entry bracket (it fails with the checkpoint removed from `ExecuteWithMemoryLimit` alone), and in `GlobalErrorEventTests`: `ARejectionHandledInTheSameCheckpointFiresNeitherEvent`, `AHandlerAttachedInALaterTaskFiresRejectionHandled`, `ARejectionHandledInALaterTimerTaskFiresBothEvents`, `AHandlerTheUnhandledRejectionListenerAttachesCompletesNoPair`, `ARejectionAListenerCausesIsAnnouncedAtTheNextCheckpoint` and `CancellingUnhandledRejectionMeansNoRejectionHandledLater`.

Five existing tests pinned the old cadence and now pin the new one — each became a *stronger* statement rather than a weaker one:

- `PromiseTests.AwaitedRejectionCaughtInsideAsyncFunctionIsTrackedThenHandled` asserted the internal-continuation path fires both operations. It now asserts **silence**, which pins the same thing harder: the `"handle"` call is what takes the promise out of the list, so an internal continuation that stopped making it would announce a caught rejection.
- `StreamRejectionTrackingTests.AFailedPipeStillReportsItsOwnRejection` reported five events (two announce-then-handle pairs the comment argued were unavoidable, plus the real one). It reports **one**.
- `FetchEventTests` — `…WaitUntilPromiseIsNotAnnouncedASecondTime` becomes `…NotAnnouncedAtAll`.
- `DiagnosticsTests` — `ARejectionHandledInTheSameTurnIsReportedAndThenMarkedHandled` becomes `…IsNotReportedAtAll`.
- `ExceptionSessionTests` — `ARejectionHandledOnTheSameLineIsReportedAndAtOnceRevoked` becomes `…IsNeverReported`, which is what Chrome does.

Runs, all Release: full `Jint.Tests` net10.0 **12 231 tests, 0 failed**; `Jint.Tests.DevTools` net8.0 + net10.0 **782, 0 failed**; `Jint.Tests.PublicInterface` net10.0 **3 603, 0 failed**; full `Jint.Tests.Browser` net8.0 + net10.0 **2 415, 0 failed**; solution build clean.

## wpt

Engine lane, `Jint.Tests.Wpt`: **680 cases, 0 failed** before and after, and no exclusion row went stale — that lane's driver fails a run in which an entry matches no failing test, so the unchanged green is the check.

Browser lane: the one row that named this defect is `custom-elements/CustomElementRegistry.html`, and it is in `NotVendored` — the file is not in the corpus, so there is nothing to un-exclude. Its reason is now spent, so the row moves to the "not vendored, and the cause has gone" group beside the four `document.createEvent` rows, with the reason rewritten to say so. **Vendoring it is deliberately left out of this PR**, for the reason `Jint.Tests/Wpt/AGENTS.md` gives those four: it moves the census's Documents and Tests columns, which the change that fixes an engine does not. The upstream directory this issue also names, `html/webappapis/scripting/processing-model-2/unhandled-promise-rejections/`, is out for an unrelated reason that has not changed (an iframe, a shared worker, a service worker and a wptserve `.py` handler). Browser lane before and after: **2 415 tests, 0 failed**, census unmoved.

## Docs

`docs/v5-migration.md` §4.122, since the timing is observable to any host with a tracker, a sink or a listener; `DiagnosticEvent.RejectionHandled`, `Jint/WebApi/AGENTS.md` and `README.md`'s diagnostics section all documented the divergence and now document the cadence.


lahma:
Rebased onto `main` after #3733 landed. Two things it changes in this PR:

- The migration row is now **§4.121** — #3733 took §4.120.
- #3733 added `Engine.CleanUpAfterRunningScript`, HTML's real *clean up after running script*, whose checkpoint has "notify about rejected promises" as its step 4. This PR deliberately does **not** run the notification from there, and `NotifyAboutRejectedPromises` now says why: HTML's step 4 only *drains* the list at a checkpoint and fires the events from a task it queues, so a rejection a listener leaves unhandled is announced after the dispatch either way — firing one inline from a listener boundary would put an event dispatch inside another dispatch to reach the same answer. Nothing is lost by waiting, because the queue running dry and the end of the host entry are both downstream of every listener boundary.

Re-ran everything on the new base, all Release: full `Jint.Tests` net10.0 **12 230, 0 failed**; `Jint.Tests.DevTools` net8.0 + net10.0 **782, 0 failed**; `Jint.Tests.PublicInterface` net10.0 **3 603, 0 failed**; full `Jint.Tests.Browser` net8.0 + net10.0 **2 403, 0 failed**; solution build clean.

PR #3745, opened 2026-09-03
XMLHttpRequest: an asynchronous timeout is a task on the event loop, so a long task holds it off instead of being interrupted by it

### LINKED ISSUE #3627: XMLHttpRequest's async timeout is a wall-clock deadline, not a task, so it fires during a long task where a browser's cannot
Found while landing #3626 (`XMLHttpRequest`), part of #3575; `xhr/xhr-timeout-longtask.any.js` is recorded in `Vendor/README.md`'s not-vendored table with this reason.

A browser's `XMLHttpRequest.timeout` is a **task**: it is queued on the event loop and cannot fire while a long-running script task holds the thread, so a request whose response arrives during a 300 ms busy loop that started before the 100 ms timeout still succeeds. Jint's async `timeout` is a wall-clock deadline on a CLR cancellation source, so it fires during the busy loop and the request is reported as timed out.

Fix: make the **async** `timeout` ride the engine's timer queue (`TimerQueue`, the same lane `setTimeout` uses), so it fires only when the loop gets a turn, and keep the **sync** timeout CLR-side (a sync request holds the thread by definition). Then vendor the file and retire the not-vendored row. Small; `Jint/WebApi/Xhr/XhrOperation.cs` is the one file.

lahma:
Fixes #3627

## What was wrong

A browser's `XMLHttpRequest.timeout` is a **task**: it is queued on the event loop and cannot fire while a long-running script task holds the thread. Jint's asynchronous `timeout` was a wall-clock deadline on a `CancellationTokenSource`, so it fired *inside* a long task and cancelled the transport mid-read. `xhr/xhr-timeout-longtask.any.js` — a 150 ms `timeout`, a response 100 ms out and a 200 ms busy loop entered from the same task as `send()` — was a not-vendored row for exactly that reason, because whether it passed depended on how loaded the machine was.

Pinned first, against unfixed code:

```
Expected engine.Evaluate("log.join('|')").AsString() to be "load|loadend", but "timeout|loadend" differs near "tim" (index 0).
```

## What changed

`XhrOperation` puts an asynchronous request's `timeout` on the engine's own `TimerQueue` — the lane `setTimeout` rides — instead of on a cancellation token source. The entry's task is [send()](https://xhr.spec.whatwg.org/#dom-xmlhttprequest-send)'s "set this's timed out to true and terminate this's fetch controller" followed by [handle errors](https://xhr.spec.whatwg.org/#handle-errors), whose first step — *"if xhr's send() invoked is false, then return"* — is what makes a deadline the loop reaches only after the response was delivered settle nothing a second time. A timer is promoted only once the job queue has run dry (`EventLoop.RunAvailableContinuations`), so the completion job the transport queued during the busy loop wins, and the deadline behind it is a no-op.

Re-arming from the `timeout` setter still measures from when the request was sent, which is the attribute's own rule (*"it will still be measured relative to the start of fetching"*): the old entry is cancelled — O(1), discarded when it surfaces — and a new one scheduled for what remains. `WebApiFeatures.XmlHttpRequest` joins `WebApiRegistration.NeedsTimerQueue` so the queue exists on an engine that named the interface and no timers. The entry does not consult `MaxActiveTimers`, exactly as `requestIdleCallback`'s own timeout does not: it is not a timer a script registered, and `Options.WebApi.Fetch.MaxConcurrentRequests` is what bounds how many can exist at once.

**What deliberately did not change.** The synchronous request keeps its CLR-side deadline, which is the shape the specification gives it too — its arm *pauses* until the response or the timeout, where the asynchronous arm waits in parallel — and so does the host's `Options.WebApi.Fetch.Timeout` (30 s by default), which is what still lets an engine nobody pumps go of its socket. The consequence worth stating: an engine that is never pumped no longer honours the script's `timeout`, only the host's.

## Tests

- `XmlHttpRequestTests.ALongTaskHoldsOffTheTimeoutSoTheResponseStillArrives` — the wpt file as a unit test, over a stub handler: `timeout = 50`, a response 150 ms out, a 600 ms busy loop from the same task as `send()`. Fails on unfixed code with the message above.
- The stub handler grew a `Delay` that honours the cancellation token it is handed, so the unfixed run fails deterministically rather than by timing.
- `TheTimeoutAttributeFiresTheTimeoutEvent` (unchanged assertions) still pins that a transport which never answers settles as `timeout|loadend`; its doc comment now says where the deadline lives.

## wpt, before and after

`xhr/xhr-timeout-longtask.any.js` is vendored and its `_notVendored` row retired. The whole driver is green with `JINT_WPT_CENSUS=1`: **681 passed, 0 failed**. The census moved by exactly the new file — XMLHttpRequest 43 → 44 files, 287 → 288 assertions, **not passing 8 → 8** — so no exclusion row went stale and nothing else in the corpus moved. Totals 382 → 383 files, 41,581 → 41,582 assertions, 2,932 not passing unchanged. `Jint.Tests` for `WebApi|Wpt|Temporal|Intl|MigrationGuideTests|AgentInstructionFileTests|SpecCitationTests`: 4,123 passed. `Jint.Tests.PublicInterface`: 3,582 passed.

## Left out

`xhr/abort-after-timeout.any.js` stays not vendored: it asks for `?pipe=trickle(d1)`, a wptserve directive the driver's server does not implement — unrelated to this. `docs/v5-migration.md` gets no new section: `XMLHttpRequest` is new in v5 and §5.13 is its entry, so the sentence there that said both deadlines were CLR-side was amended instead.


lahma:
The first shape of this raced the network: both the unit test and the re-vendored `xhr-timeout-longtask.any.js` needed a delayed response to arrive before a busy loop ended, which a loaded runner cannot promise, and macOS, Windows and ARM all reported the deadline winning while Linux passed. Fixed in `c622873b9`: the unit test gates the transport and lets the deadline come due while nothing pumps, which is what a long task looks like from the engine's side, so the assertion is pure ordering. It fails against the wall-clock deadline and passes against the timer-queue task, verified both ways. The wpt file goes back on the not-vendored list with its reason rewritten: the deadline is no longer the machine-dependent half, but the response's arrival within the busy window still is.

Issue #3523, opened 2026-08-31
Temporal: a persian month addition is not additive at the end of the BCL table


lahma:
Found while validating #3520. It is present on `main` (`f6fa93b`) and unchanged by the work there, which reproduces it exactly.

## What happens

```js
const P = s => Temporal.PlainDate.from(s).withCalendar('persian');

P('9999-06-01').add({ months: 8 });                                   // +010000-02-01
let d = P('9999-06-01'); for (let i = 0; i < 8; i++) d = d.add({ months: 1 });
d.toString();                                                          // +010000-01-31
```

Eight months in one step and eight single steps land a day apart. Steps +1 to +7 agree exactly; the two
spellings part at +8 and stay a day apart after it (+9 is `+010000-03-02` against `+010000-03-01`).

## Why

`PersianCalendar.MaxSupportedDateTime` is ISO 9999-12-31, which falls *inside* Persian year 9378 — so as
far as the BCL table is concerned that year holds **ten** months: `GetMonthsInYear(9378)` answers 10 and
`GetDaysInMonth(9378, 11)` throws. Jint reckons every Persian year as twelve months
(`NonIsoCalendars.GetMonthsInYear`, `PersianDateToIso`), so a date in month 11 or 12 of 9378 is placed by
`PersianAlgorithmicToIso` instead, whose day numbering is one off the table's there.

A step that *stops* inside month 11 of 9378 therefore takes the arithmetic answer and carries it forward,
while a step that flies over that month lands on a year the table places and takes the table's. That is
the whole of the discrepancy.

The same shape exists at the other partial year in a band — Hebrew 5343, where `MinSupportedDateTime` sits
inside the year — but there only `ToDateTime` refuses, and no answer differs today.

## Scope

Persian 9378 months 11 and 12, which is ISO 9999-03-22 to 9999-12-31: the last nine months the .NET table
covers, at the very end of `DateTime`'s own range.

## Possible fixes

- Ask the table how many months a year holds rather than assuming twelve, so a partial year is recognised
  as one; or
- narrow the trusted band to whole years, so a year the table covers only partly is answered entirely by
  the arithmetic reckoning. #3520's fix adds `NonIsoCalendars.SupportsYear`, which is the one place that
  decides which reckoning answers for a year, so either fix has a single site.

Either one changes answers for those nine months, so it wants its own migration-guide row and the same
equivalence sweep.

PR #3751, opened 2026-09-03
Temporal: the persian calendar extends into proleptic years on the 33-year cycle, so the ends of Temporal's range land in the right Persian year

### LINKED ISSUE #3604: Persian eraYear is off by one at the calendar's minimum date, and three extreme-dates exclusions hang on it
Found while executing #3591: `intl402/Temporal/ZonedDateTime/prototype/withCalendar/extreme-dates.js` was on that issue's candidate list but does not pass — and its failure is sharper than its exclusion comment suggested. The `year` assertion already passes; only `eraYear` fails, off by exactly one at the persian calendar's minimum supported date:

```
persian minimum supported date: eraYear result: Expected SameValue(«-272443», «-272442»)
```

The whole file runs in `"UTC"`, so no time-zone or ICU data is involved — this is Jint's own era arithmetic at the calendar's lower bound. Two near-identical siblings (`PlainDate` and `PlainDateTime` `withCalendar/extreme-dates.js`) remain excluded and presumably share the defect; whoever fixes it should expect to remove all three exclusions together (6 test262 cases).

`git diff` over `Jint/Native/Temporal/` between the #3591 measurement base and current `main` is empty, so nothing recent regressed it — it has been failing at least since the exclusions were added. The restored exclusion entry (#3603) names the assertion.

lahma:
Fixes #3604

## What was wrong — and it is not the era

`System.Globalization.PersianCalendar` spans ISO 622-03-22 to 9999-12-31 (Persian 1 to 9378) and derives those years astronomically. Everything outside them is answered by an arithmetic rule, because no ephemeris covers a quarter of a million years — and the rule Jint carried was the **2820-year cycle** (Reingold–Dershowitz / Birashk), which is not the one the platform's `persian` calendar is. ICU's `PersianCalendar` uses the **33-year cycle**: a year's first day is `365 × (year − 1) + floor((8 × year + 21) / 33)` days after the epoch, and a year is a leap year when `floorMod(25 × year + 11, 33) < 8`.

Over Temporal's own range the two drift about two months apart, so both ends came back in the wrong Persian year:

| ISO | Jint, before | test262 / ICU |
| --- | --- | --- |
| `-271821-04-20` | `-272443-M11-08` | `-272442-M01-10` |
| `+275760-09-13` | `275139-M09-14` | `275139-M07-12` |

**The issue's reading of the symptom was wrong, and worth recording.** It reported that only `eraYear` failed and that `year` already passed. `TemporalHelpers.assertPlainDateTime` asserts `era`, then `eraYear`, then `year` — so the run stopped at `eraYear` and never reached `year`. `persian` has a single era, `ap`, so `eraYear` **is** `year` (`NonIsoCalendars.CalendarEraYear`), and both were off by the same one. The era arithmetic was never involved.

Against unfixed code, all six cases of the three files fail identically on `eraYear`, expecting `-272442` and getting `-272443`; the fourth file, which asserts the way back, fails by 61 days.

## What changed since this was parked

The parked draft failed four `Jint.Tests` cases on every CI leg — `linux`, `linux - ARM`, `macos` and `windows` alike, with byte-identical actual values — while the local run passed. The comment on the draft read that as the platform's calendar data differing per runner. It is not: the four are `HebrewMonthSteppingTests` cases whose expected persian values were recorded under the 2820-year cycle and lie outside `PersianCalendar`'s window, so changing the rule legitimately moved them. The local run passed because the filter that was run (`~Temporal|~Intl|~Calendar|…`) does not match `HebrewMonthSteppingTests`. Five platforms agreeing to the day is what says so.

The remedy asked for stands anyway, and is the substance of the second commit: **the hand-off is now Jint's to place rather than the platform's.**

- `PersianToCalendarDate` compared against `PersianCalendar.MinSupportedDateTime` / `MaxSupportedDateTime`, and `PersianDateToIso` asked the same question by catching the `ArgumentOutOfRangeException` `ToDateTime` raises outside them. So "outside the table" was whatever calendar data the runner shipped, and a runtime that moved its window by a day would have moved every proleptic answer with it — with the only symptom a date a quarter of a million years away landing somewhere else on one leg.
- It is now an ISO window this repository states: **622-03-22 through 9999-12-31**, which is Persian 1-01-01 through 9378-10-13, held as `PersianTableFirstJdn` / `PersianTableLastJdn` and, for the direction that asks in Persian fields, `PersianTableHolds`. The `try`/`catch` goes with it.
- The four persian expectations past ISO 9999 in `HebrewMonthSteppingTests` move with the reckoning underneath them. A step is still taken the same way and the two spellings of it still agree on every case; the answer they agree on is now the cycle's. The class remarks say so and point at the boundary tests.

## The delegation stays, and here is the measurement

The honest question was whether delegating still earns its keep — if the cycle agreed with `PersianCalendar` throughout the window, using the cycle everywhere would be simpler than a hand-off. It does not agree. Measured over the 9,377 whole years the two share, identically on `net472` and on .NET 10:

| | |
| --- | --- |
| year-starts that differ | **4,144 of 9,377**, by −1 to +3 days |
| days placed in a different Persian month | **1,514,413 of 3,425,164** |
| disagreeing years per millennium AP | 66 · **19** · 69 · 128 · 241 · 446 · 798 · 1000 · 1000 |
| years 1300–1500 AP | **none** — no date anybody keeps moves |

The table derives a year from the true vernal equinox at Tehran and is what makes a real Persian date right; the cycle is what every other Temporal implementation answers a proleptic date with. Both are needed, so both stay, and the region comment now records where and why they differ.

Making `persian` ICU's calendar *throughout*, rather than only outside the window, is a much larger change with a much larger blast radius and is still not this fix.

## Tests

`Jint.Tests/Runtime/PersianCalendarBoundaryTests.cs` (15 cases) is what makes the window hold, and nothing in it is a date written down because a run once produced it:

- `ThePlatformsPersianTableCoversExactlyTheWindowJintStates` asserts the four numbers against the platform — the ISO window, the Persian fields at each end, and that `ToDateTime` accepts up to 9378-10-13 and refuses past it — so a runtime that ever disagrees fails **there**, naming the reason, instead of silently moving the proleptic years.
- `TheHandOffHappensOnTheDayTheWindowNamesAndOnNoOther` sweeps every day for four hundred either side of each edge: inside, the answer must be the table's; outside, the cycle's. Moving either constant by one day fails it and two of its neighbours — checked by doing exactly that.
- `TheTwoRulesDisagreeAtBothEdgesOfTheWindow` pins that the seam is a seam, so the sweep is asserting a choice and not a coincidence.
- `AProlepticYearIsAsLongAsTheCyclesLeapRuleSaysAndTheNextStartsThere` and `EveryMonthOfAProlepticYearIsWhereTheCyclePutsIt` assert Jint against the 33-year cycle **written out in the test file** — a test that asked the engine what the engine's arithmetic ought to be would say nothing — over 1,090 proleptic years and every day of five of them, in both directions. That cycle is integer arithmetic on a fixed epoch, so its answers are the same on every runner by construction.
- `TheYearTheWindowStopsInsideIsStillAWholeYear` keeps #3523's answer pinned where the two rules meet.

`NonIsoCalendarTests.ThePersianCalendarPlacesTheEndsOfTemporalsRangeWhereTheStandardLibrariesDo` pins both extremes of Temporal's range with test262's own numbers; against unfixed code it fails with `Expected value to be -272442.0, but found -272443.0`. `ThePersianCalendarStillPlacesTheYearsItsTableCovers` pins Nowruz 1403 (2024-03-20) and 22 Bahman 1357 (1979-02-11), and passes on both sides of the change, which is what says the table-backed range is untouched.

## Exclusions removed

Four files, eight cases:

- `intl402/Temporal/PlainDate/prototype/withCalendar/extreme-dates.js`
- `intl402/Temporal/PlainDateTime/prototype/withCalendar/extreme-dates.js`
- `intl402/Temporal/ZonedDateTime/prototype/withCalendar/extreme-dates.js`
- `intl402/Temporal/ZonedDateTime/from/extreme-dates.js` — the issue expected three; this one was excluded in the same block, fails on the same date before the change and passes after, so it goes too

`intl402/Temporal/PlainDateTime/from/extreme-dates.js` stays, and now carries the reason it actually fails for: one of its extremes raises `Date is outside valid range` from `PlainDateTime.from`, which is a range check rather than a calendar. `PlainYearMonth/from/extreme-unsupported-dates.js` stays as it was.

## Runs

- **Test262: 102,581 passed, 0 failed, 107 skipped** — the whole suite, after the rebase onto `main`.
- `Jint.Tests` `net10.0`, whole project: **12,282 passed, 0 failed**, 5 skipped.
- `Jint.Tests` `net472`, the calendar and stepping classes: 160 passed — the second target the window was measured on.
- `Jint.Tests.PublicInterface` `net10.0`: 3,590 passed.
- Solution build in Release: clean.

No migration-guide row: nothing public moved, and the dates that changed are outside the range any `PersianCalendar`-backed answer ever covered.

### DIFF
--- a/Jint.Tests.Test262/Test262Harness.settings.json
+++ b/Jint.Tests.Test262/Test262Harness.settings.json
@@ -212,24 +212,21 @@
     "intl402/Temporal/PlainDate/prototype/daysInMonth/basic-islamic-umalqura.js",
     "intl402/Temporal/PlainDate/prototype/daysInYear/basic-islamic-umalqura.js",
     "intl402/Temporal/PlainDate/prototype/inLeapYear/basic-islamic-umalqura.js",
-    "intl402/Temporal/PlainDate/prototype/withCalendar/extreme-dates.js",
+    // Still out, but on a range check rather than a calendar: one of its extremes raises
+    // "Date is outside valid range" from PlainDateTime.from. The persian rows it shares with the
+    // withCalendar files pass (https://github.com/sebastienros/jint/issues/3604).
     "intl402/Temporal/PlainDateTime/from/extreme-dates.js",
     "intl402/Temporal/PlainDateTime/prototype/daysInMonth/basic-islamic-umalqura.js",
     "intl402/Temporal/PlainDateTime/prototype/daysInYear/basic-islamic-umalqura.js",
     "intl402/Temporal/PlainDateTime/prototype/inLeapYear/basic-islamic-umalqura.js",
-    "intl402/Temporal/PlainDateTime/prototype/withCalendar/extreme-dates.js",
     "intl402/Temporal/PlainMonthDay/from/islamic-umalqura-month-codes.js",
     "intl402/Temporal/PlainYearMonth/from/extreme-unsupported-dates.js",
     "intl402/Temporal/PlainYearMonth/prototype/daysInMonth/basic-islamic-umalqura.js",
     "intl402/Temporal/PlainYearMonth/prototype/daysInYear/basic-islamic-umalqura.js",
     "intl402/Temporal/PlainYearMonth/prototype/inLeapYear/basic-islamic-umalqura.js",
-    "intl402/Temporal/ZonedDateTime/from/extreme-dates.js",
     "intl402/Temporal/ZonedDateTime/prototype/daysInMonth/basic-islamic-umalqura.js",
     "intl402/Temporal/ZonedDateTime/prototype/daysInYear/basic-islamic-umalqura.js",
-    "intl402/Temporal/ZonedDateTime/prototype/inLeapYear/basic-islamic-umalqura.js",
-    // Only the persian minimum date still fails here, on eraYear alone: -272443 for the -272442
-    // the file asserts, while its `year` of -272442 is already right.
-    "intl402/Temporal/ZonedDateTime/prototype/withCalendar/extreme-dates.js"
+    "intl402/Temporal/ZonedDateTime/prototype/inLeapYear/basic-islamic-umalqura.js"
 
   ],
   // Timing-sensitive suites: serialise the whole generated test method instead of
--- a/Jint.Tests/Runtime/HebrewMonthSteppingTests.cs
+++ b/Jint.Tests/Runtime/HebrewMonthSteppingTests.cs
@@ -1,4 +1,4 @@
-#nullable enable
+﻿#nullable enable
 
 namespace Jint.Tests.Runtime;
 
@@ -25,6 +25,16 @@ namespace Jint.Tests.Runtime;
 /// a regression rather than a matter of taste
 /// (<see href="https://github.com/sebastienros/jint/issues/3520"/>).
 /// </para>
+/// <para>
+/// <b>The four Persian values past ISO 9999 moved once.</b> Not because a step is taken differently — the
+/// two spellings still agree on every one of those cases — but because the reckoning underneath them
+/// became the 33-year cycle rather than the 2820-year one, which is what every other Temporal
+/// implementation answers a proleptic <c>persian</c> date with
+/// (<see href="https://github.com/sebastienros/jint/issues/3604"/>). Those years lie outside
+/// <c>PersianCalendar</c>'s window, so nothing but integer arithmetic on a fixed epoch decides them; where
+/// that window ends, and that it is Jint's to state rather than the runner's, is
+/// <see cref="PersianCalendarBoundaryTests"/>.
+/// </para>
 /// </remarks>
 public class HebrewMonthSteppingTests
 {
@@ -40,7 +50,7 @@ public class HebrewMonthSteppingTests
     [TestCase("hebrew", 100_000, "+010085-03-13")]
     [TestCase("hebrew", 300_000, "+026255-08-10")]
     [TestCase("hebrew", 3_000_000, "+244556-01-22")]
-    [TestCase("persian", 3_000_000, "+251999-10-18")]
+    [TestCase("persian", 3_000_000, "+251999-12-13")]
     public void ABulkMonthAdditionDoesNotWalkYearByYear(string calendar, int months, string expected)
     {
         var engine = new Engine(options => options.LimitStatements(2));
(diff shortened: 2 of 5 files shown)


lahma:
Marked draft: this is not ready, and the reason is worth writing down.

`AStepAcrossTheEndOfTheTableLandsWhereItAlwaysDid("persian", …)` fails on `linux - ARM` and `windows` while Linux and the local run pass, and the values differ by a month at year 10001 and by nine days at year 43666. That is the platform's calendar data talking: the change delegates to `System.Globalization.PersianCalendar` inside its supported range and takes the 33-year cycle outside it, but where that range ends is ICU's answer, and ICU is not the same version on every runner. So "outside the table" is a different set of dates per platform, and a test that names absolute strings cannot hold.

The fix has to make the boundary ours rather than the platform's: pick the ISO window the delegation covers as a constant we state, and compute everything outside it from the cycle, so the same input gives the same answer on every runner. The 33-year cycle itself and the epoch correction look right — test262's persian and extreme-date files pass everywhere, including the four exclusions this retires — so it is only the hand-off that is unpinned.

Parked rather than abandoned; it goes back in the queue for the next free agent slot.

lahma:
**Correcting my earlier comment: the platform diagnosis was wrong.** I said the four failures were ICU-version dependent. They are not. All four legs reported byte-identical values, and .NET's `PersianCalendar` is a managed astronomical computation rather than an ICU-backed table, so its window never varied by runner. What actually failed were `HebrewMonthSteppingTests` cases whose Persian expectations had been recorded under the old 2820-year cycle at ISO dates past 9999 — stale witnesses that the rule change legitimately moved, and which the local test filter did not select, which is why the local run looked clean.

The boundary work was done anyway and stands on its own: the hand-off is now ISO 622-03-22 through 9999-12-31 as named constants rather than the platform's own min and max plus a catch-all, so the same input gives the same answer everywhere. Keeping the delegation was measured, not assumed: over the whole window the 33-year cycle and `PersianCalendar` disagree on 4,144 of 9,377 year starts, and none between 1300 and 1500 AP, so replacing it with the cycle would have moved real dates. Out of draft.

Issue #3712, opened 2026-09-03
DOM binding generator: a nullable DOMString parameter, and a value-returning hooks form


lahma:
Two things the DOM binding generator (`tools/dom-bindings/`) cannot express, found while C6 (#3709) built `customElements`.

1. **A nullable `DOMString?` parameter.** `getAttributeNS(null, name)` answers `null` because a nullable string parameter goes through `TypeConverter.ToString`, which turns JavaScript `null` into the string `"null"`. This is the single largest cause of failures in the vendored `custom-elements/` suites (every row of `attribute-changed-callback.html`) and has nothing to do with custom elements: `getAttributeNS`, `setAttributeNS`, `hasAttributeNS`, `createElementNS`, `getElementsByTagNameNS` and every other namespaced member take a nullable namespace. The generator should read AngleSharp's `String?` (nullable annotations) as WebIDL `DOMString?` and pass `null` through.
2. **A value-returning `hooks` form.** A `hooks` entry can only return `undefined`, so `createElement`, `createElementNS` and `cloneNode` had to be declared through `skip` + `additions` (a re-implementation rather than a hook around the generated member). A hook that may answer the value keeps the generated member and the override table smaller.

Recorded rather than left as TODOs, per the campaign's rule.

Issue #3656, opened 2026-09-02
DOM binding generator: hook-backed and added members, [LegacyUnforgeable], and IWindow returns


lahma:
Found while landing #3648 (the `Jint.Browser` page runtime), part of #3575.

Three things the page runtime had to put on **instances** because the binding generator (`tools/dom-bindings/`) cannot express them on prototypes: `document.defaultView` and `document.currentScript` (getter hooks the runtime answers, AngleSharp's own values being wrong or absent — see AngleSharp#1310-ish findings), and `Location`'s `href` setter, `assign`, `replace`, `reload`, `toString` (`[LegacyUnforgeable]` in the IDL, and members AngleSharp's `ILocation` does not carry in the shape the DOM wants). Also, `iframe.contentWindow` is `undefined` rather than `null` because the conversion table drops `IWindow`-returning members entirely.

Ask, for the generator: an `overrides.json` `hook` kind that emits a prototype accessor/method whose body calls an `IDomHostHooks` member (the seam R3 already uses for parser members), a way to *add* a member the AngleSharp interface lacks (name, kind, hook), an `unforgeable` flag that emits own non-configurable properties at wrap time, and a conversion-table entry that maps an `IWindow` return to `null` (or to the window value once frames script). Then move the five members onto the prototypes and pin `Object.getOwnPropertyNames(document)` as empty.

lahma:
Done in substance by #3776, which said `Part of #3656` rather than `Fixes` on purpose. Closing by hand, and recording **why the three remaining items were refused rather than deferred** — otherwise the issue title reads as three unbuilt features and the work gets rediscovered.

**1. The generator's `unforgeable` flag.** The only two members that would need it are `Window`'s, and `Window` is an `excludedInterfaces` entry — so the flag would have shipped as a branch nothing exercises. A generator flag with no live case is worse than no flag: it looks supported and has never run.

**2. The `additions`-with-a-`hook` form.** `defaultView` was its sole candidate, and the `IWindow` conversion #3776 added turns it into an ordinary generated member instead. If an addition ever needs to reach a hook, it can already call one from its `body`.

**3. `Location`'s unforgeable members.** Checked rather than assumed: `LocationInstaller` already defines them with `PropertyFlag.OnlyEnumerable` — own, non-configurable, non-writable, enumerable — which **is** what `[LegacyUnforgeable]` prescribes. Moving them onto a prototype would make them *less* correct, not more.

[...]

Issue #3636, opened 2026-09-02
JsObjectShape has no symbol-keyed members, so generated prototypes cannot carry Symbol.iterator or Symbol.toStringTag


lahma:
Found while landing #3634 (generated DOM bindings), part of #3575.

`JsObjectShape.Builder` has no symbol-keyed member, so a generated prototype cannot carry `Symbol.iterator` (or `Symbol.toStringTag` as a real property, or `Symbol.unscopables`). The bindings therefore put `Symbol.iterator` on every collection **instance**, which is wrong in two observable ways: `NodeList.prototype[Symbol.iterator]` is `undefined`, and each instance pays an own property where a browser has none.

Ask: symbol-keyed members on `JsObjectShape.Builder` (method and accessor, same flags model), so `Symbol.iterator` for the array-like collections and `Symbol.toStringTag` for every interface prototype live on the prototype like everything else. Additive; the WebIDL attribute rules in `Jint/WebApi/AGENTS.md` apply (`@@iterator` is writable/configurable, non-enumerable; `@@toStringTag` configurable only).

PR #3781, opened 2026-09-03
DOM bindings: a union parameter takes both its arms, and a form and a select answer by index and by name

lahma:
Sits on **#3776** (issue #3656) → **#3761** (#3712) → **#3753** (#3636). Review bottom-up.

## What was wrong

```js
el.append('text')                 // TypeError  (a browser inserts a text node)
select.add(optgroup)              // TypeError  (a browser accepts it)
form[0], form.username, select[0] // undefined
```

## What changed

**`(Node or DOMString)...`.** AngleSharp's signature is `INode[]`, and the string half of the union needs a document to create a `Text` node against — which a static member body had no way to reach. It had one all along: the *receiver* is in scope in every emitted body. `DomConvert.NodeOrTextRest` stringifies a non-node argument and creates the node in the receiver's node document; the generator selects it for any variadic `INode` parameter, which is `append`, `prepend`, `before`, `after` and `replace` across `Element`, `Document`, `DocumentFragment`, `CharacterData` and `DocumentType`.

**The two union-typed operations.** `select.add` and `HTMLOptionsCollection.add` are `(HTMLOptionElement or HTMLOptGroupElement)`, which AngleSharp models as two overloads sharing one `[DomName]`, so one arm always lost. They are `skip` + `additions` entries over `DomUnionMembers` now, dispatching on the argument's interface most-derived-first, with an argument that is neither arm still Web IDL's conversion failure. A `skip` also suppresses the duplicate-declaration diagnostic — a member the table has taken over is a decision, not news — so the generator now reports **no** diagnostics at all, and `DomBindingsStalenessTests` asserts emptiness, which is a stronger claim than the two it used to pin.

**`form[0]`, `form.username`, `select[0]`.** A node wrapper is a `JsEventTarget` and carried no property projection; it cannot become an `ArrayLikeObject` either, because the tree-dispatch lane keys on the node wrapper and the cache keeps exactly one per node — a form that became a collection would stop being an `EventTarget` the dispatcher can walk. So `Collections/DomIndexedNodeObject` is a node wrapper with the interface's *generated* accessor projected on top, behind three overrides (`GetOwnProperty`, `ProbeOwnProperty`, `GetOwnPropertyKeys`) that all read the accessor at the same instant — which is what keeps `hasOwnProperty` and `Object.getOwnPropertyNames` agreeing about one object, the obligation `Jint/Native/Object/AGENTS.md` states. Overriding `GetOwnProperty` and **not** `Get` keeps the receiver on the engine's ordinary property-access lane, so an inherited member still resolves in one probe and the prototype-method cache still serves it.

`HTMLFormElement`'s supported property names are computed rather than read from metadata: HTML makes them the `name` content attributes and the ids of the form's listed elements, and `name` is not an IDL member of `IElement` at all, so no `[DomName]` heuristic could have found it. It is the one named getter over elements in the surface, so the rule is written once in the emitter.

## Tests

`Jint.Tests.Browser/DomIndexedNodeTests` — six tests: a form by index and by name (including by id, and identical to what `form.elements` answers), a select by index and through the `Array.prototype` generics, the coherence of every way of asking about a projected property plus read-only indices and expandos, that a projected node is still a node whose event path works, both arms of both unions plus the failure for neither, and strings becoming text nodes through `append`/`prepend`/`before` and on a `DocumentFragment`.

Regenerated with `JINT_DOM_BINDINGS=update`; staleness test green. `Jint.Tests.Browser` 1225 net10.0 / 1222 net8.0 green — **the wpt browser lane needed no exclusion or census edit for this PR**. The `Shape|MigrationGuide|AgentInstructionFile|SpecCitation` filter (210) and `Jint.Tests.SourceGenerators` (71) green. No engine API changed, so no migration-guide row.

## What was left out

- **The three items are one commit, not three.** They share one regenerated artefact (`DomShapes.*.g.cs`, `DomInterfaces.g.cs`, `DomCollectionAccessors.g.cs`), so splitting them would have produced intermediate commits whose checked-in bindings did not match their emitter — which is exactly what the staleness test exists to forbid. The commit body separates the three.
- **A found divergence, not fixed here:** AngleSharp spells DOM's `ChildNode.replaceWith` as `replace`, so `el.replaceWith(x)` is `undefined` and `el.replace(x)` is the member. It is three `additions` rows to alias, and it is not on this issue's list; recorded for the lead.
- A form's named properties are the elements' `name` and `id`; HTML's rule also scopes them to *listed* elements, and AngleSharp's indexer decides what the form contains. Whatever it contains is what enumerates, which is one source of truth rather than two.

### DIFF
--- a/Jint.Browser/Dom/AGENTS.md
+++ b/Jint.Browser/Dom/AGENTS.md
@@ -44,7 +44,7 @@ version of AngleSharp nobody references.
 | --- | --- |
 | `excludedInterfaces` | An interface the runtime owns instead. Today: `IWindow` (campaign item R1). |
 | `manual` | An interface whose shape is hand-written in `DomManualShapes`. Today: `IHtmlCollection<T>`, whose generic invariance keeps a member body from naming its receiver. A manual interface contributes no members to any closure — its children inherit them through the prototype chain. |
-| `skip` | A member whose AngleSharp implementation must not be projected: navigation (`location.assign`, `location.href`'s setter), the parser (`document.open`/`close`/`load`), `document.createEvent` whose AngleSharp `Event` must never reach script, `DOMImplementation.createHTMLDocument` whose title AngleSharp makes required, and the six the events bridge re-declares because AngleSharp's own do nothing — see the divergence table. Every one of them is re-declared in `additions`, so a skip here is a *replacement* and not an absence. `half: "setter"` skips the write half only. |
+| `skip` | A member whose AngleSharp implementation must not be projected: HTML's two union-typed operations (`select.add`, `HTMLOptionsCollection.add`, whose `(HTMLOptionElement or HTMLOptGroupElement)` AngleSharp models as two overloads sharing one `[DomName]`, so one arm always lost — `DomUnionMembers` dispatches on both, and the entry is also what turns the collision from a diagnostic into a decision), navigation (`location.assign`, `location.href`'s setter), the parser (`document.open`/`close`/`load`), `document.createEvent` whose AngleSharp `Event` must never reach script, `DOMImplementation.createHTMLDocument` whose title AngleSharp makes required, and the six the events bridge re-declares because AngleSharp's own do nothing — see the divergence table. Every one of them is re-declared in `additions`, so a skip here is a *replacement* and not an absence. `half: "setter"` skips the write half only. |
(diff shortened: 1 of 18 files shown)


sebastienros:
Superseded by #3872, which adapts this work to current main, preserves Marko Lahma's authored commits, and carries the DOM union and indexed form/select integration through the full browser matrix. #3872 is now merged. Thank you for the original implementation.

Issue #3768, opened 2026-09-03
Eleven Node and Element members the DOM standard requires are not in the bindings — 182 wpt rows


lahma:
The browser lane's new DOM suites reach for eleven members that are simply not there. Each failure is `Property '…' of object is not a function` before the test can say anything about behaviour, so they are cheap to find and cheap to fix, and together they are **182 failing tests across 18 documents**.

### The members

| Member | Standard | Rows | Where AngleSharp stands |
| --- | --- | ---: | --- |
| `ChildNode.replaceWith` | [DOM §4.2.7](https://dom.spec.whatwg.org/#dom-childnode-replacewith) | 34 | `INode` has no member for it |
| `Element.getAttributeNode` / `setAttributeNode` / `getAttributeNodeNS` / `setAttributeNodeNS` / `removeAttributeNode` | [DOM §4.9](https://dom.spec.whatwg.org/#dom-element-getattributenode) | 29 | `IElement` has `Attributes`, and these are the `Attr`-node spellings |
| `ParentNode.replaceChildren` | [DOM §4.2.6](https://dom.spec.whatwg.org/#dom-parentnode-replacechildren) | 24 | absent |
| `Element.insertAdjacentElement` / `insertAdjacentText` | [DOM §4.9](https://dom.spec.whatwg.org/#dom-element-insertadjacentelement) | 17 | absent; `insertAdjacentHTML` is hooked and present |
| `Element.toggleAttribute` | [DOM §4.9](https://dom.spec.whatwg.org/#dom-element-toggleattribute) | 9 | absent |
| `Element.hasAttributes` | [DOM §4.9](https://dom.spec.whatwg.org/#dom-element-hasattributes) | 14 | absent |
| `Node.isSameNode` | [DOM §4.4](https://dom.spec.whatwg.org/#dom-node-issamenode) | 8 | absent (legacy, but normative) |
| `Element.getAttributeNames` | [DOM §4.9](https://dom.spec.whatwg.org/#dom-element-getattributenames) | 4 | absent |
| `Element.webkitMatchesSelector` | [DOM §4.9](https://dom.spec.whatwg.org/#dom-element-matches) | 3 | the standard requires the alias |
| `NodeList.keys` / `values` / `entries` / `forEach` | WebIDL iterable | 4 | see sebastienros/jint#3636 |
| `XMLDocument` | [DOM §4.5](https://dom.spec.whatwg.org/#xmldocument) | 1 | not a name at all |

[...]

Issue #3641, opened 2026-09-02
GarbageCollectionTests' source-text retention measurements flake on the macOS CI leg


lahma:
Two occurrences in two days on the `macos` CI leg, both on PRs that touch no engine code, both green on the other six legs and on re-run:

- 2026-09-01, #3586 (a new package, zero files under `Jint/`): `GarbageCollectionTests` — "Expected retained to be greater than 10000000L because RetainFunctionSourceText did not keep the source text alive".
- 2026-09-02, #3634 (bindings package, zero files under `Jint/`): `PreparedScriptsDoNotRetainSourceTextByDefault` — "Expected notRetained to be less than 5000000L … Not retained: 8.0 MB, Source: 19.1 MB, found 8438280".

#3435 already moved these measurements to per-phase deltas, and they still read a fraction of the expected saving on macOS: the runner's GC returns a different `GetTotalMemory` residue than Linux/Windows, and 8 MB against a 19 MB source is exactly the shape a coupled-residue reading has. Two flakes in two days on a leg that costs a re-run each time is enough to act on.

Options, in order of preference: measure with `GC.GetTotalAllocatedBytes(precise: true)` deltas (allocation, not residency, is what "retained a copy" changes) or `GC.GetGCMemoryInfo().HeapSizeBytes` after `GC.Collect(2, Forced, blocking: true, compacting: true)` with a `GC.WaitForPendingFinalizers()` sandwich; widen the thresholds to a ratio with a generous margin; or `[Platform(Exclude = "MacOsX")]` on the two absolute assertions with the reason written down. The test's purpose (the retained-text claim) survives any of them.

PR #3784, opened 2026-09-03
Document creation: createCDATASection and DOMImplementation.createDocument are DOM's, so a document that has neither reports its tests

### LINKED ISSUE #3766: document.createCDATASection and DOMImplementation.createDocument are missing, and 36 wpt documents report nothing because of it
Two members of the DOM's document-creation surface are not projected into script, and between them they stop **thirty-six** vendored web-platform-tests documents from registering a single test. They are the largest single return available in the browser lane's corpus: one member lands thirty-one documents.

### Which tests

`dom/common.js` is the fixture builder that `dom/ranges/` and half of `dom/traversal/` load, and its file scope calls `document.createCDATASection`. With the member absent the script throws before the first `test()` runs, so testharness reports `ERROR` for the whole document and no per-test exclusion can name it. All of these are `NotVendored` rows today:

* `createCDATASection` — **31 documents**: `dom/ranges/Range-mutations-*.html` (10), `Range-cloneContents`, `Range-cloneRange`, `Range-collapse`, `Range-commonAncestorContainer`, `Range-compareBoundaryPoints`, `Range-comparePoint`, `Range-deleteContents`, `Range-extractContents`, `Range-insertNode`, `Range-intersectsNode`, `Range-isPointInRange`, `Range-selectNode`, `Range-set`, `Range-surroundContents`, `dom/traversal/NodeIterator.html`, `NodeIterator-removal.html`, `TreeWalker.html`, `dom/nodes/Node-compareDocumentPosition.html`, `Node-contains.html`, `Node-properties.html`, `MutationObserver-textContent.html`.
* `DOMImplementation.createDocument` — **5 documents**: `dom/nodes/Document-createAttribute.html`, `DocumentType-remove.html`, `Node-textContent.html`, `append-on-Document.html`, `prepend-on-Document.html`; and it is 24 more failing rows inside documents that *do* report (`Document-adoptNode.html`, `Node-isSameNode.html`, `Node-replaceChild.html`, `Element-tagName.html`, `Node-cloneNode-XMLDocument.html`, …).

`dom/nodes/Document-createCDATASection.html` is a case and its one test fails: it asks that `createCDATASection` throw `NotSupportedError` on an **HTML** document, which is the shape below.

### What the standard says

[`Document.createCDATASection`](https://dom.spec.whatwg.org/#dom-document-createcdatasection) is on `Document`
unconditionally. Its step 1 is "if this is an HTML document, then throw a `NotSupportedError`" — so on an HTML
document the *correct* answer is a `NotSupportedError`, not a `TypeError` for a member that is not there. That
distinction is the whole of what unblocks the 31 documents: `dom/common.js` calls it inside a `try`, and a
`NotSupportedError` is caught where a missing-member `TypeError` at that point is not.

[`DOMImplementation.createDocument`](https://dom.spec.whatwg.org/#dom-domimplementation-createdocument) creates
an `XMLDocument`. `Jint.Browser`'s bindings project AngleSharp's `IImplementation`, which has
`CreateHtmlDocument` and `HasFeature` and no `CreateDocument` at all — it is already recorded as an AngleSharp
gap in `Jint.Browser/Dom/AGENTS.md`. Since this package builds no XML document, the honest options are a
narrow implementation over the pieces AngleSharp does have (`XmlDocument` exists in AngleSharp.Xml) or a
refusal that at least is the right *kind* of refusal.

### Smallest repro

```js
document.createCDATASection('x');            // TypeError: Property 'createCDATASection' … is not a function
                                             // a browser: NotSupportedError, because this is an HTML document
document.implementation.createDocument(null, 'root', null);
                                             // TypeError: Property 'createDocument' … is not a function
```

### Why it is worth doing first

The exclusion table in `Jint.Tests.Browser/Wpt/WptBrowserExclusions.cs` currently declines twenty-eight of
`dom/ranges/`'s forty-five documents for this one reason, so the browser lane runs seventeen Range documents
where it could run forty-five. Adding `createCDATASection` — even as nothing but the standard's own
`NotSupportedError` for an HTML document — turns thirty-one not-vendored rows into cases in a single change.

this browser").

lahma:
Fixes #3766.

> **Merging note.** #3779 has landed, so this is now one commit on `main`. It touches `WptBrowserExclusions.cs` and the `README.md` census table, which #3783 and #3785 also touch; whichever of the three merges first, the other two need a re-census (`JINT_WPT_BROWSER_CENSUS=update`) rather than a manual edit — the figures below are measured against `main` before any of them.

## What was wrong

Two members of DOM's document-creation surface are not in AngleSharp at all, so nothing was projected:

* [`Document.createCDATASection`](https://dom.spec.whatwg.org/#dom-document-createcdatasection) — a CDATA section is `AngleSharp.Xml`'s `IXmlCDataSection`, and that assembly is not in the binding generator's pin.
* [`DOMImplementation.createDocument`](https://dom.spec.whatwg.org/#dom-domimplementation-createdocument) — `IImplementation` is `createHTMLDocument`, `createDocumentType` and `hasFeature`, and nothing else.

**An absent member is the *wrong refusal*, not merely a missing one.** `createCDATASection`'s first step on an HTML document is a `NotSupportedError`; a caller guarding with `try`/`catch` catches that, where a `TypeError` for a member that is not there reaches it before the guard can mean anything. `dom/common.js` — the fixture builder the whole of wpt's `dom/ranges/` and half of `dom/traversal/` load — reaches both members before its first `test()`, which is why **thirty-six documents reported nothing at all**.

## What changed

Both are `overrides.json` `additions` with their bodies in `Dom/Views/DomViewMembers`, so the interfaces stay one generated shape and the `.g.cs` moves only by regeneration (`JINT_DOM_BINDINGS=update`).

**`createCDATASection`** is DOM §4.5's five steps in order: WebIDL converts the argument, then an HTML document is a `NotSupportedError`, then data containing `]]>` is an `InvalidCharacterError` — checked *before* the node exists, which is what makes the order the standard's rather than AngleSharp's `Data` setter's — and otherwise `IXmlDocument.CreateCDataSection`. Every non-HTML document this package can make is one of those: `new Document()` and every `DOMParser` XML type go through `XmlParser`.

**`createDocument`** is DOM §4.5.1's algorithm over the pieces AngleSharp does have — the empty XML parse `new Document()` already uses, `createElementNS` for the document element (so a bad qualified name is refused where the standard refuses it), and DOM's own append for the doctype and then the element, in the standard's order. `[LegacyNullToEmptyString]` is honoured: `null` is the empty string and `undefined` is still the five letters of `"undefined"`.

**Two steps it cannot take, both now in `Jint.Browser/Dom/AGENTS.md`'s divergence table:** `Document.ContentType`'s setter is not public, so a document made here answers `application/xml` whatever namespace it was asked for; and the result is wrapped as `Document`, because `XMLDocument` is a name no assembly the generator reads declares. That second one is #3768's row and is what most of the newly-registered failures say.

## The tests that pin it

`Jint.Tests.Browser/DocumentCreationTests.cs`, seven tests over both members: the HTML refusal, the `]]>` refusal, the node a real XML document gives back, adopting one into the page the way `dom/common.js` does, the empty and the fully-specified `createDocument`, and WebIDL's own argument rules. Against the unfixed code all seven fail:

```
Failed ACdataSectionCanBeAdoptedIntoThePage
Failed CreateCdataSectionOnAnHtmlDocumentIsANotSupportedError
Failed CreateCdataSectionOnAnXmlDocumentMakesACdataSection
Failed CreateCdataSectionRefusesDataThatWouldCloseTheSection
Failed CreateDocumentAppliesWebIdlsOwnArgumentRules
Failed CreateDocumentMakesAnEmptyXmlDocument
Failed CreateDocumentTakesANamespaceAQualifiedNameAndADoctype
Failed!  - Failed: 7, Passed: 0, Skipped: 0, Total: 7
```

(the first is `TypeError: Property 'createCDATASection' of object is not a function`). After: `Passed! - Failed: 0, Passed: 7`.

## The exclusion rows retired

Thirteen rows across nine documents, all `NeedsXmlDocuments`, all of which the runner reported as covering passing tests:

| Document | Row(s) retired |
| --- | --- |
| `Document-adoptNode.html` | `Adopting*` (3 tests) |
| `Document-createCDATASection.html` | `*` — the document's one test, which is the `NotSupportedError` |
| `DOMImplementation-createDocument.html` | `*)` — the file's title test |
| `Node-cloneNode-document-with-doctype.html` | `*createDocumentType` |
| `Node-insertBefore.html` | `* is.`, `* now.`, `Should check the*` |
| `Node-isEqualNode.html` | `*too` |
| `Node-normalize.html` | `*.` |
| `Node-replaceChild.html` | `* is.`, `* now.`, `If the*thrown`, `Should check the*` |
| `Range-adopt-test.html` | `Parented*` |

## The census, and the one ceiling this raises on purpose

`DOMImplementation-createDocument.html` builds its own table of **434 cases inside its first test**, and the builder called the missing member — so the file reported **two** tests and the other 432 were never registered at all. They are registered now and 348 of them fail: `XMLDocument is not defined`, a document with a `location`, `characterSet` as `utf-8` rather than `UTF-8`, and eighteen refusals that pick a `NamespaceError` where DOM names an `InvalidCharacterError`. Eight globs name them, four `NeedsXmlDocuments` and four `NeedsTriage`, and the document's minimum-test floor goes 2 → 434.

So this is the one change that uses `JINT_WPT_BROWSER_CENSUS=update-raising-the-ceiling`, and it is used for what that spelling is for — **the failures are not new, only newly counted**:

| Suite | Documents | Tests | Not passing |
| --- | ---: | ---: | ---: |
| `dom/nodes/` before | 159 | 4,364 | 2,000 |
| `dom/nodes/` after | 159 | **4,796** | **2,333** |
| `dom/ranges/` before | 17 | 78 | 53 |
| `dom/ranges/` after | 17 | 78 | **52** |
| lane total before | 336 | 6,211 | 2,704 |
| lane total after | 336 | **6,643** | **3,036** |

## The thirty-six not-vendored rows

They keep their standing and change their *reason* to name a cause that is spent — `"not vendored: it called createCDATASection … at file scope, which now exists"` — which is exactly what this table already does for the four `dom/events/` documents that were waiting on `document.createEvent`. **Vendoring them is a change of its own**, because it moves the census's `Documents` and `Tests` columns, which the change that fixes an engine deliberately does not; that is the table's own rule and `Wpt/README.md` now says which thirty-six are waiting on it. `Document-createElement-namespace.html`'s reason loses its `createDocument` half and keeps the iframe one, which is the real remaining blocker.

## Verification

* `dotnet test Jint.Tests.Browser -c Release -nr:false` with `JINT_WPT_BROWSER_CENSUS=1` — green on both legs: **1414 passed** (net10.0), **1411 passed** (net8.0), 5 skipped each (the Playwright client suite, which needs `JINT_BROWSER_CLIENTS`).
* All 348 wpt browser cases pass, including `DomBindingsStalenessTests` — the `.g.cs` diff is the emitter's own.
* `dotnet test Jint.Tests -c Release -f net10.0 -nr:false --filter "…MigrationGuideTests|…AgentInstructionFileTests|…SpecCitationTests"` — 11 passed.


lahma:
The macOS leg failed on `PreparedScriptsDoNotRetainSourceTextByDefault`, which is the known heap-retention flake on that runner (#3641) and has nothing to do with this change: it measures an absolute `GC.GetTotalMemory` difference, and the macOS leg reads a few megabytes of residue as retention. Re-triggered with a proven identical tree.

PR #3788, opened 2026-09-03
Node and Element: ten members the DOM standard requires are in the bindings, so a page that reaches for one gets an answer

### LINKED ISSUE #3768: Eleven Node and Element members the DOM standard requires are not in the bindings — 182 wpt rows

lahma:
Fixes #3768.

> **This pull request stacks on [#3785](https://github.com/sebastienros/jint/pull/3785)** (`fix/3767-domtokenlist`), which is the first of the two commits below: `NodeList`'s value iterator is the same `Dom/Collections/DomIterableMembers` entry `DOMTokenList`'s is, and duplicating it would be a second answer to one WebIDL declaration. Merge #3785 first and this becomes one commit. It also touches `WptBrowserExclusions.cs` and the `README.md` census table, which #3783 and #3784 touch too — whichever lands first, the others need a re-census (`JINT_WPT_BROWSER_CENSUS=update`) rather than a manual edit.

## What was wrong

Eleven members the DOM standard requires were not projected at all, and every one fails the same way before a test can say anything about behaviour: `Property '…' of object is not a function`. **182 failing tests across eighteen documents**, and each is on the shortest path of an ordinary page — a client library that feature-detects `toggleAttribute` takes a slower path when it is missing, and one that does not throws.

## What changed

Ten of them land here, all as `overrides.json` `additions` with their bodies in the new `Jint.Browser/Dom/DomElementMembers`, so every interface stays one generated shape.

| Member | What it needed |
| --- | --- |
| [`ChildNode.replaceWith`](https://dom.spec.whatwg.org/#dom-childnode-replacewith) | **only a name**: `IChildNode` carries `[DomName("replace")]`, so the mixin projected a member the standard has not got. A `skip` plus an alias on the three interfaces it lands on (`Element`, `CharacterData`, `DocumentType`) |
| [`getAttributeNode`](https://dom.spec.whatwg.org/#dom-element-getattributenode), `getAttributeNodeNS`, `setAttributeNode`, `setAttributeNodeNS`, `removeAttributeNode` | AngleSharp's own `INamedNodeMap`, under WebIDL's *other* spelling (`getNamedItem`), plus the one refusal that is not there: DOM removes an attribute **node**, so one the element does not hold is a `NotFoundError` rather than a silent miss on a matching name |
| [`insertAdjacentElement`](https://dom.spec.whatwg.org/#dom-element-insertadjacentelement), `insertAdjacentText` | DOM's four-position switch over `InsertBefore` and `AppendChild`, ASCII case-insensitive, a `SyntaxError` for anything else, and `null` — not a refusal — for a position that needs a parent the element has not got |
| [`toggleAttribute`](https://dom.spec.whatwg.org/#dom-element-toggleattribute) | the same given-versus-not-given `force` `DOMTokenList.toggle` has, plus step 2's lower-casing for an HTML element in the HTML namespace |
| [`hasAttributes`](https://dom.spec.whatwg.org/#dom-element-hasattributes), [`getAttributeNames`](https://dom.spec.whatwg.org/#dom-element-getattributenames), [`isSameNode`](https://dom.spec.whatwg.org/#dom-node-issamenode), [`webkitMatchesSelector`](https://dom.spec.whatwg.org/#dom-element-matches) | one expression each; the alias reaches the same `DomSelectorMembers.Matches` that makes a bad selector a catchable `SyntaxError` |
| `NodeList.keys`/`values`/`entries`/`forEach` | WebIDL's value iterator, through the same `extend` entry `DOMTokenList`'s `iterable<>` uses |

**Two are deliberately not here, and both are argued in `Jint.Browser/Dom/AGENTS.md`'s divergence table and filed as [#3789](https://github.com/sebastienros/jint/issues/3789):**

* **`replaceChildren`.** DOM's *replace all* is one operation and therefore one mutation record. AngleSharp's `ReplaceChild` with a `DocumentFragment` records the removal and **no addition at all** — measured: `a:+0:-1 | a:+0:-1 | b:+0:-1` where the standard's shape is `…| b:+2:-1` — and `ParentNode-replaceChildren.html`'s last test is an `async_test` that waits for that record before it calls `done()`. Every arrangement of append, remove and replace gives four or five records where it asserts three, so implementing the member turns a document that reports thirty-one results into a `TIMEOUT`, and this lane's rule is that a document must report. It is worth its own change, over the mutation records rather than over the member.
* **`XMLDocument`.** AngleSharp builds the same `XmlDocument` for an XML parse *and* for `new Document()`, which DOM makes a plain `Document` — so the CLR type `DomTypeMap` keys on cannot decide the interface. It is `HTMLFrameSetElement`'s shape with no local name to select on, and `AngleSharp.Xml` is outside the generator's pin besides.

## The tests that pin it

`Jint.Tests.Browser/DomNodeMemberTests.cs`, twelve tests over the ten members. Against the unfixed code **all twelve fail**:

```
Failed ANodeListCarriesWebIdlsValueIterator
Failed HasAttributesAndGetAttributeNamesReadTheAttributeList
Failed InsertAdjacentPutsANodeAtEachOfTheFourPositions
Failed InsertAdjacentRefusesAnUnknownPositionAndAnswersNullWithNoParent
Failed IsSameNodeComparesIdentityAndNotEquality
Failed RemovingAnAttributeNodeTheElementDoesNotHoldIsANotFoundError
Failed ReplaceWithSwapsTheChildForItsArguments
Failed TheAttrNodeFamilyReadsWritesAndRemoves
Failed TheNamespacedAttrNodeMembersUseTheNamespace
Failed ToggleAttributeLowerCasesTheNameOfAnHtmlElementsAttribute
Failed ToggleAttributeTellsAForceThatWasGivenFromOneThatWasNot
Failed WebkitMatchesSelectorIsTheAliasTheStandardRequires
Failed!  - Failed: 12, Passed: 0, Skipped: 0, Total: 12
```

After: `Passed! - Failed: 0, Passed: 12`.

## The exclusion rows retired, and the census

**Forty-one rows across thirteen documents.** Six documents pass whole now — `Attr-prefix.html` keeps one row, and `ChildNode-replaceWith.html`, `Element-hasAttributes.html`, `Element-insertAdjacentElement.html`, `Element-insertAdjacentText.html`, `Node-isSameNode.html`, `insert-adjacent.html`, `Node-childNodes.html` and `remove-next-sibling-during-replace-with.html` lose their blanket `"*"`. Seventeen rows of `custom-elements/reactions/`'s `ChildNode.html`, `Element.html` and `Node.html` go with them.

**Thirteen narrower rows take their place**, for what the members made *visible* rather than fixed — every one a defect this corpus could not reach before:

* `IChildNode.Replace` converts its arguments before it checks whether the child has a parent, so a call on a parentless child moves them out of the tree the standard leaves them in; and it refuses outright when the child is among its own replacements (6 rows).
* An `Attr` write reaches the attribute observer **without its new value**, so a custom element's `attributeChangedCallback` is handed `null` (4 rows).
* A parser-inserted namespaced attribute records no prefix; `importNode`'s `deep` shapes; three `attributes.html` assertions.

Eighteen more rows move to the `(Node or DOMString)` cause row they always belonged to — `replaceWith` joins `before`, `after`, `append` and `prepend` there now that it exists.

Census, `JINT_WPT_BROWSER_CENSUS=update` (both figures measured on top of #3785):

| Suite | Tests | Not passing before | after |
| --- | ---: | ---: | ---: |
| `dom/nodes/` | 4,364 | 1,350 | **1,280** |
| `custom-elements/reactions/` | 255 | 217 | **200** |
| **total** | 6,211 | 2,042 | **1,955** |

## Verification

* `dotnet test Jint.Tests.Browser -c Release -nr:false` with `JINT_WPT_BROWSER_CENSUS=1` — green on both legs: **1432 passed** (net10.0), **1429 passed** (net8.0), 5 skipped each (the Playwright client suite, which needs `JINT_BROWSER_CLIENTS`).
* All 348 wpt browser cases pass, `DomBindingsStalenessTests` included — the `.g.cs` diff is the emitter's own.
* `dotnet test Jint.Tests -c Release -f net10.0 -nr:false --filter "…MigrationGuideTests|…AgentInstructionFileTests|…SpecCitationTests"` — 11 passed.


lahma:
Rebased onto main (its base #3785 merged) and pushed as `3b43ac31c`, with two things worth flagging for review.

**The census was regenerated, not merged.** These pull requests all rewrite the same table, so the numbers were taken from the corpus with `JINT_WPT_BROWSER_CENSUS=update` rather than hand-merged: the lane now reports 340 documents, 6,228 tests and 1,954 not passing.

**A second commit moves the AngleSharp divergence register out of `Jint.Browser/Dom/AGENTS.md`.** That file was 233 bytes over its 32 KiB budget and the instruction-file test failed. The register had reached 48 rows and every finding adds one, so what pushed the file over was data rather than instruction. The rule an agent must read before editing stays where it was, and points at `Jint.Browser/Dom/divergences.md`, which is not budgeted; the instruction file is 19,699 bytes now. Every pull request in flight that adds a divergence row will conflict here once, and the resolution is to move the row into the register.

Issue #3770, opened 2026-09-03
HTML content-attribute reflection is not implemented: 22,028 of 56,660 wpt assertions fail, and the suite cannot be vendored because of it


lahma:
Vendoring `html/dom/` into the browser lane ran HTML's reflection conformance suite — ten generated documents covering every content attribute of every HTML element. It is the largest single body of conformance data available to this package, and **22,028 of its 56,660 assertions fail**.

The ten documents are a `NotVendored` row rather than cases, and the reason is unusual enough to be worth stating: they are **not slow** (22.5 s for the set, 7.3 s for the largest, well inside the lane's 30 s per-document deadline) and they report cleanly. What stops them is the artefact. The smallest table of patterns that covers those failures and no passing test is **over four thousand rows**, and every one of them says the same thing. One issue saying it once is the better record.

### The measurement

| Document | Tests | Failing | Time |
| --- | ---: | ---: | ---: |
| `html/dom/reflection-text.html` | 10,202 | 3,360 | 2.6 s |
| `html/dom/reflection-embedded.html` | 8,922 | 3,774 | 7.3 s |
| `html/dom/reflection-forms.html` | 8,271 | 2,160 | 2.8 s |
| `html/dom/reflection-tabular.html` | 6,116 | 3,552 | 2.2 s |
| `html/dom/reflection-sections.html` | 5,604 | 2,189 | 1.7 s |
| `html/dom/reflection-grouping.html` | 5,358 | 2,006 | 1.7 s |
| `html/dom/reflection-misc.html` | 4,877 | 1,866 | 1.5 s |
| `html/dom/reflection-metadata.html` | 3,110 | 1,218 | 1.0 s |
| `html/dom/reflection-obsolete.html` | 2,621 | 1,483 | 1.0 s |
| `html/dom/reflection-forms-weekmonth.html` | 1,579 | 420 | 0.7 s |
| **total** | **56,660** | **22,028** | **22.5 s** |

### What is missing

`Jint.Browser/Dom` projects AngleSharp's properties as they are. [HTML §2.6.1](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflecting-content-attributes-in-idl-attributes) defines what an IDL attribute that *reflects* a content attribute must do, and none of it is implemented:

[...]

PR #3799, opened 2026-09-03
Reflection: the text-level and grouping-content documents are cases, so 15,560 more assertions of HTML §2.6.1 run

lahma:
Part of #3770, and **stacked on #3796** — that pull request is the machinery and the first document, and this one is only its table growing. Review it first; the diff here is what is left once it merges.

## What this adds

Twenty-five more rows in `overrides.json`'s `reflected` list, and the two documents they make green:

| Document | assertions | failing before | failing now |
| --- | ---: | ---: | ---: |
| `html/dom/reflection-text.html` | 10,202 | 3,360 | **0** |
| `html/dom/reflection-grouping.html` | 5,358 | 2,006 | 38 |

`reflection-text.html` needed only eleven rows of its own for 3,360 assertions, because #3796's fifteen were mostly the **global** attributes every element carries (`dir`, `lang`, `tabIndex`, `autofocus`, `inputMode`, `enterKeyHint`) and this document is thirty-three elements sharing them.

## Three of the rows are findings rather than absences

Most of the twenty-five name an attribute AngleSharp has no member for — `a.ping`, `a.coords`, `hr.noShade`, `li.type`, `br.clear` and the rest of HTML §16.2's obsolete-but-still-reflected set. Three are members that exist and answer wrongly:

* **`a.type` has a setter that writes nothing.** `a.type = 'x'` left the content attribute exactly as it was, so the IDL attribute and the content attribute could disagree permanently.
* **`ol.start` was parsed with .NET's integer rules, not HTML's.** U+000B VERTICAL TAB is whitespace to `char.IsWhiteSpace` and *not* to HTML's rules for parsing integers, so `<ol start="&#11;7">.start` answered 7 where a browser answers the default, 1.
* **`q.cite` and `blockquote.cite` answered the raw attribute value**, where HTML reflects `cite` as a URL. The same divergence `script.src` had; both elements are `HTMLQuoteElement`, so it is one row.

## What is left in `reflection-grouping.html`, and why it is not reflection

`<dl>`. The pinned assemblies have **no `IHtmlDListElement` and no `[DomName("HTMLDListElement")]`**, so a `<dl>` is an `HTMLElement` here and there is no interface for a `reflected` row to put `compact` on — putting it on `HTMLElement` instead would give the member to every element, which is worse than not having it. Declaring the interface by local name is what `DomManualInterfaces` already does for `<frameset>`; doing it for `<dl>` is a change of its own. The divergence table records it, along with `<dt>` and `<dd>` being `IHtmlListItemElement` where HTML gives them a plain `HTMLElement`.

**Thirteen exclusion rows rather than one `dl.compact: *`**, because one of the thirty-eight passes by accident and the two-sided rule will not have it covered: `IDL set to true` asserts `hasAttribute('compact')` is true, and the attribute is still there from the `setAttribute()` half that ran before it — the expando the assignment really made is invisible to that assertion.

## The census ceiling is raised, deliberately

| | before | after |
| --- | ---: | ---: |
| `html/dom/` documents | 6 | **8** |
| `html/dom/` tests | 4,962 | **20,522** |
| `html/dom/` not passing | 72 | **110** |
| **total tests** | **11,105** | **26,665** |
| **total not passing** | **2,039** | **2,077** |

`Not passing` only ever goes down, so this took `JINT_WPT_BROWSER_CENSUS=update-raising-the-ceiling` — the spelling that exists to make a raise something somebody said out loud. The raise is `<dl>`'s thirty-eight, all named in the exclusion table under one cause, against 15,560 assertions the lane did not run at all before. It is the shape that spelling is for: a corpus arriving with new failures, not a bad run being re-baselined.

## What is not vendored, and what each of the seven now needs

The not-vendored rows carry it. Two are worth naming here because their remaining causes are **not** reflection either, and both were found by running the documents rather than guessed:

* **`reflection-metadata.html`** (1,218 failing at the measurement, 624 with these rows in) needs `link.nonce` and `style.nonce`, which are HTML's `[[CryptographicNonce]]` internal slot rather than a reflected attribute — an IDL set must *not* write back to the content attribute — and it has one row that cannot pass at all: **AngleSharp.Css refuses an unparseable media query from `setAttribute`**, so `style.setAttribute('media', '5%')` raises where the DOM says an attribute is a string. It also exposed a third: AngleSharp's `a.protocol`/`host`/`pathname` ignore `<base href>` while `element.BaseUri` honours it, and the suite computes its own URL expectations from the former.
* **`reflection-sections.html`** (2,189 → 489) needs `document.dir`, `bgColor`, `fgColor`, `linkColor`, `vlinkColor` and `alinkColor`, each of which reflects an attribute of **another** element — the document element, or the body — which the `reflected` list cannot express yet, plus `[LegacyNullToEmptyString]`.

`Jint.Tests.Browser` is green on both target frameworks with `JINT_WPT_BROWSER_CENSUS=1`, `dotnet build -c Release` is clean, and `Jint.Tests`' `Wpt` / `MigrationGuideTests` / `AgentInstructionFileTests` / `SpecCitationTests` pass. The four vendored files are byte-identical to upstream at the pin.


sebastienros:
Integrated on current `main` in #3871, preserving the original commit and authorship. #3871 has merged, so I’m closing this superseded PR. Thank you for the implementation.

Issue #3647, opened 2026-09-02
Vendor/README.md's drift-verification figures (359 files, 48 directories) are stale; the claim should be produced by the check, not typed


lahma:
Noticed while landing #3639 (part of #3575).

`Jint.Tests/Wpt/Vendor/README.md`'s manual drift-verification section still claims "359 files, 48 directories" verified against upstream at the pin; the corpus is now 438 files in 55 directories (the fetch, xhr and FileAPI suites landed since the figures were written), and the section was already stale before this campaign. The claim is a *verification statement*, so it should not be edited by hand to the new numbers: the fix is to re-run the full hash comparison against upstream at the pinned commit (`git hash-object` per file against the upstream blob ids, which is what #3639 did for its one file) and record the figures the run produced, or to replace the manual figures with a test that does the comparison and prints them (`JINT_WPT_CENSUS=update`-style), so the number can never drift again.

Issue #3339, opened 2026-08-24
The WPT census turns a flaky fetch test into a hard CI failure on unrelated PRs


lahma:
`WptCensusTests.TheInventoryTableMatchesWhatTheCorpusMeasures` compares the measured corpus totals against `Vendor/README.md`'s inventory table by **exact string equality**, including the "Not passing" column. It runs on the PR workflow's Windows leg under `JINT_WPT_CENSUS`.

That column is not stable. On #3335 — a PR that unseals the `Intl` default providers and cannot reach the network — the census failed with:

| | stated | measured |
|---|---|---|
| Fetch (`fetch/api/` ×5) | 154 not passing | **168** |
| **total** | **2,968** | **2,982** |

Every other row was byte-identical, and `Files` and `Assertions` matched exactly (293 / 40,983). Fourteen extra fetch assertions failed, on a runner that was also producing `vstest.console` connect timeouts and 30-second test262 timeouts in the same window.

So a timing-sensitive suite makes an unrelated PR red, and the failure presents as "the inventory table is out of date" — which reads like the author forgot to re-census, and invites exactly the wrong fix (running `JINT_WPT_CENSUS=update` and committing a *worse* number as the new baseline). That would silently ratchet the not-passing count upward, which is the opposite of what `Jint.Tests/Wpt/AGENTS.md` wants from this artefact.

Options worth weighing:

1. Hold `Files` and `Assertions` by equality (they are deterministic) but treat "Not passing" as a **ceiling** rather than an equality — a rise fails, a fall tells you to re-census.
2. Keep equality but exclude the fetch rows from the enforced comparison until they are deterministic, and say so in the table.
3. Make the fetch suite deterministic — if the 154 baseline failures are already "expected", the flaky 14 are a different category and should be separated from them.

Option 1 looks closest to the file's intent: it still catches a real regression, and it cannot be satisfied by re-baselining a bad run.

Seen this session on #3335. Related runner-load symptoms in the same window are noted on #3323's sibling issues.

lahma:
Hit again on [#3337](https://github.com/sebastienros/jint/pull/3337) — a PR that promotes five `ErrorConstructor` intrinsics from `internal` to `public` and touches nothing reachable from the corpus.

[...]

PR #3819, opened 2026-09-04
Web platform tests: the cause table's two columns are generated, so they cannot go stale unnoticed again

lahma:
> **Stacked on [#3812](https://github.com/sebastienros/jint/pull/3812)**, and the first commit here is that
> PR's. This is the follow-up that PR named for itself. Review only the second commit; merge #3812 first and
> this rebases to one.

## Why

#3812 repaired a table and said, in as many words, that nothing checks it. That is the whole reason it
rotted: `Wpt/README.md`'s "What the DOM corpus says about this browser" counts, per cause, **how many tests
fail and over how many documents**, and those two columns were typed. By the time anyone measured, **nine of
seventeen rows were wrong, in both directions** — one cause had grown 165 → 478 as a fix uncovered a table a
missing member had been hiding, another had shrunk 65 → 24 as its own fix landed.

A row reads exactly the same whether it is true or four merged fixes out of date. That is the failure
[#3647](https://github.com/sebastienros/jint/issues/3647) recorded for the drift-verification recipe and
[#3339](https://github.com/sebastienros/jint/issues/3339) for the census, in a third place — and the repair
is worth much less than the thing that stops it recurring.

## A cause is a value now, not a comment

`WptBrowserExclusions.Causes` is the exclusion table split into named groups; `All` is that flattened. **What
was a `// ----------------` header comment is the name of the array under it**, and every exclusion row is
untouched — the diff is 32 headers becoming 32 field declarations and nothing else. Three groups shared the
header "a frame that runs script"; two are qualified now, because a name is a key.

`EveryRowNamesACauseAndEveryCauseIsNamedOnce` holds `All` to the causes flattened *in order*, so the split
cannot lose or reorder a row the runner enforces against.

## What is generated, and what is not

`WptBrowserCauses` counts each cause against what the lane reported and renders the table beside the census —
same run, same switch (`JINT_WPT_BROWSER_CENSUS=1` checks, `=update` rewrites).

**Only the two numbers and the order are generated.** A row's prose is written by whoever worked out what the
cause was and is copied through untouched. What joins the two is a key the rendered page does not show:

```md
| 478 | 10 | [#3766](…) **An XML document, and the two members that make one.** … <!-- cause: an XML document, and the two members that make one --> |
```

The key is deliberately *not* derived from the prose: a row's wording changes when what it describes changes,
and a key that moved with it would silently start counting something else. A group renamed in the code and
not in the README fails for free, in the test that runs on every build.

## The two properties that make the arithmetic mean anything

`WptBrowserCauseTests` holds both, and without them a column is an arbitrary sum:

* **Every failing test in the lane is claimed by exactly one cause.** The runner already refuses a failing
  test no exclusion names; this adds that no two causes name the same one — the difference between a sum and
  a partition. A test two causes both claimed would be counted twice while the table still added up to
  something plausible.
* **A cause that accounts for anything in the six suites the section is about has a row, and a row accounts
  for something in them.** A cause outside them (`custom-elements/`'s own frames, say) is a real cause of a
  real failure and simply not what this section counts.

## Neither column is a ceiling

That is the one way this differs from the census, and it is stated in both places. `Not passing` bounds *how
much* fails, so it may only go down. These two say what each failure *is*, so **a cause that grew and a cause
that shrank are equally a table that has stopped describing this browser**. Both are equalities in both
directions.

## The regression check is that it changes nothing

You asked for byte-identical output on an unchanged corpus. It is:

```
$ JINT_WPT_BROWSER_CENSUS=update dotnet test Jint.Tests.Browser -c Release -f net10.0 \
    --filter "WptBrowserCauseTests|WptBrowserCensusTests"
Passed! - Failed: 0, Passed: 16, Skipped: 1

$ git diff --stat -- Jint.Tests.Browser/Wpt/README.md
 Jint.Tests.Browser/Wpt/README.md | 34 +++++++++---------------
 1 file changed, 17 insertions(+), 17 deletions(-)      # the 17 cause keys, and nothing else
```

An `update` run **rewrote not one byte of the table #3812 derived by hand.** The generator and the derivation
agree exactly, which is what makes either of them worth believing. Eight of the seventeen rows had already
come back byte for byte during that derivation as the control; now all seventeen do, on every run.

## Verified

* `dotnet test Jint.Tests.Browser -c Release` — net8.0 1,687 passed, net10.0 1,691 passed, 0 failed, with
  `JINT_WPT_BROWSER_CENSUS=1` set (so both the census and the cause table are enforced).
* `JINT_WPT_BROWSER_CENSUS=update` — no change to the README.
* `AgentInstructionFileTests` — the lane's `AGENTS.md` gains a section on the mechanism and stays inside the
  size budget.

## Two figure-carrying sentences, and the paragraph now says which is which

The check used to anchor on the words **"They arrived together,"** — the phrase that makes those numbers read
as a record of when the six suites were vendored. So the file offered a reader two figure-carrying sentences
beside a generated table with nothing distinguishing them, and the one that looked historical was the checked
one. That is a trap someone would have sprung by "correcting" a record, or by leaving a live figure stale
because it looked like history.

Fixed at the source rather than by adding a sentence that explains the ambiguity: the live sentence is now
present tense (`Across the six of them there are 207 documents and 5,247 tests, and 1,487 of those tests do
not pass`), the anchor moved with it, and the arrival figures stay beside it as an explicitly past clause —
**4,815 tests with 2,024 not passing** — with the paragraph stating that those are a record of a moment,
deliberately not re-derived, and checked by nothing.

The move was verified the only way an anchor change can be: the figure was poisoned to 1,499 and the check
failed naming both sides, then restored. A green run against a moved anchor is otherwise indistinguishable
from an anchor that matches nothing and silently checks nothing.

## What is checked but not rewritten

The paragraph's opening sentence — `207 documents and 5,247 tests, 1,487 not passing` — is **checked** and
not generated: the reconciliation compares its three numbers against the census and fails naming both sides,
but it does not rewrite the sentence, because generating prose is a different and worse job than generating a
table. The rebase onto #3814 proved the check works before anything else did: it failed with *"the sentence
above the cause table says 207 documents, 5247 tests and 1532 not passing, where the census measured 207,
5247 and 1517"*, and that number was then corrected from the run rather than by hand. The note that the rows
sum past that figure is still ordinary prose and is checked by nothing.

## What the rebase onto #3814 found, since this PR is the one that claims staleness cannot pass unnoticed

Three cause-table clauses were reported as retired by fixes that have landed. Measured on this branch, not
assumed:

* **`StaticRange` — the numbers surfaced, the prose did not.** `=update` rewrote the row from `21 | 4` to
  `6 | 4` on its own, and `=1` fails on the same row until it is written. What no check catches is the
  sentence beside it, which said the constructor "still refuses seventeen of the twenty-one": the document
  registers **17** tests and **2** of them fail (a `DocumentFragment` container and endpoints in disconnected
  trees). Corrected here from that run. **The generated columns cannot police the prose, and this is the
  first row that proves it.**
* **`createDocumentType("edi:root", …)` — did not surface, and could not.** The row's figures (`50 | 2`) are
  exactly what the run measures, so nothing is stale about them; the *prose* names a test that **does not
  exist anywhere in the corpus** — the family is `edi:` followed by a punctuation character. Rewritten from
  the run: 45 of the document's 82 names are refused as "Invalid character detected" where DOM stopped
  validating a doctype's qualified name at all.
* **The empty-name clause was the control, and then it landed.** When this was first measured, #3820 was
  still open and `HTMLCollection-empty-name.html` was 7 FAIL of 7, so the clause was correctly *not* stale.
  #3820 has since merged and this branch was re-derived behind it: the document measures **7 PASS of 7**, the
  row moved `107 | 28` -> `100 | 27` **by the generator**, and only the sentence beside it needed typing. That
  is the whole mechanism working end to end on a cause that genuinely went out — first as a control, then as
  the live case.

**Rebased onto #3761 as well, and it moved three more things the generator caught and one it could not.**
The generator rewrote the `#3712` row `101 | 5` -> `5 | 2` and the one-assertion-each row `87 | 18` -> `76 | 17`.
What it cannot police is the prose beside them, and both sentences were wrong in the same direction: the
paragraph said #3712 "has no row left" when the run finds **five**, and the row said "80 of the 101 are
`AriaMixin-string-attributes.html`" when the 101 are five. Re-derived by triaging the two documents: what
survives is the *attribute* half — `Text.nodeValue`, `Comment.nodeValue`, `ProcessingInstruction.nodeValue`,
`Text.data = null`, `Comment.data = null`. The argument half really is fixed.

And the sum sentence stopped being true in the most satisfying way: the rows now sum to **1,487**, exactly the
stated total, because the two causes that used to reach a `custom-elements/reactions/` document (80 rows of
`AriaMixin-string-attributes.html`, 4 of `Element.html`) no longer do. The overlap is gone rather than hidden,
and the sentence says so.

A second figure moved with it: the paragraph's stated total went 1,517 -> **1,510**, exactly #3820's seven
rows, and it is typed rather than rewritten for the reason the section above gives.


sebastienros:
The current-main refresh is available as #3958. It preserves all 362 exclusion entries byte-for-byte and in order, incorporates #3817's enlarged selector corpus, and adds the review fix that serializes cause/census measurement to prevent concurrent corpus traversal and README rewrites.

I opened the refreshed commit from a repository-owned branch so this contributor-owned fork branch remains untouched. I will leave this PR open until #3958 completes, then close it as superseded.

sebastienros:
Superseded by #3958, which preserved the selector exclusions and their order while generating the cause counts from the authoritative WPT data. #3958 passed all PR jobs and was squash-merged as fdc0850ddc1b0d4eca16818075f23f1ddcd71c4e. Thank you for the original contribution and the direction for this reporting improvement.

PR #3822, opened 2026-09-04
ARIA reflection: an element-reflecting IDL attribute is an attr-associated element, so ariaLabelledByElements answers elements no id could name

### LINKED ISSUE #3773: The ARIA reflection mixin is not projected, so element.role and every aria* IDL attribute is undefined — 65 wpt rows
`html/dom/aria-attribute-reflection.html` fails all 41 of its tests and `html/dom/aria-element-reflection.html` 22 of its 27, for one reason: `ARIAMixin` is not on `Element` at all.

### What happens

```js
document.body.role;                 // undefined; a browser: null
document.body.role = 'button';      // silently creates an expando; no `role` attribute is written
document.body.ariaLabel = 'x';      // the same; no `aria-label` attribute
document.body.ariaDescribedByElements;   // undefined
```

Every row reads `assert_equals: expected (string) "true" but got (undefined) undefined` or
`assert_array_equals: value is undefined, expected array`.

### What the standard says

[ARIA §9](https://w3c.github.io/aria/#ARIAMixin) puts `ARIAMixin` on `Element`: `role` plus thirty-odd
`aria*` IDL attributes, each reflecting its `aria-` content attribute as a nullable `DOMString`, and the
element-reflection members (`ariaActiveDescendantElement`, `ariaControlsElements`,
`ariaDescribedByElements`, `ariaLabelledByElements`, …) reflecting an element or a frozen array of elements
with an explicitly-set value that survives the target leaving the tree.

### Which tests

* `html/dom/aria-attribute-reflection.html` — 41/41. The string half: every `aria*` attribute round-tripping
  through `setAttribute` and the IDL attribute in both directions.
* `html/dom/aria-element-reflection.html` — 22/27. The element half, including the explicitly-set-value rules.
* `html/dom/aria-element-reflection-disconnected.html` — 2/2.
* `html/dom/aria-element-reflection-labelledby.html` is a `NotVendored` row: it reads `firstElementChild` off a
  null shadow root before it can report.
* And it is the same absence `custom-elements/reactions/AriaMixin-element-attributes.html` (16 rows) and
  `AriaMixin-string-attributes.html` (80 rows) already report, which the browser lane's README records as
  cause 4 of the custom-element corpus.

So 161 rows across the two corpora rest on one mixin.

### Where to look

AngleSharp's `IElement` has no ARIA members, so the generator emits none —
`Jint.Browser/Dom/AGENTS.md`'s `additions` list is where a member the standard puts on a generated interface
and AngleSharp's metadata cannot express goes. The string half is thirty-odd attribute reflections and is
mechanical; the element half needs the explicitly-set-value bookkeeping and is not.

The accessibility tree (`Jint.Browser/Accessibility/`) already computes roles and names from html-aam, so the
two would want to agree about what `role` means once this exists.

this browser").

lahma:
**Stacked on #3799**, which is stacked on #3796. **Only the last commit — "ARIA reflection: an element-reflecting IDL attribute is an attr-associated element…" — is for review here;** the three below it are those two pull requests and will disappear from this diff as they merge. The stack is chain B of the reflection campaign, and this shares its machinery: an element-reflecting IDL attribute is the same §2.6.1 question as a string-reflecting one, asked of an element instead of a value.

## What was missing

`Element` carries [ARIA §10.1's `ARIAMixin`](https://w3c.github.io/aria/#ARIAMixin) since #3791, but only its **string** half — `role` and the forty-four `aria-*` IDL attributes, each a view of one content attribute that holds nothing of its own. The eight members that reflect an [attr-associated element](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#attr-associated-element) cannot be that shape, and were `undefined`:

`ariaActiveDescendantElement`, `ariaControlsElements`, `ariaDescribedByElements`, `ariaDetailsElements`, `ariaErrorMessageElements`, `ariaFlowToElements`, `ariaLabelledByElements`, `ariaOwnsElements`.

A page may hand one of them an element with no `id` at all, or one whose `id` is not unique, or one about to leave the tree and come back. So the relationship has to be held **by reference**, and the content attribute stops being the state and becomes a flag.

## The three rules

1. **No content attribute is `null`** — how a page tells "no relationship" from "a relationship to nothing" (`[]`).
2. **Setting writes the empty string**, never an id, because an id could not name the element in general — and stores the reference.
3. **A stored reference is filtered at *get* time**, by the shadow-including-ancestor rule, never at set time. That is what lets a relationship survive either end leaving the tree and start answering again when it comes back, which is the entire subject of `aria-element-reflection-disconnected.html`.

HTML's **attribute change steps** drop the stored reference whenever anything else writes the content attribute. This reads that from the attribute's *value* rather than from an observer, deliberately: `IAttributeObserver` reports neither a namespaced write, nor an `Attr` node's value, nor the parser, and the value says it without any of them — the setter writes the empty string and nothing else here does.

**One divergence, stated and pinned.** A page writing that same empty string by hand keeps a reference where a browser drops it. Both spellings answer no elements from the ids; `Jint.Browser/Dom/AGENTS.md` records it and `AnEmptyStringWrittenByHandKeepsTheReferenceThatABrowserDrops` fails if it is closed by accident.

## Measured, not claimed

The issue title says 65 rows. **It is 24, across 2 documents**, which is what the corrected census (#3812) also gives — measured here with the triager before and after:

| Document | Before | After |
| --- | --- | --- |
| `html/dom/aria-element-reflection.html` | 22 FAIL, 5 PASS | **27 PASS** |
| `html/dom/aria-element-reflection-disconnected.html` | 2 FAIL | **2 PASS** |

Twelve exclusions are gone — re-derived from a real run, not picked — and the group they were in is empty and removed. The lane's `html/dom/` census figure falls **69 → 45**, a fall of exactly 24, which is the same number from the other side.

`html/dom/aria-attribute-reflection.html` and every other document in the lane are unchanged; the whole of `Jint.Tests.Browser` is green on net8.0 and net10.0, and the generated bindings are byte-identical (this is the `additions` extend form, so nothing new is emitted).

## Shape and cost

Both halves are declared onto the **one** `Element` shape through the extend form, so the prototype keeps its shared shape and its inline cache. State is a `ConditionalWeakTable` on `DomRealm` keyed on the AngleSharp element — the arrangement `CustomElementRegistry` already uses — created on first use, so an element that never has an ARIA relationship pays nothing. The references are **weak**, as the standard says.

## Two small things fixed on the way

- The root walk `getRootNode` already had is now `DomNodeMembers.Root`. The visibility rule is a question about two roots, and a second copy of that loop would be a second place to get it wrong.
- `DomConvert.NodeSequence`'s summary called its result a frozen array. It is not, and must not be — WebIDL's `sequence<T>` is a plain array; the `FrozenArray` this change does add is beside it.

## The accessibility tree reads the same relationships (#3824, folded in)

`Accessibility/` resolved `aria-labelledby` and `aria-describedby` as idrefs off the content attribute — which this change makes the **empty string** for a relationship set through the IDL attribute, so the name, the description and a `<section>`'s role would all have read "no references" for exactly the case a page went out of its way to express. Three sites, two members; the other six are read nowhere.

**The state is split by what needs an engine, not by what declared it.** `Explicit` is `WeakReference<IElement>[]` and engine-free — it moves to `AriaElementReferences`, a `static readonly ConditionalWeakTable<IElement, …>`, the pattern this package already uses six times and `Accessibility/` twice. Only the frozen array a getter last answered stays on `DomRealm`, because a JavaScript object belongs to the engine that made it. `Accessibility/` reads the engine-free half through one `internal static` door over `IElement` — the standing `Dom/Views/CssCascade` already has there, and `Accessibility/AGENTS.md` now records it. It does **not** ask the binding, because `DomRealm.Of` creates a realm if there is none, and computing an accessible name must never construct one.

**The reconciliation moved with the state, and that is the part that is not bookkeeping.** HTML's attribute change steps ran only inside the IDL getter, so `el.ariaLabelledByElements = [x]; el.setAttribute('aria-labelledby','y')` with no intervening read would have handed a direct reader of the field the reference it had just dropped. There is one door now and the field is never read through anything else — `AnIdrefWrittenOverAnIdlSetRelationshipNamesTheElementInstead` is that case.

Five tests, in `Accessibility/`; **three of the five fail without the change**, and the two that do not are the negative cases. As predicted, it gains **zero wpt rows** — `aria-element-reflection-labelledby.html` is not vendored and no accname corpus is — and the **thirteen accessibility golden files are byte-identical**, which is the outcome that says the idref resolver was not touched. The two resolvers still disagree for a detached subtree and for shadow roots; that is a separate defect and is left alone.

**On the non-atomic lazy init:** removed rather than argued. The per-engine cache table is now a `readonly` field initialised in `DomRealm`'s constructor, and the relationship state is `static readonly`, so there is no lazy initialisation left to be unsafe. One empty `ConditionalWeakTable` per realm, against two ~200-entry arrays it already allocates.

## One note for whoever owns `Wpt/README.md`

Only the census table's two numbers are touched here, by `JINT_WPT_BROWSER_CENSUS=update`, because the census test fails without them. The **causes table** is left alone as agreed — but its `#3773` row is now doubly stale: it still reads "65 | 3 | The ARIA reflection mixin is not there", and after this the cause has no rows at all.

Fixes #3773


sebastienros:
Integrated on current `main` by #3874 and merged as 43f6f8931af867f3686d507a5c2d1c6ec328e6ea. The integration preserves your original commit and author attribution while resolving the intervening HTML reflection and census changes. Thank you, @lahma.

Issue #3771, opened 2026-09-03
A page's child frame is never given an engine, so iframe.contentDocument is null and iframe.onload never fires — 540 wpt rows and 17 documents that cannot report


lahma:
The browser lane's new DOM suites make this the second-largest cause in the corpus: **540 failing tests across 13 documents**, plus **17 more documents that report nothing at all** because they wait for a load that never arrives.

It is already a category in the lane (`WptDivergence.NeedsIframeScripting`) and it was already known from `custom-elements/`, where thirty-seven documents are not vendored for it. What the DOM corpus adds is the size of it: a very large part of wpt's DOM coverage is written as "make this assertion in each of an HTML document, an XML document and an XHTML document", and the second and third of those are always frames.

### What happens

```html
<iframe src="/common/dummy.xml"></iframe>
<script>
  const frame = document.querySelector('iframe');
  frame.contentDocument;            // null
  frame.onload = () => { /* never runs */ };
</script>
```

`Jint.Browser/Runtime` parses a child frame's markup and gives it no engine and no document of its own, so
`contentDocument` is null, `contentWindow` is undefined, and no `load` event is ever dispatched at the frame
element.

### Which tests

488 of the 540 are one line — `Cannot read properties of null (reading 'documentElement')` — from
`dom/nodes/Document-createElementNS.html` (390 of its 447 failures) and `Document-createElement.html` (98 of
106), each of which runs its whole table once per document type. The rest are
`dom/nodes/node-realm-*.html`, `node-creation-realm.html`, `Node-isConnected.html`,
`dom/traversal/TreeWalker-acceptNode-filter-cross-realm.html` and `TreeWalker-realm.html`.

Seventeen documents cannot report at all and are `NotVendored` rows: `Comment-constructor.html`,
`Text-constructor.html`, `Document-URL.html`, `Document-characterSet-normalization-1.html` and `-2.html`,
`Document-createElement-namespace.html`, `Element-matches.html`, `Element-webkitMatchesSelector.html`,
`ParentNode-querySelector-All.html`, `MutationObserver-cross-realm-callback-report-exception.html`,
`Node-parentNode.html`, `Node-baseURI.html`, `attach-shadow-realm-after-adoption.html`,
`create-element-realm-after-adoption.html`, `getElementsByClassName-31.htm`,
`query-target-in-load-event.html`, `dom/traversal/TreeWalker-acceptNode-filter-cross-realm-null-browsing-context.html`.

[...]

PR #3825, opened 2026-09-04
A frame served XML gets an XML document, so the wpt XML half of createElementNS reports what it means

lahma:
Phase 1 of the three that reach the [#3771](https://github.com/sebastienros/jint/issues/3771) rows without a
second realm. The survey that decided the phases is [a comment on the issue](https://github.com/sebastienros/jint/issues/3771#issuecomment-5537117352).

## What was wrong

`<foo>Dummy XML document</foo>` served to a frame as `text/xml` came back as an **HTML** document. The text
survived — wrapped in an `<html><body>` skeleton — so `documentElement.tagName` was `HTML`, and every XML
rule a page then asked about was answered by the wrong kind of document.

AngleSharp needs its XML document factory registered to produce one, and `ParserDriver` built its
configuration with `.WithCss()` and nothing else. **Registering it is the whole change.** AngleSharp.Xml is
already referenced for `DOMParser`'s four XML content types, so this costs a service registration and no
dependency.

## Only a frame can reach it, by construction

Not by a guard that could rot: `Parse` states `text/html` for the document a navigation produces — what the
navigate rules already decided — and a top-level navigation to an XML content type is refused by
`DocumentFetch` before any parser sees it. Both halves are pinned by a test, the second one asserting the
refusal rather than a parse.

## Measured

| | before | after |
| --- | ---: | ---: |
| `dom/nodes/` not passing | 1,439 | **1,383** |
| lane total not passing | 1,855 | **1,784** |
| `Document-createElementNS.html` tests registered | 596 | 596 |
| …passing | 173 | **229** |

**56 rows, for one line.** The document registers the same 596 tests either way — worth saying because an
earlier note of mine said 205, which came from dividing by the file's own advertised case count of 447
instead of by what it measurably registers. The run is the number; 56 is the run.

## The exclusion table is re-derived from that run, not narrowed by hand

* `"createElementNS test in XML*"` covered **56 tests that now pass**, so the runner refused it. It becomes
  `"createElementNS test in XML document:*_ERR\""` — the 110 rows that reach `doc.defaultView.DOMException`,
  which a frame with no realm has no window for. Every one of their names ends in the exception it expects,
  which is exactly what separates them from the 56 that do not throw at all. Verified: **110 failing matched,
  zero passing.** That one pattern replaces the 95 exact names a mechanical carve produced.
* `"createElementNS test in XHTML*"` is **no longer about frames** and moves `NeedsIframeScripting` →
  `NeedsXmlDocuments`. `application/xhtml+xml` is routed to the HTML parser *even with the XML factory
  registered*, so all 195 fail on the trailing newline an HTML skeleton adds. That is AngleSharp's own
  content-type mapping and it is phase 2's subject.
* **29 rows join the two groups their HTML twins already sit in** — fifteen where a nullable `DOMString`
  answers the string `"null"` ([#3712](https://github.com/sebastienros/jint/issues/3712)) and fourteen
  validate-and-extract refusals. Those were always the cause; the frame was simply in front of them.

Every new pattern was checked against the run to match at least one failing test and no passing one before it
was written down, which the runner then re-checks from both sides on every build.

## Fixed in passing

The README's `#3771` row said *"none of these 540 moved"*. True when it was written, and not now — it says 56
have, and which half of `Document-createElementNS.html` is left and why.

## What this does not do

- **XHTML is still parsed as HTML** — 195 rows of this document and 49 of `Document-createElement.html`.
  Phase 2.
- **A frame still has no window**, so `contentWindow` is `null` and a frame document's `defaultView` is
  undefined — the 110 rows above, plus `Document-createElement.html`'s 98. Phase 3.
- **No realm.** The `node-realm-*`, `node-creation-realm` and cross-realm `TreeWalker` documents — about 52
  rows — still need one, and #3771 stays open for exactly that.

## Verified

* `dotnet test Jint.Tests.Browser -c Release` — net8.0 1,711 passed, net10.0 1,715 passed, 0 failed, with
  `JINT_WPT_BROWSER_CENSUS=1` set so both the census and the exclusion table are enforced.
* The two new tests were run against the **unfixed** code first: `AFrameServedXmlGetsAnXmlDocument` failed
  with `"HTML"` where it wants `"foo"`.
* `Jint.Tests` corpus, census and agent-instruction checks — 0 failed.

No public API change and no migration section, so nothing to number against §5.30/§5.31.

### DIFF
--- a/Jint.Browser/Runtime/Parsing/AGENTS.md
+++ b/Jint.Browser/Runtime/Parsing/AGENTS.md
@@ -84,6 +84,15 @@ script runs before a module script that precedes it in the document, because the
 HTML's one; and the first import map found anywhere applies to every module, because none of them could have
 resolved before the parse ended anyway.
 
+**A frame served XML gets an XML document, and only a frame can.** `Configuration` carries AngleSharp.Xml's
+document factory, so a response whose content type is an XML MIME type is parsed by the XML parser rather
+than wrapped in an HTML skeleton — without it `<foo>x</foo>` served as `text/xml` came back with
+`documentElement.tagName === "HTML"` and every XML rule a page then asked about was the wrong document's.
+The page's own document cannot reach it: `Parse` states `text/html` for what a navigation produces, and a
+navigation to an XML content type is refused by `DocumentFetch` before a parser sees it. `application/xhtml+xml`
+is **still** routed to the HTML parser — that is AngleSharp's own content-type mapping, not this file's, and
+it is what `NeedsXmlDocuments` covers in the browser lane.
+
 **`document.readyState` is the page's shadow.** `PageRuntime.ReadyState` moves `loading` → `interactive` →
 `complete` and `ParserDriver.SetReadyState` fires the `readystatechange` that goes with each, because
 `Document.ReadyState`'s setter is protected and unreachable from outside AngleSharp's assembly. AngleSharp's
--- a/Jint.Browser/Runtime/Parsing/ParserDriver.cs
+++ b/Jint.Browser/Runtime/Parsing/ParserDriver.cs
@@ -99,6 +99,15 @@ internal sealed class ParserDriver : IDisposable
         // service rather than replacing one, so AngleSharp's own observer keeps working.
         var configuration = Configuration.Default
             .WithCss()
+            // https://html.spec.whatwg.org/multipage/document-lifecycle.html#read-xml — a document whose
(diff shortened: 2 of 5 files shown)


sebastienros:
Superseded by #3873, which rebased and integrated this XML child-frame work on current `main` while preserving the original contribution and authorship. #3873 has now passed the full required CI matrix and was squash-merged as 70917bc7c9a4507b6dc37d9b776b60e38229344c. Thank you for the implementation.

PR #3826, opened 2026-09-04
XHTML in a frame is AngleSharp's content-type mapping, and the register says so rather than the next reader re-deriving it

lahma:
> **Stacked on [#3825](https://github.com/sebastienros/jint/pull/3825)**, whose commit is the first here.
> Review only the second; merge #3825 first and this rebases to one.

Phase 2 of [#3771](https://github.com/sebastienros/jint/issues/3771) went the other way, which was one of the
two allowed outcomes. **It is a write-up, not a fix**, and this is the evidence — because the alternative is
somebody spending the same afternoon on it again in six months.

## The question

#3825 left `application/xhtml+xml` as the largest single cause in the browser lane's DOM table: 195 rows of
`Document-createElementNS.html` and 49 of `Document-createElement.html`. Registering AngleSharp.Xml's factory
fixed `text/xml` and demonstrably did **not** fix this one. So: is AngleSharp's content-type → factory mapping
reachable from our side at all?

## What was measured, type by type

Against a frame, with the XML factory registered:

| content type | what a frame got |
| --- | --- |
| `text/xml` | XML document, `documentElement.tagName` `"foo"` — correct |
| `application/xml` | XML document — correct |
| **`application/xhtml+xml`** | **HTML document, `tagName` `"HTML"`** — wrong |
| `image/svg+xml` | no document at all — a separate question no vendored document asks |

So exactly one type is both wrong and loaded.

## Why it cannot be composed

`DefaultDocumentFactory.Register` and `Unregister` are **public**, so the mapping looks extensible. It is not,
and each step below was run rather than reasoned about:

1. **`Configuration.With<T>` adds a service rather than replacing one** — which the parse already relies on
   for `IAttributeObserver`, and says so in a comment. So a second `IDocumentFactory` makes AngleSharp's own
   resolution throw `Sequence contains more than one element` **on every page load**. Measured: 20+ unrelated
   tests failed identically.
2. **`Without<IDocumentFactory>()` then `With(ours)` resolves — and discards whatever `WithXml()`
   contributed.** `text/xml` and `application/xml` regressed to HTML in the same run. So `WithXml()` works by
   configuring that one service, and the service is all-or-nothing from outside.
3. **AngleSharp.Xml's XML creator is not public**, so nothing can be composed in its place. Driving
   `XmlParser` over a materialised `TextSource` ourselves produced no usable document — and would have meant
   this package *building* documents rather than choosing a parser, which is the one thing
   `Jint.Browser/AGENTS.md` says it is not for.

The remaining paths were rewriting the response's `Content-Type` before AngleSharp sees it (a lie in the
transport, and `document.contentType` would then read the lie) or reaching a shared factory instance
AngleSharp composed (not exposed, and process-wide if it were). Neither is acceptable, so neither was taken.

## The sharpest part, and what a fix would actually remove

**The same package already answers this content type correctly.** `Dom/Views/JsDomParser` — `DOMParser` —
routes `text/xml`, `application/xml`, `application/xhtml+xml` and `image/svg+xml` to `XmlParser`. So
`new DOMParser().parseFromString(s, 'application/xhtml+xml')` gives an XML document and `<iframe src>` gives
an HTML one, **for the same bytes and the same type**. That inconsistency is ours; the thing standing in the
way of removing it is upstream's mapping.

## What lands

One row in `Dom/divergences.md`, which is where an AngleSharp divergence is recorded rather than worked
around, carrying the table above, the three measured refusals, and the `DOMParser` inconsistency.

No exclusion moves: the browser lane has carried these rows as `NeedsXmlDocuments` since #3825, which is the
category for a scope decision rather than debt.

## Fixed in passing

`Dom/divergences.md` carried the `el.classList.add('')` row **three times** — two partial copies of one full
one, from three changes each appending rather than editing. One row now, the fullest.

## Verified

`dotnet test Jint.Tests.Browser -c Release` with `JINT_WPT_BROWSER_CENSUS=1` — net8.0 1,715 passed, net10.0
1,719 passed, 0 failed. Nothing moves; that is the point.

## For the maintainer

This is **AngleSharp finding #8** and it is held rather than filed, per the standing rule. If it is ever
filed, the ask is small and precise: `application/xhtml+xml` is an XML MIME type and HTML says an XML MIME
type is read by the XML parser, so `DefaultDocumentFactory`'s default mapping should send it to the XML
creator when `WithXml()` is registered.


sebastienros:
Superseded by #3890, which preserves this documentation fix on current main and retains Marko Lahma's authored commit. #3890 is now merged. Thank you for the original fix.

Issue #3701, opened 2026-09-03
Seams the Network and Fetch domains wished they had: an answerable OnResponse, WebSocket observation, request timing, off-loop commands, navigator.platform


lahma:
Five seams the `Network`/`Fetch` domain work (#3700) wished it had. Each one gates a named piece of the protocol that is deliberately left out today.

1. **An answerable `FetchObserver.OnResponse`.** Today it is a `void` notification; a response-stage pause could only ever continue unchanged, so `Fetch`'s response stage, `continueResponse`, `Fetch.getResponseBody`, `takeResponseBodyAsStream` and the `IO` domain are unimplemented rather than half-implemented. A hook that can return a substitute response (status, headers, body) the way `OnRequestAsync` returns a `FetchInterception` unlocks all five.
2. **WebSocket and EventSource handshake observation.** `Jint/WebApi/AGENTS.md` says the engine deliberately does not observe those two handshakes. `Network.webSocketCreated`/`webSocketWillSendHandshakeRequest`/`webSocketHandshakeResponseReceived`/`webSocketClosed` need `OnWebSocketHandshake`-shaped callbacks on `FetchObserver`, kept out of the request lifecycle so an open socket is not a request that never finishes (which would stop `Page.lifecycleEvent(networkIdle)`).
3. **A per-request timing record on `ObservedFetchResponse`** (DNS, connect, TLS, send, receive) — the only thing between the page and an honest `Network.ResourceTiming`; a document of zeros reads as a page that loaded instantly.
4. **A gateway hook for commands that touch no engine state.** `Fetch.continueRequest` and friends are answered on the page loop; a `DevToolsTarget.RunsOffThread(method)` seam would make "a pause never blocks the loop" unconditional instead of true with one named exception (a `<script src>` a running script inserted).
5. **A `navigator.platform` option on the engine.** `Network.setUserAgentOverride`'s `platform` reaches the *next* document only, because the navigator is built with the engine.

Also from R8b (#3699): a test that runs one WPT document and prints every result (triage today means editing the exclusion table to find out what to put in it), and a per-`Page` loop-turn counter for scheduling tests.

Recorded rather than left as TODOs, per the campaign's rule.

lahma:
Three have landed; the rest is unchanged.

[...]

PR #3827, opened 2026-09-04
The Fetch response stage and the four WebSocket events, over the two seams that now exist

lahma:
Refs #3701 — items 1 and 2, wired.

> **Stacked on #3823**, which is stacked on #3821. Review the top commit; the two below are those PRs'. Merge them first and this rebases to its own fifteen files.

Both seams were named in the manifest and in two instruction files as *absent with a reason*. The reasons are now false, so they are rewritten here rather than left to rot.

## The `Fetch` response stage

Over `FetchObserver.OnResponseAsync`. A pattern asking for `requestStage: "Response"` pauses with `responseStatusCode`, `responseStatusText` and `responseHeaders`; **`Fetch.continueResponse`** answers one, and `fulfillRequest` and `failRequest` answer one too. A pattern naming no stage still pauses the request stage alone — that is the protocol's own default, and pausing both for one pattern would double every pause a recorded client expects.

It crosses the three layers, as predicted:

- `IPageNetworkListener` grows an **ask** (`ResponseWillBeDeliveredAsync`) beside its `ResponseReceived` notification — the same ask/notify split the engine seam has, so a client that rewrote a status sees its own value in the announcement that follows.
- `PageNetworkRecorder` overrides `OnResponseAsync` and translates a `PageNetworkResponseDecision` into the engine's `FetchResponseInterception`.
- `FetchDomain` holds the pause in a **second map**: the two stages are answered with different decisions and one identifier space names both, so a command looking in the wrong one would answer "Invalid InterceptionId" for an identifier it had just handed out.

### `RunsOffThread` — I checked, and yes

**`Fetch.continueResponse` belongs there, for a reason no weaker than the other three's.** A `<script src>` a running script inserted is fetched with the loop *blocked* on the whole fetch — `ParserDriver`'s `fetch.GetAwaiter().GetResult()`, not `PumpUntil` — so the loop is held from the request stage through the body read, and the response-stage ask happens inside that window. A response-stage pause therefore holds the loop exactly as a request-stage pause does, and a `continueResponse` that had to reach the loop would deadlock on the very fetch it is releasing. It clears the bar the thread rule sets: like the other three it looks one entry up in a dictionary and completes a promise — no engine, no `JsValue`, no node.

## The four `Network` WebSocket events

Over #3823's socket observer. `PageNetworkRecorder` grows a **second** observer for them, because the engine has two seams and one class can derive from only one; it writes to no log and holds no state.

**A socket takes no entry in the request log, and its identifier is `ws-`-prefixed.** Both are the constraint the issue named, met structurally: a socket stays open for as long as the page wants it, so an entry would leave a request outstanding for that whole time and `networkIdle` would never fire — and a client must not be able to ask `getResponseBody` of a socket. `webSocketFrameSent`, `webSocketFrameReceived` and `webSocketFrameError` stay absent, honestly: the observer is told about the two handshakes and the close, and a frame never reaches it.

## What is still out, and why

**`Fetch.getResponseBody`, `takeResponseBodyAsStream` and the whole `IO` domain.** The pause has the response's *headers* while its body is still on the socket, so there are no bytes to hand a paused client without buffering them first — a budget decision (how much, whose, dropped how) rather than a hook. `Network.getResponseBody` answers a body here. **Worth an issue** rather than a guess in this PR.

## What was verified

Six tests. `Jint.Tests.Browser/DevTools/FetchDomainTests`: the response-stage pause carrying status and headers with `continueResponse` letting it through, `continueResponse` rewriting a status and headers while the server's own body still arrives, `fulfillRequest` and `failRequest` answering a response pause, and a default pattern pausing exactly once. `NetworkDomainTests`: a socket reaching a client as its own events and **not** as a request.

That last one uses a URL the context's filter refuses, which is what lets it run with no WebSocket server: a refused socket is still created and still closes, which is the pair that proves the whole chain — engine observer, recorder, listener, target, domain, client. What the two *handshake* events carry is pinned one level down in `Jint.Tests/Runtime/WebApi/WebSocketTests` over a connection double, because the connection factory is engine-internal and a browser test cannot inject one.

- `Jint.Tests.Browser` — 1 714 / 1 718, 0 failed.
- `Jint.Tests.DevTools` — 405 / 407, 0 failed (the manifest and generated-protocol currency checks).
- `Jint.Tests` — 12 300 / 12 300 / 8 492, 0 failed.
- `Jint.Tests.PublicInterface` — 3 603 / 3 613 / 2 879, 0 failed.

Manifest: `Fetch.continueResponse` under `implementedMethods`, the four socket events under `implementedEvents`, regenerated — the diff is three generated files, the rest is line-ending noise git normalizes away.

## Fixed in passing

The manifest's `Fetch` paragraph, `FetchDomain`'s class and `Matches` remarks, and `Jint.Browser/DevTools/AGENTS.md` on both lanes and on `RunsOffThread` — every one of them explained an absence that this PR removes.

### DIFF
--- a/Jint.Browser/DevTools/AGENTS.md
+++ b/Jint.Browser/DevTools/AGENTS.md
@@ -72,20 +72,32 @@ target/runtime split and the manifest are there and none of it is repeated here.
   [`Runtime/AGENTS.md`](../Runtime/AGENTS.md#the-request-log-is-the-protocols-seam-too) argues why moving them
   would deadlock the one fetch a page cannot pump through. **The three commands that release a pause run
   there too**, which is what makes "a pause never blocks the loop's answer" unconditional rather than true
-  with one exception: `PageTarget.RunsOffThread` names `Fetch.continueRequest`, `failRequest` and
-  `fulfillRequest` — they look one entry up and complete a promise, touching no engine, no `JsValue` and no
-  node, which is the bar [`Jint.DevTools/AGENTS.md`](../../Jint.DevTools/AGENTS.md#the-thread-rule) sets —
-  so a `<script src>` a running script inserted, fetched with the loop *blocked* rather than pumping, is
-  released while it blocks. `Fetch.enable` and `disable` are deliberately not named: they are the domain's
+  with one exception: `PageTarget.RunsOffThread` names `Fetch.continueRequest`, `failRequest`,
+  `fulfillRequest` and `continueResponse` — they look one entry up and complete a promise, touching no
+  engine, no `JsValue` and no node, which is the bar
+  [`Jint.DevTools/AGENTS.md`](../../Jint.DevTools/AGENTS.md#the-thread-rule) sets — so a `<script src>` a
+  running script inserted, fetched with the loop *blocked* rather than pumping, is released while it
+  blocks. **`continueResponse` is on that list for a reason no weaker than the other three's**: the loop is
+  blocked on the *whole* fetch (`ParserDriver`'s `fetch.GetAwaiter().GetResult()`), from the request stage
+  through the body read, so a response-stage pause holds it exactly as a request-stage pause does. `Fetch.enable` and `disable` are deliberately not named: they are the domain's
   own state, not a request's. The document's request carries the `loaderId` as its `requestId`, which is how
   every client tells a navigation apart.
 - **What is accepted and not effective says so, in place.** `Network.setCacheDisabled` (there is no cache)
-  and `Audits.enable` are answered because a refusal fails an ordinary connection. Two whole lanes are
-  absent with a reason rather than pending: the `Fetch` **response stage** and with it `IO` (an observer
-  cannot answer `OnResponse`, so a response-stage pause could only continue unchanged), and the **WebSocket**
-  events and `eventSourceMessageReceived` (a socket's handshake is not a fetch and is not observed; a stream
-  *is*, but as bytes rather than as the events they decode into, so its requests are in the log as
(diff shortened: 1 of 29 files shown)


lahma:
Force-pushed one follow-up commit's worth of corrections, amended into the same commit.

The PR's own story is that the three places naming these events as *absent with a reason* get their reasons rewritten. Two were missed, and both were left contradicting the same file they live in:

- `tools/devtools-protocol/manifest.json` still said "Network's WebSocket and EventSource events are absent because the engine deliberately does not observe those two handshakes" — forty lines above the four `Network.webSocket*` entries this PR adds to `implementedEvents`. Rewritten to say what is actually here (the four lifecycle events over the socket observer) and what is not (`webSocketFrame*`, because the observer is never told about a frame; `eventSourceMessageReceived`, because a stream is observed as bytes).
- `Jint.Browser/DevTools/NetworkDomain.cs`'s class doc said the same thing in prose. Same correction.

While in the manifest I also fixed one pre-existing staleness in the same comment block: "Input is dispatchMouseEvent alone. The keyboard half … is the next campaign item" — `dispatchKeyEvent`, `insertText` and `imeSetComposition` have been in `implementedMethods` since that campaign item landed. It now names the four that are answered and the nine that are not.

The manifest's whole-file digest is stamped into two generated files, so the regeneration is in the commit too: `ProtocolJsonContext.g.cs` and `ProtocolManifest.g.cs`, one header line each, `sha256:c131a95ae83a` → `sha256:ff928e6894f7`. `Jint.Tests.DevTools` protocol suite green (74/74 on both frameworks).

PR #3831, opened 2026-09-04
A child frame gets a window of its own on the page's realm, so contentWindow, defaultView and frames[i] answer

lahma:
Part of #3771, and the half of it that is not a realm.

## What the enumeration said

The issue reads as "a frame needs an engine". The [survey on #3771](https://github.com/sebastienros/jint/issues/3771) says otherwise: of the 484 rows left in that cause after #3809 and #3825, most were not waiting for a second realm at all — they were waiting for an **object**. `create_window_in_test` wants a window to resolve with, `doc.defaultView.DOMException` wants a constructor to compare a refusal against, `frames[0]` wants an indexed property. A realm is the expensive half and none of those need it.

## The shape

A frame gets a **window** and not a realm. `Runtime/FrameWindows` builds one object per frame whose `[[Prototype]]` is the page's global object, and that one line is the whole design: every interface object, every constructor and every intrinsic is inherited, `frameWindow instanceof Window` still holds, and what is *different* about a frame is shadowed by an own property — itself for `window`/`self`/`frames`, the page for `parent`/`top`, and its own `document`, `frameElement`, `length`, `name`, `origin` and `location`. The object is remembered per frame element, so `contentWindow === contentWindow` and `frames[0] === contentWindow`.

`location` is shadowed rather than inherited because inheriting it would answer the **page's** URL for a document that is somewhere else — wrong information rather than a shared object — and every write to it throws a `TypeError` that names the refusal. That last part is not a preference. Against a silent no-op, `event-global-is-still-set-when-coercing-beforeunload-result.html` waits forever for a frame `load` that a navigation was supposed to cause, and the whole file **TIMEOUTs** — the one outcome `Jint.Tests.Browser/Wpt/AGENTS.md` singles out as worse than a failure. It was a real regression during this work, and the refusal is what fixed it.

## Measured, on the browser wpt lane

| Suite | Before | After |
| --- | ---: | ---: |
| `dom/events/` | 20 | 15 |
| `dom/nodes/` | 1,345 | 1,213 |
| `dom/traversal/` | 7 | **0** |
| `dom/ranges/` | 10 | 9 |
| **total** | **1,784** | **1,639** |

−145 not passing, and the #3771 cause itself goes from **484 rows over 13 documents to 41 over 15**. Nine exclusion entries are removed because a real run says they no longer match anything failing, and `Document-createElement.html`'s XML half is narrowed from one glob into 39 measured entries. (The Documents column rises because the previous 13 counted the DOM suites only; the six `html/webappapis/scripting/` documents were always in this cause.)

**Three documents now pass trivially, and the README says so rather than counting them as evidence.** `TreeWalker-realm.html`, `TreeWalker-acceptNode-filter-cross-realm.html` and `EventListener-handleEvent-cross-realm.html` each assert that a node keeps its own realm across a hop. With one realm for both sides to be in, the corpus cannot tell that from having no second realm at all: they are green and they are not proof of anything.

## The divergence, written down rather than left to be found

`frame.contentWindow.DOMException === DOMException` is `true`, where a browser gives a frame its own constructors. It is truthful for a one-realm browser — the exception a call on the frame's document really throws *is* the page's, because the page's realm built the wrapper — and `Jint.Browser/Dom/divergences.md` records it with that reasoning, beside the `location` refusal. `contentWindow !== window` and `frames[0] === contentWindow`, which is what the corpus actually tests, do hold.

## One thing the rebase changed, not the change

The pinned AngleSharp assemblies now project `HTMLIFrameElement.contentWindow` (they did not when this was
written), so declaring it in `overrides.json` under `additions` alone is **reported by the generator and
ignored** — the page would have got AngleSharp's own `IWindow` conversion rather than this one. It is a `skip`
plus an addition now, a re-declaration exactly as its sibling `contentDocument` already was.
`TheGeneratorReportsOnlyTheTwoDiagnosticsWeKnowAbout` is what caught it, which is the diagnostic list earning
its keep: an ignored override is otherwise silent.

## Small things fixed here because they were found here

- the root `README.md` still said an `<iframe>`'s `src` is not fetched, which #3809 changed;
- four more files still said `contentWindow` is `null`: `Frame.cs`, `BrowserOptions.cs`, `docs/design/headless-browser.md` §3 and `Jint.Tests/Wpt/WptExclusions.cs`'s category doc;
- the not-vendored tables still gave "an `iframe.onload` never arrives" as their reason, untrue since #3809. Those rows are owed a re-derivation that takes vendoring the documents (a change of its own, because it moves the census columns), and the comments now say that instead of stating a spent cause as fact.

## Found and deliberately not fixed

An `<iframe>` with no `src` gets no document, where HTML gives every nested browsing context an initial `about:blank` one. AngleSharp asks the resource loader for a frame's document only when there is a source to ask about, so nothing here is refusing it — and opening a document into a context nobody navigated is not on AngleSharp's public surface. It is what the five `window-onerror-with-cross-frame-event-listeners-*` documents meet *before* they ever reach the realm (`new frames[0].Function(…)` reads a member of `undefined`), and the exclusion comment now records the measurement rather than guessing.

## Verification

`dotnet test Jint.Tests.Browser -c Release` on **net8.0** and **net10.0**, both green (1,756 / 1,761 passed), plus `JINT_WPT_BROWSER_CENSUS=1` for the census ceiling and the two-sided exclusion check. `ChildFrameTests` grew to 16 tests, three of them new: the window and its identities, `defaultView`, the `location` read and refusal, and wpt's own `create_window_in_test` helper run in the shape the corpus uses it.

Stacked on #3825 and #3826; it needs both.


sebastienros:
Superseded by #3891, which adapted this work to current `main`, preserved Marko Lahma’s contribution through co-authorship, and has now merged. Closing this older PR so the queue reflects the landed implementation.

PR #3836, opened 2026-09-04
A 401 is a question a client can answer: Fetch.authRequired and continueWithAuth, over a new observer ask

### LINKED ISSUE #3828: Fetch.getResponseBody, takeResponseBodyAsStream and the IO domain need a body-buffering budget, which is a design call rather than a hook
Split out of #3701 item 1, which #3821 and #3827 closed the seam half of. Recording the part that is genuinely undecided rather than half-implementing it.

## Why the seam did not unlock these three

#3821 made `FetchObserver.OnResponseAsync` answerable, and #3827 wired the `Fetch` **response stage** over it — `continueResponse`, and `fulfillRequest`/`failRequest` from a response pause. That is enough for a client that wants to *rewrite* a response.

It is **not** enough for a client that wants to *read* one. The response-stage pause holds the response's **headers** while the body is still on the socket, so there are no bytes to hand a paused client without buffering them first. `getResponseBody`, `takeResponseBodyAsStream` and the `IO` domain all need the bytes.

## The three open questions

1. **The budget and its unit.** How many bytes, and per request or per page? A per-request cap is simple and lets a page with many large responses cost an unbounded total; a per-page cap is honest about the page's footprint and makes one request's cost depend on its siblings.
2. **Who arms it.** `Network` already has body capture; the natural answer is that buffering is armed only while a response-stage pattern is registered, so a page nobody is intercepting pays nothing — which is the same shape as `WebSocketObserver` being wired only when a host sets one (#3823) and `CollectHttpResponseDetails` being turned on only then.
3. **What happens when a body exceeds the budget.** This is the sharp one. The buffered prefix has to be **replayed ahead of the rest of the stream**, or the cap silently becomes a truncation the client cannot see — and a client that asked for a body and got a prefix with no signal is worse off than one that was refused.

## Why it is not a hook

Every one of those is a decision about what a page is allowed to cost, not about where a callback goes. Guessing at them would produce exactly the half-implemented command #3701 exists to avoid — which is why the agent that built the seam stopped here and said so.

Related: #3701 (items 1 and 2, seam halves now landed), #3821, #3823, #3827. Tracked by #3575.

lahma:
Closes #3828's first-ranked gap. Stacks on #3833; only the last commit is this PR's.

**The only protocol gap that failed silently.** Both Puppeteer and Playwright send `Fetch.enable { handleAuthRequests: true }` unconditionally whenever they intercept — Playwright writes the flag literally in `crNetworkManager.ts`, Puppeteer likewise in `NetworkManager.ts` — and this browser accepted it and raised nothing. A `401` went to the page as an ordinary response, and the credentials a client had configured through `page.authenticate()` or `httpCredentials` were discarded with no error anywhere. Every other absent command answers `-32601`; this one answered "yes" and did nothing.

## The seam

`FetchObserver` grows a third ask beside `OnRequestAsync` and `OnResponseAsync`:

```csharp
public override ValueTask<FetchAuthDecision?> OnAuthRequiredAsync(
    ObservedFetchAuthChallenge challenge,
    CancellationToken cancellationToken)
    => new(challenge.CanProvideCredentials
        ? FetchAuthDecision.ProvideCredentials("ada", "l0velace")   // sends the hop again
        : null);                                                    // delivers the 401
```

**It is asked in `SendForStreamCoreAsync`, not `SendForStreamAsync`** — beside `OnRequestAsync` rather than `OnResponseAsync`. That is the same reasoning that made #3821 serve every lane instead of `fetch()` alone: a `401` on a document fetch, on a subresource and on an `XMLHttpRequest` is exactly the case this is for, and only `fetch()` takes the lane the final-response ask is asked in, so an ask placed there would have served the one caller that needed it least.

## Only `Basic` can be answered, and every scheme is still reported

`Basic` is the one scheme whose answer is a function of the credentials alone — one `Authorization` header, no state — while `Digest` needs a nonce exchange and `Negotiate` and `NTLM` a handshake bound to the connection, none of which a transport that hands the socket back after every response can hold.

Reporting only the answerable ones was the tempting narrow reading, and it recreates the same defect in a smaller place: a client that got no `authRequired` for a `Digest` challenge would learn nothing about why its credentials did nothing. So:

- the challenge is raised **whatever scheme it names**, which also keeps the protocol shape Chrome's;
- `ObservedFetchAuthChallenge.CanProvideCredentials` answers which kind it is **before** the observer decides;
- `Fetch.continueWithAuth` with `ProvideCredentials` for an unanswerable scheme is **refused with an error naming the scheme** — `Credentials cannot be provided for a 'Digest' challenge.` — and the `401` is then delivered.

**An ask that cannot be honoured must fail visibly, because an ask that silently does nothing is the defect this seam exists to remove.** That sentence is in the observer's own docs, in `Jint/WebApi/Fetch/AGENTS.md`, and in the domain's.

## The two bounds

**One retry per request**, not per hop (`authRetried`): an observer has one credential to offer, so a challenge on the retry is delivered rather than asked about again, and a loop is worse than a `401`. The retry is not a redirect and spends none of `MaxRedirects`.

**A `407` is not reported at all**, and `source` is always `Server`. The proxy belongs to the `HttpClient` the host supplied, and setting `Proxy-Authorization` under a handler that re-frames the request is not this engine's to do.

## Two things that were already right

**The cross-origin rule needed no new code.** `FetchTransport._crossOriginHeaderNames` already strips `authorization` when a redirect leaves the origin — [HTTP-redirect fetch](https://fetch.spec.whatwg.org/#http-redirect-fetch) step 13 — so the header this seam adds is dropped at exactly the right boundary. Stated in the PR and in the docs because the next reader will wonder.

**The precondition was checked rather than assumed.** `CreateSharedClient` sets no `HttpClientHandler.Credentials` and no `PreAuthenticate`, so a `401` really does arrive unhandled rather than being answered underneath us.

## The protocol half

`Fetch.enable`'s `handleAuthRequests`, the `Fetch.authRequired` event and `Fetch.continueWithAuth`, over a third pause map — one identifier space, three things it can name, which is the shape #3827 established for the response stage. `handleAuthRequests` is read as a flag rather than filtered by pattern, because a challenge is a property of a response and there is nothing for a URL pattern or a resource type to match on; that is the protocol's own shape too.

`Fetch.continueWithAuth` joins the four commands in `PageTarget.RunsOffThread`, for the reason they are all there: the challenge arrives on the hop's own response, with the page loop blocked on that fetch if a running script started it, and answering it looks one entry up in a `ConcurrentDictionary` and completes a `TaskCompletionSource` — no engine, no `JsValue`, no node.

## Tests, at both levels

**`WebApiFetchDocumentTests`** drives the engine seam from the embedder's side, against the real transport:

- a `Basic` challenge answered, with the retry's own `Authorization: Basic YWRhOmwwdmVsYWNl` read off the handler and the first attempt asserted to carry none;
- a declined challenge and an uninterested observer both delivering the `401`, with exactly one request sent;
- a `Digest` challenge reported as `answerable=False` and its credentials refused — one request, no invented header;
- a server that challenges twice getting exactly one retry and exactly one ask.

**`FetchDomainTests`** drives the protocol over a real page and a real origin:

- the challenge pausing with its `source`, `scheme`, `realm` and `origin`, `continueWithAuth` letting the navigation through, and the server's copy of both attempts asserted;
- `CancelAuth` delivering the `401` with nothing sent again;
- `ProvideCredentials` for `Digest` refused by name with `-32000`, and the request still released rather than left hanging on the error;
- no event at all for a client that did not send `handleAuthRequests`.

## Verification

Full solution builds clean. `Jint.Tests` 12323/12323 (net8.0, net10.0) and 8503/8503 (net472); `Jint.Tests.Browser` 1790/1790; `Jint.Tests.PublicInterface` 3627/3627 + 2889/2889 (net472), with the two `PublicApiTest` baselines updated for the three new preview members and `UndocumentedPublicApi` unchanged; `Jint.Tests.DevTools` 409/409, including the manifest and generated-protocol currency checks after regeneration.

### DIFF
--- a/Jint.Browser/DevTools/AGENTS.md
+++ b/Jint.Browser/DevTools/AGENTS.md
@@ -1,4 +1,4 @@
-﻿# Agent instructions: the page-level protocol
+# Agent instructions: the page-level protocol
 
 > **Read this when:** You are touching `Jint.Browser/DevTools/` — a page target, a page-level domain (`Page`, `DOM`,
 > `Network`, `Fetch`, `Input`, `Emulation`, `Accessibility`, `Jint`), or the request log the protocol reads.
@@ -83,7 +83,13 @@ target/runtime split and the manifest are there and none of it is repeated here.
   own state, not a request's. The document's request carries the `loaderId` as its `requestId`, which is how
   every client tells a navigation apart.
 - **What is accepted and not effective says so, in place.** `Network.setCacheDisabled` (there is no cache)
-  and `Audits.enable` are answered because a refusal fails an ordinary connection. The `Fetch` **response stage** is here: a
+  and `Audits.enable` are answered because a refusal fails an ordinary connection. **Authentication is here**:
+  `handleAuthRequests` turns it on, a `401` carrying a `WWW-Authenticate` pauses as `Fetch.authRequired`, and
+  `continueWithAuth` answers it over `FetchObserver.OnAuthRequiredAsync`. Only `Basic` can be answered, every
+  other scheme is still *reported* — being asked is how a client tells "unsupported" from "never challenged" —
+  and `ProvideCredentials` for one of those is refused with an error naming the scheme rather than accepted
+  and dropped. A `407` is a proxy's and is not reported, the proxy belonging to the context's `HttpClient`, so
+  `source` is always `Server`. The `Fetch` **response stage** is here: a
   pattern asking for `requestStage: "Response"` pauses with the response's status and headers, and
   `continueResponse`, `fulfillRequest` and `failRequest` answer one — a default pattern still pauses the
   request stage only, that being the protocol's own default, and pausing both would double every pause a
--- a/Jint.Browser/DevTools/FetchDomain.cs
+++ b/Jint.Browser/DevTools/FetchDomain.cs
@@ -49,9 +49,17 @@ namespace Jint.Browser.DevTools;
 /// budget decision rather than a hook. <c>Network.getResponseBody</c> is what answers a body here.
 /// </para>
 /// <para>
-/// <b>No authentication challenge exists here.</b> <c>handleAuthRequests</c> is accepted and
-/// <c>Fetch.authRequired</c> is never sent: nothing in this browser answers a <c>401</c> with credentials, so
-/// there is no challenge for a client to be asked about, and <c>continueWithAuth</c> is absent.
+/// <b>Authentication challenges are here.</b> <c>handleAuthRequests</c> turns them on, a <c>401</c> carrying
+/// a <c>WWW-Authenticate</c> pauses as <c>Fetch.authRequired</c>, and <c>continueWithAuth</c> answers one
+/// over the engine's <c>FetchObserver.OnAuthRequiredAsync</c>. <b>Only <c>Basic</c> can be answered</b> — it
+/// is the one scheme whose answer is a function of the credentials alone, while <c>Digest</c> needs a nonce
+/// exchange and <c>Negotiate</c> and <c>NTLM</c> a handshake bound to the connection. Every other scheme is
+/// still <i>reported</i>, because being asked is how a client tells "unsupported" from "never challenged",
+/// and <c>continueWithAuth</c> answering <c>ProvideCredentials</c> for one is refused with an error naming
+/// the scheme rather than accepted and dropped: an ask that cannot be honoured must fail visibly. A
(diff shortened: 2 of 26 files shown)


sebastienros:
Integrated on current `main` in #3868, preserving the original commit and authorship. #3868 has merged, so I’m closing this superseded PR. Thank you for the implementation.

PR #3888, opened 2026-09-07
Improve query-scoped browser geometry and add Scalar reproduction

### LINKED ISSUE #3882: Scalar OpenAPI UI operations do not render because geometry queries exhaust the page task budget
## Description

The Scalar OpenAPI UI shell loads in Jint.Browser, but its client-side operation UI does not become usable. Two OrchardCore tests fail independently with `Jint.Browser 5.0.0-preview-1976`:

1. `OpenApiTests.ScalarUIRendersOperationsUnderTenantPrefix`
   - Navigates to the Scalar page and observes the schema request.
   - `page.WaitForLoadStateAsync(LoadState.NetworkIdle)` never completes and times out after 30 seconds.
2. `OpenApiTests.ScalarUISilentlyAuthenticatesAndCallsProtectedEndpoint`
   - The page title/shell loads.
   - `.sidebar a` containing `GetEndpoint` is never created, even after the test's existing wait.

The simpler `ScalarUIIsAccessibleWhenEnabled` title test passes, so the server route and initial HTML navigation work. Top-level JSON document navigation is also fixed in preview-1976 through #3876/#3879; this issue concerns Scalar's browser-side loading/rendering path.

## Versions

- Jint/Jint.Browser/Jint.DevTools: `5.0.0-preview-1976`
- Package commit: `e77604fd70ea2b8bd841d83f6250e28cc78f48ab`
- Microsoft.Playwright: `1.62.0`
- .NET 10.0
- macOS arm64

## Isolated results

```text
ScalarUIRendersOperationsUnderTenantPrefix
System.TimeoutException: Timeout 30000ms exceeded
at Frame.WaitForLoadStateAsync(LoadState.NetworkIdle)
```

```text
ScalarUISilentlyAuthenticatesAndCallsProtectedEndpoint
Timeout 30000ms exceeded
- waiting for Locator(".sidebar a").Filter(HasText = "GetEndpoint").First
```

Each test runs in its own process. Representative durations and peak RSS were approximately 130 seconds / 1.45 GB and 215 seconds / 1.97 GB respectively.

## Expected behavior

Scalar should fetch the OpenAPI schema, settle its network lifecycle, and render the operation sidebar/DOM under Playwright connected to Jint.Browser over CDP, as it does under Chromium.

## Requested investigation

Reduce the exact OrchardCore Scalar assets and schema to a standalone Jint.Browser + Playwright fixture, capture page errors/console/request failures, identify the missing or incorrect browser behavior, and add a focused regression test for the root cause.

sebastienros:
Scalar's sidebar and API Client workflows exhausted the page task budget while geometry reads repeatedly positioned the entire document and computed unrelated CSS. Index required subject classes through AngleSharp's public selector visitor within each query, preserve native matching and full computed-style inheritance, resolve shorthand variable dependencies, and use one placement algorithm for both complete layouts and individual rectangles. Skip irrelevant declarations before allocating scoped rule wrappers. Accessibility snapshots share their cascade through accessible-name computation. Queries remain fresh across DOM, CSSOM, control-state and media changes.

Fixes #3882. Includes AngleSharp 1.8.1 / AngleSharp.Css 1.1.2 from #4045 and current main's browser changes. Full hit tests, rectangle reads, offsets and resize measurements retain one synthetic geometry model and scroll clamp. The renderer's documented layout limitations remain.

Vendor the exact Scalar.AspNetCore 2.12.9 assets and captured OrchardCore Blog schema. Real Microsoft.Playwright clients at root and tenant routes select GetEndpoint, open the GET operation and reach the API Client dialog; a public Page fixture exercises the same workflow. Keep the 30-second engine task budget and render waits, semantic locators, page-error assertions and completed-request checks. Playwright clicks span several protocol commands and page turns, so their hang bound uses TestBudgets.WedgeCeiling (two minutes). With this exact test on main's unchanged runtime, the sidebar still fails to appear within 30 seconds and PageLoop reports that its task budget elapsed. Failure diagnostics preserve the original exception even if the execution context is lost. Scalar 2.0.0 remains a separate upstream Vue diagnostic.

Run this test project's target frameworks sequentially by default: NUnit's nonparallel fixtures cannot isolate separate test hosts, and simultaneous .NET 8/.NET 10 runs contend for the task budgets. Both complete suites and engine deadlines remain enabled in local runs and every CI leg.

Validation: the focused 354-test selection covers native selector and full-declaration equivalence, flex-variable inheritance, snapshot freshness, nested rules, bounded-height measurements, and scroll clamping after document shrink. Fresh full Release browser-client suites at 434732b82 passed on .NET 8 (2,902 passed, 11 existing skips) and .NET 10 (2,905 passed, 11 existing skips), including all three Scalar workflows on each framework. All eight CI checks passed for the same commit: Linux, ARM, Windows, macOS, browser clients, host-contract verification, Native AOT, and the packaged generator.


lahma:
AngleSharp.Css 1.1.1-beta.308 — the preview carrying AngleSharp/AngleSharp.Css#244 — is now the pinned package on `main` (#3980, `b31a05fd1`), with a regression suite for the individual-transform re-entry (`Jint.Tests.Browser/Views/IndividualTransformStyleTests.cs`). So the dependency this draft was waiting on is in; a rebase should be able to drop the diagnostic-package language. I have not touched this branch.

lahma:
Follow-up to the pin note above: `main` now carries **AngleSharp.Css 1.1.1-beta.310** (#4016, `1da270dc1`), which is beta.308 plus AngleSharp.Css `8c7ec8b9` and `534c18ec` (the transform-function null guard and the `matrix()` column order). The bump surfaced one defect this browser could reach: every 2D `translate` spelling (`translate(10px)`, `translateX`, `translateY`, `translateZ`, `translate(10px, 20px)`) computed to `""` on beta.308 because `CssTranslateValue.Compute` dereferenced the axes the function did not name and `CssCascade` swallowed the `NullReferenceException`; `Views/TransformFunctionStyleTests` pins it now. The `ComputeMatrix` half is unreachable here (no transform stage in the flat box model). Still not touching this branch.

Issue #3698, opened 2026-09-02
Seams the DOM domain wished it had: protocol citations in Jint.Browser, a parsed-document observer call, symmetric extra-domain observation, layout invalidation


lahma:
Four small seams the `DOM` domain work (#3697) wished it had. None blocks anything; each is a workaround today.

1. **`ProtocolCitationTests` only scans `Jint.DevTools`.** `Jint.Browser/DevTools/*` now carries ~20 CDP citations (`https://chromedevtools.github.io/devtools-protocol/tot/<Domain>/#method-<name>`) that nothing verifies against the vendored protocol JSON. The check should cover every project that owns a domain.
2. **`IPageObserver` has no "the document is parsed" call that carries the runtime.** `DocumentCreated` fires before the parse, so the DOM domain's mutation bridge remembers the runtime there and arms itself on `Phase(Committed)`. A `DocumentParsed(PageRuntime, loaderId)` (or the runtime on `Phase`) removes the two-step dance.
3. **A target's extra domains are observed asymmetrically.** `TargetDomains.Detach` unobserves any `ITargetObserver` among `Extra` automatically, but registration does not observe them, so `PageTarget.RegisterDomains` calls `Observe` by hand. Register and unregister should be symmetric.
4. **Nothing can invalidate a layout cache cheaply.** The flat layout is recomputed per query (linear in the document) because the only mutation signal is a whole-document `MutationObserver`, which would make every page pay for records nothing reads. A document mutation counter is an AngleSharp-side seam; listed here per the campaign's rule, not filed upstream. `MouseEventState` likewise carries no `pageX`/`offsetX`, so those are computed on read (observable only by scrolling inside a listener).

Recorded rather than left as TODOs, per the campaign's rule.

## Completion plan (2026-09-13)

- [x] Verify merged foundations: #3762 (citations), #3763 (parsed document), #3777 (domain observation).
- [x] Verify merged coordinate fixes #3818, #3908 and #3927, and query-local geometry improvements #3888.
- [x] Audit supported invalidation hooks in the pinned AngleSharp 1.8.1 / AngleSharp.Css 1.1.2 dependencies; document precise gaps and a local integration design.
- [x] Define synchronous invalidation for DOM, CSSOM/stylesheets, viewport/media, selector-affecting control state, and document replacement.
[...]

PR #3908, opened 2026-09-07
Preserve mouse offsets when a listener scrolls before reading them

sebastienros:
A mouse listener that scrolls before its first `offsetY` read gets a coordinate based on the new viewport, even though the event's page position was captured when dispatch began. Compute the offset from that captured document position and the target's existing document-space box. This preserves the initial scroll position without another layout query or a cache.

Add regression cases for initially unscrolled and already-scrolled pages. Both failed on main before the fix; the Release browser event suite passes all 118 cases on net8.0 and all 118 on net10.0 (SDK 10.0.400). `git diff --check` passed.

Refs #3698. This addresses the independent scroll-before-first-read remainder after #3818. Target mutation before the first offset read and cheap layout invalidation remain open. No Scalar draft #3888 files or geometry/cache algorithms are changed.


sebastienros:
Updated to current main at exact head `8284360ea` after the rerun exposed two unrelated Windows conditions:

- The authoritative corpus count from #3901 is `dom/nodes` tests 4,802 → 4,801 and total 27,107 → 27,106; this head carries that exact correction.
- `NodeList-static-length-getter-tampered-1.html` timed out once under runner load. The same isolated case previously passed on rerun in #3907, and the implementation here does not touch NodeList or that harness.

Fresh Release validation after the update: `MouseEventCoordinateTests` plus `WptBrowserCensusTests` passed 21 with the one expected Windows-only census skip on each of .NET 8 and .NET 10. The new current-head workflow will determine merge readiness.

sebastienros:
The branch now restores the browser census to 4,802 `dom/nodes` registrations and 27,107 total. Fresh complete Windows runs on #3921 and #3928 registered 4,802 on both target frameworks. The earlier 4,801 observation came from a timed-out one-test document; the census path had counted that failed document as zero and made the incomplete run look like an exact corpus change.

The current head also merges #3932, so this rerun uses the reduced PR workflow without the native browser-tool packaging matrix. A separate fail-closed census fix is being prepared; this PR does not change timeouts, exclusions, or corpus behavior.

Issue #3898, opened 2026-09-07
Add isolated DOM binding comparison harness for D0.3


sebastienros:
D0.3 in #3575 remains unproven: the repository has generic host-prototype benchmarks, but no reproducible comparison of AngleSharp.Js reflection bindings with a hand-shaped Node/Element/Document prototype over the same AngleSharp DOM.

Implement isolated, explicitly pinned baseline and main-source candidate executables, shared offline workloads with independently expected correctness checksums, and cold setup versus warm prepared-script lanes. Record dependency/runtime/workload provenance and reuse the repository's BenchmarkDotNet measurement controls. A verification mode must exercise every lane without publishing timings; document paired idle-machine measurements and the limits of comparing different interpreter versions and binding surfaces.

Acceptance: Release builds and checksum verification pass, the measurement lanes can be enumerated, and the harness/documentation land. Gate-quality timing evidence remains a separate campaign step on an idle machine; this issue does not assert a performance win.

PR #3921, opened 2026-09-07
Fix wide BenchmarkDotNet CSV parsing in binding comparison

### LINKED ISSUE #3920: Fix binding comparison parsing of wide BenchmarkDotNet CSV reports
The first accepted D0.3 gate run from main 7f5bbcb921e7c00cab754fe50d59ef245d7d90a5 completed all eight baseline benchmark rows, then `BindingComparison/run.py` stopped with `Could not determine delimiter` before the candidate arm.

The real BDN CSV has a 661-character header and eight roughly 430-character rows. `csv.Sniffer().sniff(stream.read(4096))` receives a truncated last row, so the otherwise valid comma-delimited report appears inconsistent. The attempt produced no complete paired evidence and supports no binding performance conclusion.

Fix delimiter detection from a complete record, preserve the actual wide report structure in a regression fixture using explicitly synthetic values, cover comma/semicolon delimiters and failure handling, and rerun isolated Release verification. Resume six-round measurements from a fresh output directory after the fix lands and the idle check accepts the machine.

Follow-up to #3898; part of #3575 D0.3.

sebastienros:
A complete BenchmarkDotNet CSV report failed collection because the fixed 4,096-character delimiter sample ended in the middle of a wide row. Detect the delimiter from the complete header, preserving comma/semicolon support, and cover the actual 53-column/eight-row report structure with explicitly synthetic measurements.

Also regenerate the browser census from the authoritative Windows result after the merged replaceChildren change: dom/nodes registrations 4,802 → 4,801, total 27,107 → 27,106. The repository renderer/writer produced the correction; outcome ceilings are unchanged.

Validation: four Python parser/checksum tests pass; fresh Release verification passes all 24 checksums in each isolated benchmark arm; Release census checks pass 14 tests on each of net8.0/net10.0, with the measured Windows-only check skipped locally. No timing claim is made.

Fixes #3920. Follows #3898 and campaign #3575 D0.3.


sebastienros:
The latest authoritative Windows run registered 4,802 `dom/nodes` tests on both target frameworks and failed only because this branch had changed the recorded count to 4,801. That earlier 4,801 observation coincided with the known intermittent `NodeList-static-length-getter-tampered-1.html` timeout, so it was not a complete census.

I restored the table to 4,802 / 27,107 at exact head `45fe5fd33`. The four binding-report parser tests remain green locally. The new workflow should now validate the complete Windows census without carrying a result derived from an incomplete WPT run.

Issue #3772, opened 2026-09-03
Five smaller DOM defects the browser lane's new suites found: hasFeature, collections, XML names, CharacterData offsets, and three constructors


lahma:
## Remaining work is owned by Jint

The maintainer has directed that required browser semantics be implemented in Jint when AngleSharp leaves them outside its scope. Do not hold this parent issue or its remaining children for upstream acceptance or a package release. #3949 tracks the three residual null-namespace adoption assertions after #4045's casing fixes; #3950 tracks modern DOM name acceptance and creation. Implement and validate those local corrections over the existing DOM, retaining the divergence record and exact WPT coverage.

## Earlier report

Five bounded findings from vendoring `dom/nodes/`, `dom/collections/`, `dom/lists/`, `dom/traversal/`, `dom/ranges/` and `html/dom/` into the web-platform-tests browser lane. Each is small enough to fix on its own and each is named test by test in `Jint.Tests.Browser/Wpt/WptBrowserExclusions.cs`; the bigger causes are separate issues.

### 1. `DOMImplementation.hasFeature` is not unconditionally true — 133 rows

[DOM §4.5](https://dom.spec.whatwg.org/#dom-domimplementation-hasfeature) is one line: "return true". It is a
legacy member kept so that old feature-detection code takes the modern path. AngleSharp answers `true` for
three of the 136 pairs `dom/nodes/DOMImplementation-hasFeature.html` asks about and `false` for the rest, so a
page that gates on `hasFeature('Core', '3.0')` takes its fallback.

```js
document.implementation.hasFeature();                 // false; a browser: true
document.implementation.hasFeature('Core', '3.0');    // false; a browser: true
document.implementation.hasFeature('nonsense');       // false; a browser: true
```

The arity is wrong too: the standard's arguments are optional and calling it with none is
`1 argument required, but only 0 present.`

### 2. A collection's named and indexed properties — 113 rows, 34 documents

* The empty string is a supported property name. `dom/collections/HTMLCollection-empty-name.html` is 7/7 on
  this: [DOM §4.2.10.2](https://dom.spec.whatwg.org/#interface-htmlcollection)'s supported property names skip
  an element whose `id` or `name` is the empty string, and `''` in `collection` answers true here.
* `getElementsByTagName` matches where the standard matches nothing — 23 rows of
  `assert_array_equals: lengths differ, expected array [] length 0, got object … length 1`, mostly the
[...]

PR #3928, opened 2026-09-07
Preserve HTMLCollection expandos when matching elements appear

### LINKED ISSUE #3925: Preserve HTMLCollection expandos when matching elements appear later
Residual of #3772. An ordinary own property on an HTMLCollection must remain visible when an element with the same id/name appears later; the explicit `namedItem()` operation must still return that element.

```js
const collection = document.getElementsByTagName('span');
collection.later = 'own';
const element = document.createElement('span');
element.id = 'later';
document.body.appendChild(element);
collection.later; // currently element; expected 'own'
collection.namedItem('later'); // expected element
```

Fresh Release WPT triage on main `7f5bbcb92`, both .NET 8 and .NET 10:

* `HTMLCollection-own-props.html`: 6 PASS / 2 FAIL; both “Setting non-array index while named property doesn't exist” modes fail.
* `HTMLCollection-supported-property-names.html`: 4 PASS / 2 FAIL; both later-added-name cases fail, including a non-configurable expando.

Separate the explicit named-item lookup from the property visibility check, applying ordinary-own-property precedence consistently to reads and visible-name enumeration. Cover deleting a configurable expando, preserving a non-configurable one, descriptors/enumeration, and live named lookup.

The HTMLCollection-as-prototype assignment case and tag-name namespace/case cases remain separate #3772 residuals.

Spec: https://webidl.spec.whatwg.org/#dfn-named-property-visibility and https://dom.spec.whatwg.org/#dom-htmlcollection-nameditem.

sebastienros:
Fixes #3925. Resolves the expando/later-name portion of #3772; the prototype-assignment and namespace cases remain separate.

When a page writes `collection.later = 'own'` and then inserts an element with `id='later'`, the collection now retains its own value while `collection.namedItem('later')` returns the live element. Deleting a configurable expando reveals the element; a non-configurable expando remains intact.

The HTMLCollection wrapper separates explicit named-item lookup from ordinary property visibility and excludes stored own properties from the projected-name list, leaving their descriptors and enumeration to the ordinary property bag. Indexed access and the existing live collection remain unchanged.

Fresh Release validation:

* All 11 new regressions fail on the original code and pass with host-contract verification on .NET 8 and .NET 10. Coverage includes loose/strict assignment, id/name matching, two live collection sources, configurable/non-configurable own properties, descriptors, enumeration, and accessor side effects.
* Full Browser suite: 2,132 pass on .NET 10 and 2,128 on .NET 8; 37 expected skips each.
* Four WPT failures now pass: both assignment modes in `HTMLCollection-own-props.html` and both later-name cases in `HTMLCollection-supported-property-names.html`. Their three exact exclusion entries are removed.

Both Windows census legs in [run 34170315094](https://github.com/sebastienros/jint/actions/runs/34170315094/job/101893219326) measured only the expected four-outcome improvement: `dom/collections` not passing 5 → 1. After merging main `052b45aa9`, the repository census renderer/writer applies that recorded row and recalculates the total from 1,356 to 1,352. The shared post-#3901 corpus correction measured in [Windows run 34168450799](https://github.com/sebastienros/jint/actions/runs/34168450799/job/101894314739) is also included: `dom/nodes` tests 4,802 → 4,801, and the test total 27,107 → 27,106. This leaves the four-outcome collections improvement intact.


sebastienros:
The branch now restores the browser census to 4,802 `dom/nodes` registrations and 27,107 total. Fresh complete Windows runs on #3921 and #3928 registered 4,802 on both target frameworks. The earlier 4,801 observation came from a timed-out one-test document; the census path had counted that failed document as zero and made the incomplete run look like an exact corpus change.

The current head also merges #3932, so this rerun uses the reduced PR workflow without the native browser-tool packaging matrix. A separate fail-closed census fix is being prepared; this PR does not change timeouts, exclusions, or corpus behavior.

Issue #3947, opened 2026-09-08
Browser WPT synchronous documents can hit harness timeout under test-suite load


sebastienros:
## Status audit — 2026-09-18

The Windows isolation proposal in the reproduction discussion **was implemented and merged in [#3999](https://github.com/sebastienros/jint/pull/3999)**. `.github/workflows/pr.yml` now runs all other test projects first and Jint.Tests.Browser in its own subsequent Windows step. The unchanged harness deadlines, vendored files, exclusions and census remain in force. [#4002](https://github.com/sebastienros/jint/pull/4002) and [#4037](https://github.com/sebastienros/jint/pull/4037) also reduced host-collection work; they do not establish that every timeout is eliminated.

The original issue is **still active in narrower scope**: the unexplained short Range-document occurrences, including macOS, were not reproduced by the Windows NodeList throughput investigation. The isolation change applies only to the Windows PR leg; it is not a demonstrated all-platform fix. The next useful investigation should begin from current main and capture phase/progress evidence for an actual remaining occurrence, accounting for the already merged scheduling and throughput changes. Do not retry away failures, widen deadlines, or redo the rejected target-framework-only serialization experiment (#3973).

`dom/ranges/Range-attribute-nodes.html` has now produced the browser lane's upstream-harness `TIMEOUT` on unrelated code and in repeatable loaded local runs, while the exact document completes in well under one second when isolated.

## CI evidence

PR #3945 is an unrelated form-submission change. At exact head `aaa0afc6fc27bece0a23c29c54e662801677d466`, its Linux job ran the ordinary solution-wide Release command:

```text
dotnet test --configuration Release
```

[Run 34182477661, job 101924175767](https://github.com/sebastienros/jint/actions/runs/34182477661/job/101924175767) failed in `Jint.Tests.Browser` net8.0 with:

```text
Expected outcome.HarnessError to be <null> because
dom/ranges/Range-attribute-nodes.html must run to completion,
but found "the harness reported TIMEOUT: ".
```

The net10.0 browser suite in the same job passed. The failing net8.0 assembly ran for 3m23s beside the solution's other test projects; net10.0 ran for 2m51s. The job's other test projects were also executing concurrently.

This document and the browser WPT harness are outside #3945's diff. The document contains only synchronous `test(...)` cases. It has no async test, explicit timer, or `timeout=long` metadata.

## Local confirmation

[...]

PR #3965, opened 2026-09-08
Make Web API installation realm-bound

### LINKED ISSUE #3964: Make Web API installation and global event state realm-bound
Follow-up prerequisite for #3771. The realm-construction and bounded host-callback work is complete (#3922, #3955), lazy Web API descriptors now capture their destination realm (#3919), and document/page state is receiver-gated (#3957, #3960, #3961). The remaining Web API registration and global-event state still assumes the engine's main realm.

Exposing a child-frame global before removing those assumptions would produce a Window whose inherited event methods can reject it or dispatch through the parent listener list.

## Scope

- Add a reviewed door for installing the expanded Web API feature set into a fully initialized, same-engine `Realm`. This may begin as an internal seam if the public `Engine.WebApi.InstallInRealm(Realm)` shape is not ready to freeze.
- Parameterize global and fetch-model registration by destination realm.
- Store `GlobalEventTarget` and `InstalledSelf` per installed realm.
- Route global event functions through their captured realm.
- Remove main-realm assumptions from `EventTargetPrototype` branding, `window.event`, global `reportError`, and callback-error dispatch.
- Define live-enable behavior for every installed realm so later `WebApi.Enable(...)` calls cannot silently leave secondary realms with a stale feature set.
- Keep timers, queues, network sockets, iframe script execution, and browser DOM realm association out of this change.

## Acceptance criteria

- Reject null, incomplete, or foreign realms before mutating their globals; claim engine ownership; repeated installation is a no-op.
- Install the engine's current expanded feature set lazily and without clobbering an existing property or evaluating a host lazy override.
- Realm A and B receive their own constructors/prototypes and each observes `self === globalThis`; reading a lazy global first in either realm cannot contaminate the other.
- Global listener lists are isolated. `EventTarget.prototype.addEventListener.call(B.global, ...)` succeeds only after B is installed, and dispatch uses B.global for `event.target`, `event.currentTarget`, and `this`.
- `window.event`, callback exception reporting, and `reportError` use the destination realm's global event target. Ordinary objects and uninstalled ShadowRealm globals still fail branding.
- Worker `self` replacement remains principal-worker-only and non-clobbering.
- Restore/dispose clears state for every installed realm without retaining dead realms.
- Principal-realm registration, live enable, workers, diagnostics, public API baselines, and host-contract verification remain green.

This is the smallest safe substrate before #3771 can replace its shared-realm frame window with a real child-realm global. It deliberately does not enable iframe scripts or claim the frame-realm WPT rows.

sebastienros:
Installing Jint's Web API surface in an additional realm previously left constructors and lazy globals tied to the principal realm, while global listener state was shared with the principal global. This adds an internal, engine-owned installation door that installs the current expanded feature set into a fully initialized same-engine realm and gives each installed realm its own lazy interfaces, `self`, synthetic global event target, `window.event` slot, and error-reporting destination.

Secondary realms are tracked with weak keys and weak enumeration entries, so installation does not extend their lifetime. Live `WebApi.Enable(...)` and fetch-model registration update every installed realm; restore drops every realm's transient listeners; disposal clears the registry. Principal-only worker `self` replacement remains unchanged.

Deferred work also retains its relevant realm: promise rejection notifications use the promise's creation realm, timers use their scheduling realm, and each idle callback uses its registration realm for both `IdleDeadline` branding and callback error reporting. This keeps secondary-realm failures and callback objects on the secondary global even when the engine-owned queue is pumped from the principal realm.

Validation:

- full `dotnet build -c Release`: zero warnings and errors
- realm installation suite: 14 passed on net8.0 and 14 passed on net10.0
- full `Jint.Tests`: 12,403 passed and 6 expected skips on each target
- full `Jint.Tests.PublicInterface` with `JINT_HOST_CONTRACT_VERIFICATION=1`: 3,707 passed on net10.0 and 3,697 passed on net8.0, with 8 expected skips on each target
- affected browser observer/frame scheduling suites: 37 passed on each target

Closes #3964


sebastienros:
Admin-merge evidence: eight checks passed. The sole Windows failure is the exact #3947 browser-census flake: dom/nodes/NodeList-static-length-getter-tampered-1.html timed out under net8.0, while net10 Jint.Tests.Browser passed 2,280 tests and Test262 passed 102,585 tests. The independently reviewed head remains 67f2f02ddb44f938f50dd640fe1a44622377eb93.

PR #3970, opened 2026-09-08
Harden Test262 corpus acquisition

### LINKED ISSUE #3967: Test262 setup amplifies transient DNS failures across every generated test
The macOS job for PR #3966 failed before Test262 execution because its one-time setup could not resolve `github.com:443`:

```
System.Net.Http.HttpRequestException: nodename nor servname provided (github.com:443)
```

The setup failure was then reported against 102,692 generated tests. Other suites in the same job completed successfully, including the full browser suite, so this was unrelated to the PR and consumed a complete platform run without producing conformance evidence.

Run: https://github.com/sebastienros/jint/actions/runs/34211117701/job/102012095209

## Desired behavior

- A transient DNS/download failure should be retried with a bounded policy or avoided through the existing Test262 cache/artifact path.
- If setup still cannot obtain the pinned corpus, report one clear infrastructure/setup failure rather than one failure per generated test.
- Preserve the pinned SHA and fail closed; do not silently run a stale or partial corpus.
- Keep local/offline behavior explicit and deterministic.

## Acceptance criteria

- A test seam reproduces a transient fetch/DNS failure without contacting the network.
- A transient first failure followed by success completes setup and runs the pinned corpus.
- A permanent failure produces a concise setup error and does not expand into tens of thousands of per-test failures.
- The successful cached and fresh-download paths retain their existing SHA verification.

Focused issue searches found no existing Test262 DNS or one-time-setup failure tracker in this repository.

sebastienros:
The generated Test262 namespace setup let a transient corpus download exception escape from `OneTimeSetUp`, so NUnit copied the same infrastructure error onto all 102,692 generated cases. The runner now retries transient acquisition failures up to three times while continuing to request the generated suite's exact SHA.

Downloads go to a unique per-process staging directory. The runner computes a canonical digest over every `harness/` and `test/` path and file body, then atomically publishes the verified archive to the shared cache. Concurrent acquisitions can only accept another process's archive after independently verifying the same pinned digest. Missing, partial, corrupt, or mismatched content remains a failure, and `JINT_TEST262_OFFLINE=1` provides a deterministic cache-only path.

If acquisition still fails, the first selected generated case reports the infrastructure error and the remaining selected cases skip. This keeps full and generated-only filtered runs red without producing tens of thousands of duplicate failures.

Validation:

- Full Release Test262 run: 102,594 passed, 107 existing exclusions skipped, 0 failed (102,701 total including 9 new infrastructure tests), 2m28s.
- Fresh empty-cache acquisition through staging, digest verification, and atomic publication: passed; only the final 54.8 MB archive remained.
- Forced generated-only offline missing-cache probe: exactly one failure and one selected generated case skipped.
- Two independent acquisitions targeting one cache publish safely; the resulting archive passes an offline digest check.
- Test262 infrastructure tests: 9 passed in Release.
- Agent instruction guardian: 5 passed on net8.0 and 5 passed on net10.0 in Release.

Closes #3967

### DIFF
--- a/Jint.Tests.Test262/AGENTS.md
+++ b/Jint.Tests.Test262/AGENTS.md
@@ -13,7 +13,13 @@ Test sources live in `..\test262\test`, which you may always read, and the harne
 
 ### Updating the test262 suite
 
-Bump `SuiteGitSha` in `Jint.Tests.Test262/Test262Harness.settings.json` to the new upstream commit. The next build notices the settings-file hash changed, wipes `Generated/` and regenerates it (`dotnet tool restore && dotnet test262 generate`); `Generated/` is gitignored, so the committed diff stays one line.
+Bump `SuiteGitSha` in `Jint.Tests.Test262/Test262Harness.settings.json` to the new upstream commit. The next build notices the settings-file hash changed, wipes `Generated/` and regenerates it (`dotnet tool restore && dotnet test262 generate`); `Generated/` is gitignored.
+
+The runtime half of that pin is `CorpusContentSha256` in `State.cs`: a canonical digest of every path and
+per-file digest under `harness/` and `test/`. Update it with the reviewed corpus whenever `SuiteGitSha` changes.
+Leaving the old value fails closed and reports the downloaded tree's actual digest; never bypass that failure
+or accept a directory-only archive, because generated source proves which tests exist but not which corpus
+content the runner opened.
 
 A bump is a **code** change, not just a pin change. An upstream normative change can turn a feature that passes today red tomorrow, so read the commits in the range rather than only the new test files — `git log --oneline <old>..<new>` in a test262 checkout, plus the `features.txt` diff for newly added features.
 
--- a/Jint.Tests.Test262/README.md
+++ b/Jint.Tests.Test262/README.md
@@ -5,6 +5,13 @@ dotnet tool restore
 dotnet test262 generate
 ```
 
+The test run downloads the archive for the generated suite's exact `SuiteGitSha` into a unique
+staging directory, verifies the pinned corpus content digest, and atomically publishes it to the
+cache. Transient acquisition failures are retried up to three times. Set `JINT_TEST262_CACHE` to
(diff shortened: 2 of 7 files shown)


sebastienros:
The first corrected-head Windows run exposed a real test-lifetime defect rather than a product failure: three success-path acquisition tests deleted their temporary cache directory while the returned `Test262Stream.Options.FileSystem` still owned the zip handle. Unix permits unlinking the open archive; Windows correctly reported `IOException` for `test262-0123456789abcdef.zip`.

Head `657f5edf0` now deterministically disposes every archive-backed test file system before deleting its cache directory. `Zio.IFileSystem`'s public API implements `IDisposable`; no acquisition behavior changed. The focused Release acquisition fixture passes 9/9, and `git diff --check` is clean. The pushed head restarts CI; the failed Windows run is https://github.com/sebastienros/jint/actions/runs/34219253427/job/102038235467.

sebastienros:
The remaining Windows failure was an archive lifetime conflict in concurrent acquisition. Publication now keeps the first validated archive: later publishers validate and reuse it instead of replacing a file backing an active reader. Invalid online cache entries are rechecked before replacement, and the concurrency regression verifies that the first publisher remains in place. Test cleanup joins both acquisitions and disposes all successful results even when one acquisition fails.

Validation: all 10 focused Test262 infrastructure tests pass in Release on .NET SDK 10.0.400, with the repository package versions unchanged. Independent review completed; Windows CI must confirm this fix before merge. Auto-merge remains disabled while that result is pending.

sebastienros:
Validated head `ee49a07d63f665c1c94c7b427a792cd8131ab0a1` against run 34249983707. Windows Test262 now passes 102,595 tests, with 107 skips and no failures; this includes all ten acquisition infrastructure regressions. Seven other CI jobs are green.

The sole failing job is Windows Browser net8.0: the known #3947 `dom/nodes/NodeList-static-length-getter-tampered-1.html` harness timeout and its two dependent census failures. Browser net10.0 passes. No changed Test262 code or regression test failed. Merging under the maintainer's authorization to bypass confirmed unrelated flaky failures. #3974 adds diagnostics for the remaining WPT timeout.

PR #3972, opened 2026-09-08
Align :any-link with link history states

### LINKED ISSUE #3969: Browser: align :any-link with :link and :visited hyperlink states
Follow-up found while independently reviewing #3968.

The page selector factory now gives `:link` / `:visited` the intended bounded HTML behavior, but `:any-link` still delegates to AngleSharp and disagrees with those states:

```js
const anchor = document.createElement('a');
anchor.setAttribute('href', '');
anchor.matches(':link');      // true
anchor.matches(':any-link');  // false

const link = document.createElement('link');
link.setAttribute('href', '/style.css');
link.matches(':link');        // false
link.matches(':any-link');    // true
```

A customized built-in HTML anchor follows the corrected `:link` path but still misses `:any-link`. An SVG `<a href="">` is also currently unmatched.

Selectors defines `:any-link` as the union of `:link` and `:visited`. Jint has no visited-history model and intentionally treats every qualifying hyperlink as unvisited, so those three selectors must remain mutually consistent even when `:visited` matches nothing.

## Acceptance criteria

- `:any-link` matches exactly the elements that match either `:link` or `:visited` under Jint's history model.
- HTML anchors and areas with a present `href`, including the empty string and customized built-ins, have consistent `:link` / `:any-link` results.
- Missing `href`, ordinary elements, and HTML `<link href>` retain their living-spec result after the behavior is verified against the current HTML definition.
- SVG hyperlink elements are covered explicitly, including empty `href` and namespace behavior.
- `matches`, prefixed `webkitMatchesSelector`, `querySelector`, and `querySelectorAll` agree.
- Add focused tests and update any affected WPT exclusions/census from measured results.

Focused issue searches found no existing `:any-link` tracker in this repository. This is separate from #3946's fixed 18-row `:link` / `:visited` slice and should not delay #3968.

sebastienros:
Selectors defines `:any-link` as exactly `:is(:link, :visited)`, but AngleSharp's default left it inconsistent with Jint's corrected link-history states: empty-`href` HTML hyperlinks missed `:any-link`, `<link href>` matched it, and SVG hyperlinks were not recognized.

Build `:any-link` from Jint's `:link` and `:visited` selectors, and extend the same no-history partition to SVG `a` elements carrying either unnamespaced `href` or namespaced `xlink:href`. SVG's nested-link rule is applied with a direct parent walk, including for detached subtrees, without allocating traversal state. Focused tests cover HTML anchors and areas, customized built-ins, HTML `link`, SVG element and attribute namespaces, empty and missing attributes, dynamic attribute changes, nested SVG links under HTML and SVG hyperlinks, detached clones, `matches`, `webkitMatchesSelector`, `querySelector`, and `querySelectorAll`. The AngleSharp divergence register now describes the full selector family.

The three existing Selectors API WPT documents still pass against the current exclusion table, so their census and cause counts do not change.

Closes #3969

Validation:

- `dotnet test Jint.Tests.Browser/Jint.Tests.Browser.csproj -c Release --filter 'FullyQualifiedName~Jint.Tests.Browser.Parsing.PagePseudoClassSelectorTests'` — 6/6 on both net8.0 and net10.0
- `dotnet test Jint.Tests.Browser/Jint.Tests.Browser.csproj -c Release --filter 'Name~Element-matches.html|Name~Element-webkitMatchesSelector.html|Name~ParentNode-querySelector-All.html'` — 3/3 on both net8.0 and net10.0


sebastienros:
Windows failed only on the documented #3947 suite-load flake: `dom/nodes/NodeList-static-length-getter-tampered-1.html` reported the upstream harness `TIMEOUT` in net8 and therefore made the fail-closed census report three failures. The same job's net10 browser suite passed 2,286 tests, all non-browser suites passed, and the other seven required jobs passed. This PR changes `:any-link` matching and its focused tests; it does not touch NodeList, the timed-out document, the WPT harness, timers, or census machinery.

The exact independently reviewed head `5bf0e873602a0a4655935409cfd6cd2b7bf117ec` is safe to admin-merge under the maintainer's confirmed-flake policy. Windows job: https://github.com/sebastienros/jint/actions/runs/34218278790/job/102035130539

PR #3975, opened 2026-09-08
Recognize dialog in input formMethod reflection

sebastienros:
Input `formMethod = 'DiAlOg'` currently reads back as `get`. HTML's form submission attributes define `dialog` for both input and button submitters; the reflection table incorrectly gave only the button that keyword. Add it to the input row and regenerate the binding, preserving the original attribute spelling and existing missing/invalid defaults.

The earlier week/month implementation was superseded by #3983, #3986 and subsequent reflection fixes. This PR now contains only the remaining keyword correction and regression tests for both interfaces.

Validation: freshly compiled Release .NET 8 and .NET 10 reflection, generated-binding currency and the unchanged week/month WPT document: 76 passed per framework. All eight CI checks pass at `e5a745084`, including Windows and both browser-client frameworks. No WPT changes, exclusions or census changes.

Specification: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-fs-formmethod


lahma:
Checked this against current `main` (`0a6d10255`) before continuing #3770's tail, so the residue is on the record:

- **Already on main** via the reflection chain (#3983, #3986): the vendored `reflection-forms-weekmonth.html` + `elements-forms-weekmonth.js`, its exact-count entry (1,579, nothing excluded), all eleven `HTMLInputElement` `reflected` rows (`formAction` under the name `documentUrlWhenEmpty`, also applied to `button.formAction` and `form.action`), and both integer-range rules (`InRange` at the unsigned setter and in `SetLimited`, plus the saturating parser).
- **Salvaged into new PRs**: the `setterOnly` generator flag is #4020 (its first application is `<meter>`'s six setters, which retires the 18-row "double written with .NET's number format" cause), and the `documentUrlWhenEmpty` read is #4021 (it answered AngleSharp's document address, so after `history.pushState` `formAction` disagreed with `document.URL`; it now reads `PageRuntime.DocumentUrl`), together with this PR's two tests (`AFormActionUrlDefaultsToTheOwningDocumentUrl`, `AnUnsignedSetterWritesItsDefaultAboveTheReflectedRange`).
- **Deliberately not carried**: `dialog` on `input.formMethod` — the vendored table (`elements-forms.js`, and the weekmonth copy) gives `input.formMethod` `["get","post"]`; only `button.formMethod` takes `dialog`, and main matches. Also not carried: `setterOnly` on `input.width`/`height` — the projected getter (`DisplayWidth`) answers −5 for `width="-5"` and 0 for `3.5`, so restoring it would let an `unsigned long` go negative; #4020 fixes the README sentence that claimed otherwise instead.

So this PR can be closed once #4020/#4021 land; nothing in it is lost.

sebastienros:
Superseded on main by #3983 and #3986, with subsequent reflection and document-URL corrections in #4011, #4020 and #4021. The week/month WPT document is already vendored with all 1,579 assertions enforced; the unsigned setter fallback, formAction document URL, and setter-only dimension reflection are implemented there. Closing the obsolete branch rather than reintroducing its earlier reflection model.

PR #3985, opened 2026-09-08
Fix quirks-mode class-name matching in browser bindings

### LINKED ISSUE #3899: getElementsByClassName does not fold ASCII case in quirks mode
**Ownership decision (2026-09-08): implement this behavior in Jint.Browser.** AngleSharp [#1319](https://github.com/AngleSharp/AngleSharp/issues/1319) and [PR #1321](https://github.com/AngleSharp/AngleSharp/pull/1321) were closed by the maintainer because extending runtime quirks-mode behavior is outside the intended scope. The user has explicitly authorized Jint to own this behavior. This supersedes the earlier dependency-blocked status and the previous comment advising against a local implementation; no AngleSharp release is required.

The remaining class-name WPT failure on main `3744d5832` is `dom/nodes/getElementsByClassName-14.htm`, subtest `document.getElementsByClassName(): case-insensitive (quirks mode)`.

With no doctype (quirks mode), the document has `<html class="a A">` and `<body class="a a">`. `document.getElementsByClassName("A a")` must match both elements using ASCII case-insensitive matching; it currently excludes body. The neighboring Unicode case-sensitivity subtest passes and must stay passing.

DOM's [class-name collection algorithm](https://dom.spec.whatwg.org/#concept-getelementsbyclassname) requires ASCII case-insensitive matching in quirks mode and identical matching otherwise. The generated bindings delegate class matching to AngleSharp. #3897 fixes collection liveness by repeating that query, but preserves this separate matching defect.

A focused fix should retain live collection behavior and cover ASCII case folding in quirks mode, case sensitivity in standards mode, and Unicode characters such as U+212A KELVIN SIGN remaining distinct from ASCII K/k. Remove only the existing `getElementsByClassName-14.htm` exclusion after its two subtests pass.

sebastienros:
## Summary

Apply quirks-mode class-name matching in Jint.Browser.

## Details

In a quirks document, `getElementsByClassName('A a')` now matches both `<html class="a A">` and `<body class="a a">`. Both Document and Element use one query over AngleSharp's tree and class tokens, preserving live HTMLCollection reads. Queries split on ASCII whitespace once; comparisons fold only ASCII in quirks mode and remain exact in standards, limited quirks and XML documents. Each live read uses the root's current node document, including after adoption.

AngleSharp [#1319](https://github.com/AngleSharp/AngleSharp/issues/1319) and [PR #1321](https://github.com/AngleSharp/AngleSharp/pull/1321) were closed as outside the intended runtime quirks scope. The project explicitly chose to own this behavior in Jint. The divergence record and binding guidance now record that decision.

Remove only the `getElementsByClassName-14.htm` exclusion, retaining its two-subtest minimum. Generated Document/Element hooks already share the implementation and regenerate without code changes. CSS selector quirks matching is outside this change.

## Linked issue

Fixes #3899

## Test plan

- Final focused regression tests against the unchanged hook: four expected quirks failures, eight passes.
- Release .NET 8 and .NET 10: 394 passing checks per target across DOM binding tests, all 35 class-name WPT documents, generated-binding currency/pins, and WPT table structure. The two opt-in measured census/cause checks are skipped locally.
- All 12 focused tests pass, covering all three HTML modes, XML, Unicode, token boundaries, live mutations, root exclusion, named lookup, conversion once and adoption.
- Windows CI must supply the measured census/cause changes before this draft is ready. Existing numeric tables remain unchanged pending that evidence.

## Breaking change?

No public API change. Quirks-mode class queries intentionally gain ASCII case-insensitive matches.


sebastienros:
Superseded by #3992, which implements the same Jint-owned ASCII quirks matching and incorporates the mode, Unicode, live collection, conversion and adoption regressions. Main also removes the affected WPT exclusion and reconciles its census (#4024). Closing this duplicate rather than merging an older competing implementation.

PR #3995, opened 2026-09-08
Two override lists the binding needed: a merged interface and @@unscopables

lahma:
**Stacked on [#3991](https://github.com/sebastienros/jint/pull/3991)** — the diff to review is the second
commit, `Two override lists the binding needed`. Merge that one first.

## What and why

The rest of the scripting/events tail of the browser wpt census: `html/webappapis/scripting/events/`,
`dom/lists/`, `dom/ranges/` and `dom/events/`. Twelve more rows retired, two new `overrides.json` lists, and
three exclusion reasons that were measurably wrong.

### 1. A DOM prototype carries no `@@unscopables`

[WebIDL §3.7.5](https://webidl.spec.whatwg.org/#es-unscopable) puts an unscopable object on the interface
prototype object of every interface with an `[Unscopable]` member, and this binding emitted none: an
extended attribute is not something AngleSharp's metadata can carry.

**Mechanism.** `overrides.json` gains an `unscopables` list — the eight members DOM marks (`ParentNode`'s
`append`, `prepend`, `replaceChildren`; `ChildNode`'s `before`, `after`, `replaceWith`, `remove`; and
`Element`'s `slot`, all confirmed against `w3c/webref`'s `dom.idl`), listed per **projected** interface
because neither mixin is an interface this binding emits. Every name is checked against a member that
interface really declares — an inherited one would put the key on the wrong prototype, since
`with (element)` consults `Element.prototype`'s object and never `Node.prototype`'s. The object is a
per-realm slot rather than a constant: WebIDL makes one per interface prototype object, and a page may write
to it (`compile-event-handler-symbol-unscopables.html` does exactly that). The property is
`{ writable: false, enumerable: false, configurable: true }` and the object's `[[Prototype]]` is null.

**It is worth more than the one document that named it.** An event handler content attribute's scope chain is
object environments carrying the *withEnvironment* flag
([HTML](https://html.spec.whatwg.org/multipage/webappapis.html#getting-the-current-value-of-the-event-handler)),
so before this `<div onclick="remove()">` called the **element's** `remove` where the standard resolves the
global's — six rows of `dom/nodes/remove-unscopable.html` say so.

### 2. `DOMSettableTokenList` is an interface the standard removed

DOM merged it into `DOMTokenList` in 2016, but `ISettableTokenList` still carries
`[DomName("DOMSettableTokenList")]` — so `iframe.sandbox`, `output.htmlFor` and `link.sizes` answered
`[object DOMSettableTokenList]`, were not `instanceof DOMTokenList`, and the removed interface was a global.

**`excludedInterfaces` cannot say it**: excluding the interface drops the three members whose CLR type names
it, and `undefined` is worse than the wrong brand. The new `mergedInterfaces` list keeps them —
`ModelBuilder.LookupInterface` answers with the target's model, so only the definition, the interface object
and the type-map candidate go, and a value that implements the merged interface is wrapped as the one the
standard says it is. The merged interface's one member goes with it and is not a loss: `Value` is the
*serialized* token set where [DOM's `value`](https://dom.spec.whatwg.org/#dom-domtokenlist-value) is the
attribute verbatim, and the inherited `DOMTokenList.value` is already the standard's.

### 3. Three exclusion reasons were wrong, and measuring them is the rest of this change

* **`Range-in-shadow-after-the-shadow-removed.html` was filed as a `Range` defect and is a *lane* gap.** The
  document's shadow mode comes from `<meta name="variant" content="?mode=open">`, which upstream's runner
  appends to the URL and a case here — a vendored path and nothing else — does not. So it runs with
  `location.search` empty, calls `attachShadow({mode: null})` and fails WebIDL's enum conversion before
  either test reaches its subject: *"the provided value 'null' is not a valid enum value"* is what both rows
  actually report. `Dom/RangeInShadowTests` measures the subject directly — both modes, both removals — and
  all four pass. **Serving a document once per declared variant is a change to the lane** (it moves the
  census's `Tests` column and makes a case a path plus a query), so it is recorded rather than done;
  `dom/events/handler-count.html` is the only other vendored document that declares any.
* **`Range-adopt-test.html`'s two rows were `NeedsXmlDocuments` and are not about XML documents.** A live
  range whose container is adopted into *any* other document stops being adjusted by that container's
  removing steps. Measured in both directions: `createHTMLDocument()` strands a range exactly as
  `createDocument(null, null)` does, and a range made *after* the adoption is adjusted correctly. AngleSharp
  keeps the live ranges on the `Document` and `adopt` does not move them — see the upstream list below.
* **`compile-event-handler-lexical-scopes-form-owner.html`'s row said `window.customElements` was missing.**
  It is not, and the file's other three rows pass. The failure is that a `formAssociated` custom element is
  not a *form-associated element* without `ElementInternals`, so it takes no part in `form.elements` and no
  form owner reaches the handler's scope chain: the handler answers the string `"global_elements"` where the
  file expects an `HTMLFormControlsCollection`.

**And the two `dom/lists/` rows that remain are neither `DOMTokenList` nor validation**, which is what their
group was called: each asks for `relList` on an `<a>` created **outside the HTML namespace**, and AngleSharp
declares seven SVG interfaces with no `ISvgAnchorElement` and no MathML element interface at all — so there
is no `SVGAElement` and no `MathMLElement` for the member to be on. The `<dl>`/`HTMLDListElement` gap in a
second namespace.

## Failing-first evidence

Measured by reverting the production files (`DomUnscopables.cs` removed, `Generated/` and `tools/dom-bindings/`
restored) with the tests kept — **9 of 27 fail**:

| Test | Unfixed |
| --- | --- |
| `UnscopablesTests.EachInterfacePrototypeListsItsOwnUnscopableMembers` (×5) | `TypeError: Cannot convert undefined or null to object` |
| `UnscopablesTests.TheUnscopableObjectHasTheAttributesWebIdlGivesIt` | `Cannot read properties of undefined (reading 'writable')` |
| `UnscopablesTests.AnUnscopableMemberDoesNotShadowAGlobalInAHandlerScope` | `function remove() { [native code] }` where the standard resolves the global |
| `UnscopablesTests.APageCanWriteToTheUnscopableObject` | `TypeError: Cannot convert undefined or null to object` |
| `DomTokenListTests.EverySettableTokenListIsADOMTokenList` | `[object DOMSettableTokenList]/true/false` |

**Two of the new tests deliberately pass on unfixed code, and say so on themselves.**
`Dom/RangeInShadowTests` (4 cases) exists to prove the engine is right and the lane is not — it is the
evidence behind the corrected exclusion reason, not a fix. And
`DomTokenListTests.ASettableTokenListKeepsTheStandardsValueRatherThanTheSerializedOne` is a no-regression
guard: the merge drops the interface's own `value` member, and this pins that `sandbox.value` still answers
the attribute verbatim rather than the serialized token set.

## Exactly what was verified

```
dotnet build -c Release                                            # solution, 0 errors
JINT_DOM_BINDINGS=update dotnet test Jint.Tests.Browser -c Release --filter DomBindingsStalenessTests
dotnet test Jint.Tests.Browser -c Release                          # 2339 (net10.0) / 2335 (net8.0) passed, 0 failed
dotnet test Jint.Tests -c Release                                  # 12419 (net8/net10), 8509 (net472), 0 failed
dotnet test Jint.Tests.PublicInterface -c Release                  # 3719 / 3709 / 2936 passed, 0 failed
JINT_WPT_BROWSER_CENSUS=1 dotnet test Jint.Tests.Browser -c Release --filter Wpt   # 401 passed, 0 failed
dotnet test Jint.Tests -c Release --filter AgentInstructionFileTests               # the size caps
```

Twelve rows retired: six of `dom/nodes/remove-unscopable.html`, three of
`compile-event-handler-symbol-unscopables.html`, three of `dom/lists/DOMTokenList-coverage-for-attributes.html`.
Census: `dom/nodes/` −6, `dom/lists/` 5 → 2, `html/webappapis/scripting/events/` 5 → 2, total 1,924 → 1,912
on the current base (the parent commit having already taken it from 1,931), and the DOM section's live
sentence 1,586 → 1,577. The cause table was regenerated the same way; three groups were renamed to what they are and
the DOM section's live sentence went 1,568 → 1,559. **Nothing was hand-edited** — the numbers and the row
order come from `JINT_WPT_BROWSER_CENSUS=update`.

The generator reports **no diagnostics**, which is what checks every name in both new lists against the
pinned assemblies: `mergedInterfaces` verifies the merged type exists and its target is projected, and
`unscopables` verifies each name is a member the interface really declares. Both were seen to fire — the
first draft of the `unscopables` table put `DocumentFragment.after` on the list and the run named it.

## What was deliberately left out

* **`<meta name="variant">` support in the lane.** It would make a case a path plus a query, move the
  census's `Documents`/`Tests` columns and touch the shared corpus reader; two vendored documents declare
  variants. Recorded as its own cause row instead, with the engine-level proof beside it.
* **`ElementInternals` and form-associated custom elements** — the one `NeedsTriage` row left in these two
  suites, and a feature rather than a fix.
* **The remaining `dom/events/` rows are not mine and are left exactly as they are**: four
  `EventTarget-dispatchEvent.html` rows are `NeedsMoreEventInterfaces` (`DragEvent`, `StorageEvent`, the two
  device events) and belong to whoever is doing the event-interface constructors; eight
  `Event-dispatch-single-activation-behavior.html` rows are `AssertsWhatNothingRequires`; two are the frame
  realm ([#3771](https://github.com/sebastienros/jint/issues/3771)) and one is `NeedsLayout`. I added the
  measured symptom to the two frame-realm rows' comment (`location.href` refused; `frames[0].onerror` is
  `Illegal invocation` because a frame's window carries the page's `Window.prototype` and the IDL setter's
  brand check does not recognise it) and changed nothing else there.
* **`Element.moveBefore`** is not in the `unscopables` list: the pinned assemblies do not project it, so an
  entry would be a diagnostic.

## Upstream (AngleSharp) findings

Two, both measured here and recorded in `Jint.Browser/Dom/divergences.md`. **No issue was opened anywhere.**

1. **A live range whose container is adopted into another document stops being adjusted by that container's
   removing steps.** [DOM §4.2.3](https://dom.spec.whatwg.org/#concept-node-remove) adjusts every live range
   whose boundary points are in the removed node's tree, and a range has a **root**, not a document — so
   adopting the tree elsewhere changes nothing about which ranges a later removal must adjust. AngleSharp
   keeps the live ranges on the `Document` and `adopt` does not move them. Reproduction: build
   `container > element` detached, `range.selectNode(element)`, `otherDoc.body.appendChild(container)`, then
   `container.removeChild(container.firstChild)` — `endOffset` stays `1` where the standard collapses it to
   `0`. A range created *after* the adoption is adjusted correctly, and `createHTMLDocument()` behaves the
   same as `createDocument(null, null)`, so it is adoption and not XML.
2. **There is no `ISvgAnchorElement` and no MathML element interface in the pinned assemblies**, so an `<a>`
   created in the SVG or MathML namespace is a bare `Element` and SVG 2's `SVGAElement.rel`/`relList` cannot
   be projected onto anything. The same shape as the already-recorded `HTMLDListElement`, `HTMLDirectoryElement`,
   `HTMLFontElement` and `HTMLFrameElement` gaps, and unreachable from this side for the same reason: an
   element's interface is fixed by AngleSharp's own internal factories.

## Base

`40201bb32` (this stack's first commit), on `bd8058d55` — *Reflection: the form controls' attribute table
(#3987)*.


lahma:
Superseded, with each part accounted for on `main`:

- the `@@unscopables` override list landed through #3997 (`ee725067c`), which implements the same `unscopables` list in `overrides.json` plus `DomUnscopables`; the two PRs were written in parallel and one mechanism is enough;
- `mergedInterfaces` for `ISettableTokenList` was resolved the other way by #3992 (`8ea3bd236`): the interface is excluded and `DomTypeMap` falls through to `DOMTokenList`, which retired the same three `dom/lists` brand rows this branch targeted, with `area.ping` and `td.headers` reflected as the strings HTML declares;
- the corrected reasons for `Range-adopt-test.html` (a live range stranded on adoption, AngleSharp's) and the `<meta name=variant>` documents are on main through #3990 and are completed by #3998.

Nothing here is dropped: the data: URL processor and the response-URL fragment from the first half of this work merged as #3991. Thank you for the measurement that corrected four cause rows.

Issue #3813, opened 2026-09-04
NodeList.length cannot be redefined because a DOM collection is an ArrayLikeObject, and moving it off the instance costs the fast lane


lahma:
Split out of #3772 item 2, because it is not a DOM defect and does not belong with the others.

`dom/nodes/NodeList-static-length-getter-tampered-*.html` (six documents) expects `length` to be an accessor on `NodeList.prototype`, redefinable like any other, and expects `Array.prototype.indexOf` over a `NodeList` whose getter has been tampered with to answer `-1`. Today `Cannot redefine property: length` and the tampered read answers `0`.

## Why it is not a one-line fix

`DomCollectionBase` documents `length` on the instance as a **deliberate deviation**, inherited from the engine's `ArrayLikeObject`. That base is what puts `list[i]`, `for..of`, spread and the `Array.prototype` generics on the one-callback-per-element lane, and the DOM binding generator deliberately does **not** emit `length` onto a collection shape *because* of it. So this is a change under [`Jint/Native/Object/AGENTS.md`](https://github.com/sebastienros/jint/blob/main/Jint/Native/Object/AGENTS.md) with a measurable fast-lane cost, not a DOM fix — which is exactly the kind of decision that file exists to gate.

## What the decision actually is

Whether a DOM collection can keep the `ArrayLikeObject` lane while exposing `length` as a prototype accessor a script may redefine, and if not, what the lane is worth against six wpt documents.

## Note on the row count

The six `NodeList-static-length-getter-tampered-*.html` documents are currently `NotVendored`, so they are **not in the corpus** — #3772's headline count for item 2 includes rows nothing runs. Vendoring them is part of this issue, not a prerequisite someone else owes.

Recorded rather than left as a TODO, per the campaign's rule. Tracked by #3575.

lahma:
Costed options note. No code; the measurement section says what I could not settle by reading.

## The decision is narrower than "can the lane survive"

The indexed lane never touches the `length` *property*. `list[i]` resolves through `TryGetIndex` and the C# `Length`; the `length` property is a separate branch in each override. **Moving `length` off the instance costs the indexed lane nothing.**

What the six documents actually test is a different thing: that `Array.prototype` generics read length through `[[Get]]`, so a tampered getter is observed. Three sites bypass that, and one of them is per-element.

| # | Site | What it does |
| --- | --- | --- |
[...]

PR #4002, opened 2026-09-09
Read a host collection's inherited length without invoking the accessor

lahma:
Draft, because the lead measures before this is a real PR. **No benchmark or profiler was run for this branch**; everything below that is not a test result comes from reading the path, and is labelled as such.

The first half of what [#3947](https://github.com/sebastienros/jint/issues/3947) called "a throughput campaign over `for (j = 0; j < obj.length; j++) obj[j]` on a host collection", and the shape [#3813](https://github.com/sebastienros/jint/issues/3813) left open after B+C landed.

## 1. Which lane a host collection is in today, by reading the path

`obj` is a `Jint.Browser` collection wrapper — `DomCollectionObject` or `DomHtmlCollectionObject<T>`, both `DomCollectionBase : ArrayLikeObject`. Two reads per iteration, in very different shape.

**`obj[j]` is already on the narrow lane and there is little left to take.** `JintMemberExpression.GetValue`'s computed-read branch tests `baseValue is ArrayLikeObject` and calls `ReadIndex` → `TryGetIndex` directly: no `Reference` rent, no key object, no descriptor. What remains is the wrapper cache — `DomRealm.Wrap`/`WrapNode` is one `ConditionalWeakTable` probe, and that probe **is** the wrapper-identity rule (`Jint.Browser/Dom/AGENTS.md`), so it cannot be skipped. Nothing here touches it.

**`obj.length` was the expensive half, and it was expensive twice over.**

`DomCollectionBase` overrides `OwnsLength => false`, because WebIDL puts `length` on the interface prototype as an accessor. That is correct and stays correct — it is what makes a redefined getter observable, and it is the deliberate divergence `Jint.Browser/Dom/AGENTS.md` records. But it sorted the read into the slowest lane the engine has for a name:

1. `ArrayLikeObject` carries `InternalTypes.OrdinaryGet | OwnValueHook`, so the read lands in `JintMemberExpression.ReadFromNonPlainReceiver` → `TryGetOwnPropertyValue("length")`, which answers **false** — `OwnsLength` is false, `length` is reserved from the named projection, and the ordinary property bag is empty;
2. → `ReadAfterOwnMiss` → the prototype-method inline cache, which hits;
3. → `UnwrapJsValue` on a `GetSetPropertyDescriptor`, i.e. **a full JavaScript function call**: `ClrFunction.Call` with its native stack-headroom check (plus a `try`/`catch` unless CLR exceptions bubble), then `DomFailures.Guard`'s closure and its `try`, then `DomBindings.Bind<INodeList>`'s brand check, then the AngleSharp read, then `DomConvert.Number` → a fresh `JsNumber`.

The second cost is larger and invisible from the first: because `JintBinaryExpression.NumericConstantComparisonLane.TryReadLengthBound` recognised only `JsArray` and `JsString`, **a host collection made the whole loop-test lane decline**. `j < obj.length` therefore also materialised `j` as a `JsNumber` out of its slot and ran the generic `JintBinaryExpression` less-than, where the identical loop over a plain array keeps both operands raw doubles and allocates nothing. That is most of the difference between "a few ns" and "~200 ns".

None of this is DOM-specific — it is `ArrayLikeObject` + `OwnsLength => false`, so **every embedder that put `length` on a prototype was in the same lane**. `Jint.Tests.PublicInterface/HostArrayLikeObjectTests.AHostCanPutLengthOnItsPrototypeAndAllLengthConsumersObserveIt` is the test that has been pinning that arrangement, and it still passes unchanged.

## 2. The lane, and why it is invisible

One new hook on `ArrayLikeObject`:

```csharp
protected virtual ObjectInstance? PristineLengthGetter => null;
```

A collection that opted out of `OwnsLength` hands back the `length` getter **its prototype was created with**. The engine then answers a `length` read from `Length` — no descriptor, no accessor invocation — while, and only while, four facts hold:

| guard | what it catches |
| --- | --- |
| `_propertiesVersion` unchanged since validation | an own `length` on the instance — `NodeList-static-length-getter-tampered-1` |
| `GetPrototypeOf()` still the validated prototype | `Object.setPrototypeOf` — `…-tampered-2` |
| the prototype's `_propertiesVersion` unchanged | `Object.defineProperty(NodeList.prototype, 'length', …)` — `…-tampered-3` |
| the resolved `length` getter still **is** `PristineLengthGetter` | a redefinition that installs a different accessor, including one computing the same answer |

That is deliberately the prototype-method inline cache's own guard set — receiver identity, receiver version, direct-prototype identity, holder version — written per *instance* rather than per call site, because the sites that read a collection's length are not all interpreter nodes (`GetLongLength` is reached once per element by the array iterator). Validation additionally refuses an exotic receiver or holder, so a `Proxy` anywhere never has a trap bypassed; refuses a holder whose version cannot witness the name (`ObjectInstance.VersionWitnessesOwnProperty`, moved out of `JintMemberExpression` so both lanes share one predicate); and probes the receiver for an own `length` through `ProbeOwnPropertyChecked`.

Why the shortcut cannot diverge from `[[Get]]`: the guard says the accessor resolving for `length` is the one the host declared, and the host's declaration says invoking that accessor answers `Length`. Reading `Length` is then the same read. **With host-contract verification on, every read that takes the lane invokes the accessor and compares** (`AssertPristineLengthAgreesWithTheAccessor`, re-entrancy-guarded so a JavaScript getter reading `this.length` cannot recurse) — so an embedder running its suite with `JINT_HOST_CONTRACT_VERIFICATION=1` is the checker, which is what this repository's browser leg then does over roughly 76 M reads per tampered document (§5).

The **verdict is cached in both directions**. Positive, so a hit is three comparisons; and negative, so a collection whose accessor a page *has* redefined does not re-walk its prototype chain on every read of a loop. That second half is why the tampered case must not regress, and it has a benchmark row of its own.

`OwnsLength` at its default needs no guard and gets the lane unconditionally: `length` is an own data property this class answers from `Length` before the property bag or the prototype chain is consulted, and neither `[[Set]]` nor `[[DefineOwnProperty]]` can replace it, so fast and slow were always the same read.

### Where it is consumed

- `ArrayLikeObject.GetLength()` / `GetLongLength()` — where `ArrayLikeObjectOperations` (every read-mode `Array.prototype` generic) and the array iterator arrive, the latter **once per element** of a `for..of` / spread / `Array.from`.
- `JintBinaryExpression`'s `identifier < base.length` loop-test lane — the shape in the reproduction, and the one that also gives the counter back its unboxed slot read.
- `JintMemberExpression.ReadFromNonPlainReceiver`, behind a build-time `_readsLengthName` flag so every other member node pays one field test and no type check.

Nothing about `list[i]`, the descriptor model, the key enumerations, `Delete`/`DefineOwnProperty` or wrapper identity changed. **No engine type derives from `ArrayLikeObject`**, so nothing test262 exercises can reach the new code at all.

### The `Jint.Browser` half

`DomRealm.PrototypeOf` captures the `length` getter **at the moment it creates a collection prototype** — before a page can have touched it; any later and a tampered getter could be recorded as the pristine one — and `DomCollectionBase.PristineLengthGetter` hands that object back. It costs one materialised `ClrFunction` per collection prototype per engine and does not move `_propertiesVersion`. Only `DomWrapperKind.Collection` and `HtmlCollection` are asked, because only those are `ArrayLikeObject`s: `DomIndexedNodeObject` (`form`, `select`) is a node wrapper and reads its `length` the ordinary way, untouched.

The claim it makes holds because every generated collection accessor reads the same AngleSharp member the interface's `length` attribute does, and `HTMLCollection.prototype.length` is literally the wrapper's `Length`.

## 3. Expected order of magnitude, from reasoning — not measurement

Per iteration of `for (var j = 0; j < obj.length; j++) obj[j]`, the lane removes:

- **one JavaScript function call** — `ClrFunction.Call` + the DOM guard closure + a brand check — and the `JsNumber` it produced;
- **the own-property probe in front of it** (`IsArrayIndex` over the string `"length"`, the projected-name test, an ordinary property-bag miss) and the prototype-cache validation behind it;
- **the whole generic binary-comparison path**, because the loop-test lane can now engage: `j` stays a raw `double` in its slot instead of being materialised, and the compare is one instruction instead of `JintBinaryExpression`'s dispatch.

What replaces it: three field comparisons, one `GetPrototypeOf()`, two virtual calls (`OwnsLength`, `Length`) ending in an AngleSharp field read.

So the loop **test** should go from something on the order of a hundred-plus nanoseconds to single digits, and the iteration as a whole from the ~200 ns #3947 measured towards whatever `obj[j]` alone costs — the computed-read lane, `TryGetIndex`, the AngleSharp indexer and one `ConditionalWeakTable` probe. I would expect a large majority of the iteration removed, and I would **not** expect the gap to the plain-array control to close: the wrapper-cache probe is the price of the collection being live, and it stays.

The tampered rows should not move. **If `InheritedLengthTampered` regresses, the negative-verdict cache is not doing its job** and that is the signal to stop rather than to accept a cost.

## 4. Failing-first, and each guard broken on purpose

Three new suites, and the split between them is the point: every *semantic* test below would still pass if the lane never engaged at all.

- **`Jint.Tests/Runtime/ArrayLikeObjectLengthLaneTests.cs`** asks the object which read answered. An owned `length` needs no guard; the lane engages only when the host declares the accessor; each of the three tamper shapes disarms it and two of them re-arm on restore; a redefinition that computes the *same* answer disarms it permanently; both verdicts are cached; an exotic (`Proxy`) prototype is refused.
- **`Jint.Tests.PublicInterface/HostArrayLikeInheritedLengthTests.cs`** makes the same claims from outside the assembly, in script: every `length` consumer (`list.length`, `Array.prototype.indexOf`, spread, `JSON.stringify`, the indexed loop) answers the tampered value; live growth and shrinkage between *and during* reads; a `Proxy` in front of the collection runs its traps; a host that declares nothing keeps the accessor running on every read (counted in script); an owned `length` is unchanged.
- **`Jint.Tests.Browser/Dom/CollectionLengthTests.cs`** covers ten real collection interfaces (static and live `NodeList`, `HTMLCollection` live and static, `DOMTokenList`, `NamedNodeMap`, `StyleSheetList`, `CSSRuleList`, `HTMLFormControlsCollection`, `HTMLOptionsCollection`) × the three tamper shapes, plus: the accessor still has WebIDL's attributes, name and arity and `length` is still not an own property; an indexed own property is still refused and changes no length; the collection is still live; `c[0] === c[0]` and one wrapper per node across a whole loop; a `Proxy` runs its traps; and every collection interface's captured getter **is** the one its prototype declares.

**Two guards were removed on purpose and watched to fail**, because a guard that matches nothing is invisible and looks green:

| guard weakened to | what went red |
| --- | --- |
| `…Get is null` instead of `ReferenceEquals(…Get, pristine)` | 20 `CollectionLengthTests` cases, 2 `ArrayLikeObjectLengthLaneTests`, 3 `HostArrayLikeInheritedLengthTests` |
| `_propertiesVersion == _lengthGuardVersion` → `true` | 10 `CollectionLengthTests.AnOwnLength…` cases, 1 `ArrayLikeObjectLengthLaneTests` |

Both were then restored and the suites re-run green.

## 5. What was verified

| | |
| --- | --- |
| `Jint.Tests` `net8.0` | 12409 passed, 0 failed |
| `Jint.Tests.PublicInterface` `net8.0` | 3718 passed, 0 failed; new tests also green on `net472` |
| `Jint.Tests.Browser` `net8.0` | 2408 passed, 0 failed; the new file also green on `net10.0` |
| `Jint.Tests.Test262` `net10.0` | 102593 passed of 102702. Two `staging/sm/expressions` cases failed on a run that shared the machine with a full-solution build; the whole staging class then passed **2741/2741** on its own. That is the documented load-flake family, and it is unreachable from here: no engine type derives from `ArrayLikeObject`, so test262 never enters the new code |
| the six wpt documents | `NodeList-static-length-getter-tampered-{1,2,3}` and their three `-indexOf-` siblings all pass |
| the same six under `JINT_HOST_CONTRACT_VERIFICATION=1` | pass — roughly 76 M reads per document, each comparing the lane's answer against the accessor it skipped |
| `DomPrototypeTests`, `WebIdlPropertyAttributeTests`, `HtmlCollection*`, `NamedNodeMapVisibilityTests` under verification | pass |
| `AgentInstructionFileTests`, `MigrationGuideTests`, `SpecCitationTests` | pass |

Public surface: one added line in each of the five `PublicApiTest_*.verified.txt` baselines — `protected virtual ObjectInstance? PristineLengthGetter { get; }` — and nothing else, with the migration-guide row written in this PR (§5.35).

**Observed wall-clock on a shared machine, offered as a signal and explicitly not as a measurement**: the six documents ran 13/10/11 s and 4/4/4 s here, against the 17/15/16 s and 6–10 s [#3947 recorded on an idle 5950X](https://github.com/sebastienros/jint/issues/3947). Only the paired run decides anything.

## 6. Benchmark rows to run — paired, before/after, default job

```
dotnet run -c Release --project Jint.Benchmark\Jint.Benchmark.csproj -- --filter "*HostCollectionLengthBenchmark*"
dotnet run -c Release --project Jint.Benchmark\Jint.Benchmark.csproj -- --filter "*HostArrayLikeBenchmark*"
```

`Jint.Benchmark/HostCollectionLengthBenchmark.cs` is new — 400 elements, one engine per row, restricted to the public surface, `[MemoryDiagnoser]`:

| row | what it is | expectation |
| --- | --- | --- |
| `InheritedLength` | the WebIDL collection with the accessor declared | **the row this is about** |
| `InheritedLengthUndeclared` | the same collection with nothing declared | the *before* picture, in the same process |
| `InheritedLengthTampered` | the same collection after `defineProperty(proto, 'length', …)` | **must not regress** — the negative-verdict cache |
| `OwnedLength` | the `ArrayLikeObject` default | should improve a little, from the loop-test lane |
| `PlainArray` (baseline) | the identical loop over a `JsArray` | the floor; the residual gap is the wrapper cache |

`HostArrayLikeBenchmark` is unchanged but its `ArrayLike` rows move (`IndexedLoop` reaches the loop-test lane, `ForOf` reaches `GetLongLength`), so it wants a paired run too. `PlainHost` and `JsArrayCopy` are its control rows and must not move.

Per `Jint.Benchmark/AGENTS.md`: `measure-paired.ps1` between two worktrees, default job, a row is a regression only when its CI excludes zero, and never a `--job short` number.

## 7. Left out, and doubts

- **`obj[j]` is untouched.** The remaining per-iteration cost is the `ConditionalWeakTable` probe in `DomRealm.Wrap`, and that probe is the wrapper-identity rule. Making it cheaper is a separate change with a different risk profile.
- **The hook is a host promise, verified rather than proved.** The engine cannot check statically that the declared getter reads `Length`; it checks it on every read under host-contract verification, and `CollectionLengthTests.EveryCollectionInterfaceCapturedItsOwnDeclaredAccessor` plus the three tamper suites cover the binding. An embedder that declares a getter computing something else gets a wrong answer in a non-verifying build — the same shape of obligation `HasIndex`, `TryGetOwnPropertyValue` and `ProbeOwnProperty` already carry, documented the same way.
- **Four fields per `ArrayLikeObject` instance** (a reference, two `uint`s, a `bool`), paid by every collection including one that never declares the hook. A per-call-site cache would have avoided them but could not have served `GetLongLength`, which is the per-element site.
- **`Jint/Native/Object/AGENTS.md` is now 109 bytes under its cap.** The next agent to add there has almost no room; the honest fix is a trap/remedy split of the `ArrayLikeObject` bullet, which did not belong in a performance PR.
- **The `-indexOf-` documents read `length` once per `indexOf` call**, not per element, so their improvement should be small. If the numbers say otherwise the explanation is somewhere else and worth chasing.

Base SHA: `8ea3bd236978b75f90d2543a21b4bcf84a774fea` (`upstream/main`, "DOM collections, the ChildNode mutation order, and three binding leftovers (#3992)"), rebased onto it immediately before pushing.


lahma:
Measured by the lead on an otherwise idle machine (5950X, Windows, Release, BenchmarkDotNet default job), paired A B B A between a worktree at `main` `da29c12f8` and this branch rebased onto it (`98227e38e`).

`HostCollectionLengthBenchmark` (400 elements per loop; the class is new here, so its *before* picture is the `InheritedLengthUndeclared` row, an `ArrayLikeObject` written without the hook, measured in the same process):

| Row | run 1 | run 2 |
| --- | ---: | ---: |
| `PlainArray` (floor) | 34.65 µs | 37.00 µs |
| `OwnedLength` | 34.57 µs | 34.71 µs |
| `InheritedLength` (the lane) | **35.29 µs** | **35.32 µs** |
| `InheritedLengthUndeclared` (the old path) | 103.81 µs | 115.76 µs |
| `InheritedLengthTampered` (negative verdict) | 132.90 µs | 106.18 µs |

So an inherited `length` now costs what an own one costs, and both cost what a plain array costs: the old path was about 3× that. The tampered row stays where the old path is, as designed; its two runs straddle the undeclared row, which reads as noise on a 100-µs row rather than a regression.

`HostArrayLikeBenchmark` (pre-existing rows, 10,000 elements), main vs branch, both orientations:

| Row | main | branch |
| --- | ---: | ---: |
| `IndexedLoop / ArrayLike` | 1,009 / 1,026 µs | 1,010 / 1,016 µs |
| `ForOf / ArrayLike` | 1,059 / 1,114 µs | 1,064 / 1,072 µs |
| `IndexedLoop / PlainHost` | 1,512 / 1,551 µs | 1,475 / 1,517 µs |
| `IndexedLoop / JsArrayCopy` | 998 / 1,007 µs | 1,021 / 1,006 µs |

No row moves outside its run-to-run envelope. Marking the draft ready and landing it.

PR #4004, opened 2026-09-09
document.all is an HTMLAllCollection, and it carries [[IsHTMLDDA]]

lahma:
## What and why

`document.all` was projected as an ordinary `HTMLCollection`. Measured on `main` before this change:

| | before | after (and what a browser answers) |
| --- | --- | --- |
| `typeof document.all` | `"object"` | `"undefined"` |
| `Boolean(document.all)` | `true` | `false` |
| `document.all == null` / `== undefined` | `false` | `true` |
| `document.all === undefined` | `false` | `false` |
| `document.all('x')` | `TypeError: all is not a function` | the element |
| `document.all.item('x')` | index-only `item`, arity 1 | name **or** index, arity 0 |
| `document.all['divwithname']` (a `<div name>`) | the element | `undefined` |
| `document.all['foo']` matching two elements | the first element | a new live `HTMLCollection` |
| `document.all instanceof HTMLCollection` | `true` | `false` |

Twenty years of browser sniffing reads the wrong branch from the first three rows.
[HTML §4.13.2.3](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#htmlallcollection)
makes `HTMLAllCollection` a standalone interface, and
[ECMAScript Annex B.3.6](https://tc39.es/ecma262/#sec-IsHTMLDDA-internal-slot) gives its one instance the
`[[IsHTMLDDA]]` internal slot. Both halves are here. #3997's agent left this out as needing "a hand-written
collection accessor plus an override mechanism to select one, and `[[IsHTMLDDA]]` is missing too" — the slot
turned out to be present but unreachable, and the override mechanism is two fields.

## Mechanism

**1. The engine seam.** `InternalTypes.IsHTMLDDA` existed, but its three behaviours were written on
`Jint.Native.IsHTMLDDA` — the class behind test262's `$262.IsHTMLDDA` — so a second bearer could only have
reproduced Annex B by hand. They belong to the slot now:

* `ObjectInstance.ToBoolean` reads the flag (B.3.6.1);
* `JsValue.IsLooselyEqual` reads it (B.3.6.2) — `JsNull`/`JsUndefined` carry the mirror halves and the
  interpreter already fused `x == null`, so only the object side of the **generic** path was missing;
* `JintUnaryExpression` already answered `typeof` from it (B.3.6.3).

A bearer declares the slot with the new `internal` `ObjectInstance.DeclareIsHtmlDda()` and implements none of
it. `DeclareCallable()` is its sibling, for an object with a `[[Call]]` that is not a `Function` — WebIDL's
legacy caller; call sites decide callability from `InternalTypes.Callable`, so an `ICallable` that does not
declare it is silently not callable. Both are `internal` because the specification permits the slot on an
implementation-defined object emulating `document.all` and nothing else; `Jint/Native/Object/AGENTS.md` records
the decision beside the two other "who can reach this lane" ones.

**No public API change.** The B.3.6.2 test sits in `JsValue.IsLooselyEqual` rather than as an `ObjectInstance`
override precisely so that no `protected` member is added for one internal slot — the `PublicApiTest` baselines
are untouched, so there is no `docs/guide/migrating-to-v5.md` row to write.

**2. The interface.** `Jint.Browser/Dom/Collections/DomHtmlAllCollectionObject.cs` is the wrapper (a
`DomCollectionBase`, so `document.all[i]`, `for..of` and the `Array.prototype` generics keep the engine's
one-callback-per-element lane) and `DomManualShapes.HtmlAllCollection` the prototype: `length`,
`item(nameOrIndex)` — an **optional** `DOMString`, so its `length` is 0 and an explicit `undefined` is "not
provided" — `namedItem(name)` (required, so `length` 1 and no argument is a `TypeError`), `@@iterator`, and the
legacy caller. Its supported names are HTML's rather than DOM's: an id always, a `name` only on one of the
fourteen "all"-named elements. A name several elements match answers a live `HTMLCollection`, new per call.

**3. The selection.** `overrides.json`'s `manual` entry gained two optional fields: `wrapper` names the
`DomWrapperKind` the interface's instances get, and `inherits` names what its prototype inherits, with the
**empty string** meaning `Object.prototype`. AngleSharp models `IHtmlAllCollection` as an
`IHtmlCollection<IElement>`, which is the one place its CLR hierarchy contradicts the IDL — without `inherits`
the chain runs through `HTMLCollection.prototype`. Both fields are validated against the pinned assemblies, and
a `manual` entry naming an interface the generator does not project is now a diagnostic (there was no check at
all before). Bindings regenerated with `JINT_DOM_BINDINGS=update`.

## Failing-first evidence

Every new test was run against the **unfixed** tree, in a scratch worktree at `upstream/main` with only the new
test files (and, for the wpt half, the corpus registration) copied in.

* **wpt, `html/infrastructure/common-dom-interfaces/collections/htmlallcollection.html`** — **22 of its 41
  subtests fail**: `named property getter with invalid name`, `named property getter returning collection`,
  `namedItem method with invalid name`, `namedItem method returning collection`, all nine `legacy caller *`
  rows (`all is not a function`, and `call requires 'this' to be of type ICallable`), seven `item method *`
  rows, and `collections are new live HTMLCollection instances`. All 41 pass here.
* **wpt, `html/obsolete/…/document-all.html`** — **both** of its subtests fail
  (`assert_false: expected false got true`, the `Boolean(document.all)` assertion). Both pass here.
* **`Jint.Tests.Browser/Dom/HtmlAllCollectionTests.cs`** — 8 of its 11 tests fail on the unfixed tree
  (`TheObjectCarriesTheIsHtmlDdaInternalSlot`, `TheInterfaceInheritsNothing`,
  `OnlyAnAllNamedElementContributesItsName`, `SeveralMatchesAnswerANewLiveCollectionEachTime`,
  `NamedItemLooksUpNamesAndNothingElse`, `ItemTakesANameOrAnIndexAndUndefinedIsNeither`,
  `TheLegacyCallerIsItemAndIsNotAConstructor`, `SupportedNamesAreOwnPropertiesAndNotEnumerable`). The three
  that pass there assert behaviour that was already right (`[SameObject]` identity and liveness, the
  array-index-versus-name precedence, the expando visibility rule) and are regression pins.
* **`Jint.Tests/Runtime/HtmlDdaSlotTests.cs`** — `ADeclaringObjectImplementsNoneOfIt` does not compile on the
  unfixed tree: `DeclareIsHtmlDda` does not exist. Its sibling
  `TheSlotAloneDecidesTypeofToBooleanAndLooseEquality` passes there **by design** — it is the pin that makes
  moving the three behaviours off `IsHTMLDDA` safe.
* **`Jint.Tests.Browser/Dom/HtmlCollectionIndexLikeNameTests.cs`** — both cases fail on the unfixed tree **only
  with `JINT_HOST_CONTRACT_VERIFICATION=1`**, where the projection raises
  `NameAt(0) reported '0', which belongs to the indexed collection`. Without the switch the engine drops the
  name silently, so there is no observable difference to assert; the fix is about not tripping the verifier,
  and the test says so.

## What was verified

All Release, all from this worktree, base `ee725067cbb55e55d8df1e867d9f0dc544673e84`.

| Command | Result |
| --- | --- |
| `dotnet build -c Release` (solution) | 0 errors; 1 pre-existing MSB3277 in `Jint.Tests.CommonScripts` (net472 test-platform assembly conflict, unrelated) |
| `dotnet test Jint.Tests -c Release` (net472/net8.0/net10.0) | 8,511 / 12,421 / 12,421 passed, 0 failed — includes `AgentInstructionFileTests` |
| `dotnet test Jint.Tests.Browser -c Release` (net8.0 + net10.0) | 2,465 and 2,469 passed, 0 failed — includes all 374 browser-lane wpt cases on both |
| `JINT_WPT_BROWSER_CENSUS=1 dotnet test Jint.Tests.Browser … --filter "…Wpt"` | 404 passed, 0 failed (the census, the cause table and the exclusion discipline) |
| `dotnet test Jint.Tests.PublicInterface -c Release` (net472/net8.0/net10.0) | 2,936 / 3,709 / 3,719 passed, 0 failed — the API baselines are unchanged |
| `dotnet test Jint.Tests.Test262 -c Release` | 102,594 passed; one load flake, `staging/sm/Array/toSpliced-dense.js` timing out at 30 s, which passes in 1 s on its own |
| `dotnet test Jint.Tests.Test262 … --filter "…IsHTMLDDA\|…annexB"` | 1,376 passed, 0 failed |
| `JINT_HOST_CONTRACT_VERIFICATION=1` on `Jint.Tests`, `Jint.Tests.PublicInterface`, `Jint.Tests.Browser` | 12,406 / 3,722 / 2,402 passed, 0 failed |
| `dotnet test Jint.Tests.DevTools`, `Jint.Tests.SourceGenerators`, `Jint.Tests.CommonScripts` | 413+415 / 71 / 28+28 passed, 0 failed |

The two vendored documents were byte-verified against the pin the way `Vendor/README.md` describes — their blob
SHAs at `6c7127bdd9f2cc6a3668fd9791757843e09d5a9e` are `14faa2128ed0bd638585955994f6b1fb48cb6a4f` and
`37a92bb09f41fe2cbcccc842967e56ede3abce26`, and `git hash-object` on the vendored copies gives the same two.
The engine lane's drift recipe was regenerated with `JINT_WPT_CENSUS=update` (81 directories, 904 files).

`html/dom/historical.html`'s `document.all cannot find applet` passes now — `applet` is not one of the fourteen
"all"-named elements — so #3997's `document.all*` exclusion is **deleted** rather than narrowed, together with
the paragraph that argued it, and the cause it belonged to (`a tag query's namespace and local-name identity`)
drops from 20 tests over 7 documents to 19 over 6. That file's own prose said two of its failures were somebody
else's cause; one is left, and the sentence says so. That is the only exclusion this change touches, and it is
the reason `html/dom`'s `Not passing` drops by one.

## Deliberately left out

* **Only one document is vendored out of each new suite directory.** Their siblings are about other interfaces
  — `HTMLFormControlsCollection`, `HTMLOptionsCollection`, `RadioNodeList`, `DOMStringList`, HTML's obsolete
  document colours — and vendoring one moves the census's `Documents` and `Tests` columns, so it is a change of
  its own. `Wpt/README.md` says so where the two suites are described.
* **A repeated *property* read of `document.all['name']` that several elements match answers one collection**,
  where a browser makes a new one per read. The engine verifies an own read by asking `TryGetOwnPropertyValue`
  and `GetOwnProperty` for the same key and comparing with `SameValue`, so a named getter may not manufacture a
  *fresh* value; the property lane memoizes the last name it built a sub-collection for, and never goes stale
  because that collection is a live filter over the name. `namedItem`, `item` and the legacy caller each build
  a new one, which is the part upstream's corpus checks (`assert_not_equals` across the four call sites).
  Recorded as a deliberate divergence in `Jint.Browser/Dom/AGENTS.md`.

## Upstream findings

Neither has been reported anywhere; recording them here as the repository's rules require.

* **AngleSharp — `IHtmlAllCollection` models `HTMLCollection` rather than `HTMLAllCollection`.** It declares no
  member of its own and simply extends `IHtmlCollection<IElement>`, so its string indexer implements DOM's
  named-lookup rule: `document.all.divwithname` answered a `<div name>`, a name several elements match answered
  the first rather than a collection, there is no `item(name)`, and nothing is callable. Less a defect than a
  surface it does not model — `Length`, the indexer and the enumerator are still AngleSharp's here, and the
  rest is this package's. A row is in `Jint.Browser/Dom/divergences.md`.
* **Jint (fixed here, not AngleSharp) — `DomHtmlCollectionObject` advertised index-spelling supported names.**
  [WebIDL's `[[OwnPropertyKeys]]`](https://webidl.spec.whatwg.org/#legacy-platform-object-ownpropertykeys)
  excludes a supported name that is an array index, because the indexed getter answers that key and stops. With
  `JINT_HOST_CONTRACT_VERIFICATION=1` a `<div id="0">` made `Object.keys(el.children)` raise; without it the
  engine dropped the name silently, so nothing was observably wrong and nothing had caught it.

## Base

`ee725067cbb55e55d8df1e867d9f0dc544673e84` — *@@unscopables, and what html/dom/historical.html is really
about (#3997)*. Rebased onto it after the first push; the numbers in the table above are from the run on this
base.

*(The first push was green on every leg. `windows` went red once on a known load flake — two
`NodeList-static-length-getter-tampered` documents timing out on a saturated 4-core runner, the same failure
two other branches hit earlier the same day — and passed on a re-run with no change.)*


lahma:
The `windows` leg is red on a known load flake, not on this change.

Both errors are `dom/nodes/NodeList-static-length-getter-tampered-1.html` and `-2.html` reporting `TIMEOUT` on the **net8.0** browser leg, which then makes the census refuse to render (`the web-platform-tests browser census is incomplete because these documents produced harness errors`). The diagnostics say what happened: `elapsedMs=66185.7`, `processCpuMs=123843.8`, `processors=4`, `lastReportReadyState=loading` — a 4-core runner with both browser TFM legs on it, and the document never finished loading. It is inherently the most CPU-bound document in the corpus: `support/NodeList-static-length-tampered.js` runs 50,000 x 51 collection reads per call and the file calls it 50 times, so ~125 million `nodeList.length` reads through a prototype accessor.

Three things say it is the runner and not the diff:

* the **net10.0** browser leg on the same runner passed both documents (`2,433 passed, 0 failed`);
* the same two documents timed out the same way on two other branches earlier today — runs [34292847826](https://github.com/sebastienros/jint/actions/runs/34292847826) (`fix/wpt-xml-documents-category`) and [34291566027](https://github.com/sebastienros/jint/actions/runs/34291566027) (`fix/wpt-scripting-tail-2`);
* locally on Windows the whole `dom/nodes` suite passes on net8.0 (`168 passed, 0 failed`, 1 m 20 s), as does the full browser suite on both TFMs, and nothing this change touches is on that document's path — `NodeList` is a `DomCollectionObject` with a generated accessor, not the `DomHtmlCollectionObject` whose supported-name list moved.

Re-running the job.

Issue #4013, opened 2026-09-09
Performance investigation backlog: what the browser and DevTools campaign measured but did not fix


lahma:
A collection point for performance findings that surfaced while landing the September browser, DevTools and reflection work, each with the measurement that found it. None of these was fixed in the PR that found it, either because it was out of that PR's scope or because it needs a profile rather than a reading of the code. The intended next step for each is a run under the ultra profiler on an idle machine, then a benchmark-gated change (paired before/after, default BenchmarkDotNet job, both orientations), per `Jint.Benchmark/AGENTS.md`.

## Interpreter and engine

- [ ] **Indexed reads over a host collection still pay a wrapper lookup per element.** #4002 removed the `length` call from `for (j = 0; j < list.length; j++) list[j]` (35 µs per 400-element loop, the plain-array floor, down from 104-116 µs), but `list[j]` itself still goes through `DomRealm.Wrap`'s `ConditionalWeakTable` probe on every read, which is the wrapper-identity rule. The three `dom/nodes/NodeList-static-length-getter-tampered-*.html` documents (~76 M iterations each) went from 15-17 s to 10-13 s on net8.0 and are still the slowest documents in the lane by a factor of five (#3947, the reproduction comment). A per-collection or per-realm identity cache that does not hold the DOM alive is the candidate; measure the CWT probe's share first.
- [ ] **net8.0 runs the same interpreted loop about 30 % slower than net10.0** (15-17 s vs 11-12 s on the documents above, same machine, same tree; #3947). Worth one profile of each runtime on the same script to see whether it is the runtime's JIT or something Jint does differently under `NET8_0`.
- [ ] **A loose-equality bit test and four guard fields per `ArrayLikeObject`** were added by #4004 and #4002 respectively; both were reasoned to be negligible and the existing `HostArrayLikeBenchmark` rows did not move, but neither was profiled on a plain-object-heavy workload. Confirm on the SunSpider/Dromaeo gate that neither shows.

## Browser runtime

[...]

PR #4027, opened 2026-09-09
Cache the element wrappers of a static NodeList, per DOM §4.2.6

lahma:
A per-collection cache of element wrappers for **static** node lists, so a repeated `list[j]` answers from an array instead of the realm's weak table.

The lead's ultra profile on [#4013](https://github.com/sebastienros/jint/issues/4013) measured the loop `for (j = 0; j < nodeList.length; j++) if (nodeList[j] === el)` over a 100-element `querySelectorAll` result — the shape `dom/nodes/support/NodeList-static-length-tampered.js` compiles. On net8.0, `DomAccessorNodeList.TryGetIndex` was **14.2 % of the page-loop thread** inclusive: `ConditionalWeakTable.FindEntry` 8.9 %, `ChkCastInterface` 0.7 %, its own body 4.4 %. [#4002](https://github.com/sebastienros/jint/pull/4002) removed the `length` read's cost; this is the element read.

## Mechanism

`nodeList[j]` went `DomCollectionObject.TryGetIndex` → `DomAccessorNodeList.TryGetIndex` → `DomRealm.WrapNodeValue` → `WrapNode` → `ConditionalWeakTable.TryGetValue` on `DomRealm._wrappers`, **on every read of every index**, so a 100-element scan re-derived a hundred wrappers it had already derived on the previous pass.

Three pieces:

* **`DomStaticNodeList`** (`Jint.Browser/Dom/Collections/`) — an `INodeList` that holds the matched nodes in an `INode[]` copied at construction. It is the mirror image of the existing `DomLiveHtmlCollection`, which exists because AngleSharp answers a snapshot where DOM wants a live collection; here the standard wants the snapshot, and the binding owns one anyway so that *staticness is a property of a type*.
* **`DomStaticNodeListObject : DomCollectionBase`** — its wrapper, with an `ObjectInstance?[]` filled lazily from `DomRealm.WrapNode` on each index's first read. The array itself is allocated on the first indexed read, so a match a page only takes the `length` of costs nothing for it.
* **A `querySelectorAll` hook** (`overrides.json` → `DomHostHooks.QuerySelectorAll`), on `Document`, `DocumentFragment` and `Element`. It corrects no behaviour — AngleSharp's `QueryCollection` already is a snapshot. It exists to put DOM §4.2.6's word *"static"* where the binding can act on it. Regenerated with `JINT_DOM_BINDINGS=update`; the generated diff is three lines.

## Where static and live are known — the seam

They were **not** distinguished before this change, and that is the whole difficulty: **nothing about an `INodeList` says whether it is live.** All three of these reached `DomCollectionObject` + `DomAccessorNodeList`:

| producer | AngleSharp type | live? |
| --- | --- | --- |
| `node.childNodes` | `AngleSharp.Dom.NodeList` | **live** |
| `control.labels` | `DomLabelNodeList` (ours) | **live** |
| `x.querySelectorAll(s)` | `AngleSharp.Dom.QueryCollection` | static |

Deciding it from the target's CLR type was the tempting shortcut and is a cliff: `QueryCollection` is `internal sealed`, and the only public fact separating it from the live `NodeList` is that it also implements `IHtmlCollection<IElement>` — a coincidence of AngleSharp's class hierarchy, not a promise about liveness, and one a future AngleSharp could break in silence. So the seam is expressed the other way round: the binding constructs the snapshot itself, from the one operation whose specification says the result is static, and the type it constructs is what the cache keys on.

**Live collections are deliberately untouched by this PR**, and `OnlyASelectorMatchTakesTheStaticLane` pins that: `childNodes` and `labels` keep the accessor-driven wrapper and keep reporting the tree they currently have. A per-index memo over a membership that moves would answer the wrong node, and a removed node it had cached would be pinned for the collection's life. Caching a live collection would need a membership version to key on, and that is a separate change.

## Why it is safe

Each claim is verifiable in the code:

* **Identity cannot move.** The cached value *is* the canonical wrapper: it comes from `DomRealm.WrapNode`, which is the one door onto `DomRealm._wrappers` (`Jint.Browser/Dom/DomRealm.cs`), so the array is a memo of a lookup and never a source of objects. `list[0] === list[0]`, `list[0] === document.querySelector(…)` and `listA[0] === listB[0]` hold by construction.
* **It adds no retention.** A static `NodeList` holds strong references to its nodes — `QueryCollection` holds a `List<IElement>`, `DomStaticNodeList` holds the `INode[]` copied from it — and the wrapper reaches them through its own `DomTarget`. `DomRealm._wrappers` is a `ConditionalWeakTable` keyed on the node, so **a node's wrapper is already kept alive exactly as long as the node is**. The cache therefore names only objects the list was keeping reachable anyway, and the array dies with the wrapper.
* **It is per collection-wrapper instance.** An `ObjectInstance` belongs to one engine and one realm, so the cache lives on the wrapper (per engine) while the node array lives on the snapshot (engine-free). Nothing is shared between engines.
* **`Delete`, `DefineOwnProperty`, `Set`, both key enumerations and the probe are still `ArrayLikeObject`'s**, derived from `Length` and `TryGetIndex`, and `HasIndex` is overridden to agree with `TryGetIndex` at the same instant. Verified on every read under `JINT_HOST_CONTRACT_VERIFICATION=1`.

The wrapper also answers the other half of the profile: the target is a typed `INode[]` field rather than `object` + a per-read interface cast, and there is no `DomCollectionAccessor` virtual call, because this shape has exactly one interface behind it. `DomCollectionAccessor`'s own `object` parameter is unchanged and has to stay — one process-shared singleton serves every engine and every CLR target type of its interface, which is why there is one collection wrapper class rather than twenty.

## Tests

`Jint.Tests.Browser/Dom/StaticNodeListTests.cs`, 11 cases. Green on **net8.0 and net10.0**, and again under **`JINT_HOST_CONTRACT_VERIFICATION=1`** (2688 / 2684 passing, 0 failed, on each leg in each configuration).

**Failing-first.** Only one of them fails against the unfixed code, and that is the honest result rather than a gap: the change is meant to move nothing observable, so the identity, removal, out-of-range and tampered-length tests are regression guards that pass on both sides. Verified by reverting the three generated call sites and re-running: `OnlyASelectorMatchTakesTheStaticLane` failed, the other ten passed.

The two retention tests were checked the other way, since they also pass on both sides. A deliberate leak — a `static List<object>` holding each cache array, introduced, run and reverted — failed `DroppingTheListAndThePageCollectsTheNodesAndTheirWrappers` **and** `ACacheDoesNotOutliveTheListInsideALivePage`, so both have teeth against exactly the mistake they are there to catch. They are reachability questions rather than heap readings, for the reason `Jint.Tests.Runtime.GarbageCollectionTests` records, and each has a control that keeps the page and the list and requires the same objects to survive.

What they cover: identity across repeated reads and expandos; a cached element against `querySelector`, `getElementById`, the tree and a second list; a removed node still answering the same wrapper from the list that matched it, beside the live `childNodes` that does change; out-of-range (`undefined` from an index, `null` from `item()`, `in`/`hasOwnProperty`/`Object.keys`); `instanceof NodeList`, `@@toStringTag`, `Symbol.iterator`, spread, `Array.from`, `forEach` and `Array.prototype.indexOf`; and the tampered-`length` shape of the six wpt documents in miniature, in both the hand-written-loop and the `Array.prototype.indexOf` spellings.

The six `dom/nodes/NodeList-static-length-getter-tampered*.html` documents were run before and after through the browser lane and pass in both. **Their wall-clock times are not a measurement** and are not quoted here: they are a test-runner figure from a machine that was also compiling, on a document whose work is dominated by a harness. The lead owns measurement.

## Benchmark row — written, not run

`Jint.Benchmark/BrowserNodeListBenchmark.cs`. `Jint.Benchmark` **can** reference `Jint.Browser` — it already does, for `BrowserLayoutBenchmark` — so the row lives in the same project as every other row rather than in a browser-side benchmark project (there is none; `tools/browser-comparison` measures other engines).

It runs the wpt loop shape over a real `Page`: 100 spans, the match at index 50, `break` on the match, 100 passes an operation. Four rows, one `Page` and therefore one engine and one document each, warmed with that row's own loop and nothing else, page construction and the HTML parse outside the measurement:

* `StaticNodeList` — the row the change is about.
* `StaticNodeListTamperedLength` — the same list after the page defines an own `length` accessor answering the same count, so the pristine-length lane must decline and the loop still does the same work. What is left in the delta is the element read.
* `LiveNodeList` — the same loop over `childNodes`; the control that must not move.
* `PlainArray` — the baseline and the floor, over `Array.from(list)`.

**No benchmark and no profiler was run.** The four row scripts were each checked for correct answers (index 50, length 100) by evaluating them once on a real page through `jint-browser eval`, which is a correctness check and not a timing.

## Doubts

* The row sizes (`Passes = 100`, 100 elements) are a guess at "long enough to measure, short enough to repeat" and may want adjusting once the lead sees the first table.
* `DomStaticNodeList` copies the match into an `INode[]` rather than holding AngleSharp's collection. That is a deliberate second array per `querySelectorAll` — beside the `List<IElement>` the match already built — bought because holding the collection would make the staticness claim rest on the hook being its only caller, and `DomLiveHtmlCollection` is itself an `IHtmlCollection<IElement>` that would compile there.
* `MutationRecord.addedNodes` / `removedNodes` are static per DOM too and are **not** covered here; they keep the accessor-driven wrapper.
* The live lane still pays the `object` + interface cast per read. Making `DomCollectionAccessor` typed would touch twenty generated classes and is not attempted here.
* `Jint.Tests.Browser/Wpt/README.md` still says the six tampered documents "are not vendored"; they are vendored and are cases (`MinimumTests`) since #3875. Not touched — it is someone else's sentence to correct.

Refs #4013 and #3947.

https://claude.ai/code/session_01WKytThti6mniJepuE8P691

### DIFF
--- a/Jint.Benchmark/BrowserNodeListBenchmark.cs
+++ b/Jint.Benchmark/BrowserNodeListBenchmark.cs
@@ -1,0 +1,147 @@
+using System.Text;
+using BenchmarkDotNet.Attributes;
+using Jint.Browser;
+
+namespace Jint.Benchmark;
+
+/// <summary>
+/// The element read of the loop <c>dom/nodes/support/NodeList-static-length-tampered.js</c> compiles —
+/// <c>for (j = 0; j &lt; list.length; j++) if (list[j] === el)</c> — over a <b>real page</b>, so that
+/// <c>list[j]</c> is a DOM projection rather than a host object written for a benchmark.
+/// </summary>
+/// <remarks>
+/// <para>
+/// <b>Why it is here and not beside <see cref="HostCollectionLengthBenchmark"/>.</b> That class measures the
+/// same loop's <c>length</c> read with a hand-written <c>ArrayLikeObject</c>, which is the right shape for a
+/// question about the engine. This one is about what the <i>binding</i> does per element — the wrapper lookup
+/// behind <c>DomAccessorNodeList.TryGetIndex</c> — and no host object written here can stand in for it,
+/// because the cost is a <c>ConditionalWeakTable</c> probe against the realm's wrapper table.
+/// <c>Jint.Benchmark</c> already references <c>Jint.Browser</c> for <see cref="BrowserLayoutBenchmark"/>, so
+/// the row lives in the same project as every other row rather than in a browser-side benchmark project of
+/// its own — there is none, and <c>tools/browser-comparison</c> measures other engines rather than this one.
+/// </para>
+/// <para>
+/// <b>What each row is.</b>
+/// </para>
+/// <list type="bullet">
+/// <item><description>
+/// <see cref="StaticNodeList"/> — the loop over a <c>querySelectorAll</c> result, which DOM §4.2.6 makes a
+/// static <c>NodeList</c>. This is the row the change is about.
+/// </description></item>
+/// <item><description>
+/// <see cref="StaticNodeListTamperedLength"/> — the same list after the page has defined an own <c>length</c>
+/// accessor answering the same count, which is the half of the wpt documents that makes the pristine-length
+/// lane decline. The loop does the same work, so what is left in the delta is the element read.
+/// </description></item>
+/// <item><description>
+/// <see cref="LiveNodeList"/> — the same loop over <c>childNodes</c>, which is live and deliberately keeps the
+/// accessor-driven wrapper. It is the control that must not move.
+/// </description></item>
+/// <item><description>
+/// <see cref="PlainArray"/> — the floor and the baseline: the identical loop over the
+/// <c>Array.from(list)</c> of the same elements, so the gap is the price of the list being a projection.
+/// </description></item>
+/// </list>
+/// <para>
+/// <b>Engine isolation.</b> One <see cref="Page"/> — and therefore one engine and one document — per row,
+/// built in <c>[GlobalSetup]</c> and given only that row's own fixture, so a row's number never depends on
+/// which sibling warmed which lane first. Page construction and the HTML parse stay outside the measurement;
+/// what the measured call adds over the loop itself is one mailbox round trip per operation, which is the
+/// same term in every row including the baseline.
+/// </para>
+/// </remarks>
+[MemoryDiagnoser]
+public class BrowserNodeListBenchmark
+{
+    /// <summary>
+    /// 100 elements, which is what the wpt helper builds and the size the profile on
+    /// sebastienros/jint#4013 was taken at.
+    /// </summary>
+    [Params(100)]
+    public int Count { get; set; }
+
+    /// <summary>
+    /// 100 passes of the inner loop per operation. The wpt document runs 50,000 and takes seconds; this is
+    /// the same shape at a size a benchmark can repeat.
+    /// </summary>
+    private const string Passes = "100";
+
+    /// <summary>
+    /// The loop, verbatim in shape from the wpt helper: the match is at index 50, and the <c>break</c> is
+    /// what makes an operation about half the list.
+    /// </summary>
+    private const string Loop = $$"""
+        (function () {
+            var index = -1;
+            for (var i = 0; i < {{Passes}}; i++) {
+                for (var j = 0; j < list.length; j++) {
+                    if (list[j] === el) { index = j; break; }
+                }
+            }
+            return index;
+        })()
+        """;
+
+    private Browser.Browser _browser = null!;
+    private Page _staticList = null!;
+    private Page _staticListTampered = null!;
+    private Page _liveList = null!;
+    private Page _plainArray = null!;
+
+    [GlobalSetup]
+    public async Task Setup()
+    {
+        _browser = new Browser.Browser();
+
+        _staticList = await CreatePageAsync("var list = document.getElementById('root').querySelectorAll('span');");
+
+        _staticListTampered = await CreatePageAsync(
+            "var list = document.getElementById('root').querySelectorAll('span');"
+            + $"Object.defineProperty(list, 'length', {{ get: function () {{ return {Count}; }} }});");
+
+        _liveList = await CreatePageAsync("var list = document.getElementById('root').childNodes;");
+
+        _plainArray = await CreatePageAsync(
+            "var list = Array.from(document.getElementById('root').querySelectorAll('span'));");
+    }
+
+    /// <summary>
+    /// One page holding <paramref name="bind"/>'s <c>list</c> and the element the loop looks for, warmed with
+    /// this row's own loop and nothing else.
+    /// </summary>
+    private async Task<Page> CreatePageAsync(string bind)
+    {
+        var page = await _browser.NewPageAsync();
+        await page.SetContentAsync(Document(Count));
+        await page.EvaluateAsync<double>(bind + " var el = list[50]; 0;");
+        await page.EvaluateAsync<double>(Loop);
+        return page;
+    }
+
+    /// <summary>A document of <paramref name="count"/> spans under one parent, and nothing else.</summary>
+    private static string Document(int count)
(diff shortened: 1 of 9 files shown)


lahma:
## Measured, and it is not ready: the control moved

Idle box (the CPU gate held under 12 % with nothing building for three consecutive samples), default BenchmarkDotNet job, paired and interleaved **A B B A**, one process per case. The two arms are this branch and **its own parent commit** (`737efc745`) rather than current `main`, with `BrowserNodeListBenchmark.cs` copied untracked into the base arm — the benchmark ships inside the commit under test, so the arms differ by exactly the cache.

| row | base | with the cache | delta | the two pairs |
| --- | ---: | ---: | ---: | --- |
| `StaticNodeList` | 720.1 µs | 656.6 µs | **−8.8 %** | −10.1 %, −7.5 % |
| `StaticNodeListTamperedLength` | 1,077.3 µs | 1,030.5 µs | −4.4 % | −5.2 %, −3.4 % |
| `LiveNodeList` (control) | 715.5 µs | 751.0 µs | **+5.0 %** | +5.9 %, +4.0 % |
| `PlainArray` (floor) | 678.0 µs | 666.1 µs | −1.8 % | −2.7 %, −0.7 % |

Allocation did not move: every row reports 13.05 KB in both arms.

**Read it against the floor rather than against the clock.** Same-binary drift on this box was −4.5 % / −1.7 % (`StaticNodeList`) and −5.3 % / −3.4 % (`PlainArray`) between the two runs of one build, so a single row's absolute delta is barely above its own noise. The ratio each run measures *within itself* is not:

| | base | with the cache |
| --- | ---: | ---: |
| `StaticNodeList` ÷ `PlainArray` | 1.058, 1.067 | 0.977, 0.994 |
| `StaticNodeListTamperedLength` ÷ `PlainArray` | 1.594, 1.583 | 1.553, 1.540 |

So the claim the change was written to make holds: **a repeated indexed read of a static `NodeList` went from about 6 % above the plain-array floor to level with it**, in both runs, and the tampered-length row improves less because a JavaScript `length` getter per iteration dominates it — which is what the profile on #4013 said too.

**And the control regressed.** `LiveNodeList` is the row that must not move, it is the *least* noisy row in the set (same-binary spread +0.3 % / −1.4 %), and both cross-arm pairs put it about 5 % slower. That is a real signal, not drift, and it needs a mechanism before this can leave draft. The hypothesis to test first: `DomStaticNodeListObject` is a second `DomCollectionBase` implementation, so the `TryGetIndex` / `Length` call sites the live lane goes through may have stopped being monomorphic. If that is it, the fix is a shape question rather than a cache question.

Nothing in the diff changes; the numbers do not license the merge yet. Staying in draft.

lahma:
Landed as part of #4037 (`f944418b0`), which carries this commit plus the shape fix on top of it, so this draft is closed rather than merged. Nothing here is lost.

What the two measurements said, for the record: the cache does what it was written to do — `StaticNodeList ÷ PlainArray` fell from 1.042 to 0.971 with the cache and 0.964 with the fix, in two independent sessions, so a repeated indexed read of a static `NodeList` now sits at the plain-array floor instead of above it. The control question that kept this in draft — a 5 % move on `LiveNodeList` in the first session — did not reproduce in the second, whose own same-binary spread on that row was too wide to resolve it either way. #4037 removes the mechanism that could have caused it regardless, at a cost of one field load, a test and a not-taken branch on the live read.

The remaining follow-ups are on #4013: `NodeList.prototype.item(i)` still goes through the weak table rather than the memo, and live collections are deliberately uncached because caching one needs a membership version to key on.

lahma:
Superseded by #4037, which contains this commit.

PR #4037, opened 2026-09-09
Keep the static NodeList's memo inside the wrapper the live ones use

lahma:
**Supersedes the branch behind #4027**, which stays in draft: this is #4027's own commit, unmodified, plus **one** commit on top that fixes the control-row regression the lead measured on it. Take the branch whole, or fold `fe377bd` into #4027 — the fix is exactly `git diff perf/static-nodelist-wrapper-cache..perf/live-lane-shape`, six files, and touches nothing #4027 does to `querySelectorAll`.

The lead's table on #4027: `StaticNodeList` −8.8 % and level with the in-run `PlainArray` floor (1.058/1.067 → 0.977/0.994), and `LiveNodeList` — the control — **+5.0 %**, both pairs agreeing (+5.9 %, +4.0 %) on the least noisy row in the set.

## The mechanism, and how it was settled

**The hypothesis was right, and it is a devirtualization question rather than a virtual-call-count one.**

Every `list[i]` a page makes arrives at the interpreter as one virtual `ArrayLikeObject.TryGetIndex`, at a single call site inside `JintMemberExpression.GetValue` (`Jint/Runtime/Interpreter/Expressions/JintMemberExpression.cs`, the host array-like arm of the computed-read lane). That site is shared by every engine in the process, and the JIT devirtualizes and inlines it from a **class profile holding one guess**. Before #4027 a live `NodeList` and a selector match were both `DomCollectionObject`, so the guess could not be wrong. #4027 gives the selector match a second wrapper, and the two `NodeList` producers then compete for that one guess.

Read by disassembly, not by reasoning. `DOTNET_JitDisasm=Jint.Runtime.Interpreter.Expressions.JintMemberExpression:GetValue` plus `DOTNET_JitDisasmDiffable=1` and `DOTNET_JitDisasmSummary=1`, on a throwaway harness that replays `BrowserNodeListBenchmark`'s four pages in its own construction order (static, tampered, live, array) and then hammers the live loop — a probe, no BenchmarkDotNet and no profiler. Arms: this branch's `4f36c51` (= #4027) and its parent.

**What the second class does when the static list warms the site first** (`Jint.dll` is byte-identical between the arms — #4027 touches only `Jint.Browser` — so every difference below is the JIT's):

| | Tier1 `GetValue`, the array-like read |
| --- | --- |
| parent commit | `cmp qword ptr [r15], <DomCollectionObject>` → inlined body → `call [DomAccessorNodeList:TryGetIndex]`, in the **hot** block |
| #4027 | `cmp qword ptr [r15], <DomStaticNodeListObject>` → inlined body → `call [DomRealm:WrapNode]`; the live collection falls through to a bare `call [rax+0x28] ArrayLikeObject:TryGetIndex` in a **cold** block, with nothing devirtualized and nothing inlined |

Two more states of the same site, same probe:

* **Mixed profile** (forced with `DOTNET_TC_CallCountingDelayMs=0`, which puts instrumented tier-0 live across the four warm-ups): neither candidate clears the likelihood threshold, the JIT emits the bare `call [rax+0x28] ArrayLikeObject:TryGetIndex` and **both** lanes lose the inline. On the parent commit under the same knob, the site is monomorphic and keeps its guess.
* **Live-dominated profile** (the default tiering delay in this probe, where instrumentation happens to start after the four pages are built): both arms produce **byte-identical** Tier1 code for `GetValue`, and for `ArrayLikeObject.ReadIndex`, `ArrayLikeObject.TryReadLength`, `JintBinaryExpression+NumericConstantComparisonLane.TryReadLengthBound`, `DomCollectionObject.TryGetIndex`, `DomRealm.WrapNode` and `DomRealm.WrapNodeValue`.

**So this needs stating plainly: I could not catch the flip happening in the probe's own timing.** In the ordering the probe reproduces, tiering promotes `GetValue` after the four pages are built and the guess lands on the live wrapper in both arms. What #4027 changes is not that the live lane *is* deoptimized in every process — it is that the live lane's codegen **became a function of which list warmed the shared site first and when tiering fired**, where before it was fixed by construction. A lost inline plus a lost devirtualization on ~5,100 element reads per operation is the right order of magnitude for the +5 % the lead measured (≈ 7 ns per read), and it is the only mechanism found that both is specific to the live lane and is deterministic per binary.

### What was eliminated

* **The benchmark's own fixture.** `BrowserNodeListBenchmark.cs` is the same file in both arms (copied untracked into the base), and the `LiveNodeList` page never calls `querySelectorAll` — its `list` is `#root.childNodes`, 100 spans, and its `el` is `list[50]`, in both arms. The cross-contamination is between the *four pages in one process*, not inside the row.
* **`DomHostHooks`' vtable.** `QuerySelectorAll` is declared after every other virtual in the class, so no existing slot moves; and nothing on the live element read reaches a hook.
* **Allocation and GC.** The lead's own table: 13.05 KB on every row in both arms. The two extra arrays per static list are in the `LiveNodeList` process *and* in the `PlainArray` process, and `PlainArray` did not move.
* **Anything below the read.** `DomAccessorNodeList.TryGetIndex`'s Tier1 body differs between arms by register allocation only (`r13` used, 48-byte frame, 691 → 701 bytes) — and two runs of the *same* binary differ in **73** Tier1 method bodies where the two arms differ in 65, so summary-level codegen churn of that size is run-to-run PGO noise, not a property of the binary.

## The fix

The smallest shape that removes the mechanism is to not add the second class.

`DomStaticNodeList` stays exactly as #4027 wrote it — staticness is still a property of a *type*, constructed only by the `querySelectorAll` hook, which is where DOM §4.2.6's word "static" is. What goes is `DomStaticNodeListObject`. The memo becomes a branch inside `DomCollectionObject`, entered from a `readonly INode[]? _nodes` field that holds the snapshot's nodes for a static list and is `null` for every other target. The element read is #4027's body verbatim, moved to `TryGetStaticIndex`, marked `NoInlining` so that what the interpreter inlines on every `list[i]` stays the size it is today.

Why this one, against the other candidates:

* **Sealing** — `DomCollectionObject` is already sealed and it does not help: the call site's static type is `ArrayLikeObject`, so the dispatch is guarded-devirtualized either way.
* **A per-collection `DomCollectionAccessor`** carrying the memo — moves the same problem one level down onto `_accessor.TryGetIndex`, which the live lane also goes through.
* **A non-virtual read path for the static shape** — there is none to give it; the engine dispatches virtually by construction.
* **A field on `DomStaticNodeList` instead of on the wrapper** — one field cheaper, but it puts per-engine objects on the AngleSharp-side type and weakens #4027's own retention argument for a byte count. Not worth it.

`Length` takes the same test, so a static list still answers its own count without the accessor and its interface call.

**What the live lane pays, exactly**, read off the two Tier1 listings side by side:

```
  parent commit                          this branch
  cmp  [r15], <DomCollectionObject>      cmp  [r15], <DomCollectionObject>
  jne  <cold>                            jne  <cold>
                                         mov  rdx, [r15+0x70]     ; _nodes
                                         test rdx, rdx
                                         jne  <static block>
  mov  rcx, [r15+0x60]   ; _accessor     mov  rcx, [r15+0x60]
  ...                                    ...
  call [DomAccessorNodeList:TryGetIndex] call [DomAccessorNodeList:TryGetIndex]
```

Three instructions on the element read — one field load off a line already being read, a `test`, one not-taken `jcc` — and two on the `length` read (`cmp gword ptr [rbx+0x70], 0` and one not-taken `jcc`, inside `ArrayLikeObject.TryReadLength`, whose Tier1 body goes 384 → 407 bytes). Nothing else in the live lane's codegen moves.

And the property that was lost comes back: with the fix, **both** orderings — static-first and the benchmark's own — produce a Tier1 `GetValue` whose hot block is the inlined `DomCollectionObject.TryGetIndex`. Static-first, which on #4027 left the live lane on a cold bare virtual call, now keeps it inlined and merely takes the not-taken branch.

## What I want measured, and why

The same four rows, same protocol (idle box, default job, paired **A B B A**, one process per case), **three** arms rather than two, because two cannot tell "the control came back" from "the control never moved in this process":

1. `737efc745` — #4027's parent, benchmark copied untracked. The base the lead already has.
2. `4f36c51` — #4027 as it stands (this branch's first commit).
3. `fe377bd` — this branch's head.

Read arm 3 against arm 1 on **two** questions, both as ratios inside each run rather than as clocks:

* **`LiveNodeList ÷ PlainArray` must return to arm 1's value.** That is the whole deliverable. Arm 1 measured 715.5/678.0 = 1.055; arm 2 measured 751.0/666.1 = 1.127. Arm 3 should read 1.055 ± the row's own spread, and the +5 % should be gone. If it is *not* gone, the answer is that the cause was never the devirtualization and the next suspect is native code layout — which would be consistent with everything above, is deterministic per binary, and would make the two-instruction cost this branch adds the only thing it bought.
* **`StaticNodeList ÷ PlainArray` must stay below 1.0.** Arm 2 measured 0.977/0.994 against arm 1's 1.058/1.067. Arm 3 keeps #4027's cache and adds one non-inlined call per cached read, so it should land in the same neighbourhood; a meaningful walk back toward 1.05 would say the `NoInlining` on `TryGetStaticIndex` is too conservative and should come off.

`StaticNodeListTamperedLength` and `PlainArray` are along for the ride; the tampered row is still dominated by a JavaScript `length` getter per iteration.

No benchmark and no profiler was run for this branch. The evidence above is disassembly and reading.

## Tests

`Jint.Tests.Browser`, green on **net8.0 and net10.0**, and again under **`JINT_HOST_CONTRACT_VERIFICATION=1`** on each: 2,754 / 2,750 passing, 0 failed. Full solution builds clean in Release with no C# warnings.

`OnlyASelectorMatchTakesTheStaticLane` is rewritten to pin both halves of the seam, since the CLR type is no longer what separates them: the **target** of a selector match is a `DomStaticNodeList` and `childNodes`/`labels`' targets are not, and all five collections are the one wrapper class — which is now itself the thing under test, with the reason in the test's own remarks. It is still the one test in the file that fails against the unfixed code: verified by pointing the hook back at `Realm.Wrap`, rebuilding and re-running — it fails, the other ten pass.

## Doubts

* The strongest one is above and stated in the open: the probe reproduces the benchmark's page order but not, as far as I could observe, the tiering moment that makes the guess flip. The fix removes the mechanism whether or not that moment is what the lead's box hit; the re-measurement is what settles it.
* `DomCollectionObject` grows by two reference fields (16 bytes) for every collection wrapper on a page, to serve one collection shape. Cheaper than the alternative that put them on the snapshot, and cheaper than the class it deletes, but it is a cost paid by `classList`, `attributes` and `childNodes` too.
* `NoInlining` on `TryGetStaticIndex` is deliberately conservative — it protects the control row at the cost of one direct call per cached read on the static row. If the measurement says the static row gave back more than a percent, that attribute is the first thing to try removing.
* A static list's `NodeList.prototype.item(i)` still goes through `DomRealm.WrapNodeValue` and the weak table rather than the memo, exactly as on #4027. Only the indexed getter is memoized.
* `MutationRecord.addedNodes` / `removedNodes` are static per DOM too and are still not covered; they keep the accessor-driven read. Adding them is now a one-line hook plus `WrapStaticNodeList`, with no new type.

Refs #4013 and #4027.

https://claude.ai/code/session_01WKytThti6mniJepuE8P691


lahma:
## Three arms, and the control effect does not reproduce

Same protocol as the first session, with the third arm this branch's report asked for: idle-gated box, default job, one process per case, six runs interleaved in both orientations (**A B C C B A**). All three arms share the base `3d99378f2`, so they differ only by the two commits.

- **A** = `3d99378f2`, the branch's own base, with `BrowserNodeListBenchmark.cs` copied untracked
- **B** = `4f36c5119`, the cache as #4027 wrote it
- **C** = `fe377bd24`, this PR's head, the memo moved inside `DomCollectionObject`

Ratios against the `PlainArray` floor measured **inside each run**, the two runs of each arm side by side:

| row | A1 | A2 | B1 | B2 | C1 | C2 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `StaticNodeList` | 1.065 | 1.019 | 0.961 | 0.981 | 0.954 | 0.974 |
| `StaticNodeListTamperedLength` | 1.804 | 1.554 | 1.493 | 1.461 | 1.476 | 1.497 |
| `LiveNodeList` | 1.065 | 1.048 | 1.105 | 1.029 | 1.087 | 1.138 |

**The static win is real and survives the fix.** Mean ratio `StaticNodeList ÷ PlainArray`: **A 1.042, B 0.971, C 0.964**. That is the second session to put the cached static list at or below the plain-array floor and the uncached one above it, and C is not worse than B, so `NoInlining` on `TryGetStaticIndex` costs nothing measurable. The tampered row agrees: A 1.679, B 1.477, C 1.486.

**The control effect is unresolved, and this session cannot resolve it.** Look at the same-binary spread — two runs of *one* build:

| row | A | B | C |
| --- | ---: | ---: | ---: |
| `StaticNodeList` | −1.8 % | +7.6 % | +0.9 % |
| `StaticNodeListTamperedLength` | −11.6 % | +3.1 % | +0.2 % |
| `LiveNodeList` | +1.0 % | −1.9 % | +3.5 % |
| `PlainArray` | +2.6 % | +5.4 % | −1.2 % |

`LiveNodeList`'s own ratio swings 1.105 → 1.029 within arm B and 1.087 → 1.138 within arm C. A row whose two same-binary runs disagree by 5 to 7 points cannot answer a 5-point question. Taken at face value the means say B +1.0 % and C +5.3 % against A, which would mean the regression vanished and the *fix* caused one — and the disassembly says the fix adds a load, a test and a not-taken branch. So the honest reading is that this session measured noise on that row, not that the first session was wrong and not that it was right.

**What that means for the two PRs.** Both stay in draft. The cache is worth having on the evidence of two sessions; the control question needs a quieter box than either of these, and the first session's tight `LiveNodeList` spread (+0.3 % / −1.4 %) shows such a box exists.

One caveat I should have stated earlier, since it bounds the method rather than the result: BenchmarkDotNet gives each row its own process, so a ratio taken between two rows of one run cancels machine-wide drift but not per-process variance. It is still the best reading available here, and it is why the argument rests on the sign agreeing across four independent runs rather than on any single number.

The shape argument for this PR stands on its own regardless: one wrapper class per interface family, because the interpreter has a single array-like call site whose class profile holds one guess. That is worth landing whether or not the 5 % was real, and `Jint.Browser/Dom/AGENTS.md` now says so.

PR #4046, opened 2026-09-12
Live DOM collections: one walk per read, and no allocation

lahma:
## What the profile said

A workload looping `for (i = 0; i < c.length; i++) c[i]` over four live collections —
`getElementsByClassName`, `getElementsByTagName`, `form.elements`, `children` — on a 1,500-element
document, profiled under ultra at 8190 Hz on `net10.0`:

`AngleSharp.Dom.NodeExtensions.GetDescendantsAndSelf.MoveNext` is **33.6 % inclusive** of the page-loop
thread, and the inverted call tree splits its callers three ways — three whole document walks per loop
iteration:

| caller | share of that subtree | what it is |
| --- | ---: | --- |
| `JintMemberExpression.GetValue` (direct) | 36.3 % | the `_collection.Length` **bounds pre-check** inside `DomHtmlCollectionObject.TryGetIndex` |
| `NumericConstantComparisonLane.TryReadLengthBound` | 35.3 % | the loop's own `i < c.length` |
| `DomLiveHtmlCollection.get_Item` | 28.3 % | the actual `c[i]` |

Around it, `CastHelpers.IsInstanceOfAny` + `StelemRef_Helper` + `IsInstanceOfInterface` +
`ChkCastInterface` were 19.3 % self — `StelemRef_Helper` alone 6.1 %, which is the array-covariance store
check on `Stack<INode>.Push` inside `GetDescendantsAndSelf`. That is the exact cost AngleSharp's own
internal `ElementTreeEnumerator` was written to avoid, and which `GetDescendantsAndSelf` does not use.

## The mechanism

**Nothing here memoizes.** These collections are live by specification and every read still re-runs the
filter against whatever the tree now is. What changes is that a read is **one** walk rather than two or
three, and that the walk **allocates nothing**. A generation-keyed memo is a different change that depends
on an AngleSharp change landing first, and it is not this one.

### 1. Stop asking for `Length` before indexing

`DomHtmlCollectionObject.TryGetIndex` read `_collection.Length` for a bounds check and then
`_collection[index]`. Every `IHtmlCollection<T>` that wrapper is handed is a **lazy view over a tree walk**,
not a list:

* AngleSharp's `HtmlCollection<T>` (`children`, and the snapshot collections) holds an `IEnumerable<T>`
  whose `Length` is `_elements.Count()` and whose indexer is `GetItemByIndex`, a linear scan;
* its `HtmlFormControlsCollection` (`form.elements`) is a `Where` over the form-control descendants of the
  **whole document**, re-run per read;
* the binding's own `DomLiveHtmlCollection` re-runs its filter.

So the pre-check ran the entire query a second time on every element read. It is gone: the index walk
already knows when it ran out, and running out **is** the bounds answer. `HasIndex` follows, and
`DomHtmlAllCollectionObject` (`document.all`, also a lazy view) gets the same treatment.

### 2. Make the walk itself cheap

`DomElementWalker` is a **struct** depth-first pre-order walk over the public `INode` surface —
`ChildNodes`, its indexer, and `Parent` — with an index stack whose first 16 levels live in an
`[InlineArray]` inside the struct. It is the design AngleSharp's internal `ElementTreeEnumerator` arrived
at, and its comment gives the reason: an index stack stores value types, pays no covariance check, and is
bounded by the tree's *depth* rather than its *width*. That type is internal to AngleSharp and cannot be
referenced, so this is the same shape written against the public surface, and rooted at a node that need
not itself be an element (these queries are called on documents and fragments too).

It replaces `root.Descendants<IElement>().Where(…)`, which per read was a `Stack<INode>` plus its backing
array, a covariance store per push, three chained iterators (`GetDescendantsAndSelf`, `Skip(1)`, `OfType`)
plus the `Where` — four virtual `MoveNext` calls and an interface type test per node — and a delegate
invocation per element.

`DomElementFilter` is the membership test as an object with a method rather than a `Func<IElement, bool>`.
That is not tidiness: the predicate depends on something read from the tree *at the moment of the read* —
`getElementsByClassName`'s comparison is ASCII case-insensitive exactly while the root's node document is
in quirks mode, and a root adopted into another document takes that document's mode with it — so a lambda
was a closure allocated per read. `BeginRead` refreshes that state once per read instead of once per
element, which is the same answer the closure gave.

### 3. Stop the named lane building three lists per read

`DomCollectionAccessor` gains `HasSupportedName(target, name)` — the membership half of `SupportedNames`,
which the generated accessors now implement as a scan. It is on the **read** path: `el.attributes.foo` asks
WebIDL's named-property-visibility question before the named getter runs, and answering it meant
materializing every attribute name into a `List<string>`, then a second filtered list, then scanning that
with the `Contains` overload taking a comparer — which is LINQ's, so it allocated an enumerator per
candidate. Three allocations and a walk of the whole map for a question about one member.

Two more things in the same area:

* the generated `SupportedNames` re-evaluated `collection.Length` **in the loop condition** and indexed
  inside the loop. For `HTMLFormElement` that `Length` is its form-control collection's, a `Count()` over a
  filtered walk of the document — so listing *n* names ran that query *n* times. The count is now read once
  into a local.
* the duplicate check in both `VisibleNames` (`HTMLCollection` and `HTMLAllCollection`) is a
  `HashSet<string>` rather than the same LINQ `Contains`: O(n²) comparisons and 2n allocations became O(n)
  and one.

## A spec fix that came with it

`HasEveryClass` read `element.GetAttribute("class")`. DOM's
[classes](https://dom.spec.whatwg.org/#concept-class) are `classList`'s token set, and DOM §7.1 reads that
attribute by *"getting an attribute value given null namespace and the local name `class`"* —
whereas [`getAttribute`](https://dom.spec.whatwg.org/#dom-element-getattribute) matches a **qualified**
name. So an attribute put in a namespace under the spelling `class` was one of the element's classes as far
as this algorithm was concerned, while `className` and `classList` both said it was not, and an element
carrying both would have answered the namespaced one's value. It now reads
`element.GetAttribute(null, AttributeNames.Class)`, which is what the standard asks for and what every
other reader of an element's classes already used.

`ClassNameCollectionTests.NamespacedAttributeSpelledClassIsNotOneOfTheElementsClasses` is the test.
**It fails on the old lookup** — and only on its last assertion: `getAttribute('class')` finding the
attribute, `className` being empty and `classList.length` being 0 all passed before the fix, which is
precisely the disagreement. The vendored wpt corpus does not cover the distinction (no
`getElementsByClassName` document uses `setAttributeNS`), so the test is ours.

It is also the cheaper read — the qualified form ASCII-folds the name it is given on every call for an
element in the HTML namespace — but correctness is why it changed.

## What did not change, and why

* **No cache and no invalidation signal.** Live means live; each read re-queries.
* **No second wrapper class.** `Jint.Browser/Dom/AGENTS.md` records why
  ([#4027](https://github.com/sebastienros/jint/pull/4027)): every `list[i]` reaches the interpreter as one
  virtual `ArrayLikeObject.TryGetIndex` devirtualized from a class profile holding one guess. The different
  read strategy is a field and a branch inside the class that already serves the interface, the same shape
  `DomCollectionObject` uses for the static `NodeList`.
* **The generated `TryGetIndex` keeps its `Length` pre-check.** It was checked. The interfaces that reach
  it are either indexable in constant time (`INodeList`, `INamedNodeMap`, `ITokenList`, `ICssRuleList` …),
  where a length probe is a field read and the cheapest possible bounds answer, or — `IHtmlFormElement`,
  `IHtmlSelectElement` — walk-backed but **not enumerable at all** on their public surface, so there is no
  one-pass form to emit. A blanket "enumerate instead" rule would turn `childNodes[i]` from O(1) into
  O(i), which is the `LiveNodeList` control row below. The walk-backed collections all arrive at the
  hand-written `DomHtmlCollectionObject`, which is where the change went.
* `Jint.Browser/Layout/**`, `Dom/Views/CssCascade.cs`, `ReadOnlyStyleDeclaration.cs`, `ResolvedStyle.cs`
  and `Accessibility/**` are untouched.

## Tests

* `dotnet build -c Release` — clean, `TreatWarningsAsErrors` on.
* `dotnet test -c Release Jint.Tests.Browser/Jint.Tests.Browser.csproj` — 2852 / 2849 passed (net10.0 /
  net8.0), 0 failed, 38 skipped. That includes the web-platform-tests browser lane and its census, and the
  generator round-trip (`DomBindingsStalenessTests`). **The wpt census figures did not move**: no exclusion
  or census file changed, and the pass count rose by exactly the one test added here.
* `dotnet test -c Release Jint.Tests/Jint.Tests.csproj` — 12426 / 12426 / 8516 passed (net10.0 / net8.0 /
  net472), 0 failed.
* The generated file was regenerated with `JINT_DOM_BINDINGS=update`; no `.g.cs` was hand-edited.

Failing before the fix: `NamespacedAttributeSpelledClassIsNotOneOfTheElementsClasses`, as described above.

## To be measured

`Jint.Benchmark/BrowserHtmlCollectionBenchmark` is new — the suite had **no** row for an `HTMLCollection`,
for `getElementsBy*`, for `form.elements` or for `children`. One `Page` per row built in `[GlobalSetup]`
and warmed with only its own script, `[MemoryDiagnoser]`, a `childNodes` control row the change does not
touch, and an `Array.from` floor as `[Benchmark(Baseline = true)]`.

```
dotnet run -c Release --project Jint.Benchmark\Jint.Benchmark.csproj -- --filter "*BrowserHtmlCollectionBenchmark*"
```

`ClassName`, `TagName`, `FormElements` and `Children` are the rows this change is about;
`LiveNodeList` is the control that must not move, and `PlainArray` is the floor.

**No figure is quoted here.** The rows were executed once on a contended machine purely to prove they run
end to end; those numbers are not reportable and are not repeated.

### DIFF
--- a/Jint.Benchmark/BrowserHtmlCollectionBenchmark.cs
+++ b/Jint.Benchmark/BrowserHtmlCollectionBenchmark.cs
@@ -43,13 +43,13 @@ namespace Jint.Benchmark;
 /// children. Shallow, so it isolates the second walk from the cost of the walk itself.
 /// </description></item>
 /// <item><description>
-/// <see cref="LiveNodeList"/> — the identical loop over <c>childNodes</c>, whose elements are the same
-/// hundred nodes. It is a <c>NodeList</c>, so it reaches the generated accessor and a constant-time indexer,
-/// and nothing in this change touches that lane. <b>The control that must not move.</b>
+/// <see cref="LiveNodeList"/> — the same loop over <c>childNodes</c>, whose elements are the same hundred
+/// nodes. It is a <c>NodeList</c>, so it reaches the generated accessor and a constant-time indexer, and
+/// nothing in this change touches that lane. <b>The control that must not move.</b>
 /// </description></item>
 /// <item><description>
-/// <see cref="PlainArray"/> — the floor and the baseline: the identical loop over the <c>Array.from</c> of
-/// the same elements, so the gap is the price of the collection being a live projection.
+/// <see cref="PlainArray"/> — the floor and the baseline: the same loop over the <c>Array.from</c> of the
+/// same elements, so what a row has above it is the price of the collection being a live projection.
 /// </description></item>
 /// </list>
 /// <para>
@@ -59,8 +59,29 @@ namespace Jint.Benchmark;
 /// drives the same <c>ArrayLikeObject.TryGetIndex</c> call site, whose class profile holds one guess, and a
 /// shared engine would hand whichever row ran first a monomorphic read and the rest a cold indirect call.
 /// Page construction and the HTML parse stay outside the measurement; what the measured call adds over the
-/// loop itself is one mailbox round trip per operation, which is the same term in every row including the
-/// baseline.
+/// loop itself is one mailbox round trip per operation.
+/// </para>
+/// <para>
+/// <b>The pass counts differ per row, deliberately, and must not be tidied into one constant.</b> That
+/// round trip is an <i>additive</i> per-invocation cost with its own scheduling jitter, so it does not
+/// cancel against the baseline the way a multiplicative one does — see <b>"A row through
+/// <c>Page.EvaluateAsync</c> must amortise the mailbox round trip"</b> in
+/// <a href="AGENTS.md"><c>Jint.Benchmark/AGENTS.md</c></a>, which carries the rule and the paired run that
+/// established it. The four target rows walk the document and cost milliseconds at ten passes, so ten is
+/// all they need. The two <i>control</i> rows are the ones the rule bites: <c>childNodes</c> is indexed in
+/// constant time and a plain array more so, so at ten passes each was a few microseconds of work behind a
+/// round trip an order of magnitude larger — a row that could not have detected a regression in what it
+/// controls for, and that measured as multimodal (<c>MValue</c> 3.56 on <see cref="PlainArray"/>) because
+/// what it was mostly reporting was thread scheduling. Their counts are set so that every row is at least
+/// in the low milliseconds and the round trip is a per-cent-level term.
+/// </para>
+/// <para>
+/// <b>What that costs, and it is the one trap here: a row is comparable to itself across builds, and to no
+/// other row.</b> Because the counts differ, <c>Ratio</c> is not the price of a live projection per read —
+/// it is that price times a pass-count ratio. The counts are chosen so the six means land within about
+/// 1.7–2.5 ms of each other, which keeps the table readable, but that near-equality is arranged rather than
+/// measured and means nothing on its own. Read each row against the same row on the other build; a paired
+/// run (<c>measure-paired.ps1</c>) does exactly that and is the right instrument for this class.
 /// </para>
 /// </remarks>
 [MemoryDiagnoser]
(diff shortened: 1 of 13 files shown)


lahma:
## Measured

Paired A/B, `Jint.Benchmark/measure-paired.ps1`, **3 rounds with the order alternating every round**, default job, one `Page` per row, on an idle machine (CPU verified under 5 % before the run, nothing else building). Baseline is this branch's own merge base with the benchmark file copied in untracked, so the two arms differ by exactly the change.

| row | median Δ | 95 % CI | |
| --- | ---: | --- | --- |
| `ClassName` | **−74.3 %** | [−76.7, −74.0] | faster |
| `TagName` | **−65.7 %** | [−68.2, −59.7] | faster |
| `Children` | **−39.8 %** | [−41.4, −39.8] | faster |
| `FormElements` | **−35.5 %** | [−38.8, −34.8] | faster |
| `LiveNodeList` *(control)* | −1.9 % | [−3.4, +2.0] | no change |
| `PlainArray` *(floor)* | +0.6 % | [−4.4, +7.0] | no change |

Allocation, which is deterministic and so quoted from a single round:

| row | before | after | |
| --- | ---: | ---: | --- |
| `ClassName` | 3,862.05 KB | **36.68 KB** | −99.0 % |
| `TagName` | 3,682.44 KB | **12.77 KB** | −99.7 % |
| `FormElements` | 339.52 KB | 239.88 KB | −29.3 % |
| `Children` | 12.81 KB | 12.78 KB | — |
| `LiveNodeList` *(control)* | 12.80 KB | 12.81 KB | — |
| `PlainArray` *(floor)* | 12.80 KB | 12.81 KB | — |

The allocation column is the clearest statement of what changed: a read used to allocate a `Stack<INode>`, a `Skip`, an `OfType` and a `Where` per query, so a loop over a live collection allocated megabytes. `DomElementWalker` allocates nothing, and what is left on `ClassName`/`TagName` is the element wrappers themselves.

`FormElements` keeps most of its allocation because the remaining churn is AngleSharp's — `HtmlFormControlsCollection` is a captured LINQ query re-executed per member access. That is upstream's to fix and partly already is: [AngleSharp#1345](https://github.com/AngleSharp/AngleSharp/pull/1345) (merged) removes the double enumeration behind its named lookup.

## How to read the table, and one thing it cannot tell you

**The `Ratio` column BenchmarkDotNet prints is meaningless across rows in this class**, and the table above deliberately does not use it. Pass counts differ per row — chosen so one mailbox round trip is amortised rather than measured — so a cross-row ratio is the per-read price multiplied by a pass-count ratio. Every figure above is a row compared **to itself** across the two builds. The six means cluster in 1.7–2.5 ms by arrangement, not by measurement.

That per-row sizing was added during review, and it is why the control rows can be believed. Before it, every row ran one query per `Page.EvaluateAsync`, and `EvaluateAsync` is a mailbox request to the page loop thread — an **additive** ~60–70 µs constant with its own scheduling jitter, which does not cancel against a baseline row the way multiplicative drift does. Measured on the sibling selector benchmark, that made rows whose code the change could not touch read +19.5 % and +18.6 %. `PlainArray` here was 78.78 µs with **MValue 3.556** — multimodal — for perhaps 5–15 µs of actual script. The rule now lives in `Jint.Benchmark/AGENTS.md`.

Merging on the strength of this table plus nine green checks.

PR #4047, opened 2026-09-12
Invert :target's per-candidate document resolution to an O(1) filter

lahma:
## What the profile said

Issue #4013's capture — `querySelectorAll` for `:target`, `:default`, `:indeterminate` and `:valid` over a 2,000-element document, 60 rounds, a document URL whose fragment names no element — put `:target` at roughly 77% of it:

| leaf | self |
| --- | ---: |
| `PagePseudoClassSelectorFactory+TargetSelector.Find` | 33.2 % |
| `AngleSharp.Dom.Element.get_NextElementSibling` | 22.4 % |
| `CollectionExtensions.GetElementById` | 8.5 % |
| `Element.get_Id` | 5.9 % |
| `CastHelpers.IsInstanceOfInterface`/`ChkCastInterface` | 7.3 % |
| `TargetSelector.Match` | 2.0 % |
| **`UrlParser.PathState`** | **0.1 %** |

The issue's own description — that the cost is `document.Url` being re-parsed with `UrlParser.Parse` on every element — is wrong. That parse is a tenth of a percent. The real cost is `TargetSelector.Find`: resolving ["the indicated part of the document"](https://html.spec.whatwg.org/multipage/browsing-the-web.html#the-indicated-part-of-the-document) — `document.GetElementById(fragment)`, and on a miss a scan of `document.All` for a legacy named anchor — ran **once per candidate element** `querySelectorAll` tests, not once per query. That turns an O(document size) resolution into an O(document size²) query, doubled again on a fragment whose percent-decoded form differs from the raw one.

## The mechanism, and what changed

`:target` matches at most one element in a document, so the fix inverts the test: instead of resolving the indicated element and then comparing every candidate to it, it asks an O(1) question about the candidate first, and only runs the document-wide resolution to confirm the (at most one, usually zero) candidate that passes:

- a candidate whose `Id` equals the fragment (or its percent-decoded form) can be the indicated element;
- a candidate that is an HTML `<a>` whose `name` attribute equals the fragment (raw or decoded) can be it;
- every other candidate cannot be, and is rejected with no document access at all.

`CouldBeIndicated` is exactly a necessary condition for `Find`'s two-phase resolution (id lookup before the legacy named-anchor scan, raw fragment before the percent-decoded one) to ever answer that candidate — `Find` itself is untouched, only called less often. On a document whose fragment names nothing (the profile's own worst case), this turns N document scans into zero; on one where it does, into one.

## What did not change, and why

- **`UrlParser.Parse(document.Url)` still runs once per candidate.** It's 0.1% of the capture; avoiding it would need caching the parsed URL across `Match` calls, which needs invalidation the moment a script changes `document.Url` — not worth contorting the code for, per the profile's own numbers.
- **`PagePseudoClassSelectorFactory`'s other three document/tree-wide predicates are left alone**, each investigated on its own evidence:
  - `DefaultSelector.IsTheDefaultButtonOfItsForm` walks the whole document tree per candidate submit button (via `HtmlFormOwner.ControlsOf`), because HTML's default-button algorithm is over the whole tree, not the form's own subtree — a control can be explicitly associated to a form via `form="id"` from anywhere in the document, ahead of the form itself in tree order. There's no O(1) necessary condition (unlike `:target`, a form can legitimately have many submit buttons, so "is a submit button" doesn't bound the match to one candidate) and no smaller-subtree walk that stays correct.
  - `IndeterminateSelector.TheRadioButtonGroupOfHasACheckedMember` walks the whole tree per unchecked, named radio button looking for a checked group member. A radio group's members can be anywhere in "the same tree," so there's no way to shrink the walk's scope, and every unchecked radio in a group re-asks the same question.
  - `ValidityStateSelector`'s fieldset arm re-walks a fieldset's whole subtree per query, so nested fieldsets multiply: an outer fieldset's evaluation subsumes an inner one's, and the inner one is independently re-evaluated when it is itself queried. A recursive, subtree-skipping walk was considered and rejected: it doesn't reduce the total node-visits summed across every fieldset's own evaluation (each still visits its own subtree once), it just moves where the visits happen.

  In all three cases, the only route to a real reduction is a memo — "does form F's tree have an owned submit button candidate whose position is P," "does group (owner, name) have a checked member," "does fieldset F's subtree have a failing candidate" — invalidated whenever a mutation could change the answer. That is out of scope for a pure cost change with no caching, per this stream's brief, so they are left as-is with this reasoning recorded here rather than a change without a benchmark behind it.

## Tests

- No test failed against the unfixed code — this is a pure cost change with unchanged behavior, verified by proof (`CouldBeIndicated` is exactly the necessary condition for `Find`'s two possible answers) and by the existing suite staying green, including `ChildFrameTests.AFramesLocationReadsItsOwnUrlAndRefusesAWriteOutLoud`'s `:target` case (duplicate IDs, a shadow-tree clone, a `DocumentFragment`-parented clone, a disconnected clone).
- Added `PagePseudoClassSelectorTests.TargetMatchesTheIndicatedElementInHtmlsOrderOrNothing`, pinning the id-lookup-first, then-legacy-anchor, then-percent-decoded ordering the new O(1) filter must stay a necessary condition for. It passes identically against the pre-change code (confirmed by temporarily reverting), since it pins semantics rather than a bug fix.
- `dotnet build -c Release` (whole solution): clean.
- `dotnet test -c Release Jint.Tests.Browser/Jint.Tests.Browser.csproj`, including the wpt browser lane, with `JINT_WPT_BROWSER_CENSUS=1` set: all green on both net8.0 and net10.0 (2853/2850 passed, 36 skipped each, the rest opt-in Playwright-driver cases), and the census figure did not move.

## To be measured

`Jint.Benchmark/BrowserSelectorBenchmark.cs` (new): one `Page` per row over a shared ~2,000-element document — `TargetMissing` (fragment naming no element, the profile's worst case), `TargetPresent` (fragment naming an element), `DefaultButton`, `Indeterminate`, `Valid`, and `Control` (a plain `.item` class selector this change touches nothing about, `[Benchmark(Baseline = true)]`).

Filter for the lead's gate run:

```
dotnet run -c Release --project Jint.Benchmark\Jint.Benchmark.csproj -- --filter "*BrowserSelectorBenchmark*"
```

I ran it once with `--job short` per the campaign brief, purely to confirm it executes (machine shared with other agents, so no number here is quotable): both the full class and an isolated `Control`-only run produced clean `WorkloadActual` iterations with sane, monotonic-looking numbers and no exceptions from the benchmark code itself — the two runs I let go were only cut short by my own outer wall-clock guard, not by a failure.


lahma:
Resized the benchmark per feedback from the paired run: every row ran its query once per `Page.EvaluateAsync` invocation, so the cheap rows (`Control`, `DefaultButton`, `Valid`, `TargetPresent`) were measuring the mailbox round trip to the page loop thread, not the selector — that round trip is additive and does not cancel against a baseline row, which is why untouched rows swung up to +19.5% between builds with identical code for them.

Pushed a second commit: each row now loops its query inside the script and sums every pass (so nothing is optimised away), with the pass count sized per row — 1 for the two rows already well above the round trip (`TargetMissing`, `Indeterminate`), 500 for the four that are cheaper than the round trip itself. Also added a rule to `Jint.Benchmark/AGENTS.md` ("A row through `Page.EvaluateAsync` must amortise the mailbox round trip") recording the measured evidence, since the existing `Passes` constants in the sibling browser benchmarks carried this reasoning nowhere written down.

lahma:
## Measured

Paired A/B, `measure-paired.ps1`, 3 rounds with order alternating, default job, one `Page` per row, idle machine. Baseline is this branch's merge base with the benchmark copied in untracked, so the arms differ by exactly the 43 lines of `TargetSelector`.

| row | before | after | median Δ | 95 % CI | |
| `TargetMissing` | 112.48 ms | **864.7 µs** | **−99.2 %** | [−99.23, −99.13] | faster |
| `TargetPresent` | 456.71 ms | 383.40 ms | **−16.1 %** | [−21.97, −15.62] | faster |
| `Indeterminate` *(untouched)* | 97.63 ms | 97.33 ms | −0.3 % | [−2.37, +17.37] | no change |
| `DefaultButton` *(untouched)* | 50.97 ms | 51.80 ms | +0.2 % | [−4.19, +1.63] | no change |
| `Control` *(untouched)* | 19.38 ms | 22.09 ms | +4.7 % | [+2.14, +13.99] | *see below* |
| `Valid` *(untouched)* | 49.43 ms | 58.97 ms | +17.3 % | [+14.85, +19.31] | *see below* |

`TargetMissing` is **130× faster**: a fragment naming no element used to run `getElementById` plus a full `document.All` scan once per candidate, and now runs the resolution at most once per query.

## The two rows that read slower are not a regression, and here is why rather than an assurance

`Control` and `Valid` exercise `.item` and `:valid` matching, which the diff never enters. Four independent things say the readings are measurement:

1. **Allocation is byte-identical between the arms on exactly those rows.** `Valid` 31,123.14 KB → 31,123.05 KB; `Control` 23,739.22 KB → 23,739.29 KB. Allocation is deterministic and does not share timing's noise, so identical bytes means identical work executed. A row doing more work would allocate differently.
2. **An A/A calibration puts the effect inside the floor.** Same worktree as *both* arms — identical binaries, identical layout — 3 rounds, `Valid` alone: median −3.8 %, **95 % CI [−4.36, +17.89]**, sign 1/3. The same binary compared against itself reaches +17.9 % on this row at this round count, so +10.7 % (isolated) and +17.3 % (in-class) are not distinguishable from noise.
3. **The other two untouched rows are flat.** `DefaultButton` +0.2 % and `Indeterminate` −0.3 %. A codegen or layout effect from adding a method to this file would not select only the two allocation-heavy rows.
4. **`Valid` and `Control` are the allocation-heavy rows** (31 MB and 24 MB per operation), which is where GC timing variance lives — and `Indeterminate`, at 14 KB, is the quietest row in the table and the flattest.

So this table proves the win and **does not** prove the absence of a sub-18 % regression on those two rows. I would rather state that than imply a clean sweep: `Valid` cannot serve as a control at this resolution, and making it one is a benchmark change, not a change to this PR. Its own 500-pass allocation of 384 MB on `TargetPresent` is a separate lead worth a look.

Isolated re-measure of `Valid` (own process, nothing else in the run): +10.74 % [+4.39, +15.41] — which is what sent me to the A/A rather than to the diff.

Merging on the two win rows, the identical-allocation evidence, and nine green checks.

PR #4048, opened 2026-09-12
The member-read inline cache can serve a holder deeper than the direct prototype

lahma:
The member-read inline cache could only remember a holder that was the receiver's **direct** prototype.
Anything deeper re-walked the prototype chain and probed each level with `GetOwnProperty`, on every read,
forever. This teaches the cache to span a chain, and proves the three facts such an entry has to keep proving.

## What the profile said

A workload of 16 M iterations of six plain DOM property reads (`el.nodeName`, `nodeType`, `tagName`,
`parentNode`, `firstChild`, `ownerDocument`), profiled under ultra at 8190 Hz on `net10.0`, steady state:

| leaf | self |
| --- | ---: |
| `ObjectInstance.Get` | 26.5 % |
| `ObjectInstance.UnwrapFromGetter` | 9.9 % |
| `JintMemberExpression.ReadAfterOwnMissUncached` | 9.2 % |
| `JintForStatement.TightForBody` | 8.7 % |
| `JintMemberExpression.GetValue` | 6.1 % |
| `TryEnsureSufficientExecutionStack` + `StackGuard.ProbeStackHeadroom` | 4.9 % |

`ObjectInstance.Get` is 28.8 % self / 60.1 % inclusive, and **107,572 of its 107,620 inclusive samples are
reached from `ReadAfterOwnMissUncached`** — 99.96 %.

## The mechanism

`ReadAfterOwnMissUncached` populated the cache only when the property was an own property of the receiver's
direct prototype; its own comment said "deeper chains fall to the slow `Get`". Every DOM member lives three to
five links up — `nodeName` and `parentNode` are on `Node.prototype`, read off an `HTMLParagraphElement` wrapper
through `HTMLElement.prototype` → `Element.prototype` → `Node.prototype` — so the cache never fired for any DOM
property read, and each read walked the chain doing a `GetOwnProperty` at every link.

**This is not a DOM problem.** Any `class C extends B extends A` reading a getter or method declared on
`A.prototype` from a `C` instance paid exactly the same.

## What changed

`JintMemberExpression` records, alongside the receiver/holder pair it already kept, the links **strictly
between** them — each with the `_propertiesVersion` it carried when the entry was created — in
`_cachedProtoChain`. It is `null` for a direct-prototype entry, which is still the common shape and still costs
what it cost: one predictable null test on top of what the check already did. A warm deep hit then validates
with a short loop of reference-plus-integer comparisons instead of N × `GetOwnProperty`.

### The correctness argument

An entry may serve a later read only if it can still prove all three of:

1. **The receiver has no own property of that name.** Unchanged: `_propertiesVersion` plus the `OrdinaryGet`
   exemption, which holds only because `ReadFromNonPlainReceiver` re-establishes a host receiver's own miss
   with a real probe (or a `false` from `TryGetOwnPropertyValue`) before every consult. Renamed to
   `CanCacheAgainstReceiverVersion`, since it is now the receiver's rule alone.
2. **No object between the receiver and the holder has gained an own property of that name.** The new
   obligation. Each intermediate link is re-proved on every hit by two facts: it is still the object that
   occupied that position — so nothing was re-pointed by `[[SetPrototypeOf]]`, inserted or removed — and its
   `_propertiesVersion` is unmoved. **The receiver's exemption does not extend to it:** nothing re-establishes
   anything about a link three levels up, so `ObjectInstance.VersionWitnessesOwnProperty` is required of every
   intermediate outright, exactly as it is of the holder. A host prototype therefore cannot sit in the middle
   of a cached chain any more than it can hold one. The predicate needed no change — what an intermediate has
   to keep proving is that the name is still *absent* from it, which is precisely the question it answers.
3. **The holder still owns it, with the same descriptor.** Unchanged.

I checked that the chain the profile was taken on actually clears all three, rather than assuming it: a
`Jint.Browser` element wrapper is a `DomNodeObject : JsEventTarget : ObjectInstance` built through the
*internal* constructor, so it carries neither `OrdinaryGet` nor `ExoticGet` and its own-property set (script's
expandos) is engine storage its version witnesses; a `DomObject` wrapper reaches the *protected* constructor
and is `OrdinaryGet`, so it takes the receiver exemption and re-establishes its own miss per read; and every
DOM prototype is a `JsObjectShape` instance — a `SharedShapeObject`, `BuiltinShapeMode`, which
`VersionWitnessesOwnProperty` carves out precisely because it keeps its whole own-property set in engine
storage and versions it. So the four links from an `HTMLParagraphElement` wrapper to `Node.prototype` are all
recordable, and the entry this PR makes possible is one the DOM bindings can actually hold.

**Exotic links are refused, and so are value-hook links.** The walk tests
`ExoticGet | OwnValueHook` *before* probing a link and, on either, hands the rest of the read to that link's own
`Get` with the original receiver — which is exactly the continuation an ordinary `[[Get]]` makes at that point.
So a `Proxy`, a `JsTypedArray`, an `IteratorResult`, a `JsArguments`, a `JsError`, a `ModuleNamespace`, an
interop `ObjectWrapper`, a `NamespaceReference` or an `ArrayLikeWrapper` anywhere on the chain resolves through
its own algorithm and is never cached past. `PrototypeMethodCacheTests.EveryGetOverrideIsClassifiedForThePrototypeMethodCache`
— the existing tripwire that forces every new `Get` override to be classified — now governs intermediate links
too, and its comment says so.

**`[[SetPrototypeOf]]` on any link kills the entry.** It moves no version, so it is caught by identity alone:
re-pointing the receiver fails the first comparison, re-pointing an intermediate fails the next one (or the
holder comparison, when the object at that position is unchanged and only what it points at moved), and
splicing a new object in front fails the first. A holder's own prototype is irrelevant — the walk stops there.

**The depth bound is 8 intermediate links** (`MaxCachedPrototypeChainLinks`), which bounds two things only: the
array one warmed site allocates, and the comparisons a warm hit runs. Past it the read still resolves — the walk
hands the rest of the chain to the deepest link it reached — it simply resolves uncached, every time. The bound
doubles as the cycle guard the previous straight-line code did not need.

**What the cache now retains** is the receiver, the holder, the descriptor and up to eight intermediate
prototypes. That is a wider retained set but not a new kind of hazard — those objects are prototypes, which the
constructor that declared them already roots — and `Jint/Runtime/Interpreter/AGENTS.md`'s retention gotcha is
updated to say so.

### The probe counts moved, and that is a result

`HostObjectProbeCountTests` and `HostNamedPropertyProbeCountTests` pinned "a name absent everywhere costs the
ordinary host **2** probes", documented in `Jint/Native/Object/AGENTS.md` as the one count the
`PropertyAccessSemantics` derivation could not improve: the probe that establishes a miss cannot also produce a
value, so the read *ended* in a `Get` that re-probed the receiver before walking the prototype.

The word that was wrong is *ended*. To cache a holder deeper than the direct prototype the member lane has to
walk the chain itself, and a walk that is already several links in has no reason to restart at the receiver. So
the lane now completes the read it started: a hit unwraps the holder's descriptor, a miss that reaches a null
`[[Prototype]]` answers `undefined`, and `ObjectInstance.Get` is re-entered only for the parts of the chain the
lane refused to walk. **The absent row is now 1, the same probe a hit costs** (3 rather than 4 with host-contract
verification on). The host is asked exactly what it was asked before — once instead of twice.

`HostNamedPropertyProbeCountTests` had already written down the fix: *"removing it means changing the
interpreter's shared own-miss fallback rather than anything here."* That is what happened. Its
`ANameAbsentEverywhereCostsTwoProjections` is renamed to `…CostsOneProjection`, and
`AMissingOwnPropertyReadIsNoCheaperThanAHit` to `AMissingOwnPropertyReadCostsTheSameAsAHit`. Both files, and
`Jint/Native/Object/AGENTS.md`, carry the argument rather than a quietly edited number. Every other pinned
figure — own hit 1, prototype hit 1 warm or cold, member-call base 1, and the whole zero-descriptor value-hook
column — is unchanged.

## What I did not change

- **The `in` operator.** `ObjectInstance.HasProperty` is a recursive `ProbeOwnPropertyChecked` walk with no
  cache at all, which is why `HostPrototypeShapeBenchmark`'s `AbsentNameMissOverChain` and `DeepHitOverChain`
  rows — both `('name' in obj)` in a loop — will not move for this PR whatever it does to member reads. They
  *should* learn the same trick: the same three facts would validate a cached boolean, and the absent case is
  the one an entry would help most because it is the one that never warms today. It is a second change with its
  own benchmark row and its own correctness argument, not a rider on this one.
- **The holder rules and `VersionWitnessesOwnProperty`.** One predicate, asked of one more kind of object.
- **The write lane, the shape lanes, the string-method cache and the `ObjectWrapper` cache.** Untouched.
- **A DOM benchmark row.** This is an engine change and the engine-only rows demonstrate it without a host
  object, a wrapper or a binding generator in the picture.

## A separate finding: the stack probe is paid twice per accessor invocation

`TryEnsureSufficientExecutionStack` + `StackGuard.ProbeStackHeadroom` is 4.9 % of that workload, and the reason
is that a getter invocation probes the native stack **twice**, back to back, with only a virtual dispatch
between them and no frame growth in between:

`ObjectInstance.UnwrapFromGetter` → `Engine.Call(ICallable, …)` (`Jint/Engine.cs:5083`) → the callee is a
`Function`, so `Engine.Call(Function, …)` → `CallNativeFunction` → **`_stackGuard.EnsureNativeStackHeadroom()`**
(`Jint/Engine.cs:5176`) → `ClrFunction.Call` → **`_engine._stackGuard.EnsureNativeStackHeadroom()`**
(`Jint/Runtime/Interop/ClrFunction.cs:65`) → the delegate.

Neither site is removable on its own, which is presumably why both exist:

- 51 of the 56 `Function.Call` overrides in the tree do **not** probe (every built-in constructor,
  `EvalFunction`, `FunctionPrototype`, `TypeReference`, the WebApi functions, …) and rely on
  `CallNativeFunction`'s probe.
- Only 5 do — `ClrFunction`, `HostFunction`, `BindFunction`, `DelegateWrapper`, `ScriptFunction` (plus
  `JsProxy`) — and they need their own, because `ICallable.Call` is invoked directly from ~25 sites that never
  go through `Engine.Call` (`IteratorPrototype`, `ArrayConstructor`, `CallbackInvoker`, `GroupByHelper`, …).

So the fix is not a deletion but a way for `CallNativeFunction` to know the callee probes for itself — a cached
`bool` on `Function`, set by those five, turning one `TryEnsureSufficientExecutionStack` call into one
predictable branch on the hottest accessor path. `InternalTypes` has no flag bits left (see the comment at the
bottom of the enum), so it would have to be a field. Left for a second stream: one change per PR.

## Tests

New in `Jint.Tests/Runtime/PrototypeMethodCacheTests.cs`, thirteen tests (fifteen cases) covering exactly the three facts:
a name gained by each intermediate level in turn, a name deleted from one, `[[SetPrototypeOf]]` on a middle
link, a link spliced in front, a holder redefined and deleted at a distance, a deep getter's `this` binding and
two receivers through one warmed site, a `Proxy` in the middle whose trap must run on every read, an array
element appearing on an intermediate under an index-like name, a chain past the depth bound, an absent name
that later becomes present at two different depths, a null-`[[Prototype]]` chain, and the deep form of a
method call.

**They were verified to police what they claim** by breaking each guard on purpose:

| guard removed | tests that fired |
| --- | --- |
| the intermediate `_propertiesVersion` comparison | 5, including all three shadow-by-level cases |
| the intermediate identity comparison | `ALinkInsertedInFrontOfACachedChainIsHonoured` |
| `VersionWitnessesOwnProperty` on intermediates | `AnArrayElementAppearingOnAnIntermediateLinkShadowsTheDeepHolder` |

One of them found a bug in itself first: the shadow-by-level cases were originally written against a getter on
`Base.prototype`, and `Middle.prototype.tag = …` through a setter-less inherited accessor silently does nothing
in sloppy mode. That is correct ECMAScript; the test now shadows a data property.

Everything run in Release, all green:

| run | result |
| --- | --- |
| `dotnet test -c Release` (whole solution) | `Jint.Tests` 12441 / 12441 / 8531 (net10 / net8 / net472), `Jint.Tests.PublicInterface` 3728 / 3718 / 2945, `Jint.Tests.Browser` 2851 / 2848, `Jint.Tests.DevTools` 415 / 413, `Jint.Tests.CommonScripts` 28 / 28, `Jint.Tests.SourceGenerators` 71, `BrowserComparison.Tests` 19 — 0 failed anywhere |
| `Jint.Tests.Test262` | **102,595 passed, 0 failed, 107 skipped** — conformance unmoved, which is the check a property-lookup change owes |
| `JINT_HOST_CONTRACT_VERIFICATION=1` on `Jint.Tests` | 12441 / 12441 / 8531, 0 failed |
| `JINT_HOST_CONTRACT_VERIFICATION=1` on `Jint.Tests.PublicInterface` | 3741 / 3731 / 2958, 0 failed |

The two verification legs are the ones this change is exactly for: every read through the new walk is
recomputed through `Get` and through `GetOwnProperty` and compared.

## To be measured

**No figure here is quotable and none is quoted.** The machine this was developed on was running four other
agents and, at one point, three concurrent `Jint.Benchmark` processes; the `--job short` runs below were
smoke tests that the row executes, nothing more. What they do establish, because it is a gap far larger than
the noise and it reproduced in both of BenchmarkDotNet's jobs, is **the shape this change is aimed at**: on
unmodified `main` the two root-prototype rows sit well above their direct-prototype twins, the getter pair
being the clean one (the method-call pair was bimodal under the load and says nothing). That gap is the
per-level `GetOwnProperty` walk. On this branch the same smoke run puts `RootPrototypeGetterRead` next to
`DirectPrototypeGetterRead` instead of well above it, and the `Allocated` column is unchanged at 704 B on
every row — which is the control for the entry array, since a warmed site reuses it rather than allocating per
read. Treat all of that as direction, not as measurement.

Rows for the lead to run, paired, on an idle box:

```
--filter "*PrototypeChainReadBenchmark*"
--filter "*HostPrototypeShapeBenchmark*"
```

plus the full SunSpider and Dromaeo tables.

`Jint.Benchmark/PrototypeChainReadBenchmark.cs` is new and is the row that actually proves this change: four
`class` declarations, one `new`, and a loop, with no host object anywhere. `OwnPropertyRead` is the baseline
floor and a control; `DirectPrototypeGetterRead` and `DirectPrototypeMethodCall` are the comparison **and the
controls that must not move** — they are the shape that already cached; `RootPrototypeGetterRead` and
`RootPrototypeMethodCall` read the same member four links up and are the subject. It is a class of its own
rather than more rows on `HostPrototypeShapeBenchmark` because that class is parameterised on
`HostPrototypeKind` and these rows contain no host object, so every one of them would be reported twice with
identical work under two labels that mean nothing to it.

On `HostPrototypeShapeBenchmark`, `SteadyStateReads` is the control to watch (its receiver's holder is one link
away); its two chain rows drive `in` and will not move, per the section above.


lahma:
## Measured

Paired A/B, `measure-paired.ps1`, default job, idle machine, baseline = this branch's merge base with the benchmark file copied in untracked.

### The targeted rows, 6 rounds

| row | median Δ | 95 % CI | |
| --- | ---: | --- | --- |
| `RootPrototypeGetterRead` | **−30.9 %** | [−36.8, −26.1] | faster, 0/6 |
| `RootPrototypeMethodCall` | **−31.3 %** | [−34.7, −28.0] | faster, 0/6 |
| `DirectPrototypeGetterRead` *(control)* | −1.4 % | [−8.80, +4.50] | no change |
| `DirectPrototypeMethodCall` *(control)* | +6.5 % | [−9.33, +10.31] | no change |
| `OwnPropertyRead` *(floor)* | +0.6 % | [−2.88, +2.81] | no change |

A read four links up costs what a read one link up costs. Neither control moved, so the extra field test on the direct-prototype path and the extra field on the handler node are both below what this can resolve.

### The wide gate: SunSpider 26 rows + Dromaeo 24 rows, 3 rounds

**No regressions.** 44 of 50 rows no change; the movers are favourable (`ObjectArray[False,True]` −2.1 %, `CoreEval[True,False]` −0.5 %, `string-validate-input` −0.8 %), which is what a cheaper member-read lane should look like.

Six rows read "slower" at 3 rounds. Four were inside this machine's ±3–5 % same-binary spread. The two that were not were re-measured at 6 rounds, per a two-stage rule fixed before the table was read:

| row | 3 rounds | 6 rounds |
| --- | --- | --- |
| `Dromaeo.StringBase64[False,False]` | +11.07 % [+3.75, +12.06] 3/3 slower | **−0.86 %** [−3.90, +2.16] 1/6 no change |
| `SunSpider.math-partial-sums` | +2.63 % [+1.66, +3.60] 3/3 slower | **−1.30 %** [−2.58, −0.74] 0/6 **faster** |

## A warning about the instrument, which this PR paid for three times

**A 3-round paired run produced a 95 % CI excluding zero, with full sign agreement, on three rows that the change cannot affect** — the two above, plus `OwnPropertyRead`, which reads `this.value` from its own shape slot and never reaches a line of this diff (3 rounds: +2.78 % [+0.84, +3.60] 3/3 "slower"; 6 rounds: +0.55 %, no change). All three reversed at 6 rounds, one of them into a win.

A percentile bootstrap over three paired observations is close to their range, and it gives no signal that it is out of its depth. Three rounds is sound for confirming a 30 % effect and unsafe for believing a 3 % one — and worse for a *regression*, where a false positive sends a reviewer hunting through a diff that cannot contain it. Across 50 rows at 3 rounds, some spurious exclusions of zero are near-certain. `Jint.Benchmark/AGENTS.md` is getting this in #4049.

Merging: the win is large and reproducible, both controls and the floor are flat at 6 rounds, the wide gate is clean, test262 is unmoved at 102,595 passing, and all nine checks are green.

PR #4064, opened 2026-09-13
Replace the O(n²) subtree walk in three selector predicates

lahma:
## What the profile found

Re-profiling the campaign's selector workload against current `main` (with #4047's `:target` fix already in): **91% of the page-loop thread is `AngleSharp.Dom.Element.get_NextElementSibling`**. `:target` itself, which used to be 77% of that capture, is now 0.4% — fixing one quadratic exposed the next.

`Element.NextElementSibling` scans its parent's whole child list from index 0 to find *itself* before returning the next element, so it is **O(siblings)**. `PagePseudoClassSelectorFactory.NextInSubtree` walked with it, which made every subtree walk built on it **O(n²) in sibling count**.

This is ours, not AngleSharp's. AngleSharp already avoids `NextElementSibling` for traversal: its own subtree walks use the index-stack struct `ElementTreeEnumerator` (`Css/Dom/SelectorExtensions.cs`, `QueryExtensions.cs`). The only internal uses of the property are the `+`/`~` combinator helper (one step, not a walk) and the markup minifier. We were walking with the wrong tool.

## The fix

`Jint.Browser/Dom/Collections/DomElementWalker` (added by #4046) is an allocation-free depth-first pre-order struct walker over element descendants, O(1) amortised per step. `NextInSubtree` is replaced at its two call sites and then deleted, since nothing calls it anymore.

The two call sites do not have the same shape:

1. **`ValidityStateSelector`'s fieldset arm** (`AFailingCandidateIsADescendantOf`) enumerates the *descendants* of `fieldset`, not the fieldset itself — exactly what `DomElementWalker` does, so this is a direct swap.
2. **`IndeterminateSelector.TheRadioButtonGroupOfHasACheckedMember`** used to start the scan *at* `root` itself. `DomElementWalker` never yields its root, so this call site now tests `root` before the walk covers its descendants. Without that, a radio group whose tree root is itself the checked member would silently stop matching. Covered by a new regression test (`IndeterminateFindsAGroupsCheckedMemberWhenItIsTheTreesOwnRoot`) that builds a detached tree where a checked radio button is the root of another radio button's ancestor chain (via `appendChild` — the void content model is a parser rule, not a node-hierarchy one, so this is valid DOM). Verified this test fails without the root pre-check and passes with it.

Both loops keep their early-exit on the first hit.

## What was *not* changed, and why

`DefaultSelector.IsTheDefaultButtonOfItsForm` goes through `HtmlFormOwner.ControlsOf`, which already uses AngleSharp's `GetDescendants()` (its own allocating LINQ walk) rather than `NextElementSibling` — a different problem, left for a separate change if it's judged worth it.

`Jint.Browser/Layout/**` and `Dom/Views/CssCascade.cs` were not touched.

## Tests that failed before the fix

`IndeterminateFindsAGroupsCheckedMemberWhenItIsTheTreesOwnRoot` (new) fails on the pre-fix shape of `TheRadioButtonGroupOfHasACheckedMember` (confirmed locally by temporarily removing the root pre-check).

## Verification

- `dotnet build -c Release` — clean, 0 errors (one pre-existing, unrelated `net472` reference warning in `Jint.Tests.CommonScripts`).
- `dotnet test -c Release Jint.Tests.Browser/Jint.Tests.Browser.csproj` — 0 failures on both `net8.0` (2944 passed) and `net10.0` (2947 passed, +1 for the new test), including the wpt browser lane.
- `JINT_WPT_BROWSER_CENSUS=1` run of the census tests — figures agree exactly with the pinned table on both target frameworks. No behaviour moved.
- `BrowserSelectorBenchmark` magnitude check (`--warmupCount 1 --iterationCount 2 --launchCount 1`, not a number to quote): `Indeterminate` drops from the documented ~100 ms (pre-fix, O(descendants²) scan over ~2,100 elements for each of 100 unchecked radios) to ~1.6 ms — about 60x — while staying comfortably above the ~60–70 µs `Page.EvaluateAsync` mailbox round trip, so `DominantWorkPasses = 1` is still correct and needed no resizing. Updated a now-stale doc comment in `BrowserSelectorBenchmark.cs` that quoted the old ~100 ms figure.

## To be measured

`*BrowserSelectorBenchmark*` — please run this at gate quality (6 rounds). Expect `Indeterminate` to show a large win and `Control`/`DefaultButton`/`TargetMissing`/`TargetPresent`/`Valid` to be unaffected (only `Indeterminate` and the fieldset arm inside `Valid` touch the changed code, and `Valid`'s fieldset in the benchmark's document is tiny — two fieldsets and one input — so its magnitude should not move noticeably).

## Shared files touched

- `Jint.Browser/Runtime/Parsing/PagePseudoClassSelectorFactory.cs` (the fix)
- `Jint.Benchmark/BrowserSelectorBenchmark.cs` (doc-comment-only change, no pass-count or sizing change)
- `Jint.Tests.Browser/Parsing/PagePseudoClassSelectorTests.cs` (new regression test, appended at the end of the existing `:indeterminate` test)


lahma:
## Measured

Paired A/B, `measure-paired.ps1`, **6 rounds** with the order alternating, default job, idle machine. Baseline is this branch's merge base with the benchmark file copied in untracked, so the arms differ by exactly this change.

| row | before | after | median Δ | 95 % CI | |
| **`Indeterminate`** | 101.05 ms | **1.64 ms** | **−98.5 %** | [−98.58, −98.36] | 0/6 |
| `DefaultButton` | 54.98 ms | 27.79 ms | −49.5 % | [−49.72, −46.36] | 0/6 — *see below* |
| `Valid` | 54.16 ms | 53.16 ms | −2.3 % | [−19.20, +2.25] | no change |
| `Control` *(untouched)* | 24.03 ms | 23.84 ms | −1.6 % | [−6.43, +5.89] | no change |
| `TargetMissing` | 913.8 µs | 906.7 µs | +0.3 % | [−2.42, +3.29] | no change |
| `TargetPresent` | 412.12 ms | 426.76 ms | +2.5 % | [−2.75, +8.81] | no change |

**`:indeterminate` is 62× faster.** Allocation drops too (14.32 KB → 12.58 KB per operation) because the walker allocates nothing where the old walk's iterators did.

`Valid` barely moves, which is expected: its fieldset arm walks a two-element nested fieldset, so the quadratic never had room to show there. The win is where the walk is over a document-sized subtree.

## `DefaultButton` also halved, and I cannot attribute it to this diff

It is real — re-measured **alone**, 6 rounds, −48.5 % [−51.43, −47.67], 0/6 — but `:default` never enters the code this change touches. `DefaultSelector.Match` dispatches to `IsTheDefaultButtonOfItsForm` → `HtmlFormOwner.ControlsOf` → AngleSharp's `GetDescendants()`, which is a stack-based LINQ walk with no `NextElementSibling` in it. An inverted call tree over the pre-change capture confirms it: **100 % of `get_NextElementSibling` came from `IndeterminateSelector`**, and `DefaultSelector` does not appear at all.

I tried to reproduce it outside BenchmarkDotNet, replaying the benchmark's exact document and its 500 `:default` passes against both builds: **638 ms vs 641 ms — no difference.** That run is cold, though, and the benchmark row is warm (55 ms for the same work), so a standalone single run has nowhere near the resolution to see a warm-state effect. The most likely explanation is second-order codegen — deleting `NextInSubtree` changes inlining and layout for the small static helpers (`IsASubmitButton`, `IsOneOf`) that `:default`'s inner loop runs over 2,000 elements per pass — but I have not proved that, and I am not claiming it as a designed improvement.

It is a win rather than a regression, so it does not block; I would rather record it as unexplained than write a mechanism into the history that I did not verify.

## One correction

The commit says "three selector predicates". It changes **two** — `ValidityStateSelector`'s fieldset arm and `IndeterminateSelector`. `DefaultSelector` walks through `HtmlFormOwner.ControlsOf` and is deliberately untouched; its `GetDescendants()` shape is a separate item. Merging with a corrected title.

PR #4065, opened 2026-09-13
Memoize the HTML-uppercased tagName/nodeName per realm

lahma:
## What the profile found

An Ultra capture on #4013 (16M reads of six DOM properties on one element — `nodeName`, `nodeType`, `tagName`, `parentNode`, `firstChild`, `ownerDocument` — 8190 Hz, net10.0) found the accessor-invocation subtree at **50.6%** of the page-loop thread. Inside it:

| | share of that subtree |
| --- | ---: |
| `DomHostHooks.NodeName` | 15.3% (`String.Ctor(wchar[])` 6.0%, `RhpGcAlloc` 6.0%) |
| `DomHostHooks.TagName` | 14.1% (`String.Ctor(wchar[])` 6.1%, `RhpGcAlloc` 5.0%) |
| `DomRealm.WrapNode` (weak-table probe) | 10.9% |
| `StackGuard.ProbeStackHeadroom` | 9.1% |

`el.tagName`/`el.nodeName` allocate a `char[]` and a `string` (via `AsciiUppercase`) plus a `JsString` wrapper (via `JsString.Create`), on **every single read** — `NodeName` delegates to `TagName` for an element, so it pays the same cost twice over.

## The mechanism

The HTML-uppercasing DOM §element-html-uppercased-qualified-name defines is a **pure function of the qualified name**, once `TagName` has already decided the element is in the HTML namespace and its owner is an `IHtmlDocument`. `DomRealm.HtmlUppercasedTagName` memoizes the finished `JsString` per distinct qualified name, so only the first read of a given element interface computes and allocates anything; every later read of an element with the same qualified name is a dictionary lookup.

- **The uppercasing rule stays conditional.** The memo is only ever consulted from the branch that already decided to uppercase; the SVG/XML branch (`element.NamespaceUri != html` or owner is not an `IHtmlDocument`) never touches it at all, so an HTML `<a>` and an SVG `<a>` sharing a local name cannot collide — they never share a cache entry because only one of them ever reaches the cache.
- **`AsciiUppercase` is untouched** apart from `private` → `internal` so the memo can call it on a miss; its ASCII-only semantics are exactly as before (no `ToUpperInvariant`).
- **`NodeName` still delegates to `TagName`** for an element, so the two members cannot disagree — the memo lives behind `TagName`, so `NodeName` gets it for free.
- **Growth is capped at 1024 distinct qualified names** (`DomRealm.MaxCachedUppercasedTagNames`). HTML+SVG+MathML together define a few hundred element names, and even a component-heavy page's custom elements number in the tens to low hundreds — real documents are nowhere near the cap. It exists for a script that manufactures a fresh `document.createElement('x-' + i)` name every iteration purely to read `tagName`; past the cap the answer is still correct, computed the way it always was, and only the memo stops growing.
- **`JsString.CachedCreate`** (already a process-wide value cache for 2–10 character strings, sound to share because a `JsString` carries no engine affinity) is used on a cache miss instead of a hand-rolled wrapper cache, so the common (≤10 character) tag names also reuse the wrapper object itself.

## What is not changed

`DomRealm.WrapNode`'s `ConditionalWeakTable` probe (10.9% of the same subtree) and `StackGuard.ProbeStackHeadroom`'s double probe (9.1%, already tracked on #4013) are separate items, reported but not touched here.

## Benchmark

`Jint.Benchmark/BrowserPropertyReadBenchmark.cs` — `TagName`/`NodeName` are the subjects, `NodeType`/`ParentNode` are controls that must not move, and `PlainProperty` (a same-shaped plain-JS object read) is the floor. Each row loops its read (50,000 passes for the DOM rows, 2,000,000 for the much-cheaper floor row) to amortise the `Page.EvaluateAsync` mailbox round trip, per `Jint.Benchmark/AGENTS.md`.

A `--job short` smoke run (not a number to quote) confirms the benchmark measures the right shape — allocation per operation before/after this change, holding the pass counts fixed:

| Method | Allocated (before) | Allocated (after) |
| --- | ---: | ---: |
| `TagName` | 4,696.4 KB | 8.87 KB |
| `NodeName` | 4,696.4 KB | 8.88 KB |
| `NodeType` (control) | 8.52 KB | 8.52 KB |
| `ParentNode` (control) | 8.91 KB | 8.9 KB |
| `PlainProperty` (floor) | 10.21 KB | 10.13 KB |

**To be measured:** `--filter "*BrowserPropertyReadBenchmark*"` at the gate's 6 rounds.

## Tests

- `dotnet build -c Release` — solution builds clean, 0 warnings/errors in the touched projects.
- `dotnet test -c Release Jint.Tests.Browser/Jint.Tests.Browser.csproj` — net8.0: 2944 passed, 0 failed; net10.0: 2947 passed, 0 failed.
- Same, with `JINT_WPT_BROWSER_CENSUS=1` (net10.0): 2949 passed, 0 failed — the wpt census table (`tagName`/`nodeName` are heavily covered there) matched what the lane measures; no figure moved.

## Files touched

- `Jint.Browser/Dom/DomHostHooks.cs` — `TagName` calls the memo on the uppercasing branch; `AsciiUppercase` is `internal` now.
- `Jint.Browser/Dom/DomRealm.cs` — new `HtmlUppercasedTagName` memo and its cap.
- `Jint.Benchmark/BrowserPropertyReadBenchmark.cs` — new benchmark.


lahma:
## Measured

Paired A/B, **6 rounds**, order alternating, default job, idle machine. Baseline is this branch's merge base with the benchmark copied in untracked, so the arms differ by exactly this change.

| row | before | after | median Δ | 95 % CI | allocation |
| **`TagName`** | 4.386 ms | 3.718 ms | **−15.9 %** | [−18.04, −13.84] | **4,696.37 KB → 8.87 KB** |
| **`NodeName`** | 4.368 ms | 4.164 ms | **−16.5 %** | [−21.66, −6.88] | **4,696.39 KB → 8.90 KB** |
| `NodeType` *(control)* | 3.215 ms | 3.014 ms | −1.3 % | [−5.78, +3.75] | unchanged |
| `ParentNode` *(control)* | 3.727 ms | 3.748 ms | −2.3 % | [−4.53, −0.23] | unchanged |
| `PlainProperty` *(floor)* | 84.541 ms | 86.746 ms | +0.1 % | [−19.93, +2.21] | unchanged |

The allocation column is the result: a repeated `tagName`/`nodeName` read now allocates what the control rows allocate — the memo removed it rather than reduced it. `ParentNode`'s interval grazes zero on a 1/6 sign, which is noise on a control, not an effect.

Six rounds rather than three deliberately: a 3-round run in this campaign produced a 95 % CI excluding zero, with full sign agreement, on three separate rows that a change could not affect. `Jint.Benchmark/AGENTS.md` carries that now.

CI: the `linux - ARM` job failed once on a net8.0 page time-budget flake (`PageLoop: The operation's time budget elapsed`) while net10.0 passed with 2,947 tests in the same run — the net8/net10 asymmetry recorded on #3947. Re-run, clean, all nine green.

Merging.

PR #4078, opened 2026-09-15
Walk the prototype chain in a loop instead of one native frame per link

### LINKED ISSUE #4076: `StackOverflowGuard` does not cover the prototype-chain walk in `ObjectInstance.Get`: a deep `__proto__` chain still ends the process
### Pre-flight checks

### Summary

#3877 ("Guard native recursion and forwarding paths", backported to 4.x as #4007) extended `StackOverflowGuard` to native recursion that previously killed the process — `Array.prototype.flat`, `JSON.stringify`, proxy and bound-function chains, host re-entry. With the guard on, those now raise a catchable `RangeError`, which I can confirm on 4.16.2 for 45,000-deep `JSON.stringify` and `flat(Infinity)`.

One native recursion path is not covered: a property **miss** walks the prototype chain through `ObjectInstance.Get` → `GetOwnProperty` → prototype's `Get`, one native frame per link. A chain of ~20,000 objects overflows the native stack on a default 1 MB thread, and the guard never fires. The chain needs no API to build — `Object.setPrototypeOf` and `Object.create` can be removed from a sandboxed realm, but the `__proto__` key in an object literal sets the prototype at creation.

### Variations tried

| Configuration | Result |
|---|---|
| `StackOverflowGuard = true` | process ends |
| `StackOverflowGuard = true` + `LimitRecursion(64)` | process ends (no JS call is involved, so the frame counter sees nothing) |
| `StackOverflowGuard = false` | process ends |
| `StackOverflowGuard = true`, evaluated on a `new Thread(…, 256 MB)` | returns `undefined` in ~530 ms |

So the outcome depends only on the calling thread's headroom, which is the situation the option's documentation describes as the one it exists to remove ("no `catch` sees it, no exception is raised, nothing is logged, and the process is simply gone").

For contrast, on the same engine and thread:

```js
let x = []; for (let i = 0; i < 45000; i++) x = [x]; JSON.stringify(x)   // RangeError, ~200 ms ✔
let x = []; for (let i = 0; i < 45000; i++) x = [x]; x.flat(Infinity)    // RangeError ✔
```

### Suggestion

The chain walk in `ObjectInstance.Get` (and presumably `HasProperty`/`Set` on a miss, which I have not tried) looks like a candidate for the same `EnsureNativeStackHeadroom` probe #3877 placed on `flat` and the JSON serializer — or, since the chain is a linked list, an iterative walk that needs no stack at all. Either would make `x.missing` on a deep chain a `RangeError` rather than a process exit, consistent with the rest of the guard.

Happy to test a fix against the reproduction above; I have it as an xunit test (currently skipped, because it takes the test host down).

### Minimal JavaScript repro

```js
let x = {}; for (let i = 0; i < 20000; i++) x = { __proto__: x }; x.missing
```

### C# host code

```csharp
using Jint;

var engine = new Engine(o =>
{
    o.Constraints.StackOverflowGuard = true;
    o.MaxStatements(1_000_000);
});

engine.Evaluate("let x = {}; for (let i = 0; i < 20000; i++) x = { __proto__: x }; x.missing");
```

### Expected behavior

`JavaScriptException` / `RangeError: Maximum call stack size exceeded`, as for `JSON.stringify` on the same depth.

```text
The process ends with a native stack overflow (Windows exit `0xC00000FD`); nothing is thrown or logged.


Stack overflow.
   at Jint.Extensions.Hash.GetFNVHashCode(System.String)
   at Jint.Key..ctor(System.String)
   at Jint.Key.op_Implicit(System.String)
   at Jint.Native.Object.ObjectInstance.GetOwnProperty(Jint.Native.JsValue)
   at Jint.Native.Object.ObjectInstance.Get(Jint.Native.JsValue, Jint.Native.JsValue)
   at Jint.Runtime.Interpreter.Expressions.JintMemberExpression.ReadAfterOwnMissUncached(...)
   at Jint.Runtime.Interpreter.Expressions.JintMemberExpression.ReadAfterOwnMiss(...)
   at Jint.Runtime.Interpreter.Expressions.JintMemberExpression.GetValue(...)
   (ObjectInstance.Get repeats once per prototype link)
```

### Jint version

4.16.2

### .NET runtime

net10.0

### Operating system

Windows

### AOT / trimming

None

### Engine configuration

- [ ] Modules enabled (`UseModules`)
- [ ] Strict mode

### Area

None

### Regression?

_No response_

### Additional context

_No response_

lahma:
## The bug

A prototype chain is built by script, so its depth is an input. `ObjectInstance`'s `[[Get]]`, `[[Set]]` and `[[HasProperty]]` each resolved it by calling the same method on `Prototype` — one native frame per link — so

```csharp
var engine = new Engine(o => { o.Constraints.StackOverflowGuard = true; o.MaxStatements(1_000_000); });
engine.Evaluate("let x = {}; for (let i = 0; i < 20000; i++) x = { __proto__: x }; x.missing");
```

ended the process with a native stack overflow (`0xC00000FD`). Nothing was thrown, no `catch` saw it, and the engine was gone — where the same depth through `JSON.stringify` or `flat(Infinity)` already raised a catchable `RangeError` (#3877). The same shape reached the write side through `x.missing = 1`, the existence side through `'missing' in x`, and — less obviously — every identifier resolved inside `with (x) { … }`, whose object environment asks `HasProperty`.

Confirmed before the fix by running the new tests against the unfixed engine: the test host dies with `Stack overflow.` / `The active test run was aborted. Reason: Test host process crashed : Stack overflow.`, the dump showing `ObjectInstance.Get` repeated to the bottom of the stack (and `JsProxy.Get` for the proxy case). That death *is* the confirmation — an unguarded case cannot fail, it can only take the runner with it.

## The fix: walk, don't recurse

All four walks are loops now — `Get` through `GetFromPrototypeChain`, the private receiver-threading `TryGetValue` overload inline, `Set` through `SetOnPrototypeChain` (from both fast-path arms and from `OrdinarySetWithOwnDescriptor`), and `HasProperty` inline. An ordinary chain of any depth therefore **resolves** rather than failing politely: the frames are gone, not bounded.

A loop in the base class may only run the *ordinary* algorithm, and that is the whole correctness question here. Every override of these three — `ArrayInstance`, `JsProxy`, `JsTypedArray`, `ObjectWrapper` and its specialized forms, `JsArguments`, `JsError`, `IteratorResult`, `ModuleNamespace`, `ArrayLikeObject`, `NamedPropertyObject`, `NamespaceReference`, `TypeReference`, `JsStorage` — calls `base` as its own tail, i.e. the base implementation runs *on behalf of* the derived object. A walk that probed `GetOwnProperty` itself would silently skip a mid-chain override. So each walk hands the rest of the algorithm to the first link it may not walk, exactly as `JintMemberExpression.ReadAfterOwnMissUncached` already does:

* **The read side** keys that on `InternalTypes.ExoticGet | OwnValueHook`, which already meant precisely this.
* **`[[Set]]` and `[[HasProperty]]`** have no such derived flag, so they ask for the *positive* claim instead and walk only a link carrying `InternalTypes.PlainObject` — set through the internal constructor by five types (`JsObject`, `JsDate`, `GlobalObject`, `NumberPrototype`, and `Prototype`, the base of every built-in prototype), none of which overrides one. That direction is the safe one: a link the walk declines is resolved exactly as it always was, and the flag's own comment now states the obligation.

## Why a probe was rejected on the walk

`Jint/Constraints/AGENTS.md` records the double native-stack probe on the accessor-read lane as a **measured NO-GO** — an Ultra capture on #4013 put `ProbeStackHeadroom` at 17.0% of a 29.1% subtree, and a six-round paired gate found no resolvable change on the subject row or any DOM row. `ObjectInstance.Get` is hotter still. And `StackOverflowGuardTests.ExactlyTheInteropAndForwardingFunctionsProbeTheNativeStack` reads the probe sites out of the IL and pins them to exactly `BindFunction`, `ClrFunction`, `DelegateWrapper` and `HostFunction`.

The probe therefore sits **at the hand-over**, which is off the ordinary path and paid only by a chain that really does forward. **The IL-scan pin needed no change and still passes**: it enumerates `Function` subclasses' `Call(JsValue, JsValue[])`, and none of the new probe sites is one (`ObjectInstance` is not a `Function`; `JsProxy` is an `ObjectInstance` whose `Call` is an explicit interface implementation). That was checked rather than assumed — the test is green.

## The second instance: `JsProxy`

A trapless proxy forwards its whole algorithm to its target (`return target.Get(property, receiver);` and the same for every other internal method), so `new Proxy(new Proxy(new Proxy(…)))` is a recursion script controls, and nothing on those paths probed. That chain **cannot** be flattened into a loop — each hop runs trap lookup and the result invariants on the way back out — so it is probed.

**The probe goes on the forward, not at the operation's entry.** A *trapped* proxy reaches its trap through `CallTrap` → `ICallable.Call`, where the callee probes for itself: `ScriptFunction`'s four entry points carry the backstop, the interop and forwarding leaves carry theirs, and a callable proxy carries `[[Call]]`'s. So a trapped chain already raised a catchable `RangeError` before this PR — and an entry probe would have been a redundant third probe on exactly the shape real code ships, since every reactivity library (Vue 3, MobX, immer) puts a `get` trap on every object it proxies and `Jint.Tests.Browser` runs a Vue 3 TodoMVC. That is the same redundancy `Jint/Constraints/AGENTS.md` records as a measured NO-GO for the dispatcher probes. *(An earlier revision of this PR did probe at the entry; this is the correction.)*

All 24 forward sites go through `ForwardToTarget(target)`, which returns the target so each site reads as the one thing it is and a forward added later without a probe stands out beside its siblings. That is 13 internal methods (each with a trapless arm and a CLR-handler-declined arm) plus `IsSpecArray` and `ToObject`, which have no trap at all. It includes the three that forward through a property rather than a call — `Prototype` is `GetPrototypeOf()` and `Extensible` is `IsExtensible()`, both virtual and both overridden by `JsProxy`, so both recurse. **No proxy internal method has a trapless path that does not forward to the target.** `AssertNotRevoked` keeps its name and body, and `[[Call]]`/`[[Construct]]` keep their own entry probes byte-identical to `main` — they are the dispatcher for everything reached through them, and narrowing those two is a pre-existing question this change deliberately leaves alone.

## Tests

* `Jint.Tests.PublicInterface/HostNativeRecursionGuardTests` gains nine depth cases over a 20 000-link chain (read hit/miss, inherited getter, write, inherited setter, `in` hit/miss, `with` hit/miss) asserting the **answer**, plus three over a 10 000-proxy chain asserting a catchable `RangeError`. On `net472` the trapless proxy *read* takes the same carve-out the trapless proxy *call* already has, the JIT turning the forwarding call into a tail call so the read legitimately completes.
* `Jint.Tests/Runtime/PrototypeChainWalkTests` is new: a census asserting no object reachable from a built engine carries `PlainObject` while overriding `Get`/`Set`/`HasProperty` (broken on purpose once — adding the flag to `ArrayInstance` fails it naming `ArrayInstance.Get`), plus two behaviour tests that a Proxy and an array mid-chain still run their own algorithms.
* `Jint.Benchmark/PrototypeChainReadBenchmark` gains `AbsentNameRead` beside its existing control rows — the only row whose cost is the walk to the end of the chain rather than a cached hit. **No numbers were run**; the lead measures serially on an idle box.

## What was run

`dotnet build -c Release` and `dotnet test -c Release` (whole solution, exit 0): Jint.Tests 12 460 / 12 460 / 8 544 on net10.0 / net8.0 / net472, Jint.Tests.PublicInterface 3 733 / 3 743 / 2 960, Jint.Tests.Browser 3 033 + 3 036, Jint.Tests.DevTools, CommonScripts, SourceGenerators — 0 failures anywhere. `JINT_HOST_CONTRACT_VERIFICATION=1` legs for `Jint.Tests` (12 460 / 12 460 / 8 544) and `Jint.Tests.PublicInterface` (3 746 / 3 756 / 2 973): green, so the probe-count and semantics pins hold in both configurations. test262: **102 601 passed, 0 failed**, 109 skipped.

## Two things this does not close, stated rather than left to be found

**The read walk's gate is self-maintaining; the census does not pin that half.** The walk hands over on
`InternalTypes.ExoticGet | OwnValueHook`, and every type overriding `Get` either sets `ExoticGet` in its
constructor (`JsProxy`, `JsTypedArray`, `JsArguments`, `JsError`, `IteratorResult`, `ObjectWrapper`,
`NamespaceReference`) or has it derived per runtime type by `ObjectInstance.ResolvePropertyAccessSemantics`,
whose own failure mode is the conservative `ExoticGet`. So unlike the `PlainObject` claim the write and
existence walks read — which a type states by hand, which is why `PrototypeChainWalkTests` asserts it over
every reachable object — the read side cannot be got wrong merely by adding a type. The one way to defeat it
is `SetPropertyAccessSemantics`, the documented escape hatch, which already gates
`JintMemberExpression.ReadAfterOwnMissUncached` in exactly the same way: a host that declares ordinary
semantics while overriding `Get` was skipped by the member lane before this PR too.

**A chain of consecutive `ObjectWrapper`s still recurses, unprobed.** `ObjectWrapper.Get`
(`Jint/Runtime/Interop/ObjectWrapper.cs:1026`) falls through to `Prototype?.Get(property, receiver)` and then
post-processes the result for `ThrowOnUnresolvedMember`, so it is not a tail call and cannot join the loop as
it stands. A chain that alternates a wrapper with ordinary objects is fine — each ordinary link re-enters the
walk — but *N adjacent wrappers* are N native frames with no probe between them. Reaching it needs
`AllowClr`, N host objects and a script calling `Object.setPrototypeOf` on each, which is why it is not
folded in here: it is a different shape from the reported one, and adding a probe to a wrapper's member-miss
lane is a measurement question of its own rather than a free line. Filed as #4087.

## Shaped prototypes: what the gate found and the fix

The wide paired gate blocked the first two commits, and the two review comments above hold the tables. At
eight rounds, on `HostPrototypeShapeBenchmark` with `PrototypeKind=Shape`, `DeepHitOverChain` was **+8.70%**
[+6.29, +14.50] 8/8 and `AbsentNameMissOverChain` **+5.52%** [+1.56, +10.06] 7/8, while the same two rows on
`Dictionary` prototypes improved **−9.94%** and **−15.42%**. The rows that do no walking were flat, which is
the control: this was a mechanism, not drift.

### The mechanism

Both chain rows are `'name' in obj`, so the walk they exercise is `[[HasProperty]]`'s. That walk and
`[[Set]]`'s ask for the *positive* claim `InternalTypes.PlainObject`, which five constructors set by hand.
`SharedShapeObject` — what every `JsObjectShape.Instantiate` returns, i.e. **every shaped host prototype** —
overrides only `Initialize()`, yet its constructor passed `InternalTypes.Object` alone. So the walk declined
a link it was entitled to walk, handed the rest of the chain back to the recursive path *and* paid the new
flag test at every level: the old recursion plus the new overhead, which is exactly the shape of the table.
The read side never had this, because its gate (`ExoticGet | OwnValueHook`) is **derived** per type by
`ResolvePropertyAccessSemantics` and so maintains itself; the positive claim does not.

### What was flagged, and why the `BuiltinShapeMode` question answers itself

`SharedShapeObject` now declares `InternalTypes.PlainObject`. As a *link*, its `[[Set]]` and
`[[HasProperty]]` are the ordinary algorithm the loop performs — the same virtual `GetOwnProperty` /
`ProbeOwnProperty` / `GetPrototypeOf` the loop calls, the same accessor and setter dispatch, the same
receiver — in both storage modes, because it overrides none of them.

The interaction with `BuiltinShapeMode` that wanted care is **not a new one**. The three lanes that read
`PlainObject` as a *storage* claim (`ObjectInstance.Get`, `Set`, `CreateDataProperty`) each spell their test
`(_type & (PlainObject | BuiltinShapeMode)) == PlainObject`, so a shaped object is excluded from all three
for as long as its shape is installed; once `DeoptBuiltinShape` has moved every slot into `_properties` it
*is* the ordinary dictionary they assume. `Object.prototype` has been exactly that pair all along — a
`Prototype` (hence `PlainObject`) declared `[JsObject(UseShape = true)]` (hence `BuiltinShapeMode`) — and
every walk has resolved it as an ordinary link since the first commit here. `JintMemberExpression`'s three
lanes test the pair with `!= Empty`, so they already served a shaped object and are unchanged.

### The census, in both directions

`PrototypeChainWalkTests` keeps the direction it had (nothing carrying the flag overrides `Get`, `Set` or
`HasProperty`) and gains the converse (nothing overriding **none** of the named methods is unflagged), over
the same reachable census — now extended with a shaped prototype chain built through the public
`JsObjectShape` API, since no engine builds one on its own, which is why the census could not see this. The
converse uses the wider method set the flag actually promises: the walk's three plus `GetOwnProperty`,
`GetOwnPropertyKeys`, `ProbeOwnProperty`, `TryGetOwnPropertyValue` and `GetInitialOwnStringPropertyKeys`.
`SetOwnProperty` and `DefineOwnProperty` are deliberately absent — `ObjectPrototype` carries the flag and
overrides both.

Both directions were broken on purpose and watched fire:

| broken | what the failure said |
| --- | --- |
| flag removed from `SharedShapeObject` | `… but found "SharedShapeObject".` (3/3 TFMs) |
| flag added to `ArrayInstance`, which overrides all three | `… but found "ArrayInstance.Get".` (3/3 TFMs) |

The census's allowlist holds the **19** other in-box types it reaches that override none of the named methods
and do not carry the flag, with one reason for all of them: `PlainObject` is two claims in one, and taking it
also switches the three storage fast lanes on — a change to measure on its own terms rather than part of this
regression. Six are built-in-shape objects reached through the public `ObjectInstance(Engine)` constructor
(`AtomicsInstance`, `GeneratorPrototype`, `JsonInstance`, `MathInstance`, `ReflectInstance`, `TemporalNow`);
thirteen are ordinary in-box objects with their state in fields and their properties in the base storage
(`BooleanInstance`, `BooleanPrototype`, `ErrorPrototype`, `GeneratorInstance`, `IntlInstance`, `JsDataView`,
`JsMap`, `JsPromise`, `JsSet`, `JsWeakMap`, `JsWeakSet`, `NumberInstance`, `TemporalInstance`). A shaped host
prototype is the one exception and is flagged, because for it the storage half is free.

### The regressions

`AShapedHostPrototypeChainIsWalkedAndAnswersCorrectly` states the behaviour: through a three-level shaped
chain, `in` hits at the deepest level and misses a name nothing declares, `hasOwnProperty` stays false, a
write to an inherited method lands on the receiver and leaves the prototype alone, an inherited setter runs
with the receiver as its `this`, and an inherited get-only accessor refuses a strict-mode write with a
`TypeError`.

`Jint.Tests.PublicInterface` adds the direct statement that the chain is *walked* rather than recursed:
`ADeepShapedPrototypeChainIsWalkedWithoutExhaustingTheNativeStack` builds **twenty thousand** links from one
process-shared shape and asserts the answers. Against the unfixed tree 9 of its 24 cases fail with
`RangeError:Maximum call stack size exceeded` — the hand-over probe firing, once per level — on the `write`,
`in miss`, `with hit` and `with miss` routes; the read routes pass either way, which is the control, because
the read walk's derived gate already accepted a shaped link.

### Verification

`dotnet build -c Release`: 0 errors, 1 pre-existing MSB3277 warning. `Jint.Tests` 33,613 total / 0 failed /
14 skipped on net8.0, net10.0 and net472; `Jint.Tests.PublicInterface` 10,523 / 0 failed / 63 skipped on the
same three. With `JINT_HOST_CONTRACT_VERIFICATION=1`: 33,613 / 0 failed / 14 skipped, and 10,523 / 0 failed /
24 skipped. `Jint.Tests.Test262`: **102,601 passed / 0 failed / 109 skipped**, the control on `main` exactly.
Rebased onto `main` at `c3bb8420964da1027035c6a2f0b5f5ea038592ac`.

No benchmark was run for this commit. The paired gate — `HostPrototypeShapeBenchmark` stage 2, plus SunSpider
and Dromaeo — is re-run by the maintainer before merge.

Closes #4076


lahma:
## Gate result: a structural regression on shaped host prototypes — blocking

The broad paired gate (SunSpider + Dromaeo + `HostPrototypeShapeBenchmark`, 3-round stage-1 screen per the two-stage rule in `Jint.Benchmark/AGENTS.md`) found a signal that is **not** noise. Both chain-walking rows flip sign by `PrototypeKind`:

| row | `Dictionary` | `Shape` |
| --- | ---: | ---: |
| `AbsentNameMissOverChain` | **−11.97%** [−18.29, −8.20] 0/3 | **+2.80%** [+2.52, +6.14] 3/3 |
| `DeepHitOverChain` | **−5.09%** [−11.34, −4.20] 0/3 | **+8.41%** [+6.71, +10.16] 3/3 |

Perfect sign agreement in both directions, CIs well clear of this box's ±3–5% same-binary spread, and the two kinds moving *opposite* ways — which is the signature of a mechanism, not of drift.

### The mechanism, confirmed in the source rather than inferred from the numbers

`SharedShapeObject` — what `JsObjectShape.Instantiate` returns, i.e. every shaped host prototype — overrides **only `Initialize()`**. It overrides no property internal method at all and runs the ordinary algorithm throughout. But its constructor is:

```csharp
internal SharedShapeObject(Engine engine, JsObjectShape shape, ObjectInstance? prototype)
    : base(engine, ObjectClass.Object, InternalTypes.Object)
```

No `InternalTypes.PlainObject`. So at `ObjectInstance.cs:1960` and `:2129` the `[[Set]]` and `[[HasProperty]]` walks decline it:

```csharp
if ((link._type & InternalTypes.PlainObject) == InternalTypes.Empty)
```

and hand the rest of the chain back to the recursive path — after paying the new flag test at every level. A shaped chain therefore gets the old recursion *plus* the new overhead, which is exactly what the table shows.

This is the cost half of the trade-off the PR body states only in correctness terms:

> That direction is the safe one: a link the walk declines is resolved exactly as it always was

Resolved as it always was, yes — but no longer *costed* as it always was. And `PrototypeChainWalkTests` polices only the converse direction (nothing carries `PlainObject` while overriding `Get`/`Set`/`HasProperty`); nothing asserts that an object overriding none of them *does* carry it, which is how this got through.

It matters because shaped prototypes are the representation the DOM bindings are moving to — `HostPrototypeShapeBenchmark`'s own doc models it on the real ~170-prototype / ~33 MB report. The PR currently makes today's dictionary pattern faster and tomorrow's shaped pattern slower.

### Not addressed here

Whether the remedy is simply adding `PlainObject` to `SharedShapeObject` needs care rather than a one-liner: several lanes test `(_type & (PlainObject | BuiltinShapeMode)) == PlainObject`, and `SharedShapeObject` is `IBuiltinShaped`, so the interaction with `BuiltinShapeMode` has to be worked out — and the census test should gain the positive direction so this cannot recur. Both want their own gate run.

Stage-2 confirmation (`*HostPrototypeShapeBenchmark*`, 8 rounds) is queued; the code evidence above is independent of it, so I am not waiting on it to call this blocking.

Raw data: 3 rounds, alternating order, `DefaultJob`, verified-idle box, base `46729496b` vs head `64942f340`. The remaining SunSpider/Dromaeo rows clearing the stage-1 bar were `access-binary-trees` +4.24%, `StringBase64[False,True]` +4.35% and `crypto-aes` +3.22%; `StringBase64` is a row already recorded in `Jint.Benchmark/AGENTS.md` as a 3-round false positive, and the others are unmeasured at stage 2 because this finding blocks the PR regardless.

lahma:
## Stage 2 confirms it, at 8 rounds

Re-measured `*HostPrototypeShapeBenchmark*` alone, 8 rounds, alternating order, `DefaultJob`, idle box, same pair (base `46729496b` vs head `64942f340`):

| row | `Dictionary` | `Shape` |
| --- | ---: | ---: |
| `AbsentNameMissOverChain` | **−15.42%** [−17.79, −14.66] 0/8 | **+5.52%** [+1.56, +10.06] 7/8 |
| `DeepHitOverChain` | **−9.94%** [−10.80, −8.31] 0/8 | **+8.70%** [+6.29, +14.50] 8/8 |
| `BuildAndTouchPrototypes` | +0.53% [−0.80, +2.27] 5/8 | −1.22% [−3.42, +0.99] 3/8 |
| `BuildPrototypes` | −0.76% [−1.37, +2.34] 2/8 | −0.64% [−2.00, +1.17] 3/8 |
| `SteadyStateReads` | −0.12% [−5.23, +1.17] 3/8 | −0.91% [−8.43, +0.67] 3/8 |

Three things this settles.

**The regression is real and it is exactly where the mechanism says it is.** Both chain-*walking* rows on shaped prototypes are slower with the CI clear of zero, and both got larger and tighter than at three rounds rather than regressing toward it. The rows that do no walking — `BuildPrototypes`, `SteadyStateReads` — are untouched, which is the control this needed: `SteadyStateReads` resolves through the prototype-method inline cache and never reaches the walk, so a change that showed up there too would have meant something other than the gate was moving.

**The win on dictionary prototypes is bigger than the stage-1 screen suggested** — −9.94% and −15.42%, up from −5.09% and −11.97%. This PR is a substantial improvement for today's pattern. That is the reason to fix the shaped lane rather than abandon the change.

**Stage 1's borderline rows were false positives, as designed.** `BuildAndTouchPrototypes[Dictionary]` read +2.33% with 3/3 sign agreement at three rounds and is +0.53% / no change at eight. That is exactly the effect `Jint.Benchmark/AGENTS.md` documents for a wide table at three rounds, and it is why the stage-1 bar (sign agreement **and** magnitude above the box's ±3–5% floor) exists rather than the CI alone.

The remedy is unchanged from the comment above and still wants care rather than a one-liner: `SharedShapeObject` overrides nothing but `Initialize()` and so is entitled to the walk, but giving it `InternalTypes.PlainObject` has to be reconciled with the lanes testing `(_type & (PlainObject | BuiltinShapeMode)) == PlainObject` on a type that is `IBuiltinShaped`, and `PrototypeChainWalkTests` should gain the positive-direction assertion so the next type that overrides nothing cannot silently miss the flag.

Issue #3931, opened 2026-09-08
Browser FormData(form) still rejects DOM-backed form elements


sebastienros:
Found by the common `event-form` workload while implementing #3930 / campaign #3575 X4. A freshly built `Jint.Browser.Tool` at main `052b45aa9fde49ca54ebcb65529f7c54ad1a5a72` rejects `new FormData(form)` inside a real browser page:

```html
<form><input name="query" value="hello"></form>
<script>
const form = document.querySelector('form');
new FormData(form).get('query'); // expected "hello"
</script>
```

The observed exception is `FormData constructor: the form argument is not supported, there is no DOM`, originating in `Jint/WebApi/Files/FormDataConstructor.cs:54`. That guard is appropriate for the standalone engine, but the browser page has an AngleSharp DOM and HTMLFormElement wrappers. `Jint.Browser/Runtime/FormSubmitter.cs` already constructs form entries for navigation submission; the constructor needs a browser integration path rather than a second serialization algorithm.

The full common workload dispatches 100 bubbling input events and checks the serialized form value. Cold and warm runs pass on installed Chromium and official Lightpanda `1.0.0-nightly.9231+b21ec6085`; Jint stops at the FormData constructor. This is correctness-only smoke evidence, not timing data. The X4 workload is retained unchanged and comparison collection remains incomplete rather than substituting a Jint-specific script.

The [XHR living standard's FormData constructor](https://xhr.spec.whatwg.org/#dom-formdata) accepts an optional form and submitter, constructs the form entry list, and specifies submitter/type/ownership and entry-list failure handling.

Acceptance:

- A browser page's `new FormData(form[, submitter])` uses the shared HTML form-entry construction rules and returns standard entries.
- Cover the simple named input, disabled/unsuccessful controls, duplicate names, form-associated controls, submitter validation, and the `formdata` event/reentrant construction behavior.
- Preserve standalone-engine behavior where no DOM integration exists, realm ownership, and WebIDL conversion rules; keep the browser-specific hook internal unless a public contract is actually required.
- Add fresh Release browser regressions and any applicable pinned WPT coverage; rerun the unchanged X4 form workload.

sebastienros:
The constructor bridge is prepared locally at `2eb407edae62fb4c9f4dfa86a66e7e8c627cd5cd`, based on main `052b45aa9fde49ca54ebcb65529f7c54ad1a5a72`; no PR or workflow has been started while the four build slots are occupied.

[...]

Issue #3930, opened 2026-09-08
Complete X4 controlled browser comparison and gate-quality evidence


sebastienros:
## Complete — 2026-09-22

[Final report and complete text tables](https://github.com/sebastienros/jint/blob/f8260f94a3755756c11e65124e7b9bdd7ab39948/docs/benchmarks/browser-comparison-2026-09/README.md) published by merged #4149 after all nine validation checks passed (PR deployment intentionally skipped). Two independent complete hosted runs passed all accounting, correctness, idle, provenance and cleanup gates, with exact independent reanalysis. The report retains calibration noise, two CPU uncertainty changes, memory tradeoffs and hosted-environment limitations. Raw results remain local under the maintainer-approved retention scope below.

## Retention scope — maintainer decision, 2026-09-22

The maintainer explicitly removed the requirement to package or publish downloadable raw archives: the complete local results are sufficient for analysis. No binary files may be committed. Acceptance now requires the two complete independently audited runs and a published text report with complete result tables, provenance/checksums and reproduction references. Raw evidence is retained locally under the ignored `artifacts/campaign-evidence` directory; original workflow artifacts remain subject to GitHub retention. External archive publication is not a closure prerequisite. Earlier refusals remain excluded and documented.

## Implementation history — 2026-09-18

The implementation requirements below landed in [#3941](https://github.com/sebastienros/jint/pull/3941). The collector, owned adapters, Linux cgroup-v2 accounting, representative workloads, provenance/cleanup journals, balanced schedule and refusal/statistics tests are present on main. The earlier [#3931](https://github.com/sebastienros/jint/issues/3931) FormData(form) blocker is closed. The checked infrastructure boxes describe implemented capabilities, not completed comparative measurements.

The subsequently completed hosted measurements validate the real Linux cgroup backend, pinned binaries, empirical same-VM A/A calibration, six-round comparisons and independent reproduction. Physical-host contention remains unobservable; idle, accounting, correctness and completeness gates were unchanged. See the final report above for accepted evidence and limitations.

Follow-up to #3902 / merged #3907; completes the measurement infrastructure and evidence work for #3575 X4.

[...]

PR #4091, opened 2026-09-18
Document current browser comparison acceptance work

sebastienros:
The browser comparison README still described FormData(form) issue #3931 as blocking collection. That issue is closed and the controlled collector landed in #3941. Update the README to name the remaining empirical calibration, dedicated Linux collection and retained-evidence requirements tracked by #3930.

Documentation only; verified the linked issue/PR states and ran `git diff --check`. No benchmark was run and no performance claim is added.

Refs #3575, #3930.


sebastienros:
The Windows check failed in the existing `ServeCommandTests.ABrowserOpensWithATabSoAClientListingTargetsFindsOne` discovery assertion: `/json/list` returned `[]` after the ready banner. This README-only change cannot alter that behavior.

The startup race is now tracked in #4102 and fixed separately in #4104. The fix creates the initial page before awaiting `AddBrowser`, which guarantees existing targets are published before the banner. The immediate discovery assertion is retained; 12 focused Release tests pass across net8/net10. I will update this PR onto the fix after it merges, preserving the original failed run as evidence rather than hiding it with a retry.

sebastienros:
#4104 is merged with all checks passing. This documentation branch is now rebased onto that fix (`e1790099b`), and fresh CI is running. The original failed run and its identified startup race remain linked above.

PR #4112, opened 2026-09-19
Keep shallow template clones empty of template contents

### LINKED ISSUE #4108: Keep shallow template clones empty of template contents
Follow-up discovered while reviewing namespace provenance in #4106, campaign #3575.

```js
const t = document.createElement('template');
t.innerHTML = '<div><span></span></div>';
t.cloneNode(false).content.childNodes.length;
// Expected: 0. Current pinned native behavior: 1 (the div is copied shallowly).
```

AngleSharp 1.8.1 `HtmlTemplateElement.Clone(owner, deep)` always loops over its content children and calls `node.Clone(owner, deep)`, even when `deep` is false. HTML template cloning steps copy contents only when the subtree flag is set: https://html.spec.whatwg.org/multipage/scripting.html#the-template-element . This also affects a template partially contained by `Range.cloneContents()` / `extractContents()`, since those native operations shallow-clone partially contained ancestors.

#4106 preserves the native algorithm while carrying namespace metadata, including this existing shallow-copy shape; its focused partial-template tests verify that metadata transfer does not introduce a new exception after extraction. Correcting the native cloning result is separate work. Any solution must keep AngleSharp-owned nodes, preserve native ownership and range behavior, and avoid private reflection or a rival DOM algorithm. No AngleSharp issue/PR is authorized or filed.

sebastienros:
Shallow template clones currently retain copied content children, including templates shallow-cloned as partial `Range` ancestors. Clear only the detached native copy's template content before namespace transfer, wrapping, or custom-element reactions. Fully selected templates and deep copies retain native contents, and extraction preserves moved-node identity.

Depends on #4106 (the branch includes its namespace provenance work until that PR merges). This follow-up changes only shallow-template correction and its range projection. No native DOM or Range algorithm is replaced.

Validation: 134 focused namespace/template/custom-element tests passed across net8.0 and net10.0. 418 browser WPT and final template ownership checks passed on net10.0; `git diff --check` passed. Tests cover optional shallow defaults, imports, nested templates, both partial-range directions, unchanged source contents, and moved-template identity. No exclusion or census changes.

Fixes #4108.


sebastienros:
The reviewed shallow-template correction is now included in #4114 at `ee8c0d5102967aadd444d2d4f58a9be92d1a05d3`, alongside the PI range-copy fix that extends the same traversal. That combined branch is based on merged #4106 and passed 150 focused tests across net8/net10 on AngleSharp 1.8.2. Full CI is running there.

The cancelled checks on this standalone branch were deliberately deferred during dependency integration. This PR stays open until the combined replacement merges, then will close as superseded; #4108 remains linked for closure by the replacement.

sebastienros:
Superseded by #4114, now merged as deb5a651863d57bc868298924488b90bc97d1cc0 after all eight CI checks passed. The combined change includes the shallow-template correction and the interacting Range PI copy fix; #4108 is resolved by that merge.

PR #4121, opened 2026-09-19
Backport #3551 to 4.x: BigInt reads the leading plus a decimal string may carry

lahma:
Backports [#3551](https://github.com/sebastienros/jint/pull/3551) (issue [#3540](https://github.com/sebastienros/jint/issues/3540)) to `4.x`.

Base: `4.x` at `ad07ed5cbc98376a4c48e6e63f6b119e193b09e5`.

## What it fixes

[StringToBigInt](https://tc39.es/ecma262/#sec-stringtobigint) parses its argument as a `StringIntegerLiteral`, whose `StrIntegerLiteral` is either a `SignedInteger` — `DecimalDigits`, `+ DecimalDigits` or `- DecimalDigits` — or a `NonDecimalIntegerLiteral`, which takes no sign at all. `TryStringToBigInt` walked the string once before parsing it and admitted a non-digit at the front only when it was `-`:

```js
BigInt('+12');   // 4.x today: SyntaxError        with this: 12n
BigInt(' +12 '); // 4.x today: SyntaxError        with this: 12n
12n == '+12';    // 4.x today: false              with this: true
12n < '+13';     // 4.x today: false              with this: true
```

`Number('+12')` had always been `12`, and every other engine reads the sign. The sign belongs to the decimal spelling and to nothing else, so `BigInt('+0x10')`, `BigInt('-0b11')` and `BigInt('+0o17')` are still `SyntaxError`s — the `0x`, `0o` and `0b` lanes below all match at index 0, which the sign occupies — as are `'+'`, `'++12'`, `'+ 12'`, `'+12.5'` and `'+1e3'`.

## What the pick carried and what it did not

The engine hunk in `Jint/Runtime/TypeConverter.cs` applies **verbatim**: this branch's `TryStringToBigInt` is unchanged from main's at the branch point.

Dropped, because `4.x` has neither file: the `docs/v5-migration.md` entry, and the `Jint.Tests/Runtime/WhiteSpaceDefinitionTests.cs` hunk.

[...]

PR #4122, opened 2026-09-19
Backport #3565 to 4.x: a split keeps its segments across a constraint that runs script

lahma:
Backports [#3565](https://github.com/sebastienros/jint/pull/3565) to `4.x`.

## What it fixes

`String.prototype.split` with a string separator collects its segments into a scratch `List<JsString>` held on `StringExecutionContext`, reused between calls so the common case allocates nothing. That buffer is **thread**-affine, not engine-affine, and the loop that fills it is not a leaf: it calls `Engine.Constraints.Check()` every 10,000 segments, and a host-supplied `Constraint` is host code that may run script — on this engine, or on another one sharing the thread. A split reached that way cleared the very list the outer split was still filling, and the outer call then returned only the segments it had managed to add since.

```csharp
sealed class LoggingConstraint : Constraint
{
    private readonly Engine _reporter = new();

    public override void Check() => _reporter.Evaluate("'x,y,z'.split(',').length");

    public override void Reset() { }
}

var engine = new Engine(options => options.Constraint(new LoggingConstraint()));
```

```js
'a,'.repeat(30000).split(',').length;  // 4.x today: 20004      with this: 30001
```

The wrong answer is silent — a shorter array, no exception, nothing in it wrong except that most of it is missing — and it takes a split long enough to reach a constraint check, so a host that hit it hit it only on its largest inputs. Measured on this branch before the fix: **20004** where 30001 is correct.

The buffer is now rented and returned rather than taken. The owner gets the shared list, a re-entrant caller gets a private one, and the release runs in a `finally`, so a constraint that throws mid-split leaves the buffer usable rather than permanently marked in use. Returning also clears it, which matters for a second reason: the segments are `JsString.CreateSliced` views that may reference the script source they were cut from.

Both engine hunks — `Jint/Native/String/StringExecutionContext.cs` and `Jint/Native/String/StringPrototype.cs` — apply **verbatim**: both files are unchanged from main's at the branch point.

Dropped: the `docs/v5-migration.md` entry (`4.x` has no such file).

Hand work in the test, which is transcribed from NUnit to xUnit v3: it registers its constraint through this branch's `options.Constraint(constraint)` rather than main's renamed `options.AddConstraint(constraint)`. No engine API is added or changed by this PR.

## Verification

Ported tests (`HostConstraintReentrancyTests`, 2 cases), against the **unfixed** `4.x` tree and then the fixed one:

| | net10.0 | net472 |
| --- | --- | --- |
| before the fix | 1 failed / 1 passed (`…to be 30001.0, but found 20004.0`) | 1 failed / 1 passed (same) |
| after the fix | **0 failed / 2 passed** | **0 failed / 2 passed** |

Whole projects, Release:

| --- | --- | --- |
| `Jint.Tests` | 7170 passed / 0 failed / 4 skipped | 7085 passed / 0 failed / 4 skipped |
| `Jint.Tests.PublicInterface` (incl. the public-API snapshot) | 1854 passed / 0 failed / 9 skipped | 1846 passed / 0 failed / 9 skipped |
| `Jint.Tests` with `JINT_HOST_CONTRACT_VERIFICATION=1` | 7170 passed / 0 failed / 4 skipped | 7085 passed / 0 failed / 4 skipped |
| `Jint.Tests.PublicInterface` with `JINT_HOST_CONTRACT_VERIFICATION=1` | 1858 passed / 0 failed / 5 skipped | 1850 passed / 0 failed / 5 skipped |

The public-API snapshot passes untouched, which is the operative proof that nothing in `4.x`'s API moved.

test262 was run once over a local tree with this and the sibling backports merged together; the counts are in a comment below.


lahma:

```
```

| --- | --- |

PR #4123, opened 2026-09-19
Backport #3585 to 4.x: a number outside a parameter's range is no longer scored as a fit for it

lahma:
Backports [#3585](https://github.com/sebastienros/jint/pull/3585) to `4.x`.

## What it fixes

Overload scoring rates a JavaScript number against a numeric parameter by magnitude, so a number outside the parameter's range must not be proposed as a fit for it. `float`, `short`, `ushort`, `byte` and `sbyte` always gated on the value fitting; `int` and `long` did not, so `3000000000` was a perfect match for an `int` parameter and `1e300` a good one for a `long`.

A perfect score also ends the search, so the wider overload beside it was never scored at all:

```js
h.Take(3000000000);          // 4.x today: binds the int overload, then fails to convert
new IntOnlyBoxed(3e9);       // 4.x today: System.OverflowException out of Evaluate
```

On the operator and constructor lanes the failed conversion left through the host's `Evaluate` as a CLR `OverflowException` rather than as a catchable JavaScript error; on the method lane the candidate list came back holding only the candidate that cannot bind. With this change the value binds to the narrowest parameter that can actually hold it (`h.Widen(3000000000)` → the `long` overload), or resolution declines the way it already did for `byte`.

Every integral gate now asks the same `FitsInt32`/`FitsInt64` predicate that the fast numeric conversion asks, so the score and the conversion cannot disagree about what fits. The narrower gates also read the double itself rather than an `(int)` cast of it, which for an out-of-range value is framework-dependent — .NET saturates the conversion, .NET Framework does not — and the comparisons were reading whatever that happened to be.

The engine hunks in `Jint/Runtime/Interop/InteropHelper.cs` apply **verbatim**. `4.x` does not carry main's #3037–#3060 security stack, so `Jint/Runtime/Interop` diverges in general — but the numeric scoring block in `CalculateMethodParameterScore`, the `LongMaxValueExclusiveUpperBound` constant and `TryConvertNumberFast` are all unchanged from main's at the branch point, and the placement was checked against `4.x`'s own file rather than trusted to the pick. `ScoreForDifferentTypeButFittingNumberRange` occurs in exactly one block in this file, so there is no second copy to keep in step.

Dropped: the `docs/v5-migration.md` entry and the `Jint/Runtime/Interop/AGENTS.md` hunk (`4.x` has neither file).

Hand work in the test, which is transcribed from NUnit to xUnit v3: `ALoneIntConstructorRefusesRatherThanOverflowing` asserts this branch's own constructor-resolution message, `"Could not resolve a constructor for type … for given arguments"`, which main later reworded. Everything else compiled and ran as picked, including `Options.Interop.AllowOperatorOverloading` and the method-resolution message.

## Verification

Ported tests (`HostNumericRangeOverloadTests`, 20 cases), against the **unfixed** `4.x` tree and then the fixed one — the unfixed run uses the same final test file, so the counts are comparable:

| | net10.0 | net472 |
| --- | --- | --- |
| before the fix | 10 failed / 10 passed | 10 failed / 10 passed |
| after the fix | **0 failed / 20 passed** | **0 failed / 20 passed** |

Among the ten: `ALoneIntConstructorRefusesRatherThanOverflowing` failed with `Expected a <Jint.Runtime.JavaScriptException> to be thrown, but found <System.OverflowException>`, which is the shape a host sees.

Whole projects, Release:

| --- | --- | --- |
| `Jint.Tests` | 7170 passed / 0 failed / 4 skipped | 7085 passed / 0 failed / 4 skipped |
| `Jint.Tests.PublicInterface` (incl. the public-API snapshot) | 1872 passed / 0 failed / 9 skipped | 1864 passed / 0 failed / 9 skipped |
| `Jint.Tests` with `JINT_HOST_CONTRACT_VERIFICATION=1` | 7170 passed / 0 failed / 4 skipped | 7085 passed / 0 failed / 4 skipped |
| `Jint.Tests.PublicInterface` with `JINT_HOST_CONTRACT_VERIFICATION=1` | 1876 passed / 0 failed / 5 skipped | 1868 passed / 0 failed / 5 skipped |

The public-API snapshot passes untouched, which is the operative proof that nothing in `4.x`'s API moved: `FitsInt32`/`FitsInt64` are `private static` on an `internal` type.

test262 was run once over a local tree with this and the sibling backports merged together; the counts are in a comment below.


lahma:

```
```

| --- | --- |

Issue #3609, opened 2026-09-02
supportedLocalesOf-unicode-extensions-ignored times out under full-suite load on unmodified main, and every PR pays a re-run for it


lahma:
`intl402/supportedLocalesOf-unicode-extensions-ignored.js` fails the full test262 suite intermittently by hitting the engine's 30-second default timeout, while passing in isolation in ~2 seconds — on **unmodified `main`**, measured directly during #3608's verification (full suite on the clean merge base: 102,572 / 1 / 115, that one failure being this file; the branch run failed the same file in both modes; isolation passes on both trees).

This flake has now cost nearly every recent PR's verification a re-run: it appears in the reports of #3543, #3546, #3555, #3556, #3559, #3579, #3585 and #3608, always with the same signature (30–35 s under full-solution parallel load, 2 s alone). The mechanism is presumably that the test's `supportedLocalesOf` sweep over the locale set is CPU-heavy enough that, descheduled under a saturated box, it overruns the wall-clock default — the same wall-clock-versus-CPU confusion #3486 removed from the regex budgets.

Worth a deliberate fix rather than ritual re-runs. Options that suggest themselves: give the harness's engine a larger timeout for this one file (it is a legitimate long computation, not a hang), profile whether the sweep is quadratic somewhere and simply fix the cost, or run it serialized. Whoever picks it up should reproduce under load first (`staging/sm/Array/toSpliced-dense.js` and `sort_large_countingsort.js` show the same signature and may share the remedy).

PR #4124, opened 2026-09-19
Backport #3778 to 4.x: BestAvailableLocale resolves a CultureInfo only when the available-locale set cannot answer

lahma:
Backports [#3778](https://github.com/sebastienros/jint/pull/3778) (issue [#3609](https://github.com/sebastienros/jint/issues/3609)) to `4.x`.

## What it fixes

`BestAvailableLocale` interleaved two things: [the spec's own algorithm](https://tc39.es/ecma402/#sec-bestavailablelocale), which is a set lookup and a truncation, and a .NET probe that resolves each candidate to a `CultureInfo` and tries that culture's name. Resolving a tag costs a `new CultureInfo` — about 15 µs — and the cache in front of it is bounded on purpose, so a sweep over more distinct tags than it holds resolves nearly every one. Every iteration of the truncation paid for it.

Measured **on this branch**, with the sweep `intl402/supportedLocalesOf-unicode-extensions-ignored.js` performs (640 generated tags, three `supportedLocalesOf` calls each, one constructor and one matcher): **6,913** culture lookups before, **none at all** after — every tag is answered by the available-locale set, bare or truncated.

The two are now separate passes: the spec algorithm first, the culture probe only if that found nothing. **The answer cannot move.** The truncation visits prefixes longest-first, and a name the probe returns is either a prefix of the candidate — in which case the first pass reaches that same prefix, and no longer one can be available, because the available set holds .NET culture names and none of those carries a `-u-` sequence — or it is not a prefix at all, which the algorithm does not admit as an answer, and which the second pass still produces exactly as before. Casing is not a case either: the set compares ordinal-ignore-case.

`IntlUtilities.CultureLookupCount` is added (`internal`) so the test bounds a **count and not a duration**. That distinction is the whole of the issue: the cost was only ever visible as a wall-clock timeout under full-suite load, and a timing test would be the same measurement that cannot tell a regression from a busy machine.

Both engine hunks in `Jint/Native/Intl/IntlUtilities.cs` apply **verbatim**: this branch's `BestAvailableLocale` — including the `zh-` script-expansion probe, which stays in both passes because it is a set lookup — and its `GetCultureInfo` are unchanged from main's at the branch point. No doc hunks were in this PR to drop.

Hand work in the test: it is transcribed from NUnit to xUnit v3 into this branch's existing `Jint.Tests/Runtime/IntlCultureCacheTests.cs`, which is already an xUnit `[Fact]` fixture.

## Verification

Ported tests (the two new ones plus the three already there, 8 cases), against an **unfixed** `4.x` tree — the two-pass split reverted, the `CultureLookupCount` instrumentation kept so the test can run at all — and then the fixed one:

| | net10.0 | net472 |
| --- | --- | --- |
| before the fix | 1 failed / 7 passed (`…must not resolve a culture per call…, but found 6913L`) | 1 failed / 7 passed (`…found 6883L`) |
| after the fix | **0 failed / 8 passed** | **0 failed / 8 passed** |

Whole projects, Release:

| --- | --- | --- |
| `Jint.Tests` | 7175 passed / 0 failed / 4 skipped | 7090 passed / 0 failed / 4 skipped |
| `Jint.Tests.PublicInterface` (incl. the public-API snapshot) | 1852 passed / 0 failed / 9 skipped | 1844 passed / 0 failed / 9 skipped |
| `Jint.Tests` with `JINT_HOST_CONTRACT_VERIFICATION=1` | 7175 passed / 0 failed / 4 skipped | 7090 passed / 0 failed / 4 skipped |
| `Jint.Tests.PublicInterface` with `JINT_HOST_CONTRACT_VERIFICATION=1` | 1856 passed / 0 failed / 5 skipped | 1848 passed / 0 failed / 5 skipped |

The public-API snapshot passes untouched, which is the operative proof that nothing in `4.x`'s API moved: `CultureLookupCount` is `internal`.

test262 was run once over a local tree with this and the sibling backports merged together; the counts are in a comment below. Worth noting there: the two conformance failures this machine's `4.x` control carries are `intl402/supportedLocalesOf-unicode-extensions-ignored.js` strict and sloppy, timing out at ~31 s against a 30 s budget under load — this change is what removes them.


lahma:

```
```

| --- | --- |

PR #4128, opened 2026-09-19
Let the white-space census admit BigInt's leading plus, which 4.x now reads

lahma:
The two backports interacted exactly as their PR bodies predicted. #4120 ported `WhiteSpaceDefinitionTests` from #3542 with the "BigInt with a leading pad" row listing only the leading zero, because 4.x still rejected a leading `+`; #4121 then backported #3551, after which `BigInt('+12')` parses and the row's assertion — the lane accepts exactly the specification's white space and nothing else — reports `U+002B was accepted`. Each PR's CI was green against a merge ref that predated its sibling; a local run of the merged 4.x head (`f27c975ef`) is what showed it: Jint.Tests 1 failed / 7,518 passed on net10.0, 1 / 7,433 on net472, the same row on both.

Main made this row change in #3551 itself; this is that hunk, which #4121 had to leave out while the file did not yet exist on 4.x. Test-only. `WhiteSpaceDefinitionTests` passes on net10.0 and net472 with it.

Base: `f27c975ef`.


lahma:
Verification of the merged 4.x head `f27c975ef` this PR is based on, run locally after #4120–#4124 landed:

| leg | result |
| --- | --- |
| `dotnet build -c Release` | 0 errors (the one warning is the pre-existing MSB3277 in Jint.Tests.CommonScripts/net472) |
| Jint.Tests.PublicInterface net10.0 / net472 | 1,874 / 1,866 passed, 0 failed — the surface snapshot is unchanged |
| Jint.Tests net10.0 / net472 | 1 failed / 7,518 passed and 1 failed / 7,433 passed — the row this PR fixes, on both |
| Jint.Tests.Test262 | **102,501 passed / 0 failed / 183 skipped** (3 m 7 s) — two better than the 102,499 / 2 control, #4124 having retired the `supportedLocalesOf` timeouts |

With this PR's row change, `WhiteSpaceDefinitionTests` is 35/35 on both frameworks.

PR #4146, opened 2026-09-22
Expose ArrayBuffer view length-tracking metadata to hosts

PrzemyslawKlys:
## Summary

Expose the native length-tracking mode of ArrayBuffer views to hosts.

## Details

`JsValue.IsLengthTrackingArrayBufferView()` distinguishes fixed-length typed arrays and DataViews from views that track their buffer's length. They can expose identical lengths until a resizable buffer grows, so a host serializer cannot recover this distinction from the JavaScript length getters.

The extension reads native metadata without executing script, resizing the buffer, or making `JsDataView` public. It reports the construction mode even after a view goes out of bounds or its buffer is detached; callers still need their own validity checks. Non-view values and proxies return `false`.

[OfficeIMO](https://github.com/EvotecIT/OfficeIMO) uses this boundary for history-state and frame-message serialization in its isolated HTML runtime. Jint owns the view metadata; the consumer owns serialization policy and budgets. A narrow backport on its maintained 4.16 fork lets that integration progress independently of this PR and a future upstream release.

## Linked issue

No existing issue; this proposes a host API addition.

## Test plan

- [x] Added external-consumer tests in `Jint.Tests.PublicInterface` for resize, detach, growable shared buffers, fixed/empty views, proxies and spoofed objects.
- [x] Release net10.0: 20 metadata, public API and XML documentation checks pass.
- [x] Release net8.0: all 10 metadata cases pass.
- [x] Regenerated and inspected all five public API baselines; only the new method is added.
- [x] Independent read-only review of the native API and its 4.16 backport found no issues.

Validated with SDK 10.0.401 on macOS arm64. The full solution and Test262 suite were not run; this adds a host query without changing ECMAScript behavior.

## Breaking change?

No. This is an additive host API.


lahma:
Merged as 04c21209c, thanks @PrzemyslawKlys. It's small, it reads the same `LengthAuto` marker the built-in getters use, and it's documented, with the tests in the right project. All CI checks passed. (#4148 will need its migration-guide section renumbered to §5.39 now that this one takes §5.38.)

PR #4148, opened 2026-09-22
Allow hosts to reject jobs after execution context retirement

PrzemyslawKlys:
A host can retire a document while its engine is draining Promise reactions. Removing DOM callbacks or cancelling timers does not stop reactions already in Jint's queue, so they can still access retained host objects.

Add `Host.CanExecuteJob()`, checked before each queued job in the shared execution path. The default permits execution. A host returning `false` discards the job without settling its promise; the currently running script or job continues, and external operations and waits remain the host's responsibility.

[OfficeIMO](https://github.com/EvotecIT/OfficeIMO) uses this boundary to retire closed popup and detached-frame realms. Its maintained Jint 4.16 fork carries the same host contract; this PR targets upstream main without requiring that consumer to adopt Jint 5. No other upstream contribution is required first.

Validation: four external-host tests on .NET 8 and .NET 10; shared task/microtask scheduler regression; all five generated API baselines and XML documentation checks; migration-guide checks. The corresponding consumer implementation passed 59 selected lifecycle tests on each runtime and an independent review with targeted confirmation.


lahma (changes requested):
Thanks for this, @PrzemyslawKlys. The motivating case is real: a host retires a document from inside a drain, and it cannot stop the drain it is standing in. Putting the check in `Engine.RunEventLoopJob` covers all four drain sites in `EventLoop.cs` with one line. The memory-limit bookkeeping is skipped consistently, a dropped `setInterval` does not keep `TimeUntilNextScheduledWork` alive, and your tests plus the full `Jint.Tests` and `Jint.Tests.PublicInterface` suites pass here on net10.0 and net472.

Several jobs carry bookkeeping of their own, so I probed what a refused job leaves behind. All snippets below use this host:

```csharp
sealed class RetirableHost : Host
{
    volatile bool _active = true;
    public bool Active { get => _active; set => _active = value; }
    public override bool CanExecuteJob() => _active;
}
```

**1. The engine's own waits cannot end, and one of them blocks `Dispose`.**

`ModuleImportOperation` and `FetchHandlerOperation` both document a polling loop and promise that it cannot poll forever: `ObserveAbandonment` fails the operation when the restore fence strands it. A refused job strands it the same way, but nothing reaches that flag:

```csharp
var host = new RetirableHost();
var loader = new HandOffModuleLoader(); // parks each ModuleLoadCompletion until DeliverAll()
using var engine = new Engine(o => { o.UseHostFactory(_ => host); o.UseModules(loader); });

var op = engine.Modules.StartImport("m");
host.Active = false;
loader.DeliverAll();
for (var i = 0; i < 1000 && !op.IsCompleted; i++) engine.Tasks.ProcessTasks();

op.IsCompleted.Should().BeTrue(); // fails: still false after 1000 turns
```

`InvokeFetchHandler` behaves the same with `async function handler(req) { retire(); await null; return new Response('ok'); }`: `IsCompleted` is still false after 1000 turns.

`EvaluateAsync`, `UnwrapIfPromise` and `Modules.Import` wait out `PromiseTimeout` and then report `Timeout of … reached`. For `Import` the message is "An asynchronous module loader did not finish the load in time", although the loader did finish. With `PromiseTimeout` disabled the await never ends, and `Dispose` is refused while it is pending:

```csharp
var host = new RetirableHost();
var engine = new Engine(o => { o.UseHostFactory(_ => host); o.Constraints.PromiseTimeout = TimeSpan.Zero; });
ManualPromise? mp = null;
engine.SetValue("pending", new Func<JsValue>(() => { mp = engine.Tasks.RegisterPromise(); return mp.Promise; }));

var task = engine.EvaluateAsync("(async () => (await pending()) + 1)()");
host.Active = false;
mp!.Resolve(41);

engine.Dispose(); // InvalidOperationException: "...has an asynchronous operation in progress..."
// task is still pending 3 s later; only a CancellationToken passed to EvaluateAsync ends it
```

The migration note says settling host waits remains the host's job. These waits belong to the engine, though, and the host holds no promise it could settle.

Related: a constraint failure on a host `Task` is delivered as a job (`JsValue.ConvertTaskToPromise` enqueues `dispatch.Throw`). Refusing it turns an `ExecutionCanceledException` into a `PromiseTimeout` rejection. My probe faulted a `TaskCompletionSource<int>` with `new ExecutionCanceledException()` after retirement: the host saw the timeout rejection after 750 ms, while the control engine got `ExecutionCanceledException` in 3 ms.

**2. Two paths still run script after retirement.**

Both run from the drain but outside `RunEventLoopJob`. Engine built with `o.UseWebApis()`, `retire` flips `host.Active`:

```csharp
// requestIdleCallback: IdleCallbackQueue.TryRunIdleCallback invokes the callback directly
engine.Execute("var ran = false; Promise.resolve().then(() => retire()); requestIdleCallback(() => { ran = true; });");
engine.Tasks.ProcessTasks();
engine.Evaluate("ran").AsBoolean().Should().BeFalse(); // fails: true

// unhandledrejection: Engine.NotifyAboutRejectedPromises dispatches to listeners directly
engine.Execute("""
    var seen = false;
    addEventListener('unhandledrejection', () => { seen = true; });
    Promise.resolve().then(() => retire());
    Promise.reject(new Error('x'));
    """);
engine.Tasks.ProcessTasks();
engine.Evaluate("seen").AsBoolean().Should().BeFalse(); // fails: true
```

**3. A retired engine keeps taking shared resources until it is disposed.**

Take a `LockManager` shared between engines, and a lock request the engine queued before it retired. When the holder releases, the manager grants that request and adds it to the held set. The grant job is then refused, so the lock stays held, and every other engine on that manager starves until `Dispose`:

```csharp
var manager = new LockManager();
var holder  = new Engine(o => { o.UseWebApis(); o.WebApi.Locks.Manager = manager; });
var retired = new Engine(o => { o.UseHostFactory(_ => host); o.UseWebApis(); o.WebApi.Locks.Manager = manager; });
var third   = new Engine(o => { o.UseWebApis(); o.WebApi.Locks.Manager = manager; });

holder.Execute("var release; navigator.locks.request('x', () => new Promise(r => release = r));");
retired.Execute("navigator.locks.request('x', () => {});");
host.Active = false;
holder.Execute("release()");
third.Execute("var got = false; navigator.locks.request('x', () => { got = true; });");
// after 20 pump rounds of all three engines: got == false, manager.HeldCount == 1
// after retired.Dispose(): got == true
```

From reading the code (I have not probed these), the same shape shows up in three more places:
- `FetchOperation.Release()` runs inside the settle job. It is what leaves the in-flight set that `MaxConcurrentRequests` counts against, disposes the CTS, and lets a pending response body go.
- A `MessagePort`'s messages keep piling up on the endpoint queue while its arrival jobs are dropped.
- A `BroadcastChannel` subscription keeps costing every sender a job that is thrown away.

**4. Returning `true` again leaves the engine permanently degraded.**

The docs ask the host to keep returning `false`, but nothing enforces that. After false → true:
- `scheduler.postTask` never runs again: `SchedulerQueue._pumpScheduled` was left `true` by the dropped pump job. `PerformanceObserverRegistry` and `FileReadQueue` have the same flag.
- A `setInterval` whose tick was dropped never fires again, because it re-arms inside its own job (`TimerEntry.Fire`). Its id stays registered and counts against `MaxActiveTimers`.
- `StartImport` of a specifier whose completion was dropped never completes and never re-fetches, because `RemovePendingLoad` runs inside the dropped job.

The restore fence handles each of these explicitly (the queues' `Clear()` methods, `DiscardPendingLoads`, and so on).

**Suggestion.** `RestoreGlobalSnapshot` already has the machinery for "work from here on must not run": `EventLoop.NextGeneration()` plus `Engine.ResetTransientEvaluationState()`. That covers `Modules.DiscardPendingLoads`, `DiscardAtomicsWaiterDeadlines`, `DiscardPendingRejectionNotifications`, `AbandonHostStreamBridges`, and `WebApiEngineState.ResetTransientState` (timers, scheduler, fetches, sockets, ports, channels, locks, idle callbacks, observers).

Could retirement be an engine-owned, one-way operation, say on `Engine.Advanced`, that does all of that and also sets a latch checked at dequeue? The latch is what still covers work queued after retirement, like the `new` reaction in your first test. With that:
- The polled operations in 1 would fault properly for free, since both `ObserveAbandonment`s key on the generation.
- 2, 3 and 4 would be closed by construction.
- The engine's own waits could fail fast with a specific exception instead of waiting out `PromiseTimeout`.

If the hook stays as it is, I think 1 and 2 at least need fixing, and 3 and 4 need documenting, with `Dispose` named as the release.

**API shape.** `public virtual` matches `EnsureCanCompileStrings` and `InitializeShadowRealm`, but those take a realm. The closest spec step is HTML's *check if we can run script*, which [HostEnqueuePromiseJob](https://html.spec.whatwg.org/multipage/webappapis.html#hostenqueuepromisejob) performs per job realm. Jint's `Host.HostEnqueuePromiseJob(Action job, Realm realm)` already receives that realm and drops it. In this repo's `Jint.Browser`, a frame is a realm inside the page's engine, so a detached frame there could not use a per-engine hook with no arguments without stopping the whole page. A parameterless virtual is also hard to widen later. Would a realm (or a small job-context) parameter work for OfficeIMO?

**Docs.** A few places should name the new reason a job is discarded:
- the `ProcessTasks`, `Post` and `RegisterPromise` remarks, which currently name only the restore fence
- `ModuleImportOperation.IsCompleted` and `FetchHandlerOperation`, whose "a host polling this cannot poll forever" no longer holds
- the `Engine.Tasks.Post` row in `Jint.Tests.PublicInterface/AGENTS.md`'s contract table

It would also help for the XML doc to say two things. The flag may be flipped from another thread, so the host needs a volatile read. And an override that throws loses the job, since it is already dequeued, and the exception erupts from the pump.

**Tests.** The placement is right: PublicInterface for reachability, `Jint.Tests` for the internal task-budget lane. What's missing is coverage for the web APIs, the two polled operations, and interaction with `RestoreGlobalSnapshot`.

**Cost.** This adds one virtual call per job, including every promise reaction and every `await` resumption. A probe with 1,000 `await`s counted exactly 1,000 calls. It is probably small next to the dequeue, but it sits on the hottest async path, so we'll take a paired measurement on our side before merge.

Happy to share the full probe file if it helps.

